Issue #7, opened 2019-09-16
Properly implement and test client-side subscriptions


tomaka:
The `rpc_api!` macro should provide a way to start a subscription and listen to events returned by the server.

Nothing has been started here, at all.

tomaka:
The `RawClient` trait needs to change, as it doesn't provide any way of feeding us the subscription messages coming from the server.

To do so, I think we should add some sort of `request_with_keep_alive` method that generates a request object (with an ID, like with `RawServer`), and a `next_event` method that returns whenever a subscription has any incoming message.

The details of that are still a bit blurry.

Once this is done, the `Client` needs to account for these changes, but that shouldn't be too difficult (I think).

Then the `rpc_api!` macro should make subscription-related methods return a `Result<impl Stream, ...>` where the `Stream` borrows from the `Client` passed as parameter to the method.

tomaka:
After #43, we still need a way to close subscriptions.
Since closing a subscription is done by performing a request to the server, we need some brainstorming to figure out what to do in situations such as if the server refuses the closing.

niklasad1:
Closing as we have some integration tests for testing subscriptions and unsubscribing.

PR #28, opened 2019-09-26
Rework the Client to support multiple concurrent requests

tomaka:
I wanted to open this PR as a draft, but GitHub's UI seems broken.


tomaka:
Apparently the draft button worked even though it didn't seem so. However I can't edit my previous comment because the edit button doesn't work either.

tomaka:
While the code is missing lots of aspects, subscriptions are now working.

tomaka:
cc #7

PR #31, opened 2019-09-27
Access control

dpmkl:
Implemented access control using the server_utils from jsonprc. 
I have not written any tests yet as I am not sure if this implementation is desired. 

Please let me know how to proceed.


dpmkl:
[clabot:check]

http/src/lib.rs
```diff
@@ -27,6 +30,14 @@ pub async fn http_server(
     Ok(Server::new(HttpRawServer::bind(addr).await?))
 }
 
+/// Starts a [`Server`](../Server) object that serves HTTP with a whitlist access control.
+pub async fn http_server_with_acl(
+    addr: &SocketAddr,
+    allowed_hosts: Option<Vec<Host>>,
```
tomusdrw: I'd be more in favour of using an enum for this all the way down, as I found the `Option` type very confusing in the past, like:
```
allowed_hosts: None // - that means that all hosts are allowed, in contrary to what you read
allowed_hosts: Some(vec![]) // - this means that no hosts are allowed
```

http/src/utils.rs
```diff
@@ -0,0 +1,70 @@
+use hyper::{self, header};
```
tomusdrw: I think we miss pre-ambles in the entire codebase (probably that was the case for jsonrpc too)

server-utils/src/cors.rs
```diff
@@ -0,0 +1,594 @@
...
+	}
+
+	/// Creates new origin given protocol, hostname and port parts.
+	/// Pre-processes input data if necessary.
+	pub fn new<T: Into<Port>>(protocol: OriginProtocol, host: &str, port: T) -> Self {
+		Self::with_host(protocol, Host::new(host, port))
+	}
+
+	/// Attempts to parse given string as a `Origin`.
+	/// NOTE: This method always succeeds and falls back to sensible defaults.
```
tomusdrw: That was a decision we made for backward compatibiltiy afair, but I think in `jsonrpsee` we could conside enforcing a proper format of `Origin`.

I.e. now we default to `http` if no protocol is given, because the `Origin` header needs a protocol, instead we could just fail with error and force user to specify that.

@tomaka your call which direction you would prefer this project to go.

server-utils/src/hosts.rs
```diff
@@ -0,0 +1,229 @@
...
+	}
+}
+
+impl Pattern for Host {
+	fn matches<T: AsRef<str>>(&self, other: T) -> bool {
+		self.matcher.matches(other)
+	}
+}
+
+impl ::std::ops::Deref for Host {
```
tomusdrw: Do we really need this?

server-utils/src/hosts.rs
```diff
@@ -0,0 +1,229 @@
...
+/// Specifies if domains should be validated.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum DomainsValidation<T> {
+	/// Allow only domains on the list.
+	AllowOnly(Vec<T>),
+	/// Disable domains validation completely.
+	Disabled,
+}
+
+impl<T> Into<Option<Vec<T>>> for DomainsValidation<T> {
```
tomusdrw: I'd get rid of this as per other comment.

server-utils/src/lib.rs
```diff
@@ -0,0 +1,17 @@
+//! JSON-RPC servers utilities.
+#![deny(missing_docs)]
+
+#[macro_use]
```
tomusdrw: Let's properly rewrite to `edition=2018` as stated in the toml and remove this stuff.

http/src/server/background.rs
```diff
@@ -79,7 +88,13 @@ impl BackgroundHttp {
 async fn process_request(
     request: hyper::Request<hyper::Body>,
     fg_process_tx: &mut mpsc::Sender<Request>,
+    allowed_hosts: &Option<Vec<Host>>,
 ) -> hyper::Response<hyper::Body> {
+    // Process access control 
+    if !utils::is_host_allowed(&request, allowed_hosts) {
```
tomusdrw: Would be good to have a bunch of high-level tests for it.

http/src/lib.rs
dpmkl: Something like this: 
`
pub enum AccessControl {
All, 
Restricted(Vec<Hosts>),
}
`
?

dpmkl:
> Also note that this PR currently only verifies Host header and does not implement CORS.

Should CORS and allowed_hosts be treated separately or would a struct holding both make sense, as it would keep the method signature of 'process_request' more compact ? 

```Rust
pub struct AccessControl {
    pub allowed_hosts: AllowedHosts,
    pub cors_domains: Option<Vec<AccessControlAllowOrigin>>,
    pub cors_max_age: Option<u32>,
    pub cors_allowed_headers: AccessControlAllowHeaders,
}
```

If so, would we want a builder pattern for AccessControl?  

@tomaka @tomusdrw

http/src/lib.rs
dpmkl:
```Rust
/// Access control for http header 'host' 
#[derive(Clone)]
pub enum AllowedHosts {
    /// Allow requests from any host 
    Any,
    /// Allow only a selection of specific hosts
    Only(Vec<Host>),
}
```

server-utils/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+name = "jsonrpsee-server-utils"
+repository = "https://github.com/paritytech/jsonrpsee"
+version = "0.1.0"
+
+[dependencies]
+bytes = "0.4"
+globset = "0.4"
+hyper = "0.13.0-alpha.4"
+jsonrpsee-core = { path = "../core" }
+lazy_static = "1.1.0"
```
yjhmelody:
```suggestion
lazy_static = "1.4"
```
better?

dpmkl:
Please let me know what is missing to get this merged. 
The CI test is failing, which seems to be related to #50 (tokio_io).
In order to test CORS properly, I might have to introduce a "pub(crate)" to the process_request method, if you see a better way, I'm all open to feedback. 
#54 Should not be a problem, personally I prefer to run rustfmt as often as possible for the sake of readability and consistency.

tomaka:
Sorry for not getting back to you!
I'm hesitating a lot about the API, but I don't want to leave this PR forever just because I can't make up my mind. Instead let's just merge this PR and do changes on top of it later.

PR #35, opened 2019-10-07
Add license header everywhere

tomaka:
The license is MIT and that doesn't change, but the headers were missing.


tomaka:
Should we merge this? cc @folsen

tomaka:
Since nobody is answering, I'm going to merge tonight.

PR #38, opened 2019-10-08
Make notifications work with the proc macro

tomaka:
Fixes the proc macro to make notifications properly work.
From the server side, notifications are the same as requests except without the `respond` parameter.

The code of the proc macro is getting kind of messy after this PR and will need to be cleaned up.


tomaka:
Hopefully #39 will fix CI.

PR #62, opened 2019-11-18
Add missing LICENSE to server-utils and bump to 1.0.0

tomaka:


tomaka:
I realized that I should have asked you @dpmkl whether it's fine to add this license for the code that you added. On the other hand, you also mostly copy-pasted Parity's code :sweat_smile:

Issue #72, opened 2019-11-21
Add a SharedClient and SharedServer


tomaka:
After some discussion with @jimpo, let's make the API more convenient to use.

As a first step, let's add a `SharedClient` and a `SharedServer`. See the sketches below.
As a second step, let's rename things around: `RawClient`/`RawServer` should become `ClientTransport`/`ServerTransport`, `Client`/`Server` should become `RawClient`/`RawServer`, and `SharedClient`/`SharedServer` should become `Client`/`Server`.

Sketches for the new API layers:

```rust
impl SharedClient<...> {
    pub async fn request<Ret>(&self, method: ..., params: ...) -> Ret;
    pub async fn subscribe(&self, method: ..., params: ...) -> Subscription;
}

impl Subscription {
    pub async fn next(&mut self) -> ...;
}

impl Drop for Subscription {
    ...  // unsubscribes
}
```

```rust
impl SharedServer<...> {
    /// Registers a request that the server must handle. Returns an error if already registered.
    pub fn register_request<Params, Ret>(&self, method: ...) -> RequestHandler<Params, Ret>;
    pub fn register_subscription<Notif>(&self, subscribe_method: ..., unsubscribe_method: ...) -> Subscription<Notif>;
}

impl<Params, Ret> RequestHandler<Params, Ret> {
    /// User must call this repeatedly and answer the requests.
    pub async fn next(&mut self) -> ...;
}

impl<Notif> Subscription<Notif> {
    pub async fn send_all(&self, to_send: Notif) { ... }
}
```

tomaka:
Since we can't implement `SharedClient`/`SharedServer` in a `no_std` environment, we shouldn't put them in `jsonrpsee-core` but in a different crate.

tomaka:
A nice property of `SharedClient`/`SharedServer` is that they don't need any template parameter.

PR #75, opened 2019-11-25
Add a SharedServer

tomaka:
cc #72


tomaka:
This works by spawning a background task that takes ownership of the server, and sends/receives commands from/to multiple frontend objects.
As an example, calling `register_method()` returns a `RegisteredMethod` object that can then be polled for incoming requests that use that method.

However, one of the difficulties is choosing the channel size to communicate from the background task to the front.

The `RegisteredMethod` contains a receiver for requests that the client sends. The sending side is owned by the background task. If that channel is full, the background task will have to freeze and wait for the `RegisteredMethod` to be processed. If the processing of the `RegisteredMethod` is blocked by something else, then the entire server becomes unresponsive.

Consequently, I think that the server should return an `InternalError` if a channel is full and doesn't unclog itself after a few dozen milliseconds.

expenses:
wait oops lol

tomaka:
All right, let's merge this. It's not 100% clean and tested, but so is the rest of this library. I first would like to have a good architecture/API, and then consolidate.

PR #76, opened 2019-12-08
Parse both named and positioned parameters

yihuang:
Current implementation only parsed named request paramters, which is problematic when migrating jsonrpc project.
This patch make macro generated implementation support both named and positioned paramters.

### DIFF
--- a/core/src/server/params.rs
+++ b/core/src/server/params.rs
@@ -35,11 +35,34 @@ pub struct Params<'a> {
 }
 
 /// Key referring to a potential parameter of a request.
-pub enum ParamKey<'a> {
-    /// String key. Only valid when the parameters list is a map.
-    String(&'a str),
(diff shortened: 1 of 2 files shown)


yihuang:
[clabot:check]

tomaka:
Thanks for the PR!
I don't think it's correct to change `ParamsKey` this way, though.
I think the fix should be done entirely in the proc-macro code. However after #75, that fix would be done in the `SharedServer` instead.

yihuang:
Thanks, I'll look into the new design, may I ask what's the status of development of jaonrpsee, is it coming to a release anytime soon?

tomaka:
It's mostly me working on it, and I've been distracted by other stuff.
I think the core is really robust, but I'm struggling to make it easy to use. Hopefully #72 should achieve that.

tomaka:
Closing in favour of #94!

PR #77, opened 2019-12-11
Update to stable futures

### LINKED ISSUE #50: Update to futures 0.3
Blocked on https://github.com/hyperium/hyper/pull/2007 and on `soketto` updating.

ascjones:
Closes #50.

- [x] fix tests:`no current reactor` at https://github.com/paritytech/jsonrpsee/blob/1a5a8d132daad84fc9b2f0c003b7c0ce41c9cc7d/http/src/server/background.rs#L77


tomaka:
> fix tests:no current reactor at 

You have to update to tokio v0.2.

ascjones:
Same failure still

ascjones:
I think this is why: 

https://github.com/hyperium/hyper/commit/cb3f39c2dc6340060f6b17f354f04c872a947574#diff-afb1712a59f738bc3dfe39004cb5e4b9L132

Previously using `Handle::default()` ([lazy](https://tokio.rs/docs/internals/net/#handle-current-vs-handle-default)) where now it goes to `Handle::current()` (eager) as per https://github.com/tokio-rs/tokio/pull/1666. 

So not sure how to fix this - just discovered it so will try a few things. Posting here in case you have any ideas @tomaka

http/src/server/background.rs
```diff
@@ -93,13 +93,27 @@ impl BackgroundHttp {
                 };
 
                 runtime.block_on(async move {
+                    let builder = match hyper::Server::try_bind(&addr) {
```
tomaka: We reaaaaally want `bind` to fail if `hyper::Server::try_bind` fails. That's the whole point of returning an error in `bind`.

ascjones: :man_facepalming:

http/src/server/background.rs
```diff
@@ -93,13 +99,25 @@ impl BackgroundHttp {
...
+                        }
+                        Err(err) => {
+                            log::error!("Failed to bind to address {}: {}", addr, err);
+                            let _ = addr_tx.send(Err(err));
+                            return;
+                        }
+                    };
                 });
             })
             .unwrap();
```
tomaka: While you're at it:

```suggestion
            ?;
```

ws/src/client.rs
```diff
@@ -246,10 +245,10 @@ impl<'a> WsTransportClientBuilder<'a> {
             }
         }
 
+        let (sender, receiver) = client.into_builder().finish();
+
         // If the handshake succeeded, return.
```
tomaka:
```suggestion
        // If the handshake succeeded, return.
        let (sender, receiver) = client.into_builder().finish();
```

ws/src/client.rs
```diff
@@ -246,10 +245,10 @@ impl<'a> WsTransportClientBuilder<'a> {
             }
         }
 
+        let (sender, receiver) = client.into_builder().finish();
```
tomaka:
```suggestion
```

ws/src/client.rs
```diff
@@ -246,10 +245,10 @@ impl<'a> WsTransportClientBuilder<'a> {
...
+
```
tomaka:
```suggestion
```

PR #83, opened 2020-01-08
Remove old methods on RawClient

tomaka:
We first need to update the tests and the procedural macro. This PR is more of a reminder that we should do this in the near future.


niklasad1:
RawClient has been removed, so closing this

PR #85, opened 2020-01-14
Support websockets client TLS

ascjones:
Adds support for TLS over websockets, urls now require scheme prefix: `ws://` for plain TCP and `wss://` for TLS.

- [x] Support both `Tls` and `Plain` streams in connection: requires an enum implementing `AsyncRead + AsyncWrite` e.g. https://github.com/libp2p/rust-libp2p/blob/3f968cbf92d5939238b798d9ab1d5266e3edea20/core/src/either.rs#L62
- [x] Figure out API for choosing TLS or Plain - potentially parse a  `Url` and use the protocol? `ws://` vs `wss://`
~~Add a `tls` feature?~~

### DIFF
--- a/examples/polkadot.rs
+++ b/examples/polkadot.rs
@@ -43,8 +43,7 @@ struct Header {
 
 fn main() {
     async_std::task::block_on(async move {
-        //        let mut raw_client = jsonrpsee::ws_raw_client("ws://127.0.0.1:9944")
-        let mut raw_client = jsonrpsee::ws_raw_client("wss://kusama-rpc.polkadot.io:443")
+        let mut raw_client = jsonrpsee::ws_raw_client("wss://kusama-rpc.polkadot.io")
             .await
             .unwrap();
         let v = System::system_name(&mut raw_client).await.unwrap();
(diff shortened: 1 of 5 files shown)


ws/Cargo.toml
```diff
@@ -19,4 +19,7 @@ log = "0.4"
 pin-utils = "0.1.0-alpha.4"
 serde = { version = "1.0.40", features = ["derive"] }
 serde_json = "1.0.40"
-soketto = { git = "https://github.com/paritytech/soketto", rev = "a09bfb69af5650f9294e0e21bdda707e7d124b2c" }
+soketto = "0.3.1"
+# todo: add TLS feature for these?
+webpki = "0.21"
```
tomaka: This is alphabetically ordered

ascjones: Yeah this was in case I decide to add a feature for it. Not sure it's necessary though

examples/polkadot.rs
```diff
@@ -43,7 +43,10 @@ struct Header {
 
 fn main() {
     async_std::task::block_on(async move {
-        let mut raw_client = jsonrpsee::ws_raw_client("127.0.0.1:9944").await.unwrap();
+        //        let mut raw_client = jsonrpsee::ws_raw_client("ws://127.0.0.1:9944")
```
tomaka:
```suggestion
```

Let's remove this.

PR #87, opened 2020-01-17
Fix panic when dropping subscription

ascjones:
The subscription is already removed when the channel is closed and unsubscribe triggered: https://github.com/paritytech/jsonrpsee/blob/6f6866fc2da36124cf93f437685daf26758b8c7b/src/client.rs#L343

So panics on the `unwrap()` when unsubscribe confirmed.


tomaka:
Oh crap, that's probably some work-in-progress code that I should have removed.
But please remove the `remove` entirely. `Unsubscribed` should never happen.

Issue #67, opened 2019-11-18
Clarify the client error situation


tomaka:
What happens when an error happens on the client side?
Everything in `Client` and `RawClient` regarding the behaviour after errors should be properly documented.

tomaka:
Note: this issue was created at a time when the `TransportClient` was named `RawClient` and `RawClient` was named `Client`.

The questions in play here is: how do we indicate whether this is an error about a single request, or an error on the entire connection? Is the transport's state still valid after an error has been produced? Everything should be strongly typed and documented.

tomaka:
Unfortunately there's a big design flaw in that there's no way to associate a request with a response in the `TransportClient`. That means that for example an HTTP client that opens one connection per request cannot report that a specific request resulted in a 4xx or 5xx error.

niklasad1:
This is now fixed by having the following:
- `http`, the background thread is removed and we just return propogate the error to the user if that happens (it's "synchronous-like")
- `ws` when an error is received we terminate the connection and the client becomes useless, the error message is not great which could be improved but "hey you can read the logs" :roll_eyes: 

However, the documentation could be more explicit regarding this but closing anyway.

PR #88, opened 2020-01-27
Don't just swallow RawClient error, log it

ascjones:
Logs an error from `RawClient` instead of just swallowing it.

I converted this tool to `jsonrpsee` https://github.com/kianenigma/offline-phragmen/pull/8, and at first it was just waiting indefinitely on a request.

Turns out it was failing to deserialize because of the `serde_json` `arbitrary_precision` feature being enabled. This took me a while to find out, would have been much quicker if it just printed this error.

rel: #67

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ members = [
 [dependencies]
 async-std = "1.2.0"
 futures = "0.3.1"
+log = "0.4"
 parking_lot = "0.9"
 jsonrpsee-core = { path = "core" }
 jsonrpsee-http = { path = "http", optional = true }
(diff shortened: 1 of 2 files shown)


tomaka:
I'm kind of hesitant. This is unfortunately work-in-progress code that I haven't had time to finish, and it's always been intended that the error gets propagated out to the user. When that's implemented, we should remove the logging.

However in the meanwhile I'm fine with adding that.

ascjones:
I briefly considered propogating the error, but no time to do it properly. How best to propogate the error, given we don't have access to the request id in this case? 

Happy to do it in a spare moment if you can describe what you had in mind.

PR #94, opened 2020-02-21
Parse Positional Params

HCastano:
We have a PR (paritytech/parity-bridges-common#1) that depends on this branch, so we want to try and get it into master if possible. I see that #76 addresses something similar, but there's a mention that it should be done using `SharedServer` instead. This PR doesn't do that at the moment, but if that's something that needs to be changed let me know and I can work on it.


tomaka:
The change is entirely self-contained in the proc macro, which is a good way to do it.

PR #95, opened 2020-02-28
Deny intra_doc_link_resolution_failure on travis only

tomaka:
No specific reason to do that now, but it seems like a slightly better practice to me.


tomaka:
For the record, the CI doesn't actually check this at the moment.

PR #96, opened 2020-02-28
Makes the crate's API way more understandable

tomaka:
Moves a lot of things around to make the crate more understandable when looking at the `cargo doc` output.

The split between modules is no longer `core`/`ws`/`http`/`client`/`server`, but rather `transport` (containing `ws` and `http`)/`raw`/`common`/`client`/`server`.


tomaka:
If someone wants to review this, please let me know and I'll wait. If nobody manifest themselves I'll merge quickly.

PR #102, opened 2020-04-11
Support multiple runtimes

vorot93:
This PR adds a special shim that encapsulates all runtime facilities used by jsonrpsee. This is to abstract over any async runtime that has I/O, task spawning and TCP networking drivers. `async-std` and `tokio` implementations are provided.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,8 @@ license = "MIT"
 edition = "2018"
 
 [dependencies]
-async-std = "1.2.0"
+async-std = { version = "1.2.0", optional = true }
+async-trait = { version = "0.1" }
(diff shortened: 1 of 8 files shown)


vorot93:
Why was it not reviewed back when it was still fresh?

Also, there is no "multiple versions" of Tokio - 1.0 has been released.

dvdplm:
> Why was it not reviewed back when it was still fresh?

Manpower mostly, I apologise.

> Also, there is no "multiple versions" of Tokio - 1.0 has been released.

This makes no sense at all. There are large code bases out there still on 0.1, 0.2 and 0.3. Some even use all of the above.

PR #105, opened 2020-05-11
implement simple error propagation for subscriptions

dt665m:
Propagating errors is currently not implemented.  This PR tries to enable Client Subscriptions to signal downstream that a channel has been closed due to errors in the inner RawClient.

- On RawClient error, clear all active subscriptions, which drops all the mpsc senders.
- Implement future's Stream trait for Client::Subscription so that the inner mpsc receiver can be polled to completion. ```Ready::(None)``` when the mpsc sender is dropped.

The stream trait was implemented such that current libs depending on the Subscription.next() API, such as substrate-subxt, wont break.  Users can choose to poll the Subscription as a wrapped stream to get notified of the closed internal mpsc receiver or use subscription.next() future, which will loop with "pending" forever on error.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,8 +7,8 @@ license = "MIT"
 edition = "2018"
 
 [dependencies]
-async-std = "1.6.2"
-bs58 = "0.3.0"
+async-std = "1.2.0"
+bs58 = "0.3.1"
 fnv = "1.0"
 futures = "0.3.4"
 futures-timer = "3.0.2"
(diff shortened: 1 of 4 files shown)


tomaka:
Sorry for not responding to that. There are caveats deeper in the client regarding error propagation that need to be fixed as well, and I'll need to come back to this and look in the details.

dt665m:
@tomaka no worries, this isn't a complete implementation of error handling as the I didn't have much time to look deeper into the RawClient.  For my use case, I just needed a way to be notified if an error occurred the temporary solution was to close the channel.

niklasad1:
Fixed by https://github.com/paritytech/jsonrpsee/pull/180

PR #109, opened 2020-06-09
Fix hot loop in HTTP client background thread

### LINKED ISSUE #107: HTTP background thread consumes 100% CPU
The reason is that [`FuturesUnordered::poll_next()`](https://github.com/paritytech/jsonrpsee/blob/62da58d12321355aac76975bfedfbc3ab0fe08e3/src/transport/http/client.rs#L222) always returns `Poll::Ready(None)` if it is empty (at the start), or when all its futures have completed. So we instantly wake up [here](https://github.com/paritytech/jsonrpsee/blob/62da58d12321355aac76975bfedfbc3ab0fe08e3/src/transport/http/client.rs#L231) and proceed to the next iteration.

Bonus issue: it seems that completed futures are never removed from `pending_requests`.

svyatonik:
closes #107 
as suggested in https://github.com/paritytech/jsonrpsee/issues/107#issuecomment-641328259

### DIFF
--- a/src/transport/http/client.rs
+++ b/src/transport/http/client.rs
@@ -24,7 +24,7 @@
 // IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 // DEALINGS IN THE SOFTWARE.
 
-use crate::{common, raw::client::RawClient, transport::TransportClient};
+use crate::{common, transport::TransportClient};
 
 use futures::{channel::mpsc, channel::oneshot, prelude::*};
 use std::{fmt, io, pin::Pin, thread};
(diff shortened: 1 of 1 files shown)


src/transport/http/client.rs
```diff
@@ -73,7 +73,16 @@ impl HttpTransportClient {
...
+            .spawn(move || {
+                let client = hyper::Client::new();
+                background_thread(requests_rx, move |rq| {
+                    // cloning Hyper client = cloning references
+                    let client = client.clone();
+                    async move {
+                        let _ = rq.send_back.send(client.request(rq.request).await);
+                    }
+                })
+            })
```
tomaka, line 85: The polling fix looks good, but I don't understand this change? Why not leave `hyper::Client::new()` in the background thread?

svyatonik: To test that `background_thread` is actually doing something - see `background_thread_is_able_to_complete_requests`.

PR #112, opened 2020-06-29
Adds a ws_server function at the root, and remove some async qualifiers

tomaka:
There was a flaw in the WebSocket server: the transport doesn't implement `Sync`.
But the trick is: there's no fundamental reason why the transport should have to implement `Sync`. In the `Server`, it is completely isolated in a separate task.

It turns out that, since the `RawServerRequest` object borrows `RawServer`, any `await` while a `RawServerRequest` is alive requires `&'a RawServer` to be `Send`, which in turn requires `RawServer` to be `Sync`.

To fix that, I removed the `async` qualifier of the methods on `RawServerRequest`. They were here "just in case, for the future" and not for any actual reason.

I also added a `ws_server` function at the crate root, to prove that the WebSocket server should now compile.

### DIFF
--- a/examples/http.rs
+++ b/examples/http.rs
@@ -55,10 +55,10 @@ fn main() {
             match request {
                 Health::SystemName { respond, foo, bar } => {
                     let value = format!("{}, {}", foo, bar);
-                    respond.ok(value).await;
+                    respond.ok(value);
                 }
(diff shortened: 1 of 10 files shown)


tomaka:
Since I suppose nobody's going to review this, I'm just going to merge immediately.

Issue #117, opened 2020-07-02
Q: meaning of `self.responses.next()` returning `None`?


dvdplm:
When making a rpc call with `client.request("<rpc method name>", params).await` I'm seeing several errors like this:

```
[2020-07-01T15:03:49Z ERROR jsonrpsee::transport::http::client] JSONRPC http client background thread has shut down
[2020-07-01T15:03:49Z ERROR jsonrpsee::client] Client Error: Inner(Http(Custom { kind: Other, error: "background thread is down" }))
[2020-07-01T15:03:49Z ERROR jsonrpsee::transport::http::client] JSONRPC http client background thread has shut down
[2020-07-01T15:03:49Z ERROR jsonrpsee::client] Client Error: Inner(Http(Custom { kind: Other, error: "background thread is down" }))
[2020-07-01T15:03:49Z ERROR jsonrpsee::transport::http::client] JSONRPC http client background thread has shut down
[2020-07-01T15:03:49Z ERROR jsonrpsee::client] Client Error: Inner(Http(Custom { kind: Other, error: "background thread is down" }))
```

The error comes from [here](https://github.com/paritytech/jsonrpsee/blob/master/src/transport/http/client.rs#L142), where `self.responses.next()` is returning a `None` and we end up in the error case; the `RawClientEvent` received from the server ([here](https://github.com/paritytech/jsonrpsee/blob/master/src/raw/client.rs#L335)) looks like this: `Response { request_id: RawClientRequestId(0), result: Ok(Null) }`. 
When I log at `trace` level I (unsurprisingly?) see several calls to `next_responses()` being performed until the request receives a response. The RPC call is performed correctly so the errors seem spurious. Running my code in release mode makes the problem go away (no errors printed); adding a pause of 1ms after the call to `client.request()` also "fixes" it.

I am uncertain of what the meaning of a `None` is but I imagine it simply means that there is no response to read, which is not an error?

Here's the code I run:

```
async fn insert_signing_key(url: &str) -> Result<(), Box<dyn std::error::Error>> {
    let client = jsonrpsee::http_client(url);
    let params = Params::Array(vec![to_value("sgx!")?, to_value("//Alice")?, to_value("0x123")?]);
    client.request("author_insertKey", params).await.map_err(Into::into)
}
```

I guess I have several questions:

1. is this a bug in jsonrpsee or is the problem rather on the server (a substrate node).
[...]

PR #118, opened 2020-07-02
Distinguish null result from error

dvdplm:
Closes https://github.com/paritytech/jsonrpsee/issues/117

When a remote endpoint replies with e.g. `{"jsonrpc":"2.0","result":null,"id":1}` we get a `None` from `next_repsponse()`. This is not an error so the current implementation is a bit confusing.
`common::Response` (and Serde) doesn't parse `null` as a valid JSON document (see https://stackoverflow.com/questions/8526995/is-null-valid-json-4-bytes-nothing-else) so this PR replaces `null` with `[]` which does parse. Not elegant but perhaps better than the status quo?

### DIFF
--- a/src/common/response.rs
+++ b/src/common/response.rs
@@ -346,3 +346,23 @@ fn should_parse_empty_response_as_batch() {
     );
     assert_eq!(deserialized2.unwrap(), Response::Batch(vec![]));
 }
+
+#[test]
+fn single_response_with_nulls_deserialize() {
+    use serde_json;
+    use serde_json::Value;
+
+    let dbr = r#"{"jsonrpc":"2.0","result":null,"id":1}"#;
+
+    let deserialized: Response = serde_json::from_str(dbr).unwrap();
+    assert_eq!(
+        deserialized,
+        Response::Single(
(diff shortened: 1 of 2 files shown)


tomusdrw:
It doesn't feel right: `None` probably just means that the stream is over, I don't see how `null` get's converted to `None` in that stream.

@svyatonik can you take a look? You have much better understanding of the codebase than I do.

svyatonik:
I think Pierre has most of knowledge here :) But imo Tomek is right. My understanding is:`StreamExt::next()` returns `None` if `Stream::poll_next()` returns `Ready(None)`. Usually it means that the stream is finished. But here we deal with `FuturesUnordered`, which may also return `Ready(None)` when it is empty (and it is empty initially, or when all futures have completed). We have had the similar problem in #107.

Afaiu, the most straightforward solution would be to replace `FuturesUnordered` with channel here - i.e. whenever response is received, background thread sends it to the channel. And the `next_response` is just waiting for the first response it sees. I may be wrong, or Pierre may suggest a better solution - just sharing my thoughts :)

PR #123, opened 2020-08-19
Use unbounded channel instead of UnorderedFutures

### LINKED ISSUE #117: Q: meaning of `self.responses.next()` returning `None`?

svyatonik:
closes #117 

This fixes 100% CPU usage when using http client.

### DIFF
--- a/src/transport/http/client.rs
+++ b/src/transport/http/client.rs
@@ -26,7 +26,7 @@
 
 use crate::{common, transport::TransportClient};
 
-use futures::{channel::mpsc, channel::oneshot, prelude::*};
+use futures::{channel::mpsc, prelude::*};
 use std::{fmt, io, pin::Pin, thread};
 use thiserror::Error;
 
@@ -50,10 +50,10 @@ pub struct HttpTransportClient {
     /// Sender that sends requests to the background task.
     requests_tx: mpsc::Sender<FrontToBack>,
     url: String,
-    /// Receives responses in any order.
-    responses: stream::FuturesUnordered<
-        oneshot::Receiver<Result<hyper::Response<hyper::Body>, hyper::Error>>,
-    >,
+    /// Responses receiver.
+    responses_rx: mpsc::UnboundedReceiver<Result<hyper::Response<hyper::Body>, hyper::Error>>,
+    /// Responses transmitter
+    responses_tx: mpsc::UnboundedSender<Result<hyper::Response<hyper::Body>, hyper::Error>>,
 }
 
 /// Message transmitted from the foreground task to the background.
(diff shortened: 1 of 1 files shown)


niklasad1 (approved):
I have backported this to the v2 and it indeed fixes the issue.

niklasad1:
Fixed by https://github.com/paritytech/jsonrpsee/pull/180

PR #131, opened 2020-10-15
[ws server]: don't close connection when `deserialization` fails

niklasad1:
The WebSocket server closes the connection when the deserialization of a single request fails, which this PR fixes.

In a follow-up PR, we should try to document the error conditions of `fn per_connection_task` and perhaps differentiate success and error with a `Result<_>` instead of returning just Vec.

### DIFF
--- a/src/ws/transport.rs
+++ b/src/ws/transport.rs
@@ -349,9 +349,8 @@ impl WsTransportServerBuilder {
 
 /// Processes a single connection.
 //
-//
-// TODO: document this function it is quite hard to understand the outcome when it returns.
-// For example it returns when it receives an error and terminates all pending tasks in the connection.
+// TODO: document this function it is quite hard to understand the outcome when it returns `Vec<WsRequestId>`
+// both when an error or if the actual connection was terminated.
 async fn per_connection_task(
     socket: TcpStream,
     next_request_id: Arc<atomic::AtomicU64>,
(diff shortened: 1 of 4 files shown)


src/ws/transport.rs
```diff
@@ -425,14 +424,18 @@ async fn per_connection_task(
                             crate::common::Version::V2,
                         ))
                         .expect("valid JSON; qed");
+
+                        // deserialization failed and then client not alive then close the connection.
```
niklasad1, line 428:
```suggestion
                        // deserialization failed and the client is not alive then close the connection.
```

src/ws/transport.rs
```diff
@@ -425,14 +424,18 @@ async fn per_connection_task(
...
+                        // deserialization failed and the client is not alive then close the connection.
                         if let Err(e) = sender.send_text(&response).await {
                             log::warn!(
                                 "Failed to send: {:?} over WebSocket transport with error: {:?}",
                                 response,
                                 e
                             );
+                            return pending_requests;
+                        } else {
+                            continue;
```
niklasad1, line 437: Instead of returning here i.e, closing the connection, continue the loop, and spawn new futures to wait for.

maciejhirsz (approved):
LGTM, just a small nit-pick

src/ws/transport.rs
```diff
@@ -458,15 +463,27 @@ async fn per_connection_task(
...
+                    Some(Err(e)) => {
+                        log::error!(
+                            "send request={:?} to frontend failed because of {:?}, terminating the connection",
+                            request_id,
+                            e,
+                        );
+                        return pending_requests;
+                    }
+                    // The future wasn't ready.
+                    // TODO(niklasad1): verify if this is possible to happen "in practice".
```
niklasad1, line 479: I'm unsure if this is the right action if the `future` wasn't ready then we terminate the connection currently.

It might be possible that the receiving side ("frontend") is busy doing other things so it might be better just to ignore the request then notify the client and continue process the connection further.

PR #133, opened 2020-10-16
[dependencies]: upgrade tokio to v0.3

niklasad1:
The PR upgrades tokio to `v0.3` but `hyper v0.13` still requires `tokio v0.2` so it requires a compat layer which causes us to have both tokio `v0.2 and v0.3` in the dependency graph.

Nothing urgent, we can wait on hyper to update or just accept a bigger dependency tree for now.


maciejhirsz:
Is there a rough ETA for hyper running on new tokio?

dvdplm:
> Is there a rough ETA for hyper running on new tokio?

There's a ticket up. First step is to port `h2` which is likely a bit involved because of the changes to `AsyncRead` and `AsyncWrite` traits; the new tokio implements https://github.com/rust-lang/rfcs/pull/2930

So I suspect it might be a while.

maciejhirsz (approved):

niklasad1:
It seems like [hyper v0.14](https://github.com/hyperium/hyper/issues/2352) is quite close to a release, so let's wait for it.

niklasad1:
Let's migrate directly to `tokio 1.0` instead

PR #134, opened 2020-10-19
[ws client]: add tests

niklasad1:
- [x] WebSocket client tests
- [x] DummyWebSocketServer for testing (only support one request/subscription per client because `id` is hardcoded and it is monotonically increased in the client starting 0, thus the second request/subscription will fail with `UnknownRequestId`)


test-utils/src/types.rs
```diff
@@ -72,3 +78,168 @@ impl WebSocketTestClient {
...
+                    if let Err(e) = sender.send_text(&r).await {
+                        log::warn!("send response to subscription: {:?}", e);
+                    }
+                }
+            }
+            ws = next_ws => {
+                // Got a request on the connection but don't care about the contents.
+                // Just send out the pre-configured hardcoded responses.
+                if let Some(Ok(_)) = ws {
+                    match &mode {
```
niklasad1, line 228: this matching is dumb in loop because `mode` is known before the loop is entered but..... I didn't want to have more ugly if's or refactor to two different functions.

src/client/ws/tests.rs
```diff
@@ -1 +1,125 @@
...
+#[tokio::test]
+async fn response_with_wrong_id() {
+	env_logger::init();
+	let server =
+		WebSocketTestServer::with_hardcoded_response("127.0.0.1:0".parse().unwrap(), internal_error(Id::Num(99_u64)))
+			.await;
+	let uri = to_ws_uri_string(server.local_addr());
+	let client = WsClient::new(&uri, WsConfig::default()).await.unwrap();
+	let err: Result<jsonrpc::JsonValue, Error> = client.request("say_hello", jsonrpc::Params::None).await;
+	assert!(matches!(err, Err(Error::TransportError(e)) if e.to_string().contains("background task closed")));
```
niklasad1, line 124: This is error is quite bad, we just terminate the background thread and the [frontend](https://github.com/paritytech/jsonrpsee/blob/v2/src/client/ws/client.rs#L182-#L185) will receive https://docs.rs/futures/0.3.8/futures/channel/oneshot/struct.Canceled.html.

We could introduce another `oneshot channel` to send error in or find the pending request/subscription and send an error on that channel.

test-utils/src/helpers.rs
```diff
@@ -64,6 +64,24 @@ pub fn internal_error(id: Id) -> String {
...
+/// NOTE: works only for one subscription because the subscription ID is hardcoded.
+pub fn server_subscription_id_response(id: Id) -> String {
+	format!(
+		r#"{{"jsonrpc":"2.0","result":"D3wwzU6vvoUUYehv4qoFzq42DZnLoAETeFzeyk8swH4o","id":{}}}"#,
+		serde_json::to_string(&id).unwrap()
+	)
+}
+
+/// Server response to a hardcoded pending subscription
+pub fn server_subscription_response(result: Value) -> String {
```
dvdplm, line 78: Can these functions be `const` these days?

dvdplm (approved):

src/client/ws/tests.rs
```diff
@@ -1 +1,124 @@
...
+		WebSocketTestServer::with_hardcoded_response("127.0.0.1:0".parse().unwrap(), internal_error(Id::Num(0_u64)))
+			.await;
+	let uri = to_ws_uri_string(server.local_addr());
+	let client = WsClient::new(&uri, WsConfig::default()).await.unwrap();
+	let response: Result<jsonrpc::JsonValue, Error> = client.request("say_hello", jsonrpc::Params::None).await;
+	assert_error_response(response, jsonrpc::ErrorCode::InternalError, INTERNAL_ERROR.into());
+}
+
+#[tokio::test]
+async fn subscription_works() {
```
niklasad1, line 98: NOTE, this doesn't test that subscription is unsubscribed when the subscription is dropped.

The `MockedServer` needs then parse the responses and maintain the requests so I prefer to have the `MockedServer` as simple as possible for now.

test-utils/src/helpers.rs
niklasad1: https://docs.serde.rs/serde_json/fn.to_string.html is not const and I think `format!` doesn't work in const context.

PR #136, opened 2020-10-22
[ws server]: parse subscription ID for unsubscription instead of hardcoding `JsonValue::Null`

### LINKED ISSUE #135: [server]: `unsubscribe_call` doesn't close subscriptions as intended
When an `unsubscription call` is received by the [Server](https://github.com/paritytech/jsonrpsee/blob/master/src/server.rs#L458) then `RawServerSubscriptionId::from_wire_message` tries to parse it from `hardcoded JsonValue::Null` where the `subscription ID` should be parsed instead. Thus, RawServerSubscriptionId::from_wire_message always fails to parse the subscription ID

This causes dropping a `registered subscription` not work with one exception when the entire client is dropped then it works via [SubscriptionsClosed](https://github.com/paritytech/jsonrpsee/blob/master/src/server.rs#L482-#L493)

niklasad1:
See https://github.com/paritytech/jsonrpsee/issues/135 for a proper description.


src/ws/server.rs
```diff
@@ -432,22 +434,25 @@ async fn background_task(mut server: RawServer, mut from_front: mpsc::UnboundedR
...
+                                if let Some(client_pos) = clients.iter().position(|c| *c == sub_id) {
+                                    clients.remove(client_pos);
+                                }
+
+                                if let Some(s_u_id) = active_subscriptions.remove(&sub_id) {
+                                    debug_assert_eq!(s_u_id, *sub_unique_id);
+                                }
+                            }
+                        }
+                        Err(_) => log::error!("Unsubscription of method=\"{}\" failed; The subscription ID must passed as the first argument of Array or \"subscription\" name of Object, got={:?}", request.method(), request.params()),
```
maciejhirsz, line 454: Indentiation is off (tabs vs sapces), otherwise looks fine.

niklasad1: nice catch, I must have messed it up during merging but still weird that rustfmt didn't fix it.

maciejhirsz (approved):

PR #137, opened 2020-10-22
[server raw params]: fix debug implementation

niklasad1:
Use debug implementation of `common::Params` instead of doing something
similar that doesn't work properly

I'm not sure whether the actual iterator implementation is useful or not, it was just used for the debug implementation AFAIK

The actual debug implementation after this PR is not super nice it looks like:
```
'Params { params: Array([Null, Bool(true), Number(-1), Number(4), Number(2.3), String("hello"), Array([Number(0)]), Object({"key": String("value")}), Array([])]) }'
```

I was thinking of another implementation using a stack to create nicer looking params with  as I think "this" was indented from the start but decided to keep it as simple as possible, if we really hate the debug format we can revisit this later.


src/http/raw/params.rs
```diff
@@ -94,7 +94,7 @@ impl<'a> IntoIterator for Params<'a> {
 
 impl<'a> fmt::Debug for Params<'a> {
```
dvdplm, line 95: Can't this be derived?

src/http/raw/params.rs
```diff
@@ -146,14 +146,22 @@ enum IterInner<'a> {
     Array(std::slice::Iter<'a, serde_json::Value>),
 }
 
+#[derive(Debug)]
+pub enum Entry<'a> {
```
dvdplm, line 150: Why can't this be in `common`?

niklasad1: It is only used for the iterator out to distinguish between `No Params`, `Array` or `Object/Map` which is only is used by `raw::Params`. With that said I don't think we the iterator implementations anymore because it was only used for the debug implemenation but it might still be useful.

I think the intention was to have `common::Params` as serialize/deserialize and `raw::Params` as wrapper with some convince methods to simplify its use. I'm in favor of moving `raw::Params` to `common` but it couldn't come up with a good name for it, so left it to fix in another PR.

dvdplm (approved):

PR #140, opened 2020-10-25
fix(ws server): sub/unsubscribe to same method should generate an error

niklasad1:
Subscribe and unsubscribe to the same method should generate an error, which this commit fixes.

This bug was introduced by myself in https://github.com/paritytech/jsonrpsee/commit/fc87889de2615dbb3d0cf2d91a306f016d48df2d

### DIFF
--- a/src/ws/server.rs
+++ b/src/ws/server.rs
@@ -236,13 +236,17 @@ impl Server {
 		);
 		{
 			let mut registered_methods = self.registered_methods.lock();
-			if registered_methods.contains(&subscribe_method_name)
-				|| registered_methods.contains(&unsubscribe_method_name)
-			{
+
+			// NOTE: optimized for that both methods are not registered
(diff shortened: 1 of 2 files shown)


src/ws/server.rs
```diff
@@ -236,13 +236,17 @@ impl Server {
 		);
 		{
 			let mut registered_methods = self.registered_methods.lock();
-			if registered_methods.contains(&subscribe_method_name)
-				|| registered_methods.contains(&unsubscribe_method_name)
-			{
+
+			// NOTE: optimized for that both methods are not registered
+			// For example if the strings are equal this will be slower than just comparing the
+			// strings.
```
dvdplm, line 242:
```suggestion
			// NOTE: We optimise for the case where neither method is registered.
			// This means that if the strings are equal this will be slower than just comparing the
			// strings.
```

dvdplm (approved):

PR #141, opened 2020-10-26
[ws client]: resultify API and fix register_subscribe

niklasad1:
* The commit changes the API to return `Err` when it's possible and to not ignore underlying errors.
* Fix that `fn subscribe` doesn't accept the subscription and unsubscription to be the same which causes
errors in the server.

### DIFF
--- a/src/client/ws/client.rs
+++ b/src/client/ws/client.rs
@@ -66,9 +66,9 @@ pub enum Error {
 	/// RawServer responded to our request with an error.
 	#[error("Server responded to our request with an error: {0:?}")]
 	Request(#[source] common::Error),
-	/// Subscription and unsubscription methods are the same.
-	#[error("Subscription error, subscribe_method={0}, unsubscribe_method={1}")]
-	SubscriptionMethod(String, String),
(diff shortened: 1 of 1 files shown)


src/client/ws/client.rs
```diff
@@ -58,14 +58,20 @@ pub struct Subscription<Notif> {
...
 	/// Networking error or error on the low-level protocol layer (e.g. missing field,
 	/// invalid ID, etc.).
 	#[error("Networking or low-level protocol error: {0}")]
 	TransportError(#[source] Box<dyn error::Error + Send + Sync>),
 	/// RawServer responded to our request with an error.
 	#[error("Server responded to our request with an error: {0:?}")]
 	Request(#[source] common::Error),
+	/// Subscription and unsubscription methods are the same.
+	#[error("Subscription error, subscribe_method={0}, unsubscribe_method={1}")]
+	SubscriptionMethod(String, String),
```
niklasad1, line 71: fix messy description

niklasad1:
Superseeded by #144

PR #144, opened 2020-10-29
refactor: resultify API + some crate reorg

niklasad1:
This is a little too big for my taste sorry about that but it does the following:

1. Return errors when it's possible and implement some additional error types (didn't touch `raw` implementations which need to fixed or merged with client)
2. The repo structure didn't provide a good place for shared error types for the client and server so I renamed `common` -> `types` and tried to distinguish concrete JSON-RPC v2 specification types from the others with a separate module for that. I think common is too vague.
3. Moved some shared wrapper types from raw/http &  raw/ws -> jsonrpc_v2/wrapped.
4. Removed some needless subscription code from `http` and `http/core`, should be quite simple to merge `raw` and `server/client` for http because it so stripped down now.

The naming could probably be better but it is hard :)


src/ws/tests.rs
```diff
@@ -1,15 +1,18 @@
...
+use std::net::SocketAddr;
+
 use futures::channel::oneshot::{self, Sender};
 use futures::future::FutureExt;
 use futures::{pin_mut, select};
 use jsonrpsee_test_utils::helpers::*;
 use jsonrpsee_test_utils::types::{Id, WebSocketTestClient};
-use serde_json::Value as JsonValue;
-use std::net::SocketAddr;
+// use serde_json::Value as JsonValue;
```
niklasad1, line 15:
```suggestion
```

src/ws/tests.rs
```diff
@@ -232,7 +235,7 @@ async fn register_methods_works() {
 #[tokio::test]
 async fn register_same_subscribe_unsubscribe_is_err() {
 	let server = WsServer::new("127.0.0.1:0").await.unwrap();
-	matches!(server.register_subscription("subscribe_hello".to_owned(), "subscribe_hello".to_owned()), Err(()));
+	matches!(server.register_subscription("subscribe_hello".to_owned(), "subscribe_hello".to_owned()), Err(Error::AlreadyRegistered(_)));
```
maciejhirsz, line 237: This needs to be `assert!(matches!(...))`

dvdplm (commented):
A bit too much to review in detail but I agree with the overall sentiment!

niklasad1:
Ok, if I merge this?

dvdplm:
@niklasad1 go for it

PR #150, opened 2020-11-01
Improve HTTP client background thread

### LINKED ISSUE #107: HTTP background thread consumes 100% CPU

AtkinsChang:
Improve the change of pr #109 which fix #107

There is a function `StreamExt::for_each_concurrent` implemented almost same algorithm but in lower level
It is more elegant / efficient to use it imo

The max concurrent level is set to `4` according to the channel buffer size.

cc @tomaka

### DIFF
--- a/src/client/http/transport.rs
+++ b/src/client/http/transport.rs
@@ -21,6 +21,8 @@ use crate::common;
 use futures::{channel::mpsc, prelude::*};
 use thiserror::Error;
 
+const REQUEST_PARALLELISM: usize = 4;
+
 /// Implementation of a raw client for HTTP requests.
 pub struct HttpTransportClient {
 	/// Sender that sends requests to the background task.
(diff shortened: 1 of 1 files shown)


niklasad1 (approved):
Thank you, looks good to me.

We should probably re-iterate why `4` is chosen and if that's good `default` it should be user-configurable too but nothing that needs to addressed in this PR.

dvdplm (approved):

PR #155, opened 2020-11-06
ci(benches): syncronous and concurrent roundtrips

niklasad1:
Improve benchmarks to take concurrent requests into account.

### DIFF
--- a/benches/benches.rs
+++ b/benches/benches.rs
@@ -11,6 +11,11 @@ use std::sync::Arc;
 criterion_group!(benches, http_requests, websocket_requests);
 criterion_main!(benches);
 
+fn concurrent_tasks() -> Vec<usize> {
+	let cores = num_cpus::get();
+	vec![cores / 4, cores / 2, cores, cores * 2, cores * 4]
+}
+
 async fn http_server(tx: Sender<SocketAddr>) {
 	let server = HttpServer::new("127.0.0.1:0").await.unwrap();
 	let mut say_hello = server.register_method("say_hello".to_string()).unwrap();
@@ -63,7 +68,7 @@ pub fn http_requests(c: &mut criterion::Criterion) {
 				}
 			})
 		},
-		vec![2, 4, 8, 16, 32, 64, 128],
+		concurrent_tasks(),
 	);
 }
 
@@ -99,7 +104,7 @@ pub fn websocket_requests(c: &mut criterion::Criterion) {
 				}
 			})
 		},
-		// TODO(niklasad1): investigate why it only works to 8 concurrent requests.
-		vec![2, 4, 8],
+		// TODO(niklasad1): this doesn't work on my machine.
+		concurrent_tasks(),
 	);
 }
--- a/benches/benches.rs
+++ b/benches/benches.rs
@@ -6,8 +6,9 @@ use jsonrpsee::http::HttpServer;
 use jsonrpsee::types::jsonrpc::{JsonValue, Params};
 use jsonrpsee::ws::WsServer;
 use std::net::SocketAddr;
+use std::sync::Arc;
 
-criterion_group!(benches, http, ws);
+criterion_group!(benches, http_requests, websocket_requests);
 criterion_main!(benches);
 
 async fn http_server(tx: Sender<SocketAddr>) {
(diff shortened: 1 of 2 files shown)


benches/benches.rs
```diff
@@ -30,36 +31,74 @@ async fn ws_server(tx: Sender<SocketAddr>) {
...
+						let _: Result<JsonValue, _> = black_box(client_rc.request("say_hello", Params::None)).await;
+					});
+					tasks.push(task);
+				}
+				for task in tasks {
+					rt.block_on(task).unwrap();
+				}
+			})
+		},
+		vec![2, 4, 8, 16, 32, 64, 128],
```
niklasad1, line 102: TODO: this works only to 16

benches/benches.rs
```diff
@@ -30,36 +31,75 @@ async fn ws_server(tx: Sender<SocketAddr>) {
...
+		move |b: &mut Bencher, size: &usize| {
+			b.iter(|| {
+				let mut tasks = Vec::new();
+				for _ in 0..*size {
+					let client_rc = client.clone();
+					let task = rt.spawn(async move {
+						let _: Result<JsonValue, _> = black_box(client_rc.request("say_hello", Params::None)).await;
+					});
+					tasks.push(task);
+				}
```
dvdplm, line 60: Can you move this part outside the `b.iter()` block to measure only the task execution?

benches/benches.rs
```diff
@@ -30,36 +31,75 @@ async fn ws_server(tx: Sender<SocketAddr>) {
...
+				for task in tasks {
+					rt.block_on(task).unwrap();
+				}
+			})
+		},
+		vec![2, 4, 8, 16, 32, 64, 128],
```
dvdplm, line 66: I don't understand what the point of measuring this over different number of loops is? Is it to see what performance is when number of tasks is higher than the number of threads in the thread pool? Or…?

benches/benches.rs
```diff
@@ -30,36 +31,75 @@ async fn ws_server(tx: Sender<SocketAddr>) {
...
+						let _: Result<JsonValue, _> = black_box(client_rc.request("say_hello", Params::None)).await;
+					});
+					tasks.push(task);
+				}
+				for task in tasks {
+					rt.block_on(task).unwrap();
+				}
+			})
+		},
+		// TODO(niklasad1): investigate why it only works to 8 concurrent requests.
```
dvdplm, line 102: Interesting, that seems a bit worrying.

dvdplm (commented):
It's great to measure perf in a concurrent context. Just not sure I understand why it's set up like this, can you explain?

benches/benches.rs
niklasad1: It is a way to measure how an increasing number of concurrent tasks impact performance,  such as try 2 concurrent request, .... to 128 if that makes sense. That's at least the intention

Thus, tasks is requests in this context.

benches/benches.rs
niklasad1: No, we spawn `n concurrent tasks` that we need to wait for to complete.

The actual joining of `tasks` are not great because a task spawned "later" could potentially be finished before....

benches/benches.rs
niklasad1: Yeah, I think that's just that the WebSocket connection might end up in some weird state

benches/benches.rs
dvdplm: This compiles, but panics as soon as it runs:

```rust
			let mut tasks = Vec::new();
			for _ in 0..*size {
				let client_rc = client.clone();
				let task = rt.spawn(async move {
					let _: Result<JsonValue, _> = black_box(client_rc.request("say_hello", Params::None)).await;
				});
				tasks.push(task);
			}
			b.iter(|| {
				for mut task in &mut tasks {
					rt.block_on(&mut task).unwrap();
				}
			})
```

…so I guess you're right! :)

benches/benches.rs
dvdplm: Ok, I see what you're saying. Maybe it should check core count and do like `let cores = …; let batches = [cores/4, cores/2, cores, cores * 2, cores * 4]`?
It is also measuring *roundtrips* so I guess it's ok to include the client part too. We should still have benchmarks for the actual execution at some point, but maybe that should be decoupled from the async machinery.

benches/benches.rs
niklasad1: You mean the `Vec` correct? I think so but then each iteration has to clear/replace it with an empty `Vec`.

niklasad1: That should work AFAIU.

PR #156, opened 2020-11-06
[http client]: refactor with "syncronous-like" design

niklasad1:
This PR simplifies the `HTTP client` and return the errors on each request instead of keeping the request until a valid response is received. A start to address https://github.com/paritytech/jsonrpsee/issues/67, some tests are added to prove that the errors are handled correctly.

I was afraid that this would impose a performance regression that is why I added benchmarks for concurrent requests such that bloating the executor with quite big futures as `async fn send_and_wait_for_response` which is more or less synchronous instead of having a background event loop.

Another benefit is that this PR simplifies the code significantly when removing the background event loop.

This PR shows to  20-30% faster than `v2`, then it should be a no brainer if I didn't screw anything up in the benchmarks.

TODOs:
- [x] Fix failing tests in `http/raw`
- [x] Update documentation

Follow up:
- Implement batch requests
- Configurable HTTP server with max request body size
- Configurable WebSocket server w.r.t number of concurrent requests and so on.


examples/http.rs
```diff
@@ -33,7 +33,7 @@ use jsonrpsee::types::jsonrpc::{JsonValue, Params};
 const SOCK_ADDR: &str = "127.0.0.1:9933";
 const SERVER_URI: &str = "http://localhost:9933";
 
-#[async_std::main]
+#[tokio::main]
```
dvdplm, line 36: Just curious: why switch to Tokio here?

niklasad1: `hyper` only works with `tokio`, Pierre spawned a background thread around that just to support both any runtime.

If we go ahead with this change it should be clearly documented or spawn a background thread for it again.

test-utils/src/helpers.rs
```diff
@@ -1,7 +1,16 @@
...
+use hyper::service::{make_service_fn, service_fn};
+use hyper::{Request, Response, Server};
 use serde_json::Value;
+use std::convert::Infallible;
 use std::net::SocketAddr;
 
+pub const PARSE_ERROR: &str = "Parse error";
+pub const INTERNAL_ERROR: &str = "Internal error";
+pub const INVALID_PARAMS: &str = "Invalid params";
+pub const INVALID_REQUEST: &str = "Invalid request";
```
niklasad1, line 11: Note, the specification states: `Invalid Request`  but we use `Invalid request`.

https://github.com/paritytech/jsonrpc/ does the same let's keep like that for now.

src/client/http/transport.rs
```diff
@@ -1,237 +1,157 @@
...
-					async move {
-						let _ = rq.send_back.unbounded_send(client.request(rq.request).await);
-					}
-				})
-			})
-			.unwrap();
-
-		let (responses_tx, responses_rx) = futures::channel::mpsc::unbounded();
-		HttpTransportClient { requests_tx, url: target.to_owned(), responses_tx, responses_rx }
+	pub fn new(target: &str, max_request_body_size: usize) -> Result<Self, Error> {
```
niklasad1, line 26: impl &str

src/client/http/transport.rs
```diff
@@ -1,237 +1,157 @@
...
-			};
-
-			if requests_tx.send(message).await.is_err() {
-				log::error!("JSONRPC http client background thread has shut down");
-				return Err(RequestError::Http(Box::new(io::Error::new(
-					io::ErrorKind::Other,
-					"background thread is down".to_string(),
-				))));
-			}
+		let body = jsonrpc::to_vec(&request).map_err(|e| Error::Serialization(e))?;
```
niklasad1, line 37: check if need to allocate?

src/client/http/transport.rs
```diff
@@ -1,237 +1,157 @@
...
+		request: jsonrpc::Request,
+	) -> Result<jsonrpc::Response, Error> {
+		let response = self.send_request(request).await?;
+		let mut body_fut: hyper::Body = response.into_body();
+
+		let mut body = Vec::new();
+
+		while let Some(chunk) = body_fut.next().await {
+			let chunk = chunk.map_err(|e| Error::Http(Box::new(e)))?;
+			if chunk.len() + body.len() > self.max_request_body_size {
```
niklasad1, line 78: >=

src/client/http/transport.rs
```diff
@@ -1,237 +1,157 @@
...
+
+	/// Send request and wait for response.
+	pub async fn send_request_and_wait_for_response(
+		&self,
+		request: jsonrpc::Request,
+	) -> Result<jsonrpc::Response, Error> {
+		let response = self.send_request(request).await?;
+		let mut body_fut: hyper::Body = response.into_body();
+
+		let mut body = Vec::new();
```
niklasad1, line 74: with_capacity

src/client/http/transport.rs
```diff
@@ -1,237 +1,157 @@
...
+		Ok(as_json)
 	}
 }
 
 /// Error that can happen during a request.
 #[derive(Debug, Error)]
-pub enum RequestError {
+pub enum Error {
+	/// Invalid URL.
+	#[error("Invalid Uri: {0}")]
```
niklasad1, line 97:
```suggestion
	#[error("Invalid Url: {0}")]
```

src/client/http/transport.rs
```diff
@@ -1,237 +1,157 @@
...
-		// drop requests sender, asking background thread to exit gently
-		drop(requests_tx);
-		background_thread.join().unwrap();
+	fn invalid_http_url_rejected() {
+		let err = HttpTransportClient::new("ws://localhost:9933", 1337).unwrap_err();
+		assert!(matches!(err, Error::Url(_)));
+	}
+
+	#[tokio::test]
+	async fn request_limit_works() {
```
niklasad1, line 141: change to exact limit ^

src/types/client.rs
```diff
@@ -1,4 +1,20 @@
-use crate::types::jsonrpc;
+use crate::types::jsonrpc::{self, JsonValue};
+use std::fmt;
+
+/// Mismatch of the expected value.
```
niklasad1, line 4: convenience type for displaying errors

src/client/http/transport.rs
niklasad1: I was wrong it should be `>`

src/client/http/transport.rs
```diff
@@ -18,43 +20,42 @@ pub struct HttpTransportClient {
 	/// HTTP client,
 	client: hyper::Client<hyper::client::HttpConnector>,
 	/// Configurable max request body size
-	max_request_body_size: usize,
+	max_request_body_size: u32,
```
niklasad1, line 23: 2 GB max size, I changed because I don't like usize because it's implicit i.e. platform dependent.

Could be `u64` as well but I think 2 GB should be sufficient some browsers limit to 2GB (e.g, Firefox)

src/client/http/transport.rs
```diff
@@ -1,237 +1,189 @@
...
+		let body_size = read_content_length(response.headers()).unwrap_or(0);
+		let mut body_fut: hyper::Body = response.into_body();
+
+		if body_size > self.max_request_body_size {
+			return Err(Error::RequestTooLarge);
+		}
 
-			// Note that we don't check the Content-Type of the request. This is deemed
-			// unnecessary, as a parsing error while happen anyway.
+		let mut body = Vec::with_capacity(body_size as usize);
```
niklasad1, line 80: The idea is that this will be faster for big payloads that needs be re-allocated but
I haven't benchmarked this change for bigger payloads because the [HttpServer](https://github.com/paritytech/jsonrpsee/blob/master/src/transport/http/server/background.rs#L243) hardcodes the max request size to 16384. 

Thus, it's just a guess.

dvdplm: I like `with_capacity` even if it turns out it doesn't change performance at all (I suspect it doesn't and that the compiler does the right thing anyway) because it communicates to the reader that we intended to use this for the body before doing it.

benches/benches.rs
```diff
@@ -30,36 +36,75 @@ async fn ws_server(tx: Sender<SocketAddr>) {
...
+						let _: Result<JsonValue, _> = black_box(client_rc.request("say_hello", Params::None)).await;
+					});
+					tasks.push(task);
+				}
+				for task in tasks {
+					rt.block_on(task).unwrap();
+				}
+			})
+		},
+		// TODO(niklasad1): this doesn't work on my machine.
```
dvdplm, line 107: On my machine it seems to block after the first run, looks like this:

```
synchronous WebSocket round trip
                        time:   [94.082 us 94.526 us 94.977 us]
  1 (1.00%) low mild
  3 (3.00%) high mild
  1 (1.00%) high severe

concurrent WebSocket round trip/6
                        time:   [226.41 us 226.88 us 227.40 us]
  1 (1.00%) low mild
Benchmarking concurrent WebSocket round trip/12: Warming up for 3.0000 s
```

It's stuck on the last one. Is that what you see too?

niklasad1: Yes, I can take look what's going on there today.

niklasad1: It looks like a deadlock :skull_and_crossbones:

niklasad1: Found the problem https://github.com/paritytech/jsonrpsee/blob/v2/src/ws/transport.rs#L352, it essentially just allows 8 concurrent tasks to be handled per connection, and if the buffer gets full -> deadlock

niklasad1, line 107:
```suggestion
		// TODO(niklasad1): This deadlocks when more than 8 tasks are spawned.
```

PR #162, opened 2020-11-16
feat(http server): configurable request body limit

niklasad1:
This PR introduces the functionality for the user to configure `max request body size` for the HttpServer.

The actual fix is rather simple but the PR became so huge because I couldn't find a good place for sharing helper methods for the `http client` and `http server`. Because they both read `HTTP responses` and process their bodies.

Currently, there's `HttpConfig` which essentially is u32 but I'm not sure if we want more stuff in there in the future and makes the API a little bit cleaner.


src/http/transport/background.rs
```diff
@@ -136,18 +144,17 @@ async fn process_request(
...
-
 	// Proceed
 	match *request.method() {
 		// Validate the ContentType header
 		// to prevent Cross-Origin XHRs with text/plain
 		hyper::Method::POST if is_json(request.headers().get("content-type")) => {
-			let request = match body_to_request(request.into_body()).await {
-				Ok(b) => b,
+			let (parts, body) = request.into_parts();
+			let request = match http::read_http_body(&parts.headers, body, config).await {
```
niklasad1, line 153: this needs to refactored and handle errors correctly.

src/utils/http/hyper_helpers.rs
```diff
@@ -60,31 +60,32 @@ pub async fn read_response_to_body(
...
+/// NOTE: There's no specific hard limit on `content_length` in specification, so bigger values than `u64::MAX` won't be parsed successfully.
+fn read_header_content_length(headers: &hyper::header::HeaderMap) -> Option<u64> {
+	let length = read_header(headers, "content-length")?;
 	// HTTP Content-Length indicates number of bytes in decimal.
-	let length = content_length.to_str().ok()?;
-	u32::from_str_radix(length, 10).ok()
+	u64::from_str_radix(length, 10).ok()
 }
 
 /// Extracts string value of a single header in request.
```
dvdplm, line 72:
```suggestion
/// Given a header name, returns the first string value.
```

src/utils/http/hyper_helpers.rs
```diff
@@ -60,31 +60,32 @@ pub async fn read_response_to_body(
...
-pub fn read_header<'a>(req: &'a hyper::Request<hyper::Body>, header_name: &str) -> Option<&'a str> {
-	req.headers().get(header_name).and_then(|v| v.to_str().ok())
+///
+/// Returns `Some(val)` if the header contains exactly one value.
```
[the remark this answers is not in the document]
niklasad1: Not sure, let me check.

src/http/tests.rs
```diff
@@ -39,9 +40,9 @@ async fn server(server_started_tx: Sender<SocketAddr>) {
 
 		pin_mut!(hello_fut, add_fut, notif_fut);
 		select! {
-			say_hello = hello_fut => (),
-			add = add_fut => (),
-			notif = notif_fut => (),
+			_say_hello = hello_fut => (),
```
niklasad1, line 43: fixed compiler warn.

src/types/error.rs
```diff
@@ -16,29 +15,42 @@ impl<T: fmt::Display> fmt::Display for Mismatch<T> {
 	}
 }
 
-/// Error produced by the client.
+/// Error type.
```
niklasad1, line 18: Merged client and server errors.

src/types/error.rs
```diff
@@ -16,29 +15,42 @@ impl<T: fmt::Display> fmt::Display for Mismatch<T> {
...
+	#[error("Method: {0} already registered")]
+	MethodAlreadyRegistered(String),
 	#[error("Custom error: {0}")]
 	/// Custom error.
 	Custom(String),
 }
+
+/// Generic transport error.
+#[derive(Debug, thiserror::Error)]
+pub enum GenericTransportError<T: std::error::Error + Send + Sync> {
```
niklasad1, line 49: Introduced to handle `transport error` from different transports.

src/utils/http/hyper_helpers.rs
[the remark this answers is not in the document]
niklasad1: I found this:

> Multiple message-header fields with the same field-name MAY be present in a message if and only if the entire field-value for that header field is defined as a comma-separated list [i.e., #(values)]. It MUST be possible to combine the multiple header fields into one "field-name: field-value" pair, without changing the semantics of the message, by appending each subsequent field-value to the first, each separated by a comma. The order in which header fields with the same field-name are received is therefore significant to the interpretation of the combined field value, and thus a proxy MUST NOT change the order of these field values when a message is forwarded. 

Also, there are single-value headers such as `Content-Length`,

>  If a message is received that has multiple Content-Length header
   fields with field-values consisting of the same decimal value, or a
   single Content-Length header field with a field value containing a
   list of identical decimal values (e.g., "Content-Length: 42, 42"),
   indicating that duplicate Content-Length header fields have been
   generated or combined by an upstream message processor, then the
   recipient MUST either reject the message as invalid or replace the
   duplicated field-values with a single valid Content-Length field
   containing that decimal value prior to determining the message body
   length or forwarding the message.
 
Thus, I think there is header specific related and I tried to be defensive and reject multiple header values when not expected. I'm willing to change back to use https://docs.rs/hyper/0.13.9/hyper/struct.HeaderMap.html#method.get again but carefully document it.

src/http/server.rs
```diff
@@ -49,7 +50,7 @@ pub struct Server {
 	local_addr: SocketAddr,
 	/// Channel to send requests to the background task.
 	to_back: mpsc::UnboundedSender<FrontToBack>,
-	/// List of methods (for RPC queries, subscriptions, and unsubscriptions) that have been
+	/// List of methods (for RPC queries and and notifications) that have been
```
dvdplm, line 53:
```suggestion
	/// List of methods (for RPC queries and notifications) that have been
```

src/utils/http/hyper_helpers.rs
```diff
@@ -0,0 +1,131 @@
...
+///
+/// Returns `Ok(bytes)` if the body was in valid size range.
+/// Returns `Err` if the body was too large or the body couldn't be read.
+pub async fn read_response_to_body(
+	headers: &hyper::HeaderMap,
+	mut body: hyper::Body,
+	config: HttpConfig,
+) -> Result<Vec<u8>, GenericTransportError<hyper::Error>> {
+	// NOTE(niklasad1): Bigger values than `u32::MAX` will be regarded as zero here and that's very unlikely to occur.
+	// Thus, in those causes we rely on the while loop to allocate instead of `pre-allocate`.
```
dvdplm, line 43:
```suggestion
	// NOTE(niklasad1): Values bigger than `u32::MAX` will be turned into zero here. This is unlikely to occur in practise
	// and for that case we fallback to allocating in the while-loop below instead of pre-allocating.
```

src/http/server.rs
```diff
@@ -49,7 +50,7 @@ pub struct Server {
...
 	/// registered. Serves no purpose except to check for duplicates.
 	registered_methods: Arc<Mutex<HashSet<String>>>,
```
dvdplm, line 55: Random thought: maybe this `HashSet` could use `Fnv` hashing? Or do we ever expose this on the wire?

dvdplm (approved):

src/http/server.rs
niklasad1: Should be fine, it's internal in the server

src/utils/http/hyper_helpers.rs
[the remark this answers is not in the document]
dvdplm: TIL! thank you. :)

Issue #103, opened 2020-04-13
Batch Requests


insipx:
I'm wondering about the plans to implement batch requests for the client. I saw a 'Batch' variant in the request enum, however, it seems like that's not being used in `RawClient` or  the `FrontToBack`/`request` method in client.rs is for singular requests. This is my first look at the code, but would it be correct to just add a `batch_request` method to add that logic/modify the enum variants and allow front-end batch-requests?

niklasad1:
Yeah, you are correct because the serialize/deserialize is already in place for it, so it should be quite easy to implement at least in the client.

It just a matter of creating/designing a proper API for it, so I would say that it's planned to be supported.

PR #163, opened 2020-11-17
[WIP]: feat(http client): batch requests

niklasad1:
Initial draft to start a discussion regarding supporting [batch requests](https://github.com/paritytech/jsonrpsee/issues/103)

The API looks very much like `requests` i.e, the user just enters `method_name` and `params` and the library takes care of the rest. Thus, the user has no access to the `request_id` at all, and therefore when a single request fails the entire batch is thrown away because the user can't know which request failed or succeeded reliably with knowing the request ID in advance.

Thoughts?

//cc @insipx


src/client/http/client.rs
```diff
@@ -92,6 +93,57 @@ impl HttpClient {
...
+				Err(Error::Custom("Server replied with single response to a batch request".to_string()))
+			}
+			jsonrpc::Response::Notif(_notif) => {
+				Err(Error::Custom("Server replied with notification to a a batch request".to_string()))
+			}
+			jsonrpc::Response::Batch(rps) => {
+				let mut responses = Vec::with_capacity(ids.len());
+				for rp in rps {
+					let next_id = ids
+						.pop_front()
```
niklasad1, line 138: This assumes the requests in the batch received in the same order as they are sent, maybe not a valid assumption note sure.

insipx: Not sure what the server-side code for this looks like, but the jsonrpc spec (from [jsonrpc.org](https://www.jsonrpc.org/specification#batch)) says:

So in the end I guess it's an implementation detail depending on how jsonrpsee library processes the requests

Everything else LGTM though

Random question though because i'm not totally sure on the specifics, do batch requests work for HTTP/IPC/WS or is it just over http?

niklasad1: Ok, cool missed that completely. I will fix that tomorrow.

src/types/jsonrpc/id.rs
```diff
@@ -40,6 +40,26 @@ pub enum Id {
...
+	/// If the `Id` is a number, returns the associated number. Returns None
+	/// otherwise.
+	pub fn as_number(&self) -> Option<&u64> {
+		match self {
+			Self::Num(n) => Some(n),
+			_ => None,
+		}
+	}
+
+	/// If the `Id` is an String, returns the associated &str. Returns None
```
niklasad1, line 53:
```suggestion
	/// If the `Id` is a String, returns the associated &str. Returns None
```

src/types/jsonrpc/response.rs
```diff
@@ -147,9 +148,10 @@ impl Output {
 	}
 }
 
-impl From<Output> for Result<JsonValue, Error> {
```
niklasad1, line 150: This was annoying w.r.t type inference, let's use `TryFrom` instead

src/client/http/client.rs
niklasad1: After implementing requests in any order I just realized that the user can't know whether we receive the requests in order or not, thus we need to sort the responses by ID before returning them to the user.

dvdplm: Pretty sure that is not a valid assumption. 
Maybe you need to use a `BTreeMap` here between ids and requests/responses?

niklasad1: Indeed, it's already implemented with a Vec instead of BTreeMap.

The reason is that the request IDs may wrap around and we can't order them by ID then

src/client/http/client.rs
```diff
@@ -76,21 +84,73 @@ impl HttpClient {
...
-				ret.map_err(Error::Request)
-			}
-			jsonrpc::Id::Num(n) => {
-				Err(Error::InvalidRequestId(Mismatch { expected: expected_id.into(), got: (*n).into() }))
+	/// Perform a batch request towards the server.
+	///
+	/// Returns `Ok` if all requests were answered successfully.
+	/// Returns `Error` if any of the requests fails.
+	///
+	/// If more than `u64::MAX` requests are performed in the `batch` then duplicate IDs are used
```
niklasad1, line 92: Technically, I think iterators are capped at u32/u64 so this comment is probably faulty.

src/client/http/client.rs
```diff
@@ -76,21 +84,73 @@ impl HttpClient {
...
+	///
+	// TODO(niklasad1): maybe simplify generic `requests`, it's quite unreadable.
+	pub async fn batch_request<'a>(
+		&self,
+		requests: impl IntoIterator<Item = (impl Into<String>, impl Into<jsonrpc::Params>)>,
+	) -> Result<Vec<JsonValue>, Error> {
+		let mut calls = Vec::new();
+		// NOTE: `request_id` is not necessilary increasing, it might wrap around at `u64::MAX`
+		// Thus, we can't use `request_id` as ordering.
+		let mut id_lookup_table = fnv::FnvHashMap::default();
```
niklasad1, line 103: maybe use `Iterator::size_hint` to pre-allocate for this.

niklasad1:
I will open a new PR instead messed up the merge.

Issue #143, opened 2020-10-27
Deadlock when request is issued when some subscription is active


svyatonik:
Steps to repeat:
1) open a connection to the ws-server (e.g. substrate node);
2) subscribe to some notifications stream (e.g. to justifications using `grandpa_justifications`);
3) wait enough time for 5 notifications to be sent by the server (e.g. 20 blocks). DO NOT read any notifications from `jsonrpsee::Subscription`;
4) make request to the server (e.g. `chain_getHeader`) using the same client => deadlock.

The reason is that [notifications channel](https://github.com/paritytech/jsonrpsee/blob/040c4b286ea273e0e0f05ac1363d5e9665e88dd3/src/client.rs#L369) is created with buffer size = 4. When we're trying to send 5th notification using blocking `SinkExt::send().await` [here](https://github.com/paritytech/jsonrpsee/blob/040c4b286ea273e0e0f05ac1363d5e9665e88dd3/src/client.rs#L388), the background thread will stop processing `FrontToBack` requests. So if we have single-threaded client, it'll block [here](https://github.com/paritytech/jsonrpsee/blob/040c4b286ea273e0e0f05ac1363d5e9665e88dd3/src/client.rs#L170), waiting for background thread to issue request and read response. At the same time, background thread is waiting for us to read pending notifications from the channel.

niklasad1:
Right, so the simple fix is to use an unbounded channel? 

It seems tricky to pick the "right" buffer size.

tomusdrw:
IMHO have it user-configured and drop excessive notifications (they usually have an ephemeral nature anyway). If the user is interested in not missing any of them, then they can set the size of the channel high enough. We can't (and imho shouldn't) really back-pressure the server, it would presumably be dropping excessive notifications if the client can't keep up either - you can't stop the external stuff that generates these notifications from making progress really :)

niklasad1:
Fixed by #180

PR #165, opened 2020-11-24
feat(websocket client):  WIP V2 ring buffer channel

niklasad1:
Super far from being mergeable but it fixes the deadlock at least in https://github.com/paritytech/jsonrpsee/issues/143

It makes is it user-configurable to allow losses for requests and subscriptions
- ring channel => allow losses
- mpsc::channel => don't push work on the queue if the channel is full.


Cargo.toml
```diff
@@ -19,6 +19,7 @@ parking_lot = "0.11.1"
 pin-project = "1.0.1"
 jsonrpsee-proc-macros = { path = "proc-macros" }
 rand = "0.7.3"
+ring-channel = { git = "https://github.com/niklasad1/ring-channel.git", branch = "na-atomicptr" }
```
niklasad1, line 22: https://github.com/brunocodutra/ring-channel/issues/69 ^^

src/client/ws/client.rs
```diff
@@ -36,6 +39,56 @@ use futures::{
 };
 use std::{collections::HashMap, io, marker::PhantomData};
 
+enum InternalChannelSender<T> {
+	AllowLosses(RingSender<T>),
+	BlockWhenFull(mpsc::Sender<T>),
+}
+
+impl<T> InternalChannelSender<T> {
+	fn send(self, data: T) -> Result<(), Error> {
```
niklasad1, line 48: TODO: rename to `send_non_blocking` and provide an additional `send_async` or something

Cargo.toml
dvdplm: Were you planning to open a PR for this?

niklasad1: It's up to the author of the library, he doesn't seem to be very keen on using `AtomicPtr` instead of `NonNullPtr` let's see how it goes. However, the idea is to split this PR into two parts when it works properly

1. Extract `try_send/send.now_or_never` into a separate PR and handle the errors without `ring channel crate`
2. Support for ring channel for ring channel.

niklasad1:
> > It's up to the author of the library, he doesn't seem to be very keen on using `AtomicPtr` instead of `NonNullPtr` let's see how it goes. However, the idea is to split this PR into two parts when it works properly
> 
>     1. Extract `try_send/send.now_or_never` into a separate PR and handle the errors without `ring channel crate`
> 
>     2. Support for ring channel for ring channel.


I will split this into two different PRs instead

PR #166, opened 2020-11-27
fix(websocket client): drop subscriptions that can't keep up with the internal buffer size

niklasad1:
Closes https://github.com/paritytech/jsonrpsee/issues/143

The major change in this PR it's using [Sender::try_send](https://docs.rs/futures/0.3.8/futures/channel/mpsc/struct.Sender.html#method.try_send) for the internal message passing channel for each subscription within the client which will drop the channel if the buffer gets full.

Thus, I had to change the API a little bit for `Subscriptions` to behave as `Streams` in Rust that now returns `None` if the channel is full or is disconnected instead of waiting for successful responses.

The capacity for the internal buffers are now user-configurable

## NOTE
This PR doesn't use [Sender::try_send](https://docs.rs/futures/0.3.8/futures/channel/mpsc/struct.Sender.html#method.try_send) for the tasks that depend on remote responses (method calls, notifications, and register "new" subscriptions). That's pointless because the sender is "cloned" because it takes [`immutable ref`](https://github.com/paritytech/jsonrpsee/blob/v2/src/client/ws/client.rs#L130-#L134), (~~this we could maybe change to &mut self~~). Thus `Sender::clone` increases the buffer size by one and it may diverge significantly if the many requests, notifications are performed "concurrently".

TLDR; It shouldn't deadlock for those but it might allow more `buffer size` than "configured"


src/client/ws/client.rs
```diff
@@ -339,28 +379,49 @@ async fn background_task(mut client: RawClient, mut from_front: mpsc::Receiver<F
...
-						.unwrap();
+				let notifs_tx = match active_subscriptions.get_mut(&request_id) {
+					None => {
+						log::debug!("Invalid subscription response: {:?}", request_id);
+						continue;
+					}
+					Some((notifs_tx, _)) => notifs_tx,
+				};
+
+				// NOTE: This is non_blocking but doesn't depend on any external handling to finish
```
niklasad1, line 392:
```suggestion
				// NOTE: This is non_blocking but it doesn't depend on any external handling to finish
```

src/client/ws/client.rs
```diff
@@ -339,28 +379,49 @@ async fn background_task(mut client: RawClient, mut from_front: mpsc::Receiver<F
...
+				log::error!(
+					"Client Error: {:?}, <https://github.com/paritytech/jsonrpsee/issues/154>",
+					SokettoError::UnexpectedOpCode(e)
+				);
+			}
 			Either::Right(Err(e)) => {
 				// TODO: https://github.com/paritytech/jsonrpsee/issues/67
-				log::error!("Client Error: {:?}", e);
+				log::error!("Client Error: {:?} terminating connection", e);
+				break;
```
niklasad1, line 424: Unrelated, but it closes the event loop once an error is received.

src/client/ws/client.rs
```diff
@@ -45,16 +50,45 @@ use std::{collections::HashMap, io, marker::PhantomData};
...
+		}
+	}
 }
 
 /// Active subscription on a [`Client`].
 pub struct Subscription<Notif> {
 	/// Channel to send requests to the background task.
 	to_back: mpsc::Sender<FrontToBack>,
-	/// Channel from which we receive notifications from the server, as undecoded `JsonValue`s.
+	/// Channel from which we receive notifications from the server, as un-decoded `JsonValue`s.
```
niklasad1, line 88:
```suggestion
	/// Channel from which we receive notifications from the server, as undecoded `JsonValue`s.
```

src/client/ws/client.rs
```diff
@@ -225,20 +263,20 @@ where
 impl<Notif> Drop for Subscription<Notif> {
 	fn drop(&mut self) {
 		// We can't actually guarantee that this goes through. If the background task is busy, then
-		// the channel's buffer will be full, and our unsubscription request will never make it.
+		// the channel's buffer will be full, and our un-subscription request will never make it.
```
niklasad1, line 266:
```suggestion
		// the channel's buffer will be full, and our unsubscription request will never make it.
```

src/client/ws/client.rs
```diff
@@ -45,6 +50,35 @@ use std::{collections::HashMap, io, marker::PhantomData};
...
+
+#[derive(Copy, Clone, Debug)]
+/// Configuration.
+pub struct Config {
+	/// Backend channel for serving requests and notifications.
+	pub request_channel_capacity: usize,
+	/// Backend channel for each unique subscription.
+	pub subscription_channel_capacity: usize,
+	/// Allow losses when the channel gets full
+	pub allow_subscription_losses: bool,
```
niklasad1, line 65: not used, could be removed.

src/client/ws/client.rs
```diff
@@ -45,6 +50,35 @@ use std::{collections::HashMap, io, marker::PhantomData};
...
+	/// Allow losses when the request/notifications channel gets full
+	pub allow_request_losses: bool,
```
niklasad1, line 67: not used, could be removed.

src/client/ws/client.rs
```diff
@@ -205,18 +242,19 @@ impl<Notif> Subscription<Notif>
 where
 	Notif: jsonrpc::DeserializeOwned,
 {
-	/// Returns the next notification sent from the server.
+	/// Returns the next notification from the stream
+	/// This may return `None` if finished subscription has been terminated
 	///
 	/// Ignores any malformed packet.
-	pub async fn next(&mut self) -> Notif {
+	pub async fn next(&mut self) -> Option<Notif> {
```
niklasad1, line 249: This is the major change in this PR.

src/client/ws/client.rs
```diff
@@ -45,6 +50,35 @@ use std::{collections::HashMap, io, marker::PhantomData};
...
 	to_back: mpsc::Sender<FrontToBack>,
+	/// Config.
+	config: Config,
+}
+
+#[derive(Copy, Clone, Debug)]
+/// Configuration.
+pub struct Config {
+	/// Backend channel for serving requests and notifications.
+	pub request_channel_capacity: usize,
```
niklasad1, line 61: this becomes useless in practice because we clone the sender every time, see PR description for more info.

src/client/ws/client.rs
```diff
@@ -205,18 +242,19 @@ impl<Notif> Subscription<Notif>
 where
 	Notif: jsonrpc::DeserializeOwned,
 {
-	/// Returns the next notification sent from the server.
+	/// Returns the next notification from the stream
+	/// This may return `None` if finished subscription has been terminated
```
niklasad1, line 246:
```suggestion
	/// This may return `None` if the subscription has been terminated, may happen if the channel becomes full or dropped.
```

src/client/ws/client.rs
maciejhirsz: Are the sending ends of the channels going to shut down gracefully when we drop the receivers here?

niklasad1: Technically, the senders are not gracefully terminated but once the `receiver` is dropped here all senders will receive `None -> Error::Internal` once any of the operations (method call, notification, subscription) is invoked.

Thus, not possible for the user to know the exact failure reason without another channel or message (but we have logs lol)

src/client/ws/client.rs
```diff
@@ -339,28 +379,49 @@ async fn background_task(mut client: RawClient, mut from_front: mpsc::Receiver<F
...
-						.close(unsubscribe)
-						.await
-						.unwrap();
+				let notifs_tx = match active_subscriptions.get_mut(&request_id) {
+					None => {
+						log::debug!("Invalid subscription response: {:?}", request_id);
+						continue;
+					}
+					Some((notifs_tx, _)) => notifs_tx,
+				};
```
dvdplm, line 390: This is nice.

dvdplm (approved):
I have a few questions and nitpicks, but overall lgtm. :)

src/client/ws/client.rs
```diff
@@ -205,18 +242,19 @@ impl<Notif> Subscription<Notif>
 where
 	Notif: jsonrpc::DeserializeOwned,
 {
-	/// Returns the next notification sent from the server.
+	/// Returns the next notification from the stream
+	/// This may return `None` if the subscription has been terminated, may happen if the channel becomes full or dropped.
 	///
 	/// Ignores any malformed packet.
```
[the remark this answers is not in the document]
niklasad1: why not

src/client/ws/client.rs
```diff
@@ -339,28 +379,49 @@ async fn background_task(mut client: RawClient, mut from_front: mpsc::Receiver<F
...
-						.unwrap();
+				let notifs_tx = match active_subscriptions.get_mut(&request_id) {
+					None => {
+						log::debug!("Invalid subscription response: {:?}", request_id);
+						continue;
+					}
+					Some((notifs_tx, _)) => notifs_tx,
+				};
+
+				// NOTE: This is non_blocking but it doesn't depend on any external handling to finish
```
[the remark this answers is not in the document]
niklasad1: no, let's remove that completely. I meant that is non-blocking such that there is nothing necessarily that drives it to completion, thus relies on the user to `poll` the receiving end.

src/ws/tests.rs
```diff
@@ -89,72 +66,6 @@ async fn single_method_call_works() {
...
-	for _ in 0..10 {
-		for (_client, hello_sub, foo_sub) in &mut clients {
-			let hello = hello_sub.next().await;
-			let foo = foo_sub.next().await;
-			assert_eq!(hello, JsonValue::String("hello from subscription".to_owned()));
-			assert_eq!(foo, JsonValue::Number(1337_u64.into()));
-		}
-	}
-}
-
```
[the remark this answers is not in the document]
niklasad1: Sorry, some of tests here were not really "unit tests" so I moved them to `integration tests` instead.

src/client/ws/client.rs
```diff
@@ -300,9 +338,10 @@ async fn background_task(mut client: RawClient, mut from_front: mpsc::Receiver<F
 					}
 				}
 			}
+			// A subscription has been closed (could improved to be used for requests too.)
 			Either::Left(Some(FrontToBack::ChannelClosed)) => {
-				// TODO: there's no way to cancel pending subscriptions and requests, otherwise
-				// we should clean them up as well
+				//TODO: there's no way to cancel pending subscriptions and requests
+				//TODO(niklasad1): using `iter().find()` is wrong, it's guessing (could close down the wrong channel) and inefficient
```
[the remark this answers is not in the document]
niklasad1: Added an issue for it.

src/client/ws/client.rs
```diff
@@ -338,20 +327,29 @@ async fn background_task(mut client: RawClient, mut from_front: mpsc::Receiver<F
...
 			Either::Right(Ok(RawClientEvent::Response { request_id, result })) => {
 				log::trace!("[backend] client received response to req={:?}, result={:?}", request_id, result);
-				let _ = ongoing_requests.remove(&request_id).unwrap().send(result.map_err(Error::Request));
+				match ongoing_requests.remove(&request_id) {
+					Some(r) => {
+						if let Err(e) = r.send(result.map_err(Error::Request)) {
+							log::error!("Could not dispatch pending request: {:?}", e);
+						}
+					}
+					None => log::error!("Invalid response ID to request received"),
```
dvdplm, line 351:
```suggestion
					None => log::error!("No pending response found for request ID {}", request_id),
```

maybe?

Issue #169, opened 2020-12-01
[ws client]: ChannelClosed message should embed `request_id`


niklasad1:
When a subscription is closed, all the subscriptions are iterated over and returns once a closed subscription is found, which is inefficient and could close the wrong subscription if "several are closed at the same time", see https://github.com/paritytech/jsonrpsee/blob/v2/src/client/ws/client.rs#L303-#L309

Thus, it would be better to just have `ChannelClosed(RequestId)` than it could be performed in constant time and is correct.

PR #170, opened 2020-12-01
fix(ws client): embed request id in `SubscriptionClosed`

niklasad1:
Fixes https://github.com/paritytech/jsonrpsee/issues/169

I changed `ChannelClosed` to `SubscriptionClosed` because that only makes sense for registered subscriptions because `pending subscriptions` and `pending request` are blocking until a response is received and fails only if the background channel is terminated so doesn't make sense to send a message to the background thread for that.

I realize now that all those operations  `method call, notification, subscription` (which are futures) are not cancel-safe and might run out of memory, such as if someone does calls in a loop drop the futures.


src/client/ws/client.rs
```diff
@@ -340,8 +343,15 @@ async fn background_task(mut client: RawClient, mut from_front: mpsc::Receiver<F
...
-				let (notifs_tx, _) = active_subscriptions.get_mut(&request_id).unwrap();
+				let notifs_tx = match active_subscriptions.get_mut(&request_id) {
+					Some((notifs_tx, _)) => notifs_tx,
+					None => {
+						log::error!("Received notification on non-registrered subscription: {:?}", request_id);
+						continue;
+					}
+				};
+
+				// Unsubscribe if the send notification failed.
```
niklasad1, line 354:
```suggestion
				// Unsubscribe if sending the notification failed.
```

src/client/ws/client.rs
```diff
@@ -116,16 +118,15 @@ enum FrontToBack {
...
-	/// While this means that closing a request or a subscription is a `O(n)` operation, it is
-	/// expected that the volume of requests and subscriptions is low enough that this isn't
-	/// a problem in practice.
-	ChannelClosed,
+	/// When a subscription channel is closed, we send this message to the background
+	/// task in order for it to garbage collect it.
+	// NOTE: There is not possible to cancel pending subscriptions or pending requests.
+	// Those operations will be blocked until a response is received or
+	// that the background thread has been terminated.
+	SubscriptionClosed(RawClientRequestId),
```
dvdplm, line 129:
```suggestion
	/// When a subscription channel is closed, we send this message to the background
	/// task to mark it ready for garbage collection.
	// NOTE: It is not possible to cancel pending subscriptions or pending requests.
	// Such operations will be blocked until a response is received or the background
	// thread has been terminated.
	SubscriptionClosed(RawClientRequestId),
```

src/client/ws/client.rs
```diff
@@ -373,11 +369,11 @@ async fn background_task(mut client: RawClient, mut from_front: mpsc::Receiver<F
 			// Received a response on a subscription.
 			Either::Right(Ok(RawClientEvent::SubscriptionNotif { request_id, result })) => {
 				let notifs_tx = match active_subscriptions.get_mut(&request_id) {
+					Some((notifs_tx, _)) => notifs_tx,
 					None => {
-						log::debug!("Invalid subscription response: {:?}", request_id);
+						log::error!("Received notification on non-registrered subscription: {:?}", request_id);
```
dvdplm, line 374:
```suggestion
						log::error!("Received notification on unknown subscription: {:?}", request_id);
```
maybe?

dvdplm (approved):

Issue #154, opened 2020-11-03
[ws transport]: soketto occasionally returns `UnexpectedOpCode(Continue)` when reading empty WebSocket message


niklasad1:
I have noticed a difference between [`soketto v0.3` and `soketto v0.4`](https://github.com/paritytech/soketto/issues/25) where it sometimes reads an empty WebSocket message and returns `UnexpectedOpCode(Continue)`.

For the client it doesn't make much difference than just that `ERROR jsonrpsee::client::ws::client] Client Error: Inner(Ws(UnexpectedOpCode(Continue)))` would be logged, still quite annoying.

For the server, it's critical because it would terminate the connection but I haven't seen it yet at least.... it might just `soketto::Client` related but  `receive_data` and `send_data` seems to have the same implemetation.

Note: I tried to downgrade `soketto v0.3.2` but it didn't manage to get the `ws client/server to work`.

niklasad1:
As Toralf described, 

The futures in `soketto::receive_data` is not cancelable and currently, those futures are spawned in a loop from [client](https://github.com/paritytech/jsonrpsee/blob/master/src/client.rs#L286) -> [raw client::next_event](https://github.com/paritytech/jsonrpsee/blob/master/src/raw/client.rs#L333) -> [ws transport::next_response](https://github.com/paritytech/jsonrpsee/blob/master/src/transport/ws/client.rs#L223)

It was fixed in the WebSocket server: https://github.com/paritytech/jsonrpsee/pull/114

However, it will require some refactoring to convert the future into a stream because the client is used for other things too and `stream::unfold` takes ownership of the closure.

niklasad1:
//cc @maciejhirsz could it be related to "bad performance" i.e, spawning unnecessary futures (bloat the executor thread pool)?

niklasad1:
Fixed by https://github.com/paritytech/jsonrpsee/pull/180

PR #171, opened 2020-12-04
[ws client]: kill raw client

niklasad1:
Fixes https://github.com/paritytech/jsonrpsee/issues/154

The major fix is that it splits up the `RawClient` into a Sender and Receiver part such that in the event loop we don't have `destroy the borrow` which causes us to drop the future and leave Soketto in a bad state. Because Soketto's async methods are not canceling safe.

As the `Sender` and `Receiver` is separate the functionality that keeps track of the state of pending requests, pending subscription and active subscriptions are now handled by a new type `RequestManager`, that lives in background task.


src/client/ws/jsonrpc_transport.rs
```diff
@@ -0,0 +1,93 @@
+//! JSONRPC Transport module.
+//!
+//! Wraps the underlying WebSocket transport with specific JSONRPC details.
+
+use crate::client::ws::transport::WsConnectError;
+use crate::types::jsonrpc;
+
+pub async fn websocket_context(target: impl AsRef<str>) -> Result<(Sender, Receiver), WsConnectError> {
+	let (sender, receiver) = crate::client::ws::transport::new(target.as_ref()).await.unwrap();
```
niklasad1, line 9: fix the unwrap!!!!

src/client/ws/manager.rs
```diff
@@ -0,0 +1,278 @@
...
+		}
+	}
+
+	/// Remove an active subscription.
+	pub fn remove_active_subscription(
+		&mut self,
+		request_id: RequestId,
+		subscription_id: SubscriptionId,
+	) -> Option<(SubscriptionSink, UnsubscribeMethod)> {
+		match (self.requests.entry(request_id), self.subscriptions.entry(subscription_id)) {
```
niklasad1, line 138: Entry API takes ownership of the key, could do:

```rust
let request = matches!(self.request.get(&request_id), Some(Kind::Subscription(_));
let sub = matches!(self.request.get(&request_id), Some(_);
if request && sub {
   // remove request and sub.
}
```

Still have to do two lookups though :(

src/client/ws/raw.rs
```diff
@@ -1,710 +0,0 @@
-// Copyright 2019 Parity Technologies (UK) Ltd.
```
niklasad1, line 1: Removed RawClient

src/client/ws/transport.rs
```diff
@@ -275,7 +261,7 @@ impl<'a> WsTransportClientBuilder<'a> {
 		};
 
 		// Configure a WebSockets client on top.
-		let mut client = WsRawClient::new(tcp_stream, &self.host, &self.url);
+		let mut client = WsRawClient::new(BufReader::new(BufWriter::new(tcp_stream)), &self.host, &self.url);
```
niklasad1, line 264: optimization

src/ws/raw/tests.rs
```diff
@@ -1,94 +0,0 @@
-// Copyright 2019 Parity Technologies (UK) Ltd.
```
niklasad1, line 1: Removed these tests because we are testing it via `WsServer` tests anyway

tests/integration_tests.rs
```diff
@@ -163,7 +163,7 @@ async fn ws_subscription_without_polling_doesnt_make_client_unuseable() {
 		assert!(hello_sub.next().await.is_some());
 	}
 
-	// NOTE: this is now unuseable and unregistered.
+	// NOTE: this is now un-useable and unregistered.
```
niklasad1, line 166:
```suggestion
	// NOTE: this is now unuseable and unregistered.
```

src/client/ws/client.rs
```diff
@@ -118,29 +115,30 @@ enum FrontToBack {
...
 	// thread has been terminated.
-	SubscriptionClosed(RawClientRequestId),
+	SubscriptionClosed(SubscriptionId),
 }
 
 impl Client {
 	/// Initializes a new WebSocket client
 	///
 	/// Fails when the URL is invalid.
 	pub async fn new(target: impl AsRef<str>, config: Config) -> Result<Self, Error> {
```
dvdplm, line 133:
```suggestion
	pub async fn new(remote_addr: impl AsRef<str>, config: Config) -> Result<Self, Error> {
```

src/client/ws/manager.rs
```diff
@@ -0,0 +1,278 @@
...
+	/// Complete a pending request.
+	pub fn complete_pending_method_call(&mut self, request_id: RequestId) -> Option<PendingRequestOneshot> {
+		match self.requests.entry(request_id) {
+			Entry::Occupied(request) if matches!(request.get(), Kind::PendingMethodCall(_)) => {
+				let (_req_id, kind) = request.remove_entry();
+				if let Kind::PendingMethodCall(send_back) = kind {
+					Some(send_back)
+				} else {
+					unreachable!();
+				}
```
dvdplm, line 126: These `unreachable`s are a bit unfortunate; ideally it shouldn't be possible to complete anything that isn't a `Kind::PendingMethodCall` but not sure how to do that. @maciejhirsz thoughts?
 
Is returning `None` better? :/

src/client/ws/manager.rs
```diff
@@ -0,0 +1,278 @@
+//! Handles and monitors JSONRPC v2 method calls and subscriptions
+//!
+//! Definitions:
+//!     * Request ID - request ID in JSONRPC-v2 specification
+//!     (may be number, string or null but only numbers are supported currently)
+//!     * Subscription ID - unique ID generated by server
```
dvdplm, line 6:
```suggestion
//!     * Request ID - request ID in JSONRPC-v2 specification
//!     (the specs allow number, string or null but this crate only supports numbers)
//!     * Subscription ID - unique ID generated by server
```

src/client/ws/manager.rs
niklasad1: It is not "pretty" but the "state" is checked above `matches!(request.get(), Kind::PendingMethodCall(_))`  and the `unreachable!` should be impossible to happen but not encoded in the type-system.

So, seemed more intuitive to use unreachable because if that happened it's a logic bug but no strong opinions on it.

As we only have the `request ID`, I don't see how to solve it without a runtime check of the state as this does.

dvdplm: Yeah I couldn't come up with a better idea either.

maciejhirsz: `unimplemented!()` is probably a better flag

maciejhirsz: If you call `complete_pending_method_call`, but the state is not pending, that sounds like a logic bug to me. I'd rather have an error type here:

```rust
    pub fn complete_pending_method_call(&mut self, request_id: RequestId) -> Result<PendingRequestOneshot, SomeError> {
        match self.requests.entry(request_id) {
            Entry::Occupied(request) => {
                let (_req_id, kind) = request.remove_entry();
                if let Kind::PendingMethodCall(send_back) = kind {
                    Ok(send_back)
                } else {
                    Err(...)
                }
```

Is there a scenario where this method can be called on any other state (or lack of state) under normal operation?

niklasad1: Currently,  it works like this:

```rust
 let response = get_reponse();

 match {
   // response on pending subscriptions and pending requests,
   RequestResponse(response) => {
       match manager.get_status(response.id) {
           RequestStatus::PendingRequest => {
              manager.complete_pending_request(request_id);
           }
           RequestStatus::PendingSubscription => {
               manager.complete_pending_subscription(request_id);
           }
           _ => // don't care invalid request ID
      }
   }
   // response on active subscription
   SubscriptionResponse(sub) => {
      if let Some(request_id) = manager.get_request_id_by_subscription(sub.id) { 
          manager.as_active_subscription_mut(request_id).send(sub.response);
      }
   }
 }
```

I would say that we know the state expect for `active subscription` where I assumed that it must be a subscription because we have a separate message for it (the subscription ID should be unique and very unlikely to have duplicates)

However, I did this very "defensively" because I assumed it could be called on non-validated input too but probably fair to assume that the caller has to ensure that it has a valid state.

FWIW, in your example, if I'd call that with a pending subscription instead it would remove that entry in the map because it has the same `request-id` but the "wrong state".

maciejhirsz: True. The `unreachable!`s there are think otherwise I think, I'll have to think about it a bit more.

src/client/ws/client.rs
```diff
@@ -257,164 +255,205 @@ impl<Notif> Drop for Subscription<Notif> {
...
+			},
+		}
+	}
+}
+
+/// Process a response for the Server.
+///
+/// Returns Some(method, params) if the client has to send back any response.
+/// None otherwise.
+fn process_response(
```
niklasad1, line 388: I just realized that this should probably return `Result<Option<callback>, Err>` if the request was not found it need to be propagated and handled to terminate the `background thread` because `Err(e)` from the backend doesn't check request id anymore.

It's one if this function's tasks.

src/client/ws/client.rs
```diff
@@ -339,14 +343,33 @@ async fn background_task(
 					break;
 				}
 				Some(Ok(jsonrpc::Response::Single(response))) => {
-					if let Some((unsubscribe, params)) = process_response(&mut manager, response, config.subscription_channel_capacity) {
-						sender.start_request(unsubscribe, params).await.unwrap();
+					match process_response(&mut manager, response, config.subscription_channel_capacity) {
+						Ok(Some((unsubscribe, params))) => {
+							if let Err(e) = sender.start_request(unsubscribe, params).await {
+								log::error!("Failed to send unsubscription response: {:?}", e);
```
niklasad1, line 349: Note, we could terminate the client here also because the server was probably shut down if the request failed.

src/client/ws/client.rs
```diff
@@ -339,14 +343,33 @@ async fn background_task(
...
 					}
 				}
 				Some(Ok(jsonrpc::Response::Batch(responses))) => {
+					// if any request fails, throw away entire batch.
 					for response in responses {
-						if let Some((unsubscribe, params)) = process_response(&mut manager, response, config.subscription_channel_capacity) {
-							sender.start_request(unsubscribe, params).await.unwrap();
+						match process_response(&mut manager, response, config.subscription_channel_capacity) {
+							Ok(Some((unsubscribe, params))) => {
+								if let Err(e) = sender.start_request(unsubscribe, params).await {
```
niklasad1, line 364: Note, we could terminate the client here also because the server was probably shut down if the request failed.

src/client/ws/manager.rs
niklasad1: //cc @dvdplm thoughts?

niklasad1:
@dvdplm I think this is ready for review

src/client/ws/client.rs
```diff
@@ -257,164 +255,228 @@ impl<Notif> Drop for Subscription<Notif> {
...
+///
+/// Returns `Ok(_)` if the response was successful or if the error could be handled.
+/// Returns `Err(_)` if the response couldn't be handled.
+fn process_response(
+	manager: &mut RequestManager,
+	response: jsonrpc::Output,
+	subscription_capacity: usize,
+) -> Result<Option<(String, jsonrpc::Params)>, Error> {
+	let response_id = match response.id().as_number() {
+		Some(n) => *n,
```
dvdplm, line 417: Could this be `let send_back_oneshot =  manager.complete_pending_method_call(response_id).ok_or_else(|| { Error::InvalidRequestId })?;`

src/client/ws/client.rs
```diff
@@ -257,164 +255,228 @@ impl<Notif> Drop for Subscription<Notif> {
...
+		None => return Err(Error::InvalidRequestId),
+	};
+
+	match manager.request_status(&response_id) {
+		RequestStatus::PendingMethodCall => {
+			let send_back_oneshot = match manager.complete_pending_method_call(response_id) {
+				Some(send_back) => send_back,
+				None => return Err(Error::InvalidRequestId),
+			};
+			let response: Result<JsonValue, Error> = response.try_into().map_err(Error::Request);
```
dvdplm, line 427: No need for type annotations here I think.

src/client/ws/client.rs
```diff
@@ -257,164 +255,228 @@ impl<Notif> Drop for Subscription<Notif> {
...
+			let (send_back_oneshot, unsubscribe_method) = match manager.complete_pending_subscription(response_id) {
+				Some(pending) => pending,
+				None => return Err(Error::InvalidRequestId),
+			};
+
+			let json_sub_id: JsonValue = match response.try_into() {
+				Ok(response) => response,
+				Err(e) => {
+					return match send_back_oneshot.send(Err(Error::Request(e))) {
+						Err(Err(e)) => Err(e),
```
dvdplm, line 443: These erorr-in-errors feel a bit boilerplatey; maybe we can try to clean it up in a diff PR?

src/client/ws/jsonrpc_transport.rs
```diff
@@ -0,0 +1,94 @@
...
+use crate::client::ws::transport::{self, WsConnectError, WsNewDnsError};
+use crate::types::jsonrpc;
+
+/// Creates a new JSONRPC WebSocket connection, represented as a Sender and Receiver pair.
+pub async fn websocket_connection(target: impl AsRef<str>) -> Result<(Sender, Receiver), WsNewDnsError> {
+	let (sender, receiver) = transport::websocket_connection(target.as_ref()).await?;
+	Ok((Sender::new(sender), Receiver::new(receiver)))
+}
+
+pub struct Sender {
```
dvdplm, line 14: Missing docs

src/client/ws/jsonrpc_transport.rs
```diff
@@ -0,0 +1,94 @@
...
+	request_id: u64,
+	transport: crate::client::ws::transport::Sender,
+}
+
+impl Sender {
+	pub fn new(transport: crate::client::ws::transport::Sender) -> Self {
```
dvdplm, line 20: missing docs

src/client/ws/jsonrpc_transport.rs
```diff
@@ -0,0 +1,94 @@
...
+		let request = jsonrpc::Request::Single(jsonrpc::Call::Notification(jsonrpc::Notification {
+			jsonrpc: jsonrpc::Version::V2,
+			method: method.into(),
+			params: params.into(),
+		}));
+
+		self.transport.send_request(request).await
+	}
+
+	pub async fn start_request(
```
dvdplm, line 64: missing docs

src/client/ws/jsonrpc_transport.rs
```diff
@@ -0,0 +1,94 @@
...
+		&mut self,
+		method: impl Into<String>,
+		params: impl Into<jsonrpc::Params>,
+	) -> Result<u64, WsConnectError> {
+		self.start_impl(method, params).await
+	}
+
+	pub async fn start_subscription(
```
dvdplm, line 72: missing docs

src/client/ws/jsonrpc_transport.rs
```diff
@@ -0,0 +1,94 @@
...
+		// Note that in case of an error, we "lose" the request id (as in, it will never be
+		// used). This isn't a problem, however.
+		self.transport.send_request(request).await?;
+
+		Ok(id)
+	}
+
+	/// Sends a notification to the server. The notification doesn't need any response.
+	///
+	/// This asynchronous function finishes when the notification has finished being sent.
```
dvdplm, line 49: s/This asynchronous function finishes when the notification has finished being sent./This asynchronous function finishes all notification have been sent./

src/client/ws/jsonrpc_transport.rs
```diff
@@ -0,0 +1,94 @@
...
+
+		let request = jsonrpc::Request::Single(jsonrpc::Call::MethodCall(jsonrpc::MethodCall {
+			jsonrpc: jsonrpc::Version::V2,
+			method: method.into(),
+			params: params.into(),
+			id: jsonrpc::Id::Num(id),
+		}));
+
+		// Note that in case of an error, we "lose" the request id (as in, it will never be
+		// used). This isn't a problem, however.
```
dvdplm, line 41: It's not a problem because we always increase the ID right?

src/client/ws/transport.rs
dvdplm: :+1

src/client/ws/transport.rs
```diff
@@ -254,7 +241,7 @@ impl<'a> WsTransportClientBuilder<'a> {
 	}
```
dvdplm, line 241: `with_url`, `with_origin_header` and `with_timeout` are not used it seems, can delete?

src/types/error.rs
```diff
@@ -33,9 +33,15 @@ pub enum Error {
 	/// Failed to parse the data that the server sent back to us.
 	#[error("Parse error: {0}")]
 	ParseError(#[source] jsonrpc::ParseError),
-	/// Invalid id in response to a request.
-	#[error("Invalid ID in response: {0}")]
-	InvalidRequestId(Mismatch<JsonValue>),
+	/// Invalid subscription ID in.
```
dvdplm, line 36: s/Invalid subscription ID in./Invalid subscription ID.

src/types/error.rs
```diff
@@ -33,9 +33,15 @@ pub enum Error {
...
+	#[error("Invalid subscription ID")]
+	InvalidSubscriptionId,
+	/// Invalid ID in request.
+	#[error("Invalid Request ID.")]
+	InvalidRequestId,
+	/// A request with the same request ID has already been registered.
```
dvdplm, line 42: The docs string is better than the error for some of these. Why not use the doc strings?

src/client/ws/transport.rs
niklasad1: I prefer the keep those if we decide to make the `client generic` or expose more `builder options` but I could export those to "silence" the warnings.

src/client/ws/manager.rs
```diff
@@ -0,0 +1,278 @@
...
+type PendingRequestOneshot = oneshot::Sender<Result<JsonValue, Error>>;
+type PendingSubscriptionOneshot = oneshot::Sender<Result<(mpsc::Receiver<JsonValue>, SubscriptionId), Error>>;
+type SubscriptionSink = mpsc::Sender<JsonValue>;
+type UnsubscribeMethod = String;
+type RequestId = u64;
+
+/// Manages and monitors JSONRPC v2 method calls and subscriptions.
+pub struct RequestManager {
+	/// List of requests that are waiting for a response from the server.
+	requests: FnvHashMap<RequestId, Kind>,
```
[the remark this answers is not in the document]
niklasad1: yeah, and it's just a `u64` which should be considered as "short" in the context of `fnv hasher`.

src/client/ws/jsonrpc_transport.rs
niklasad1: yes, it's increased on every request and it may "wrap around" when `u64::MAX` is reached.

The comment is copy-pasted from the old code.

src/client/ws/manager.rs
```diff
@@ -0,0 +1,278 @@
...
+			Kind::Subscription(_) => RequestStatus::Subscription,
+		})
+	}
+
+	/// Get a mutable reference to underlying `Sink` in order to send messages to the subscription.
+	pub fn as_active_subscription_mut(&mut self, request_id: &RequestId) -> Option<&mut SubscriptionSink> {
+		if let Some(Kind::Subscription((callback, _))) = self.requests.get_mut(request_id) {
+			Some(callback)
+		} else {
+			None
```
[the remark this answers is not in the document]
niklasad1: I can add logs but the caller should ideally do logging because the "state` is returned back".

src/client/ws/jsonrpc_transport.rs
niklasad1: I removed the `async comment`, it should be clear as the function is `async fn`.

Before it was in a trait with `Pin<Box<dyn Future<Output = ()>>>`

src/client/ws/client.rs
```diff
@@ -257,164 +255,228 @@ impl<Notif> Drop for Subscription<Notif> {
...
+/// Process a response from the server.
+///
+/// Returns `Ok(_)` if the response was successful or if the error could be handled.
+/// Returns `Err(_)` if the response couldn't be handled.
+fn process_response(
+	manager: &mut RequestManager,
+	response: jsonrpc::Output,
+	subscription_capacity: usize,
+) -> Result<Option<(String, jsonrpc::Params)>, Error> {
+	let response_id = match response.id().as_number() {
```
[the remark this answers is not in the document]
niklasad1: yeah, with an additional ok_or :)

src/client/ws/client.rs
niklasad1: The problem is that send returns `Result<(), T>` then T is `Result<JsonValue, Error>`, 

So, Err<Ok(_)> should be unreachable

src/client/ws/transport.rs
dvdplm: I see. Or just add `#[allow(unused)]`.

src/client/ws/client.rs
dvdplm: > Maybe explicit handling the code?

Not sure what you mean, can you elaborate?

src/client/ws/manager.rs
```diff
@@ -0,0 +1,278 @@
...
+	) -> Result<(), PendingSubscriptionOneshot> {
+		if let Entry::Vacant(v) = self.requests.entry(id) {
+			v.insert(Kind::PendingSubscription((send_back, unsubscribe_method)));
+			Ok(())
+		} else {
+			Err(send_back)
+		}
+	}
+
+	/// Inserts a new active subscription
```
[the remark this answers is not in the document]
dvdplm: sigh, "Tries to insert…", my bad.

src/client/ws/manager.rs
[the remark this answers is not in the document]
dvdplm: Good pt.

src/client/ws/transport.rs
niklasad1: Fixed in https://github.com/paritytech/jsonrpsee/pull/171/commits/1f904e1a4ab2d9775832b00931e22215ee0327ff

src/client/ws/manager.rs
```diff
@@ -38,7 +38,7 @@ type RequestId = u64;
 /// Manages and monitors JSONRPC v2 method calls and subscriptions.
 pub struct RequestManager {
 	/// List of requests that are waiting for a response from the server.
-	// NOTE: `FnvHashMap` is used here because `RequestId` is known to be short key (u64).
+	// NOTE: `FnvHashMap` is used here because `RequestId` is known to be a short key (u64).
```
dvdplm, line 41: `// NOTE: `FnvHashMap` is used here because `RequestId` is not under the callers control and is known to be a short key (u64).`

src/client/ws/client.rs
niklasad1: Yeah,  I think I got a little confused by myself for a while.

What I meant is that before I did:

```rust
				match send_back_oneshot.send(Err(Error::InvalidSubscriptionId)) {
					Err(Err(e)) => Err(e),
                    // Err(Ok(_)) -> will become Ok(None) handled by wildcard but it should be an error because it can't be handled when the channel is down.
					_ => Ok(None),
				}
```

I changed it to this:

```rust
				match send_back_oneshot.send(Err(Error::InvalidSubscriptionId)) {
					Err(Err(e)) => Err(e),
					Err(Ok(_)) => unreachable!("Error sent above; qed"),
					_ => Ok(None),
				}
```

src/client/ws/client.rs
```diff
@@ -257,164 +255,225 @@ impl<Notif> Drop for Subscription<Notif> {
...
-						log::warn!("[backend]: client start subscription failed: {:?}", err);
-						let _ = send_back.send(Err(Error::TransportError(Box::new(err))));
+				}
+				// User dropped a subscription.
+				Some(FrontToBack::SubscriptionClosed(sub_id)) => {
+					log::trace!("Closing in subscription: {:?}", sub_id);
+					// NOTE: The subscription may have been closed earlier if
+					// the channel was full or disconnected.
+					if let Some(request_id) = manager.get_request_id_by_subscription_id(&sub_id) {
+						if let Some((_sink, unsubscribe_method)) = manager.remove_subscription(request_id, sub_id.clone()) {
```
niklasad1, line 336: NOTE: I missed to send a request to the unsubscribe method before.

I should write a test for this but..... not in this PR :)

src/client/ws/client.rs
```diff
@@ -257,164 +255,225 @@ impl<Notif> Drop for Subscription<Notif> {
...
+							if let Ok(json_sub_id) = jsonrpc::to_value(sub_id) {
```
niklasad1, line 337: this should be infallible but....

src/client/ws/manager.rs
```diff
@@ -0,0 +1,278 @@
...
+type PendingCallOneshot = oneshot::Sender<Result<JsonValue, Error>>;
+type PendingSubscriptionOneshot = oneshot::Sender<Result<(mpsc::Receiver<JsonValue>, SubscriptionId), Error>>;
+type SubscriptionSink = mpsc::Sender<JsonValue>;
+type UnsubscribeMethod = String;
+type RequestId = u64;
+
+/// Manages and monitors JSONRPC v2 method calls and subscriptions.
+pub struct RequestManager {
+	/// List of requests that are waiting for a response from the server.
+	// NOTE: `FnvHashMap` is used here because `RequestId` is known to be a short key (u64).
```
niklasad1, line 41:
```suggestion
	// NOTE: FnvHashMap is used here because RequestId is not under the caller's control and is known to be a short key (u64).
```

PR #174, opened 2020-12-16
WsServer refactor

maciejhirsz:
Previous ws server is moved to the `ws-old` folder.

Known issues:
- Tests aren't updated and thus don't compile.
- Benches are still slow, but that should be all in the client as the server logic is mostly a copy-paste from the microsecond range experiment I had done before (which ditched the WsClient in benches).
- Registering methods currently only permits closures with `'static` lifetime, which makes it unfortunately unsuitable for any kind of reasonable state. This needs to be narrowed down to lifetime of the server, but making guarantees here is really difficult so it might be a dead end :|.

TODOS:
- [x] Fix tests :).
- [x] Finish subscriptions on existing scaffolding
- [x] Narrowing down the lifetime from `'static` to the lifetime of the hashmap / server (might need some `PhantomData` wizardry).


niklasad1:
@maciejhirsz @dvdplm 

I merged this branch to `v2` after the WebSocket client refactoring:

```bash
synchronous WebSocket round trip
                        time:   [56.826 us 59.964 us 63.214 us]
                        change: [-97.497% -97.088% -96.545%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) high mild

concurrent WebSocket round trip/3
                        time:   [107.34 us 114.95 us 124.17 us]
                        change: [-95.438% -94.540% -93.352%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 11 outliers among 100 measurements (11.00%)
  5 (5.00%) high mild
  6 (6.00%) high severe
concurrent WebSocket round trip/6
                        time:   [112.10 us 125.48 us 143.39 us]
                        change: [-96.184% -95.591% -94.811%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  1 (1.00%) high mild
  3 (3.00%) high severe
concurrent WebSocket round trip/12
                        time:   [187.41 us 201.00 us 214.74 us]
                        change: [-93.131% -92.076% -90.762%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  4 (4.00%) high mild
concurrent WebSocket round trip/24
                        time:   [288.84 us 301.69 us 314.57 us]
                        change: [-90.763% -89.288% -87.411%] (p = 0.00 < 0.05)
                        Performance has improved.
  3 (3.00%) high mild
concurrent WebSocket round trip/48
                        time:   [525.04 us 553.64 us 583.15 us]
                        change: [-91.163% -90.038% -88.638%] (p = 0.00 < 0.05)
                        Performance has improved.
  3 (3.00%) high mild

```
:tada: :muscle:

src/ws/server.rs
```diff
@@ -21,476 +21,134 @@
 // SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
 // CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
 // OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR
-// IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+// IN background_task WITH THE SOFTWARE OR THE USE OR OTHER
```
dvdplm, line 24: This made me giggle!

dvdplm:
>     * Registering methods currently only permits closures with `'static` lifetime, which makes it unfortunately unsuitable for any kind of reasonable state. This needs to be narrowed down to lifetime of the server, but making guarantees here is really difficult so it might be a dead end :|.

I had a stab at this today and I failed spectacularly. Unsure if it's my Rust that is weak or if it cannot be done but it seems non-trivial for sure.
Taking a step back though I wonder if you could elaborate a bit on why this makes the architecture unsuitable for "any kind of reasonable state" – are you thinking about subscriptions here? I'm not sure I see how tying the closure lifetimes to the `Server` would be that different from `'static`. Data closed over in the closure would still need to hang around for the lifetime of the `Server` right?

src/ws/server.rs
niklasad1: Probably some search and replace nit :)

src/ws/server.rs
```diff
@@ -21,476 +21,134 @@
...
-	/// Local socket address of the transport server.
-	local_addr: SocketAddr,
-}
-
-/// Notification method that's been registered.
-pub struct RegisteredNotification {
-	/// Receives notifications that the client sent to us.
-	queries_rx: mpsc::Receiver<jsonrpc::Params>,
-}
+type Methods = FxHashMap<&'static str, Box<dyn Send + Sync + Fn(Option<&RawValue>, &str, &mpsc::UnboundedSender<String>) -> anyhow::Result<()>>>;
```
niklasad1, line 40: `Box<dyn Fn()>` here is syntactic sugar `Box<dyn Fn()  + 'static>` so we need to add explicit lifetime here for this to work.

Did you try that @dvdplm?

src/ws/server.rs
```diff
@@ -21,476 +21,134 @@
...
-				unsubscribe_method: unsubscribe_method_name,
-			})
-			.map_err(|e| Error::Internal(e.into_send_error()))?;
-
-		Ok(RegisteredSubscription { to_back: self.to_back.clone(), unique_id })
-	}
+    pub fn register_method<F, R>(&mut self, method_name: &'static str, callback: F)
+    where
+        R: Serialize,
+        F: Fn(&str) -> Result<R, RpcError> + Send + Sync + 'static, // TODO: figure out correct lifetime here
```
niklasad1, line 61: `Methods` type needs an explicit lifetime for that to work, seems like a dead-end because `fn start` calls `tokio::spawn`

src/ws/server.rs
```diff
@@ -21,476 +21,134 @@
...
+    pub async fn start(self, addr: impl AsRef<str>) -> anyhow::Result<()> {
+    	let addr = addr.as_ref();
+        let mut listener = tokio::net::TcpListener::bind(addr).await?;
+        let mut incoming = listener.incoming();
+
+        let methods = Arc::new(self.methods);
+
+        while let Some(socket) = incoming.next().await {
+            if let Ok(socket) = socket {
+                socket.set_nodelay(true).unwrap();
```
niklasad1, line 101: Is this redundant if `BufferReader/Writer` is used?

niklasad1 (commented):
good stuff here... it's just a matter to push this into something more "complete"  with subscriptions and so on.

personally, I think it's fine to ignore the `static closure TODO` and just go ahead.

niklasad1:
I will take care of the merge conflicts sorry

maciejhirsz:
Had to force push a new history to fix the merge conflicts.

ws-client/src/transport.rs
```diff
@@ -202,6 +202,9 @@ impl Receiver {
 	pub async fn next_response(&mut self) -> Result<jsonrpc::Response, WsConnectError> {
 		let mut message = Vec::new();
 		self.inner.receive_data(&mut message).await?;
+
+		println!("{:?}", std::str::from_utf8(&message));
```
niklasad1, line 206:
```suggestion
```

tests/src/helpers.rs
```diff
@@ -32,37 +32,28 @@ use std::net::SocketAddr;
...
-				let timeout = tokio::time::sleep(Duration::from_millis(100)).fuse();
-				futures::pin_mut!(hello_fut, timeout);
-				futures::select! {
-					_ = hello_fut => (),
-					_ = timeout => {
-						sub_hello.send(JsonValue::String("hello from subscription".to_owned())).await.unwrap();
-						sub_foo.send(JsonValue::Number(1337_u64.into())).await.unwrap();
-					}
-				}
+				tokio::time::sleep(Duration::from_millis(100)).await;
```
niklasad1, line 53: that is much nicer than select :)

tests/src/helpers.rs
```diff
@@ -32,37 +32,28 @@ use std::net::SocketAddr;
...
 		let rt = tokio::runtime::Runtime::new().unwrap();
 
-		let server = rt.block_on(WsServer::new("127.0.0.1:0")).unwrap();
-		let mut sub_hello =
-			server.register_subscription("subscribe_hello".to_owned(), "unsubscribe_hello".to_owned()).unwrap();
-		let mut sub_foo =
-			server.register_subscription("subscribe_foo".to_owned(), "unsubscribe_foo".to_owned()).unwrap();
-		let mut call = server.register_method("say_hello".to_owned()).unwrap();
-		server_started.send(*server.local_addr()).unwrap();
+		// let server = rt.block_on(WsServer::new("127.0.0.1:0")).unwrap();
```
niklasad1, line 40:
```suggestion
```

ws-server/src/types.rs
```diff
@@ -0,0 +1,83 @@
...
+	pub method: Cow<'a, str>,
+
+	#[serde(borrow)]
+	pub params: &'a RawValue,
+}
+
+#[derive(Serialize, Debug)]
+pub struct JsonRpcNotification<'a, T: Serialize> {
+	pub jsonrpc: TwoPointZero,
+	// pub id: Option<&'a RawValue>,
```
niklasad1, line 26:
```suggestion
```

ws-server/src/types.rs
```diff
@@ -0,0 +1,83 @@
...
+use serde_json::value::RawValue;
+use std::fmt;
+
+#[derive(Deserialize, Debug)]
+#[serde(deny_unknown_fields)]
+pub struct JsonRpcRequest<'a> {
+	pub jsonrpc: TwoPointZero,
+
+	#[serde(borrow)]
+	pub id: Option<&'a RawValue>,
```
niklasad1, line 14: technically not really to constrained to be a valid `request id` but I guess that's taken care of somewhere else to avoid allocations.

ws-server/src/server.rs
```diff
@@ -21,477 +21,201 @@
 // SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
 // CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
 // OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR
-// IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+// IN background_task WITH THE SOFTWARE OR THE USE OR OTHER
```
niklasad1, line 24:
```suggestion
// IN CONNECTION
```

ws-server/src/server.rs
```diff
@@ -21,477 +21,201 @@
...
+		T: Serialize,
+	{
+		for ((_, sub_id), sender) in self.subscribers.lock().iter() {
+			let msg = serde_json::to_string(&JsonRpcNotification {
+				jsonrpc: TwoPointZero,
+				method: self.method,
+				params: JsonRpcNotificationParams { subscription: *sub_id, result },
+			})?;
+
+			sender.send(msg)?;
```
niklasad1, line 83: hrm, this is `tokio::mpsc::UnboundedSender` and it's non-blocking?

ws-server/src/server.rs
```diff
@@ -21,477 +21,201 @@
...
 
-/// Pub-sub subscription that's been registered.
-// TODO: unregister on drop
-pub struct RegisteredSubscription {
-	/// Clone of [`Server::to_back`].
-	to_back: mpsc::UnboundedSender<FrontToBack>,
-	/// Value passed to [`FrontToBack::RegisterSubscription::unique_id`].
-	unique_id: usize,
+#[derive(Clone)]
+pub struct SubsciptionSink {
```
niklasad1, line 66:
```suggestion
pub struct SubscriptionSink {
```

ws-server/src/server.rs
```diff
@@ -21,477 +21,201 @@
...
-		if !self.registered_methods.lock().insert(method_name.clone()) {
-			return Err(Error::MethodAlreadyRegistered(method_name));
-		}
+		{
+			let subscribers = subscribers.clone();
+			self.methods.insert(
+				subscribe_method_name,
+				Box::new(move |id, _, tx, conn| {
+					let sub_id = {
+						const JS_NUM_MASK: u64 = !0 >> 11;
```
niklasad1, line 121: This is used because [JavaScrip` max number is 2^53-1](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER)?

ws-server/src/server.rs
```diff
@@ -21,477 +21,201 @@
...
 
-		self.to_back
-			.unbounded_send(FrontToBack::RegisterMethod { name: method_name, handler: tx })
-			.map_err(|e| Error::Internal(e.into_send_error()))?;
+		{
+			let subscribers = subscribers.clone();
+			self.methods.insert(
+				unsubscribe_method_name,
+				Box::new(move |id, params, tx, conn| {
+					let [sub_id]: [u64; 1] = serde_json::from_str(params)?;
```
niklasad1, line 142: should work?!

```suggestion
					let sub_id: u64 = serde_json::from_str(params)?;
```

ws-server/src/server.rs
```diff
@@ -21,477 +21,201 @@
...
-	/// Local socket address of the underlying transport server.
-	pub fn local_addr(&self) -> &SocketAddr {
-		&self.local_addr
-	}
+	pub fn register_subscription<T>(
+		&mut self,
+		subscribe_method_name: &'static str,
+		unsubscribe_method_name: &'static str,
+	) -> SubsciptionSink {
+		let subscribers = Arc::new(Mutex::new(FxHashMap::default()));
```
niklasad1, line 113: is a `Mutex` required here because the different closures are `Send + Sync`? Otherwise, I don't understand why a Mutex is required.

ws-server/src/server.rs
```diff
@@ -21,477 +21,201 @@
...
+				unsubscribe_method_name,
+				Box::new(move |id, params, tx, conn| {
+					let [sub_id]: [u64; 1] = serde_json::from_str(params)?;
+
+					subscribers.lock().remove(&(conn, sub_id));
+
+					let json = serde_json::to_string(&JsonRpcResponse {
+						jsonrpc: TwoPointZero,
+						id,
+						result: SubscriptionId::Num(0),
```
niklasad1, line 149: this is not intended to be hardcoded?

niklasad1: https://github.com/paritytech/jsonrpsee/blob/master/ws-client/src/client.rs#L399, in the client code we just send it out and don't care about the result but perhaps we should ^^

ws-server/src/server.rs
```diff
@@ -21,477 +21,201 @@
...
-
-		log::trace!(
-			"[frontend]: server register subscription: subscribe_method={}, unsubscribe_method={}",
-			subscribe_method_name,
-			unsubscribe_method_name
-		);
-		let unique_id = self.next_subscription_unique_id.fetch_add(1, atomic::Ordering::Relaxed);
+		while let Some(socket) = incoming.next().await {
+			if let Ok(socket) = socket {
+				socket.set_nodelay(true).unwrap();
```
niklasad1, line 169: Should we really panic if `set_nodelay` fails`? 
I guess that is a question on configuration, fine with ignoring this for now

ws-server/src/types.rs
maciejhirsz: Yup, if we get garbage in, we give garbage back, the id isn't really used anywhere internally (not in any map) so it doesn't really matter that we are more permissive than the spec.

The only thing that can happen is that we will return `null` on responses when request id was omitted.

ws-server/src/server.rs
maciejhirsz: It needs a `Mutex` because you you should be able to share a sink across threads (including clones), plus the closures internally need access to the map (to sub or unsub).

ws-server/src/server.rs
maciejhirsz: Won't work unfortunately, since params is a json array string, including the brackets. Can't use a single element tuple, so gotta use an array.

maciejhirsz:
Think this is good now, would leave unifying types between HTTP and WS server to another PR.

examples/examples/ws_subscription.rs
```diff
@@ -60,13 +60,16 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
 async fn run_server(server_started_tx: Sender<()>, url: &str) {
-	let server = WsServer::new(url).await.unwrap();
-	let mut subscription =
-		server.register_subscription("subscribe_hello".to_string(), "unsubscribe_hello".to_string()).unwrap();
+	let mut server = WsServer::new(url).await.unwrap();
 
-	server_started_tx.send(()).unwrap();
-	loop {
-		subscription.send(JsonValue::String("hello my friend".to_owned())).await.unwrap();
+	let mut subscription = server.register_subscription("subscribe_hello", "unsubscribe_hello");
```
niklasad1, line 65:
```suggestion
	let mut subscription = server.register_subscription("subscribe_hello", "unsubscribe_hello").unwrap();
```

examples/examples/ws.rs
```diff
@@ -52,12 +52,11 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 }
 
 async fn run_server(server_started_tx: Sender<()>, url: &str) {
-	let server = WsServer::new(url).await.unwrap();
-	let mut say_hello = server.register_method("say_hello".to_string()).unwrap();
+	let mut server = WsServer::new(url).await.unwrap();
+
+	server.register_method("say_hello", |_| Ok("lo"));
```
niklasad1, line 57:
```suggestion
	server.register_method("say_hello", |_| Ok("lo")).unwrap();
```

ws-server/src/server.rs
```diff
@@ -21,477 +21,296 @@
...
+
+use crate::types::{JsonRpcError, JsonRpcErrorParams};
+use crate::types::{JsonRpcInvalidRequest, JsonRpcRequest, JsonRpcResponse, TwoPointZero};
+use crate::types::{JsonRpcNotification, JsonRpcNotificationParams};
+
+type ConnectionId = usize;
+
+type RpcSender<'a> = &'a mpsc::UnboundedSender<String>;
+type RpcId<'a> = Option<&'a RawValue>;
+type Method = Box<dyn Send + Sync + Fn(RpcId, RpcParams, RpcSender, ConnectionId) -> anyhow::Result<()>>;
```
niklasad1, line 52: as I said before `Box<dyn Trait>` is syntactic sugar for `Box<dyn Trait + 'static>`, the proper way to deduce the `lifetime` is to assign a lifetime to it such as `type Method<'a> = Box<dyn Send + Sync + 'a>`,

Then I think we could get rid of the `'static` bound on the closures but fine for now.

ws-server/src/server.rs
```diff
@@ -21,477 +21,296 @@
...
-			return Err(Error::MethodAlreadyRegistered(method_name));
+fn send_error(id: RpcId, tx: RpcSender, code: i32, message: &str) {
+	let json = match serde_json::to_string(&JsonRpcError {
+		jsonrpc: TwoPointZero,
+		error: JsonRpcErrorParams { code, message },
+		id: id,
+	}) {
+		Ok(json) => json,
+		Err(err) => {
+			log::error!("Error serailizing error message: {:?}", err);
```
niklasad1, line 137:
```suggestion
			log::error!("Error serializing error message: {:?}", err);
```

ws-server/src/server.rs
niklasad1: ahh right, missed the params makes sense. Won't work for `Objects/Map` but I'm fine not supporting it for now.

niklasad1 (commented):
The PR is really nice and clean, I think `+432 -1602` concludes it quite well.

I would want to get the hardcoded request-id in `unsubscriptions` sorted out if that's not to complicated otherwise I'm all good with the changes.

ws-server/src/server.rs
maciejhirsz: It should work with maps too, you just need to parse it to a map type or another struct with derived `Deserialize` :)

tests/src/helpers.rs
```diff
@@ -31,57 +31,45 @@ use jsonrpsee_ws_server::WsServer;
 use std::net::SocketAddr;
 use std::time::Duration;
 
-use futures::channel::oneshot::{Receiver, Sender};
-use futures::future::FutureExt;
+use futures::channel::oneshot::Sender;
 
 pub fn websocket_server(server_started: Sender<SocketAddr>) {
```
dvdplm, line 36: Docs would be good.

tests/src/helpers.rs
```diff
@@ -31,57 +31,45 @@ use jsonrpsee_ws_server::WsServer;
...
+
+				sub_hello.send(&"hello from subscription").unwrap();
+				sub_foo.send(&1337_u64).unwrap();
 			}
 		});
 	});
 }
 
-pub fn websocket_server_with_wait_period(server_started: Sender<SocketAddr>, wait: Receiver<()>) {
+pub fn websocket_server_with_wait_period(server_started: Sender<SocketAddr>) {
```
dvdplm, line 61: This doesn't wait for anything anymore though does it? Maybe this should be named `websocket_server` and the other method `websocket_server_with_subs` or something?

tests/src/lib.rs
```diff
@@ -202,7 +201,6 @@ async fn ws_more_request_than_buffer_should_not_deadlock() {
 		}));
 	}
 
-	concurrent_tx.send(()).unwrap();
```
dvdplm, line 205: The comment on line 196 isn't correct anymore; we can use any number of requests I think. :)

ws-server/Cargo.toml
```diff
@@ -7,17 +7,23 @@ edition = "2018"
 license = "MIT"
 
 [dependencies]
+anyhow = "1.0.34"
 async-std = { version = "1.8.0", features = ["attributes"] }
+beef = "0.5"
```
dvdplm, line 12: :heart:

ws-server/src/tests.rs
```diff
@@ -126,24 +109,22 @@ async fn invalid_request_object() {
...
-	assert!(server.register_method("say_hello".to_owned()).is_err());
-	assert!(server.register_notification("notif".to_owned(), false).is_ok());
-	assert!(server.register_notification("notif".to_owned(), false).is_err());
-	assert!(server.register_subscription("subscribe_hello".to_owned(), "unsubscribe_hello".to_owned()).is_ok());
-	assert!(server.register_subscription("subscribe_hello_again".to_owned(), "notif".to_owned()).is_err());
+	let mut server = WsServer::new("127.0.0.1:0").await.unwrap();
+	assert!(server.register_method("say_hello", |_| Ok("lo")).is_ok());
+	assert!(server.register_method("say_hello", |_| Ok("lo")).is_err());
+	assert!(server.register_subscription("subscribe_hello", "unsubscribe_hello").is_ok());
+	assert!(server.register_subscription("subscribe_hello_again", "unsubscribe_hello").is_err());
```
dvdplm, line 116: Do you think it is actually useful to users to be able to provide the unsubscribe method name like this? It seems like a pretty niche feature and I fail to see why it's useful. Can we get rid of it and just prepend `un` to the subscription method name?

dvdplm (approved):
Good stuff!

tests/src/helpers.rs
niklasad1: good catch, I think this change makes the test `ws_more_request_than_buffer_should_not_deadlock` useless because the server would respond to all requests instantly instead of the client has to increase its own buffer to exceed the configured buffer size.

Or am I missing something? :)

ws-server/src/tests.rs
maciejhirsz: I defaulted on not changing the API too much here, but I also reckon explicit names are better than implicit conventions.

niklasad1 (approved):

PR #177, opened 2021-01-07
refactor: crate re-organization with separate crates

niklasad1:
The initial motivation behind this was to make an experiment to make the clients compile down to WASM (haven't finished that yet) and reduce the number of dependencies.

After discussing an eventual release of the client's "soon" makes this change useful which means that we don't have to release a version of the ongoing work on the servers.


dvdplm (approved):
LGTM. 

Do you think it makes sense to add a "façade" crate at some point that re-exports all the clients/servers or is there no need for that?

niklasad1:
> Do you think it makes sense to add a "façade" crate at some point that re-exports all the clients/servers or is there no need for that?

I don't have any objections to that, could simplify the useability of the library I suppose instead of having to import a bunch of crates.

PR #179, opened 2021-01-11
[ci]: github actions

niklasad1:
The motivation is to replace Travis CI with Github actions instead.

- Add dependabot
- Add clippy lint
- Add benchmarks (this is rather slow could maybe do this nightly matter or make it optional?)


dvdplm (approved):
I don't know anything about Github Actions but I strongly approve anyway.

(And yes, running benches once per day seems reasonable, not sure how to consume the results though?)

niklasad1:
Cool, I found https://github.com/rhysd/github-action-benchmark it will put the bench in Github Pages and can create alerts for regressions.

I will try to create a daily action for that.

insipx (approved):
Looks good overall

Just wondering why run dependabot only on the v2 branch? Are we changing dependabot branch on each release

niklasad1:
> Just wondering why run dependabot only on the v2 branch? Are we changing dependabot branch on each release

No, `master` is still the default branch (dependabot runs on the default branch by default) but it is unmaintained and all action happens in `v2` that's why. In other words, `v2` will become the new master soon

PR #180, opened 2021-01-13
merge v2 to master

niklasad1:


maciejhirsz (approved):

yjhmelody:
LGTM.
I hope this PR can continue to advance because jsonrpc is better split into different crates. I just want to use jsonrpc's codec and various checks directly and don’t want to use the network part, but jsonrpc-core cannot be well integrated into some projects

examples/subscription.rs
```diff
@@ -0,0 +1,71 @@
...
+		i += 1;
+	}
+
+	Ok(())
+}
+
+async fn run_server(server_started_tx: Sender<()>, url: &str) {
+	let server = WsServer::new(url).await.unwrap();
+	let mut subscription =
+		server.register_subscription("subscribe_hello".to_string(), "unsubscribe_hello".to_string()).unwrap();
```
yjhmelody, line 64: Is it better to rename this file to ws_subscription?  Though http1.1 don't suppport subscription.

http-server/src/server.rs
```diff
@@ -0,0 +1,311 @@
...
+	///
+	/// Clients will then be able to call this method.
+	/// The returned object allows you to handle incoming requests.
+	///
+	/// Contrary to [`register_notifications`](Server::register_notifications), there is no
+	/// `allow_losses` parameter here. If the handler is too slow to process requests, then the
+	/// server automatically returns an "internal error" to the client.
+	///
+	/// Returns an error if the method name was already registered.
+	pub fn register_method(&self, method_name: String) -> Result<RegisteredMethod, Error> {
```
yjhmelody, line 184: Whether register_method should provide generic parameters, such as T: Into<String>>. The same goes for some other similar methods

examples/ws.rs
```diff
@@ -0,0 +1,62 @@
...
+	Ok(())
+}
+
+async fn run_server(server_started_tx: Sender<()>, url: &str) {
+	let server = WsServer::new(url).await.unwrap();
+	let mut say_hello = server.register_method("say_hello".to_string()).unwrap();
+
+	server_started_tx.send(()).unwrap();
+	loop {
+		let r = say_hello.next().await;
```
yjhmelody, line 59: Provide a channel for each rpc method, I want to know if this programming model is reasonable

http-server/src/server.rs
```diff
@@ -0,0 +1,311 @@
...
+		if !self.registered_methods.lock().insert(method_name.clone()) {
+			return Err(Error::MethodAlreadyRegistered(method_name));
+		}
+
+		log::trace!("[frontend]: register_method={}", method_name);
+		let (tx, rx) = mpsc::channel(32);
```
yjhmelody, line 190: Whether similar parameters should be configurable, I think this kind of parameters may affect performance and availability

niklasad1:
@yjhmelody

Thanks for reviewing, your comments are legit but I will not address them in this PR instead I will create issues for them

niklasad1:
//cc @dvdplm needs another review ^^

dvdplm (approved):

PR #183, opened 2021-01-16
Add timeout to ws request

wuminzhe:


niklasad1 (changes requested):
Hi, thanks for your PR but the description/motivation behind this PR is missing.

I think it is rather opinionated to disregard requests that are not answered within 30 seconds, so in order to accept this PR, the timeout should be configurable such as an `Option<timeout>`.

However, it is possible for users to implement a timeout on top of the request future (similarly to what you did) so I'm a bit reluctant to accept this change. Can you please try to convince me why this is useful?

wuminzhe:
@niklasad1 I'm using subxt to interact with substrate-based chain. It's a long-running application. It often stops working because of waiting indefinitely. 
So, I think jsonrpsee needs to add a timeout to the WebSocket request. 
You are right, it should be optional and the timeout delay should be configurable.

niklasad1:
//cc @maciejhirsz any thoughts on this?

maciejhirsz (approved):

maciejhirsz:
This looks pretty reasonable to me.

niklasad1 (approved):
Thanks LGTM,

Note to us, we need to add this the `http client` for a more uniform API, feel free to do follow up PR :)

Long term, we would want his functionality in a [core client abstraction](https://github.com/paritytech/jsonrpsee/issues/193)

Issue #187, opened 2021-01-19
[CI]: benchmark github action doesn't work


niklasad1:
https://github.com/paritytech/jsonrpsee/actions/runs/494859119

niklasad1:
Now we run into https://github.com/rhysd/github-action-benchmark/issues/24 which doesn't work for scheduled tasks apparentely, the failed job can be found here: https://github.com/paritytech/jsonrpsee/actions/runs/502532462

It could make sense to remove the action for now and add it back when the issue is fixed.

PR #190, opened 2021-01-20
[ci]: fix benchmark workflow

niklasad1:
Fixes https://github.com/paritytech/jsonrpsee/issues/187

### DIFF
--- a/benches/benches/bench.rs
+++ b/benches/benches/bench.rs
@@ -44,7 +44,7 @@ pub fn http_requests(c: &mut criterion::Criterion) {
 	let server_addr = block_on(rx_addr).unwrap();
 	let client = Arc::new(HttpClient::new(&format!("http://{}", server_addr), HttpConfig::default()).unwrap());
 
-	c.bench_function("synchronous http round trip", |b| {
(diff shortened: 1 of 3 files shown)


benches/benches/bench.rs
```diff
@@ -44,7 +44,7 @@ pub fn http_requests(c: &mut criterion::Criterion) {
 	let server_addr = block_on(rx_addr).unwrap();
 	let client = Arc::new(HttpClient::new(&format!("http://{}", server_addr), HttpConfig::default()).unwrap());
 
-	c.bench_function("synchronous http round trip", |b| {
+	c.bench_function("synchronous_http_round_trip", |b| {
```
niklasad1, line 47: took me too much time to identify that https://github.com/rhysd/github-action-benchmark/blob/master/src/extract.ts#L181 doesn't support spaces in benchmark names ^^

dvdplm: <insert-joke-about-regex-and-now-you-have-two-problems/>

dvdplm (approved):

insipx (approved):

PR #196, opened 2021-01-27
Replace hashbrown with a Vec-based dense map, remove dependencies

maciejhirsz:
The internals of `BatchesState` now use a `Vec`-based dense map instead of a HashMap, which should be both simpler and faster, and it also means we don't have to pull in `hashbrown` or `fnv` for dependencies in `jsonrpsee-types` (should be easier on Wasm or any `no-std` environment).

Also replaced hashbrown with std `HashMap` in `ws-server`, since they are identical and server should be fine with std (we need system io after all anyway).


types/src/jsonrpc/wrapped/batches.rs
```diff
@@ -97,7 +93,7 @@ pub struct BatchesElem<'a, T> {
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
 pub struct BatchesElemId {
 	/// Id of the batch within `BatchesState::batches`.
-	outer: u64,
+	outer: usize,
```
maciejhirsz, line 96: I assume this isn't going to cause any problems in Wasm since the type is not serializable, and the other member is `usize` already.

niklasad1 (approved):

types/src/jsonrpc/wrapped/batches.rs
```diff
@@ -123,48 +119,50 @@ impl<T> BatchesState<T> {
 		// better performances in real-world situations. More brainstorming and benchmarking could
 		// get helpful here.
 
-		// Because of long-standing Rust lifetime issues
-		// (https://github.com/rust-lang/rust/issues/51526), we can't do this in an elegant way.
-		// If you're reading this code, know that it took several iterations and that I hated my
```
niklasad1, line 128: I love this comment :)

types/src/jsonrpc/wrapped/batches.rs
```diff
@@ -123,48 +119,50 @@ impl<T> BatchesState<T> {
...
 
 			let what_can_we_do = {
-				let (batch, _) = self.batches.get_mut(&batch_id).expect("all keys are valid; qed");
+				// Unwrap the entry, skipping `None`s
+				let (batch, _) = match entry {
+					Some(entry) => entry,
+					None => continue,
+				};
+
 				let is_ready_to_respond = batch.is_ready_to_respond();
```
dvdplm, line 136: This call seems to be non-free, and it seems like its result is only used for the `None` case below? If so, maybe this is more efficient:

```
				match batch.next() {
					None => {
						if batch.is_ready_to_respond() {
							WhatCanWeDo::ReadyToRespond
						} else {
							continue;
						}
					}
					Some(batch::BatchInc::Notification(n)) => WhatCanWeDo::Notification(n),
					Some(batch::BatchInc::Request(inner)) => WhatCanWeDo::Request(inner.id()),
				}
```

types/src/jsonrpc/wrapped/batches.rs
```diff
@@ -123,48 +119,50 @@ impl<T> BatchesState<T> {
 		// better performances in real-world situations. More brainstorming and benchmarking could
 		// get helpful here.
 
-		// Because of long-standing Rust lifetime issues
-		// (https://github.com/rust-lang/rust/issues/51526), we can't do this in an elegant way.
-		// If you're reading this code, know that it took several iterations and that I hated my
-		// life while trying to figure out how to make the compiler happy.
-		for batch_id in self.batches.keys().cloned().collect::<Vec<_>>() {
+		for (batch_id, entry) in self.batches.iter_mut().enumerate() {
```
dvdplm, line 122: Is it true that "entry" is always a request? Can we name it "request"?

types/src/jsonrpc/wrapped/batches.rs
```diff
@@ -123,48 +119,50 @@ impl<T> BatchesState<T> {
...
 				WhatCanWeDo::ReadyToRespond => {
-					let (batch, user_param) =
-						self.batches.remove(&batch_id).expect("key was grabbed from self.batches; qed");
+					// Here we take ownership of the entry, leaving `None` in its place, so
+					// we also must store the `batch_id` as being now vacant
+					self.vacant.push(batch_id);
+					let (batch, user_param) = entry.take().expect("entry is checked for `None`s above; qed");
+
 					let response =
 						batch.into_response().unwrap_or_else(|_| panic!("is_ready_to_respond returned true; qed"));
```
dvdplm, line 153: Why not `expect` here?

types/src/jsonrpc/wrapped/batches.rs
```diff
@@ -180,32 +178,18 @@ impl<T> BatchesState<T> {
 	/// Injects a newly-received batch into the list. You must then call
 	/// [`next_event`](BatchesState::next_event) in order to process it.
```
dvdplm, line 179:
```suggestion
	/// [`next_event`] in order to process it.
```

Would work right?

types/src/jsonrpc/wrapped/batches.rs
```diff
@@ -180,32 +178,18 @@ impl<T> BatchesState<T> {
 	/// Injects a newly-received batch into the list. You must then call
```
dvdplm, line 178:
```suggestion
	/// Injects a newly received batch into the list. You must then call
```

types/src/jsonrpc/wrapped/batches.rs
```diff
@@ -180,32 +178,18 @@ impl<T> BatchesState<T> {
...
-
-			match self.batches.entry(id) {
-				Entry::Occupied(_) => continue,
-				Entry::Vacant(e) => {
-					e.insert((batch, user_param));
-					break;
-				}
-			}
-		}
+		self.batches.insert(batch_id, Some((batch, user_param)));
```
dvdplm, line 184: Doesn't this shift all elements to the right? We know it is ok to clobber this index so could we not do better than this?

types/src/jsonrpc/wrapped/batches.rs
```diff
@@ -180,32 +178,18 @@ impl<T> BatchesState<T> {
...
 	}
 
 	/// Returns a list of all user data associated to active batches.
 	pub fn batches<'a>(&'a mut self) -> impl Iterator<Item = &'a mut T> + 'a {
-		self.batches.values_mut().map(|(_, user_data)| user_data)
+		self.batches.iter_mut().filter_map(|entry| match entry {
+			Some((_, user_param)) => Some(user_param),
+			None => None,
+		})
```
dvdplm, line 192: Could this be `self.batches.iter_mut().filter_map(|entry| entry.map(|(_, user_param)| user_param ))` or something similar?

dvdplm (changes requested):
Good stuff. I have questions and nitpicks.

types/src/jsonrpc/wrapped/batches.rs
```diff
@@ -50,14 +49,11 @@ use hashbrown::{hash_map::Entry, HashMap};
...
-	next_batch_id: u64,
-
 	/// For each batch, the individual batch's state and the user parameter.
-	///
-	/// The identifier is lineraly increasing and is never leaked on the wire or outside of this
-	/// module. Therefore there is no risk of hash collision.
-	batches: HashMap<u64, (batch::BatchState, T), fnv::FnvBuildHasher>,
+	batches: Vec<Option<(batch::BatchState, T)>>,
+
+	/// Vacant re-usable indices into `batches`. All indices here must point to a `None`.
```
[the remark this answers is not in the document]
maciejhirsz: Per idiomatic rust (and per actual use case here), when you remove something from a collection, you usually return the owned value stored at whatever the key is, so we replace the value with a `None` here.

types/src/jsonrpc/wrapped/batches.rs
maciejhirsz: This was me being clever, but this is actually a bug, it should either push if there are no vacant ids, or overwrite an existing `None` otherwise.

maciejhirsz: Yup, this also fixed the failing integration test!

types/src/jsonrpc/wrapped/batches.rs
```diff
@@ -86,7 +82,7 @@ pub enum BatchesEvent<'a, T> {
 /// Request within the batches.
 pub struct BatchesElem<'a, T> {
 	/// Id of the batch that contains this element.
-	batch_id: u64,
+	batch_id: usize,
 	/// Inner reference to a request within a batch.
 	inner: batch::BatchElem<'a>,
 	/// User parameter passed when calling `inject`.
```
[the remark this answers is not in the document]
maciejhirsz: I'm not sure if there is an intelligible description that can be provided here, that param can be whatever `T` the user of this wants, and I've yet to grok precisely what the intent is.

types/src/jsonrpc/wrapped/batches.rs
```diff
@@ -86,7 +82,7 @@ pub enum BatchesEvent<'a, T> {
 /// Request within the batches.
 pub struct BatchesElem<'a, T> {
```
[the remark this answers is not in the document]
maciejhirsz: Will rename to `BatchesRequest`. There is some more renaming to be done, since there is also a `batch` module that has a `BatchElem` type too, and I don't want to blow this PR too much.

types/src/jsonrpc/wrapped/batches.rs
maciejhirsz: Classic Pierre :)

types/src/jsonrpc/wrapped/batches.rs
maciejhirsz: TIL: you don't hyphenate when an adverb ends with "-ly"

types/src/jsonrpc/wrapped/batches.rs
```diff
@@ -109,8 +105,8 @@ impl<T> BatchesState<T> {
 	/// Creates a new empty `BatchesState`.
```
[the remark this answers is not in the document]
maciejhirsz: This is an implementation detail, and there is no reason why you couldn't go over 256 batches (there is no max length enforcement).

types/src/jsonrpc/wrapped/batches.rs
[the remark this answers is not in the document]
maciejhirsz: Actually scrap that, renamed the original `BatchElem` to `BatchRequest` too. The comment above here wasn't fitting after changes so reverted. Hope it all makes some sense now :).

types/src/jsonrpc/wrapped/batches.rs
maciejhirsz: `entry` is an option of a tuple of `(batch, user_param)`, not request per say. I reckon `entry` is a pretty recognizable term for a member of a collection/map.

types/src/jsonrpc/wrapped/batches.rs
maciejhirsz: I'm always down to remove boolean flag variables, since they usually are code smell :).

types/src/jsonrpc/wrapped/batches.rs
```diff
@@ -123,48 +119,50 @@ impl<T> BatchesState<T> {
 		// better performances in real-world situations. More brainstorming and benchmarking could
```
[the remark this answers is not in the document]
maciejhirsz: I'd opt for leaving it here rather than making an issue about something I don't yet fully grok.

types/src/jsonrpc/wrapped/batches.rs
```diff
@@ -86,7 +82,7 @@ pub enum BatchesEvent<'a, T> {
 /// Request within the batches.
```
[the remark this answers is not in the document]
maciejhirsz: Kept the comment as it was (explanation above).

maciejhirsz:
I think that's all re: grumbles on my end.

types/src/jsonrpc/wrapped/batches.rs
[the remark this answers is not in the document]
dvdplm: k, let's loop back when we do understand it better. :)

types/src/jsonrpc/wrapped/batches.rs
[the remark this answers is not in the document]
dvdplm: That's fair. What I was getting at was giving the reader an intuition for how heavy this is when creating them but maybe that's not super important.

types/src/jsonrpc/wrapped/batches.rs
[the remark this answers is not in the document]
dvdplm: Nah, I disagree. Comments like this one linger and if nobody cares enough to even file an issue it'll likely never ever be addressed. It's fair to take the time to grok it enough to write a proper ticket, or, alternatively, file the issue saying "Figure out what this comment entails: … …".

types/src/jsonrpc/wrapped/batches.rs
dvdplm: With the same logic we could name a variable `let int = some_call_that_returns_some_sort_of_integer()`, or `let string = get_input_from_keyboard()`. Variable names should strive to describe their purpose in the code, not what they are from a technical POV, i.e. have some kind of semantic value as opposed to just opaque markers. 
Here it's a bit tricky to come up with something good and terse, but given it comes from `self.batches` I think `batch` is decent enough (if you squint a bit).

dvdplm (approved):
Nice, good work here. Still some nitpicks but I won't make a fuss. :)

Issue #197, opened 2021-01-27
[websocket client]: use max_request_body_size in transport


niklasad1:
The intention was to use https://github.com/paritytech/jsonrpsee/blob/master/ws-client/src/client.rs#L63 to configure but it's not used by the transport client instead the default max size is configured by [soketto](https://docs.rs/soketto/0.4.2/src/soketto/connection.rs.html#19) which should fixed.

PR #198, opened 2021-01-28
[ws client]: use `max_payload_size` in transport

niklasad1:
This fixes https://github.com/paritytech/jsonrpsee/issues/197.

One downside is that if a message exceeds the limit then the client is terminated because it will be an error from Soketto,
We could have an internal guard in transport instead for this, thoughts?

TODO: maybe write a test for this :)


ws-client/src/client.rs
```diff
@@ -135,9 +134,10 @@ impl WsClient {
 	///
 	/// Fails when the URL is invalid.
 	pub async fn new(remote_addr: impl AsRef<str>, config: WsConfig) -> Result<Self, Error> {
-		let (sender, receiver) = jsonrpc_transport::websocket_connection(remote_addr.as_ref())
-			.await
-			.map_err(|e| Error::TransportError(Box::new(e)))?;
+		let (sender, receiver) =
+			jsonrpc_transport::websocket_connection(remote_addr.as_ref(), config.max_request_body_size)
```
dvdplm, line 138: Not sure I like this very much, it seems a bit hacky to add a new argument just for max payload size. What about passing the whole `WsConfig` here? Maybe down the road we'll need to use more config options, dunno, receive window size, throttling or rate limiting etc?

ws-client/src/transport.rs
```diff
@@ -63,6 +63,8 @@ pub struct WsTransportClientBuilder<'a> {
 	/// `Origin` header to pass during the HTTP handshake. If `None`, no
 	/// `Origin` header is passed.
 	origin: Option<Cow<'a, str>>,
+	/// Max payload size
+	max_payload_size: usize,
```
dvdplm, line 67: I think this is a better name and it'd be good to be consistent, so s/max_request_body_size/max_payload_size/ perhaps?

ws-client/src/transport.rs
```diff
@@ -276,7 +283,9 @@ impl<'a> WsTransportClientBuilder<'a> {
 		}
 
 		// If the handshake succeeded, return.
-		let (sender, receiver) = client.into_builder().finish();
+		let mut builder = client.into_builder();
+		builder.set_max_message_size(self.max_payload_size);
+		let (sender, receiver) = builder.finish();
 		Ok((Sender { inner: sender }, Receiver { inner: receiver }))
```
dvdplm, line 289: Do you think we should have a "constructor" of some sort for `(Sender, Receiver)`?

ws-client/src/client.rs
niklasad1: Indeed, good point. 
I'm also in favor of removing those very coupled methods to use the builder instead.

ws-client/src/transport.rs
niklasad1: It's already is part of a builder pattern but I can remove those coupled helper methods.

ws-client/src/client.rs
```diff
@@ -42,18 +42,16 @@ use std::{io, marker::PhantomData};
 /// Client that can be cloned.
 ///
 /// > **Note**: This struct is designed to be easy to use, but it works by maintaining a background
```
niklasad1, line 44:
```suggestion
/// > **Note**: This struct is designed to be easy to use, but it works by maintaining a background task running in parallel.
```

niklasad1:
After thinking about the `WsConfig` it could make sense to include `url` in it too and just provide Into<WsTransportClientBuilder`.
Instead of having `WsConfig` in `WsTransportClientBuilder` is confusing and inconsistent IMHO.

ws-client/src/client.rs
```diff
@@ -52,31 +51,56 @@ pub struct WsClient {
...
+	fn default() -> Self {
+		Self { request: 100, subscription: 4 }
+	}
+}
+
+/// Configuration.
+#[derive(Clone, Debug)]
+pub struct WsConfig {
+	/// URL to connect to.
+	pub url: String,
```
niklasad1, line 73: fix needless allocation if this is good enough ;)

ws-client/src/client.rs
```diff
@@ -52,31 +51,56 @@ pub struct WsClient {
...
 }
 
-impl Default for WsConfig {
-	fn default() -> Self {
+impl WsConfig {
+	pub fn default_settings_with_url(url: &str) -> Self {
 		Self {
-			request_channel_capacity: 100,
-			subscription_channel_capacity: 4,
+			url: url.to_string(),
```
niklasad1, line 92: fix needless allocation.

niklasad1:
OK, I refactored this PR completely to expose all settings in `WsConfig` to configure the `WebSocket transport` you can think of if as `From<WsConfig> for WebSocketTransportBuilder { ... }`

Thus, gives users complete control if they would want it such `handshake connection timeout` which wasn't that before.

Thoughts about this? If you think it's fine I can polish the needless allocations

ws-client/src/client.rs
```diff
@@ -35,15 +35,14 @@ use jsonrpsee_types::{
...
+use std::{borrow::Cow, convert::TryInto};
 use std::{io, marker::PhantomData};
 
 /// Client that can be cloned.
 ///
-/// > **Note**: This struct is designed to be easy to use, but it works by maintaining a background
-/// >           task running in parallel. If this is not desirable, you are encouraged to use the
-/// >           [`RawClient`] struct instead.
+/// > **Note**: This struct is designed to be easy to use, but it works by maintaining a background task running in parallel.
+/// >           task running in parallel.
```
dvdplm, line 45:
```suggestion
```

…or?

ws-client/src/client.rs
```diff
@@ -134,15 +158,16 @@ impl WsClient {
 	/// Initializes a new WebSocket client
 	///
 	/// Fails when the URL is invalid.
-	pub async fn new(remote_addr: impl AsRef<str>, config: WsConfig) -> Result<Self, Error> {
-		let (sender, receiver) = jsonrpc_transport::websocket_connection(remote_addr.as_ref())
+	pub async fn new(config: WsConfig) -> Result<WsClient, Error> {
+		let (sender, receiver) = jsonrpc_transport::websocket_connection(config.clone())
```
dvdplm, line 162: Maybe you can find a way to get rid of this clone?

ws-client/src/jsonrpc_transport.rs
```diff
@@ -3,11 +3,12 @@
 //! Wraps the underlying WebSocket transport with specific JSONRPC details.
 
 use crate::transport::{self, WsConnectError, WsNewDnsError};
+use crate::WsConfig;
 use jsonrpsee_types::jsonrpc;
 
 /// Creates a new JSONRPC WebSocket connection, represented as a Sender and Receiver pair.
-pub async fn websocket_connection(target: impl AsRef<str>) -> Result<(Sender, Receiver), WsNewDnsError> {
-	let (sender, receiver) = transport::websocket_connection(target.as_ref()).await?;
+pub async fn websocket_connection<'a>(config: WsConfig) -> Result<(Sender, Receiver), WsNewDnsError> {
```
dvdplm, line 10: Why was the liftetime needed here?

ws-client/src/manager.rs
```diff
@@ -1,9 +1,11 @@
 //! Handles and monitors JSONRPC v2 method calls and subscriptions
 //!
 //! Definitions:
-//!     * Request ID - request ID in JSONRPC-v2 specification
-//!     (the specs allow number, string or null but this crate only supports numbers)
-//!     * Subscription ID - unique ID generated by server
+//!
+//! 	- RequestId: request ID in JSONRPC-v2 specification
+//! 		> **Note**: The spec allow number, string or null but this crate only supports numbers.
```
dvdplm, line 6: What's up with the `>` and funny indentation here?

ws-client/src/manager.rs
```diff
@@ -1,9 +1,11 @@
 //! Handles and monitors JSONRPC v2 method calls and subscriptions
 //!
 //! Definitions:
-//!     * Request ID - request ID in JSONRPC-v2 specification
-//!     (the specs allow number, string or null but this crate only supports numbers)
-//!     * Subscription ID - unique ID generated by server
+//!
+//! 	- RequestId: request ID in JSONRPC-v2 specification
```
dvdplm, line 5:
```suggestion
//! 	- RequestId: request ID in the JSONRPC-v2 specification
```

ws-client/src/tests.rs
```diff
@@ -116,7 +116,7 @@ async fn response_with_wrong_id() {
 		WebSocketTestServer::with_hardcoded_response("127.0.0.1:0".parse().unwrap(), internal_error(Id::Num(99_u64)))
 			.await;
 	let uri = to_ws_uri_string(server.local_addr());
-	let client = WsClient::new(&uri, WsConfig::default()).await.unwrap();
+	let client = WsClient::new(WsConfig::default_settings_with_url(&uri)).await.unwrap();
```
dvdplm, line 119:
```suggestion
	let client = WsClient::new(WsConfig::default_with_url(&uri)).await.unwrap();
```

Maybe it could be shortened a bit, given it's used often?

ws-client/src/transport.rs
```diff
@@ -141,28 +144,9 @@ pub enum WsConnectError {
...
-	}
-}
-
-/// Creates a new WebSocket connection from URL, represented as a Sender and Receiver pair.
-pub async fn websocket_connection(remote_addr: impl AsRef<str>) -> Result<(Sender, Receiver), WsNewDnsError> {
-	let url =
-		url::Url::parse(remote_addr.as_ref()).map_err(|e| WsNewDnsError::Url(format!("Invalid URL: {}", e).into()))?;
+/// Creates a new WebSocket connection based on [`WsConfig`](crate::WsConfig) represented as a Sender and Receiver pair.
+pub async fn websocket_connection(config: WsConfig) -> Result<(Sender, Receiver), WsNewDnsError> {
+	let url = url::Url::parse(&config.url).map_err(|e| WsNewDnsError::Url(format!("Invalid URL: {}", e).into()))?;
```
dvdplm, line 149: Not really a DNS error, right?

ws-client/src/client.rs
```diff
@@ -280,7 +305,7 @@ async fn background_task(
 	mut sender: jsonrpc_transport::Sender,
 	receiver: jsonrpc_transport::Receiver,
 	mut frontend: mpsc::Receiver<FrontToBack>,
-	config: WsConfig,
+	channel_capacities: ChannelCapacities,
```
dvdplm, line 308: Maybe `ChannelCapacities` could be `ChannelConfig`? Or they could be in-lined directly here perhaps?

dvdplm (commented):
I like the direction here, I left a few nitpicks and suggestions but overall it looks good. Let me know when I should review "for real".

ws-client/src/jsonrpc_transport.rs
niklasad1: It was a leftover/nit when I experimented with `&'a str` in the WebSocketConfig but I will cleanup the code and let you know when it's ready for another round of reviews.

ws-client/src/manager.rs
niklasad1: It looked very bad in the generated docs....

ws-client/src/tests.rs
maciejhirsz: Reading it I had the same suggestion in mind, partly to make it shorter, partly to not mix language here (config | settings)

niklasad1: ok, what about `with_url`?

ws-client/src/client.rs
maciejhirsz: Using `channel` in name is leaking abstraction, I reckon `concurrent|cy_limits` would be better here.

In general though, I don't see why this needs to be a separate struct at all, I'd just have separate fields.

ws-client/src/client.rs
```diff
@@ -134,15 +157,16 @@ impl WsClient {
...
-	pub async fn new(remote_addr: impl AsRef<str>, config: WsConfig) -> Result<Self, Error> {
-		let (sender, receiver) = jsonrpc_transport::websocket_connection(remote_addr.as_ref())
+	pub async fn new(config: WsConfig) -> Result<WsClient, Error> {
+		let (sender, receiver) = jsonrpc_transport::websocket_connection(config.clone())
 			.await
 			.map_err(|e| Error::TransportError(Box::new(e)))?;
 
-		let (to_back, from_front) = mpsc::channel(config.request_channel_capacity);
+		let channel_capacities = config.channel_capacities;
+		let (to_back, from_front) = mpsc::channel(channel_capacities.request);
```
maciejhirsz, line 166: What happens when the capacity is set to `0`? Should we force the config to be `NonZeroUsize`, or default to either unbound channel (I guess that will require some extra generic mess here) or `usize::MAX`.

ws-client/src/client.rs
niklasad1: The intention was not to pass/clone to entire `Config` struct to the background thread when having `non-static` borrows, the background thread requires `static`.  However, we could just pass the `subscription capacity` and keep all fields in `Config` instead.

ws-client/src/tests.rs
dvdplm: I like `with_url`.

ws-client/src/client.rs
dvdplm: >  I'd just have separate fields.

I like this. If cloning the config ever becomes a performance problem, maybe we have bigger problems?

ws-client/src/client.rs
niklasad1: That's a great question, according to the docs

> The channel's capacity is equal to buffer + num-senders.

So, in that case, it will be `1` which I think is fine. I was thinking about the `NonZero` before but it makes the API quite annoying to use. Thoughts?

ws-client/src/transport.rs
niklasad1: Renamed the error to `WsHandshakeError`

ws-client/src/client.rs
```diff
@@ -35,48 +35,65 @@ use jsonrpsee_types::{
...
 		Self {
-			request_channel_capacity: 100,
-			subscription_channel_capacity: 4,
+			url,
 			max_request_body_size: 10 * 1024 * 1024,
 			request_timeout: None,
+			connection_timeout: Duration::from_secs(10),
+			origin: None,
+			handshake_url: From::from("/"),
+			max_concurrent_requests_capacity: 100,
```
niklasad1, line 90: maybe decrease this to something like 10?

ws-client/src/client.rs
maciejhirsz: Ah, should be fine then. I'd just note in doc comment that the actual limit is 1 + this.

ws-client/src/client.rs
maciejhirsz: 10 sounds pretty low. What's the `size_of` the messages we are passing?

niklasad1: Yeah, your hunch is correct. 
The message size is 96 bytes.

So 10 * 96  is quite reasonable, but we could to `10 * + std::mem::size_of::<FrontToBack>()` to future proof it.

ws-client/src/transport.rs
```diff
@@ -233,7 +228,6 @@ impl<'a> WsTransportClientBuilder<'a> {
 	/// Sets the timeout to use when establishing the TCP connection.
 	///
 	/// The default timeout is 10 seconds.
-	// TODO: design decision: should the timeout not be handled by the user?
```
niklasad1, line 236: Removed this comment because it's exposed to the user but these methods are not used. It's exposed via the `WsConfig`.

ws-client/src/client.rs
```diff
@@ -134,17 +154,18 @@ impl WsClient {
...
-		let (to_back, from_front) = mpsc::channel(config.request_channel_capacity);
+		let (sender, receiver) =
+			jsonrpc_transport::websocket_connection(config).await.map_err(|e| Error::TransportError(Box::new(e)))?;
 
 		async_std::task::spawn(async move {
-			background_task(sender, receiver, from_front, config).await;
+			background_task(sender, receiver, from_front, max_capacity_per_subscription).await;
 		});
-		Ok(Self { to_back, config })
+		Ok(Self { to_back, request_timeout })
```
niklasad1, line 168: removed `Config` from the client, ideally, everything should be handled by the `transport` but we have two exceptions: background thread and timeouts.

niklasad1:
@dvdplm @maciejhirsz I think it's ready for another round of review.

I removed the `config` from the `WsClient` which is only used by the transport with two exceptions:
  - `subscription capacity per subscription`, in the background thread
  - `request timeout`, in the client to terminate unanswered requests.

The request timeout makes the overall structure a little bit messy (code structure and maintainability). Ideally, I could change it back to pass Config everywhere instead but then the `Config` needs either to take `URL` as String or `& 'static str` to compile.

Alternatively, have `timeout` as a param to `fn request` instead but I prefer not because most users will probably not want to have a timeout for their request. Maybe `fn request_with_timeout`?

ws-client/src/manager.rs
niklasad1: my bad I used tabs in `doc comments`

ws-client/src/client.rs
maciejhirsz: Wait, no, that doesn't make any sense, the channel capacity is in messages, not in bytes :upside_down_face: 

I'm asking what the size of the message is to make sure the buffer isn't going to cause memory issues. If we set the default to be something like 256, then (with current shape of the message) that buffer will allocate some 24kb, which should be fine.

niklasad1: lol, I got confused thanks.

ws-client/src/client.rs
```diff
@@ -35,48 +35,68 @@ use jsonrpsee_types::{
...
+	/// Url to send during the HTTP handshake.
+	pub handshake_url: Cow<'a, str>,
+	/// Max concurrent request capacity.
+	///
+	/// **Note**: The actual capacity is `num_senders + max_concurrent_requests_capacity`
+	/// because it is passed to [`futures::mpsc::channel`](futures::channel::mpsc::channel)
+	/// and the capacity may increase because the sender is cloned when new
+	/// requests, notifications and subscriptions are created.
+	pub max_concurrent_requests_capacity: usize,
+	/// Max concurrent capacity for each subscription which might be dropped if the capacity is exceeded.
```
dvdplm, line 76: Something's odd with this phrase after "which". Do you mean to say "Max concurrent capacity for each subscription; when exceeded messages will be dropped."?

ws-client/src/client.rs
```diff
@@ -35,48 +35,68 @@ use jsonrpsee_types::{
...
+	///
+	/// **Note**: The actual capacity is `num_senders + max_subscription_capacity`
+	/// because it is passed to [`futures::mpsc::channel`](futures::channel::mpsc::channel).
```
dvdplm, line 79: Would just ```[`futures::mpsc::channel`]``` work here?

ws-client/src/client.rs
```diff
@@ -35,48 +35,68 @@ use jsonrpsee_types::{
...
+	pub max_subscription_capacity: usize,
 }
 
-impl Default for WsConfig {
-	fn default() -> Self {
+impl<'a> WsConfig<'a> {
+	/// Default `WebSocket` configuration with a specified URL to connect to.
```
dvdplm, line 84: Do we have a type called `WebSocket`?

dvdplm (approved):
Works for me.

ws-client/src/client.rs
niklasad1: I can change it `futures::channel::mpsc::channel` instead but `futures::mpsc::channel` won't work

ws-client/src/client.rs
niklasad1: Removed the "tics", ok?

ws-client/src/client.rs
niklasad1: I updated the documentation of `WsConfig` struct, ok?

maciejhirsz (approved):

PR #199, opened 2021-01-29
feat: client trait + porting client proc macros

### LINKED ISSUE #104: Unused type parameters when there is no function with a return value
Version: 0.1.0 (from crates.io)

I encountered an issue while trying to make a very small API for a test:

```rust
jsonrpsee::rpc_api! {
	Registrar {
		fn register_para(foo: i32, bar: String);
	}
}
```

I did this and it was failing with:

```
error[E0282]: type annotations needed
   --> test/parachain/tests/integration_test.rs:154:3
    |
154 |         Registrar::register_para(&mut client, 100, "{\"scheduling\":\"Always\"}").await.unwrap()
    |         ^^^^^^^^^^^^^^^^^^^^^^^^ cannot infer type for type parameter `R` declared on the enum `Registrar`
```

(with more logs you can see that the type parameter `I` is also unused)

I expanded the code and this is what I get:

```rust
enum Registrar<'a, R, I> {
    RegisterPara { foo: i32, bar: String },
}
```

I changed it to:

```rust
jsonrpsee::rpc_api! {
	Registrar {
		#[rpc(method = "tx.registrar.registerPara")]
		fn register_para(foo: i32, bar: String) -> String; // <---- return value here
	}
}
```

And now R and I are used:

```rust
enum Registrar<'a, R, I> {
    RegisterPara {
        respond: jsonrpsee::raw::server::TypedResponder<'a, R, I, String>,
        foo: i32,
        bar: String,
    },
}
```

niklasad1:
This PR introduces a client trait to support to `jsonrpsee_proc_macros` again and refactor the macros to the new client design.
As a bonus it closes #104

Not addressed in this PR:
- subscriptions in `proc macros`
- the trait could use `serde raw value` with Maciej's design instead of `Into<String>, Into<Params`

### Annoyance with the macros

FYI,
One annoyance that I found is how the enums are generated by the `proc macros` which is that all type params on the methods are bound to specific enum variant, to illustrate this:

```rust
jsonrpsee_proc_macros::rpc_client_api! {
	Health {
		fn say_hello<T, B>(t: T) -> B;
	}
}
```

expands to 

```rust
// this won't compile because the type params are not bound
// needs to be enum Health<T, B> { ... }
enum Health {
    SayHello {
        t: T,
    },
    #[allow(unused)]
    _0: B,
}
```

As you can see the entire enum needs to have that type param to compile so a little annoying. So one have to do this instead:

```rust
jsonrpsee_proc_macros::rpc_client_api! {
	Health<T, B> {
		fn say_hello(t: T) -> B;
	}
}
```

One solution,  could be collect all generic params and diff it with `type generics` on the enum and add the missing params manually.


http-client/src/client.rs
```diff
@@ -26,32 +30,33 @@ impl HttpClient {
...
+impl Client for HttpClient {
+	type Error = Error;
+	type Subscription = ();
+
+	fn notification<'a>(
+		&'a self,
 		method: impl Into<String>,
 		params: impl Into<jsonrpc::Params>,
-	) -> Result<(), Error> {
+	) -> Pin<Box<dyn Future<Output = Result<(), Self::Error>> + Send + 'a>> {
```
maciejhirsz, line 43: couldn't we avoid this with [`async-trait`](https://crates.io/crates/async-trait)

niklasad1: yes :+1:

proc-macros/src/lib.rs
```diff
@@ -89,255 +57,21 @@ pub fn rpc_api(input_token_stream: TokenStream) -> TokenStream {
...
 
 	Ok(quote_spanned!(api.name.span()=>
-		#visibility enum #enum_name #tweaked_generics {
-			#(#variants),*
-		}
-
-		impl #impl_generics #enum_name #ty_generics #where_clause {
-			#next_request
-		}
+		#visibility struct #enum_name #tweaked_generics;
```
niklasad1, line 70: this won't work for generics,

we would want to something similar to 

```rust
#visibility struct #enum_name #tweaked_generics(core::marker::PhantomData<(R, T, .... N)>);
```

`R, T, ... N` is in `#tweaked_generics` but is `<R, T>` so ideally we would convert it ti `<(R, N)>`

niklasad1: Might be easier to just use the old code to use an enum where each type param is bound to a specific variant.

proc-macros/src/lib.rs
```diff
@@ -89,37 +57,22 @@ pub fn rpc_api(input_token_stream: TokenStream) -> TokenStream {
...
 	let mut variants = Vec::new();
-	let mut tmp_variants = Vec::new();
 	for function in &api.definitions {
-		let function_is_notification = function.is_void_ret_type();
 		let variant_name = snake_case_to_camel_case(&function.signature.ident);
-		let ret = match &function.signature.output {
-			syn::ReturnType::Default => quote! {()},
-			syn::ReturnType::Type(_, ty) => quote_spanned!(ty.span()=> #ty),
+		if let syn::ReturnType::Type(_, ty) = &function.signature.output {
+			non_used_type_params.insert(ty);
```
niklasad1, line 71: return types are not bound, so handle them differently in a separate enum variant

such as: `Foo { _0: T, _1: U, .... }` I'm no proc macro expert but I realized that this may make the enum size bigger if this variant grows huge :)

types/src/traits.rs
```diff
@@ -0,0 +1,44 @@
+use crate::client::Subscription;
+use crate::error::Error;
+use crate::jsonrpc::{DeserializeOwned, Params};
+use alloc::string::String;
+use async_trait::async_trait;
+use core::fmt;
+
+/// Basic `JSONRPC` client that can make requests and notifications.
```
niklasad1, line 8:
```suggestion
/// Basic `JSONRPC` client that can make requests, notifications and subscriptions.
```

types/src/traits.rs
```diff
@@ -0,0 +1,47 @@
...
+		subscribe_method: SM,
+		params: P,
+		unsubscribe_method: UM,
+	) -> Result<Subscription<N>, Error>
+	where
+		SM: Into<String> + Send,
+		UM: Into<String> + Send,
+		P: Into<Params> + Send,
+		N: DeserializeOwned,
+		Self: Sized;
```
niklasad1, line 46: revert `Self::Sized` this trait is useless for trait objects.

types/src/traits.rs
```diff
@@ -0,0 +1,40 @@
+use crate::client::Subscription;
+use crate::error::Error;
+use crate::jsonrpc::{DeserializeOwned, Params};
+use alloc::string::String;
+use async_trait::async_trait;
+
+/// Basic `JSONRPC` client that can make requests, notifications and subscriptions.
+#[async_trait]
+pub trait Client {
```
niklasad1, line 9: one annoyance is that this trait can't be used as a trait object because of the generics (e.g, tried porting this to subxt)

we could change everything to concrete types and return `JsonValue`  instead of `T: DeserializeOwned `but it's really inelegant IMHO.

types/src/traits.rs
```diff
@@ -0,0 +1,40 @@
...
+		T: DeserializeOwned,
+		M: Into<String> + Send,
+		P: Into<Params> + Send;
+
+	/// Send a subscription request to the server.
+	///
+	/// The `subscribe_method` and `params` are used to ask for the subscription towards the
+	/// server. The `unsubscribe_method` is used to close the subscription.
+	//
+	// TODO: ideally this should be a subtrait but let's have it to simplify macro stuff for now.
```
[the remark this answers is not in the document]
niklasad1: yeah

http-client/src/client.rs
```diff
@@ -26,15 +30,15 @@ impl HttpClient {
...
-	///
-	/// WARNING: This method must be executed on [Tokio 1.0](https://docs.rs/tokio/1.0.1/tokio).
-	pub async fn notification(
-		&self,
-		method: impl Into<String>,
-		params: impl Into<jsonrpc::Params>,
-	) -> Result<(), Error> {
+#[async_trait]
+impl Client for HttpClient {
+	async fn notification<M, P>(&self, method: M, params: P) -> Result<(), Error>
```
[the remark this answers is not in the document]
niklasad1: I'm not aware of it.

http-client/src/client.rs
```diff
@@ -26,15 +30,15 @@ impl HttpClient {
...
+	where
+		M: Into<String> + Send,
+		P: Into<jsonrpc::Params> + Send,
+	{
 		let request = jsonrpc::Request::Single(jsonrpc::Call::Notification(jsonrpc::Notification {
```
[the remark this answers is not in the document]
niklasad1: https://github.com/paritytech/jsonrpsee/issues/213

types/src/traits.rs
```diff
@@ -0,0 +1,40 @@
+use crate::client::Subscription;
+use crate::error::Error;
+use crate::jsonrpc::{DeserializeOwned, Params};
+use alloc::string::String;
+use async_trait::async_trait;
+
+/// Basic `JSONRPC` client that can make requests, notifications and subscriptions.
+#[async_trait]
+pub trait Client {
+	/// Send a notification request.
```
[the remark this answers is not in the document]
niklasad1: Added a link to the `JSONRPC spec` instead, ok?

types/src/traits.rs
```diff
@@ -0,0 +1,40 @@
...
+		&self,
+		subscribe_method: SM,
+		params: P,
+		unsubscribe_method: UM,
+	) -> Result<Subscription<N>, Error>
+	where
+		SM: Into<String> + Send,
+		UM: Into<String> + Send,
+		P: Into<Params> + Send,
+		N: DeserializeOwned;
```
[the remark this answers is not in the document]
niklasad1: Updated the documentation for it ok?

http-client/src/client.rs
```diff
@@ -45,13 +49,12 @@ impl HttpClient {
...
-		params: impl Into<jsonrpc::Params>,
-	) -> Result<JsonValue, Error> {
+	async fn request<T, M, P>(&self, method: M, params: P) -> Result<T, Error>
+	where
+		T: DeserializeOwned,
+		M: Into<String> + Send,
+		P: Into<jsonrpc::Params> + Send,
+	{
 		// NOTE: `fetch_add` wraps on overflow which is intended.
 		let id = self.request_id.fetch_add(1, Ordering::SeqCst);
```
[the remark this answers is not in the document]
niklasad1: Probably not, I just enabled the safest option without further investigation.

I think using it the same request ID twice is probably fine too.

niklasad1:
@dvdplm @maciejhirsz This is ready for another round of review.

I splitted up the client trait into client (request, notif) and subscribeclient (client + subscription) and documented the proc macro stuff

maciejhirsz (approved):
Looks pretty clean to me!

types/src/client.rs
```diff
@@ -0,0 +1,91 @@
...
+	/// Send a subscription request to the server.
+	Subscribe {
+		/// Method for the subscription request.
+		subscribe_method: String,
+		/// Parameters to send for the subscription.
+		params: Params,
+		/// Method to use to later unsubscription. Used if the channel unexpectedly closes.
+		unsubscribe_method: String,
+		/// When we get a response from the server about that subscription, we send the result on
+		/// this channel. If the subscription succeeds, we return a `Receiver` that will receive
```
dvdplm, line 48:
```suggestion
		/// this channel. If the subscription succeeds, we return a [`mpsc::Receiver`] that will receive
```

types/src/client.rs
```diff
@@ -0,0 +1,91 @@
...
+		/// notifications.
+		send_back: oneshot::Sender<Result<(mpsc::Receiver<JsonValue>, SubscriptionId), Error>>,
+	},
+
+	/// When a subscription channel is closed, we send this message to the background
+	/// task to mark it ready for garbage collection.
+	// NOTE: It is not possible to cancel pending subscriptions or pending requests.
+	// Such operations will be blocked until a response is received or the background
+	// thread has been terminated.
```
dvdplm, line 57: Is this a possible DoS attack vector? I guess not, a malicious user can saturate their ws connection with subscription requests but not impact other users' connection right?

types/src/client.rs
```diff
@@ -0,0 +1,91 @@
...
+	SubscriptionClosed(SubscriptionId),
+}
+
+impl<Notif> Subscription<Notif>
+where
+	Notif: DeserializeOwned,
+{
+	/// Returns the next notification from the stream
+	/// This may return `None` if the subscription has been terminated, may happen if the channel becomes full or dropped.
```
dvdplm, line 66: long line

dvdplm, line 66:
```suggestion
	/// This may return `None` if the subscription has been terminated, may happen if the channel becomes full or  is dropped.
```

dvdplm (approved):
Agree with Maciej that this looks pretty clean. Left a few nitpicks.

types/src/client.rs
niklasad1: yeah, it will send a message to background thread which in turn makes a request to the server but it's up to the server to handle the request and connection. If the connection gets closed by the server the user has to restart the client to get it working again.

So nothing really related to this, from a malicious users point-of-view it is better to make requests to DoS a server but it might be possible to make the client less responsive/hung by continuously spawn new subscriptions and close them directly.

PR #201, opened 2021-02-05
feat(http client): support tokio02 via hyper 0.13

### LINKED ISSUE #200: [http client]: provide backward compat layer for tokio 0.x
For a while ago we updated the http client to use `hyper 0.14` which may create annoyance for users that is using on tokio 0.x/async-std because the async methods such as `request, notification and subscription` needs to be spawned directly on tokio 1.0 to work because hyper is used and can only be polled through tokio. Thus, if we could provide a compatibility for hyper/tokio that would help alot such as provide optional features for hyper0.13/tokio02, hyper14/tokio1.

The current workaround for this: is to spawn another thread in which a separate tokio 1.0 runtime is initialized, [before jsonrpsee did that for them](https://github.com/paritytech/jsonrpsee/blob/040c4b286ea273e0e0f05ac1363d5e9665e88dd3/src/transport/http/client.rs#L73-#L86) but the background thread is removed for the http client and those methods are exposed directly without the additional threads

niklasad1:
Conditional compilation mess to support `tokio 0.2` and `tokio 1.0` via feature flags,  closes #200.
Tested successfully locally, but I don't prefer to port all tests for both runtimes.

FYI, `hyper` doesn't support `tokio 0.3` that's why it's not supported and  `tokio 0.1` doesn't support `async await` which doesn't make sense for this library.

### DIFF
--- a/http-client/src/lib.rs
+++ b/http-client/src/lib.rs
@@ -1,3 +1,12 @@
+#[cfg(not(any(feature = "tokio1", feature = "tokio02")))]
+compile_error!("Either feature \"tokio1\" or \"tokio02\" must be enabled for this crate.");
+
+#[cfg(feature = "tokio1")]
+extern crate hyper14 as hyper;
+
+#[cfg(feature = "tokio02")]
+extern crate hyper13 as hyper;
+
 mod client;
 mod transport;
 
--- a/http-client/src/lib.rs
+++ b/http-client/src/lib.rs
@@ -1,8 +1,8 @@
 #[cfg(all(feature = "tokio1", feature = "tokio02"))]
-compile_error!("feature \"tokio1\" and \"tokio02\" are mutably exclusive");
+compile_error!("feature `tokio1` and `tokio02` are mutably exclusive");
 
 #[cfg(not(any(feature = "tokio1", feature = "tokio02")))]
-compile_error!("feature \"tokio1\" or \"tokio02\" must be enabled for this crate");
+compile_error!("feature `tokio1` or `tokio02` must be enabled for this crate");
 
 #[cfg(all(feature = "tokio1", not(feature = "tokio02")))]
 extern crate hyper14 as hyper;
--- a/http-client/src/lib.rs
+++ b/http-client/src/lib.rs
@@ -1,10 +1,13 @@
-#[cfg(not(any(feature = "tokio1", feature = "tokio02")))]
-compile_error!("Either feature \"tokio1\" or \"tokio02\" must be enabled for this crate.");
+#[cfg(all(feature = "tokio1", feature = "tokio02"))]
+compile_error!("feature \"tokio1\" and \"tokio02\" are mutably exclusive");
 
-#[cfg(feature = "tokio1")]
+#[cfg(not(any(feature = "tokio1", feature = "tokio02")))]
+compile_error!("feature \"tokio1\" or \"tokio02\" must be enabled for this crate");
+
+#[cfg(all(feature = "tokio1", not(feature = "tokio02")))]
 extern crate hyper14 as hyper;
 
-#[cfg(feature = "tokio02")]
+#[cfg(all(feature = "tokio02", not(feature = "tokio1")))]
 extern crate hyper13 as hyper;
 
 mod client;
--- a/utils/src/http/mod.rs
+++ b/utils/src/http/mod.rs
@@ -26,10 +26,13 @@
 
 //! HTTP Server utilities for the `jsonrpsee` library
 
+#[cfg(any(feature = "tokio1", feature = "tokio02"))]
 pub mod access_control;
+#[cfg(any(feature = "tokio1", feature = "tokio02"))]
+pub mod hyper_helpers;
+
 pub mod cors;
 pub mod hosts;
-pub mod hyper_helpers;
 
 mod matcher;
 
--- a/utils/src/http/mod.rs
+++ b/utils/src/http/mod.rs
@@ -26,13 +26,10 @@
 
 //! HTTP Server utilities for the `jsonrpsee` library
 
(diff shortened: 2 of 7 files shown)


utils/src/http/mod.rs
```diff
@@ -26,10 +26,13 @@
 
 //! HTTP Server utilities for the `jsonrpsee` library
 
+#[cfg(any(feature = "tokio1", feature = "tokio02"))]
 pub mod access_control;
+#[cfg(any(feature = "tokio1", feature = "tokio02"))]
+pub mod hyper_helpers;
```
dvdplm, line 32: dq: why do we need this? Can't we assume either one of them is present and just do without `[cfg(…)]` here?

http-client/src/lib.rs
```diff
@@ -1,3 +1,12 @@
+#[cfg(not(any(feature = "tokio1", feature = "tokio02")))]
+compile_error!("Either feature \"tokio1\" or \"tokio02\" must be enabled for this crate.");
+
+#[cfg(feature = "tokio1")]
+extern crate hyper14 as hyper;
```
dvdplm, line 5: Is this here so we can do conditional crate renaming?

http-client/src/lib.rs
```diff
@@ -1,3 +1,12 @@
...
+
+#[cfg(feature = "tokio02")]
```
maciejhirsz, line 7:
```suggestion
#[cfg(all(feature = "tokio02", not(feature = "tokio1")))]
```

Typing from the top of my head so not sure if that's 100% correct, but it should avoid weird errors in case both feature flags are enabled.

Could even go as far as add another one for when both are enabled with some more human readable `compile_error!`

niklasad1: I think an additional `compile_error!` is good, users will get some weird error `hyper` not found.

http-client/src/lib.rs
niklasad1: indeed, do have any better suggestion?

utils/src/http/mod.rs
niklasad1: in practice yes, these are helpers that use `hyper` under the hood.

niklasad1:
superseded by #204

dvdplm (approved):

http-client/src/lib.rs
kianenigma: so something like `use X as Y` won't work with a `#[cfg()]` atop?

kianenigma (approved):

http-client/src/lib.rs
```diff
@@ -1,3 +1,15 @@
+#[cfg(all(feature = "tokio1", feature = "tokio02"))]
+compile_error!("feature \"tokio1\" and \"tokio02\" are mutably exclusive");
```
dvdplm, line 2:
```suggestion
compile_error!("""feature "tokio1" and "tokio02" are mutably exclusive""");
```

http-client/src/lib.rs
```diff
@@ -1,3 +1,15 @@
...
+
+#[cfg(not(any(feature = "tokio1", feature = "tokio02")))]
+compile_error!("feature \"tokio1\" or \"tokio02\" must be enabled for this crate");
```
dvdplm, line 5:
```suggestion
compile_error!("""feature "tokio1" or "tokio02" must be enabled for this crate""");
```

http-client/src/lib.rs
dvdplm: > so something like `use X as Y` won't work with a `#[cfg()]` atop?

Nope, I just tried. Compiler gets super angry. Weird.

utils/src/lib.rs
```diff
@@ -1 +1,13 @@
+#[cfg(all(feature = "hyper13", feature = "hyper14"))]
+compile_error!("feature \"hyper13\" and \"hyper14\" are mutably exclusive");
```
dvdplm, line 2: Same as above, using `"""stuff "dsd" ddsas"""` is a bit nicer.

utils/src/lib.rs
```diff
@@ -1 +1,13 @@
...
+
+#[cfg(not(any(feature = "hyper13", feature = "hyper14")))]
+compile_error!("feature \"hyper13\" or \"hyper14\" must be enabled for this crate");
```
dvdplm, line 5: Same as above, using `"""stuff "dsd" ddsas"""` is a bit nicer.

dvdplm (approved):
lgtm modulo some nits

http-client/src/lib.rs
niklasad1: it doesn't work properly the error message becomes:

`error: compile_error! takes 1 argument
 --> http-client/src/lib.rs:5:1
  |
5 | compile_error!("""feature "tokio1" or "tokio02" must be enabled for this crate""");
`

http-client/src/lib.rs
niklasad1: yeah, we don't have any crate named `hyper` in the dependency tree anymore.

I guess another solution is that move the `#[cfg()]` in to the struct `HttpTransportClient`

http-client/src/lib.rs
dvdplm: hahah, ok, sorry for the noise then. I did run `cargo check` but I guess that's not enough.

niklasad1: right, you have to do `cargo check  -p jsonrpsee-http-client --no-default-features` or `cargo check  -p jsonrpsee-http-client --features tokio02`

Because `tokio1` is enabled by default, feature flags life you know.

niklasad1: We could do compile_error!("feature `tokio1` or `tokio02` must be enabled for this crate"); instead if you don't like the escaping (`\"`)

maciejhirsz (approved):

PR #204, opened 2021-02-10
[http client]: switch to surf instead of hyper

### LINKED ISSUE #200: [http client]: provide backward compat layer for tokio 0.x

niklasad1:
This PR removes the direct dependency to `hyper` and replaces it with `surf`.

The benefits of surf is:
1. supports multiple backends: `wasm`, `curl` and `hyper`
2. curl works for both `async-std` and `tokio`
3. makes the `HTTP client` compile for WASM.

The drawback is that surf doesn't support `tokio 1.0` via hyper yet.

So for now, `curl` is enabled by default which should work on any runtime but is significantly slower than `hyper` from my naive benchmarks.
Then additional features are introduced to enable  `wasm http client` or `hyper client`, those are mutably exclusive

Note: the hyper feature does only work on `tokio 0.2`, so users should be aware of that before enabling that feature.

Closes #200

### DIFF
--- a/http-client/src/transport.rs
+++ b/http-client/src/transport.rs
@@ -6,8 +6,9 @@
 // that we need to be guaranteed that hyper doesn't re-use an existing connection if we ever reset
 // the JSON-RPC request id to a value that might have already been used.
 
+use futures::prelude::*;
 use jsonrpsee_types::{error::GenericTransportError, http::HttpConfig, jsonrpc};
-use jsonrpsee_utils::http::hyper_helpers;
+use surf::http::{mime::JSON, Url};
 use thiserror::Error;
 
 const CONTENT_TYPE_JSON: &str = "application/json";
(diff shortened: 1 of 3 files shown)


http-client/src/transport.rs
```diff
@@ -69,13 +66,22 @@ impl HttpTransportClient {
...
-		let body = hyper_helpers::read_response_to_body(&parts.headers, body, self.config).await?;
+		let mut response = self.send_request(request).await.map_err(|e| Error::Http(Box::new(e)))?;
+
+		let length = response.len().unwrap_or(0);
+
+		if length > self.config.max_request_body_size as usize {
+			return Err(Error::RequestTooLarge.into());
+		}
+
+		let mut buffer = Vec::new();
```
maciejhirsz, line 77:
```suggestion
		let mut buffer = Vec::with_capacity(length);
```
?

maciejhirsz (approved):

insipx (approved):

PR #207, opened 2021-02-11
[http client]: support HTTPS

### LINKED ISSUE #189: [http client]: HTTPS support
It would be nice to support `HTTPS` too.

I guess we have two options either check the `URL` if it starts with `https` or provide a feature flag to enable `https (hyper_tls)`

niklasad1:
Closes #189, `HTTPS` is supported by surf which makes this quite simple to implement. The bad thing is that users can't disable it.

I found a faulty unwrap where we should return an error instead, which this PR fixes too but [surf::Error](https://github.com/http-rs/http-types/issues/88) doesn't implement `std::errror::Error`.Thus, I decided to switch to `anyhow` instead `Box<dyn std::error::Error + Send + Sync>`

### DIFF
--- a/http-client/src/transport.rs
+++ b/http-client/src/transport.rs
@@ -6,8 +6,9 @@
 // that we need to be guaranteed that hyper doesn't re-use an existing connection if we ever reset
 // the JSON-RPC request id to a value that might have already been used.
 
+use anyhow::anyhow;
 use futures::prelude::*;
-use jsonrpsee_types::{error::GenericTransportError, http::HttpConfig, jsonrpc};
+use jsonrpsee_types::{http::HttpConfig, jsonrpc};
 use surf::http::{mime::JSON, Url};
 use thiserror::Error;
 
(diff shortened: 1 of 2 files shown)


http-client/src/transport.rs
```diff
@@ -167,4 +156,17 @@ mod tests {
 		let response = client.send_request(request).await.unwrap_err();
 		assert!(matches!(response, Error::RequestTooLarge));
 	}
+
+	#[tokio::test]
```
niklasad1, line 160: maybe a little bit heavy for a test

niklasad1: I can remove it but just to demonstrate that it works :)

maciejhirsz: I'd move it to the separate `tests`, it's more of integration test than unit test.

maciejhirsz (approved):

niklasad1:
Closing this as #204 is getting reverted.

PR #209, opened 2021-02-12
[ws client]: add `wss` test + refactor `WebSocketTransport builder`

niklasad1:


ws-client/src/transport.rs
```diff
@@ -163,8 +163,8 @@ pub async fn websocket_connection(config: WsConfig<'_>) -> Result<(Sender, Recei
 	for sockaddr in target.to_socket_addrs().await.map_err(WsHandshakeError::ResolutionFailed)? {
 		let builder = WsTransportClientBuilder {
 			target: sockaddr,
-			host: host.into(),
-			dns_name: target.as_str().into(),
+			host: target.as_str().into(),
```
niklasad1, line 166: I fucked up this,  
`host` - hostname with port kusama-rpc.polkadot.io:443
`dns-name` - kusama-rpc.polkadot.io

before this PR it was the opposite which made it fail on bad DNS name.

dvdplm: I think having `host` and `dns_name` is confusing and leads to some funny looking code elsewhere:

```rust
						let dns_name = webpki::DNSNameRef::try_from_ascii_str(&self.dns_name)?;
						let tls_stream = connector.connect(&dns_name.to_owned(), socket?).await?;
```
…after all `DNSName` is just a newtype around `String`?

What I'm trying to say is that as far as I can see we only use `dns_name` in `WsTransportClientBuilder.build` which is not performance critical (and when we do use it, we do that conversion to `DNSNameRef` and then clone back to a `String`). So perhaps we can just store host and port and skip the `dns_name`?

tests/src/lib.rs
```diff
@@ -207,3 +207,10 @@ async fn ws_more_request_than_buffer_should_not_deadlock() {
 		req.await.unwrap();
 	}
 }
+
+#[tokio::test]
+async fn wss_works() {
```
dvdplm, line 212: Would be good to be complete here:
- assert that `ws://` works
- assert that omitting the port works (and uses the default)
- assert that providing a wacky port fails, bigger than `u16::MAX`, negative number, `0` etc
- assert that host name must be ascii (does it?)

dvdplm (changes requested):

ws-client/src/transport.rs
niklasad1: > So perhaps we can just store host and port and skip the dns_name?

Sounds good to me, I did least effort to get it working and the bug was introduced by the naming. Let's do it properly ^^

ws-client/src/transport.rs
```diff
@@ -144,27 +151,30 @@ pub enum WsConnectError {
 	ParseError(#[source] serde_json::error::Error),
 }
 
-/// Creates a new WebSocket connection based on [`WsConfig`](crate::WsConfig) represented as a Sender and Receiver pair.
-pub async fn websocket_connection(config: WsConfig<'_>) -> Result<(Sender, Receiver), WsHandshakeError> {
+/// Perform checks on the `WsConfig` that doesn't require any heavy operations.
+fn validate_ws_offline_settings(config: &WsConfig<'_>) -> Result<(Host, Mode, String), WsHandshakeError> {
```
niklasad1, line 155: extracted to a helper function to easily test this without having to do an actual to connection to the host.

ws-client/src/transport.rs
```diff
@@ -37,6 +37,15 @@ use thiserror::Error;
 
 type TlsOrPlain = crate::stream::EitherStream<TcpStream, TlsStream<TcpStream>>;
 
+#[derive(Clone)]
+pub struct Host(String);
```
niklasad1, line 41: Introduced a new type to avoid nasty bugs when several params has the same type.

It could be a `pub struct Host<'a>(&'a str);

niklasad1:
@dvdplm I think this is ready for another round of review :)

tests/src/lib.rs
```diff
@@ -207,3 +210,23 @@ async fn ws_more_request_than_buffer_should_not_deadlock() {
...
+
+#[tokio::test]
+async fn ws_with_non_ascii_url_doesnt_hang_or_panic() {
+	let err = WsClient::new(WsConfig::with_url("wss://♥♥♥♥♥♥∀∂")).await;
+	assert!(matches!(err, Err(Error::TransportError(_))));
+}
+
+#[tokio::test]
+async fn http_with_non_ascii_url_doesnt_hang_or_panic() {
+	let client = HttpClient::new("http://♥♥♥♥♥♥∀∂", HttpConfig::default()).unwrap();
```
niklasad1, line 229: https://docs.rs/url/2.2.0/url/enum.Host.html#variant.Domain is the reason why it's doesn't fail parsing the URL in the constructor.

niklasad1: and in the `WebSocket` case the host is resolved to a `SockAddr` because an actual connection is established in the constructor.

ws-client/src/transport.rs
```diff
@@ -277,6 +267,31 @@ impl<'a> WsTransportClientBuilder<'a> {
...
+	fn try_from(config: WsConfig<'a>) -> Result<Self, Self::Error> {
+		let url =
+			url::Url::parse(&config.url).map_err(|e| WsHandshakeError::Url(format!("Invalid URL: {}", e).into()))?;
+		let mode = match url.scheme() {
+			"ws" => Mode::Plain,
+			"wss" => Mode::Tls,
+			_ => return Err(WsHandshakeError::Url("URL scheme not supported, expects 'ws' or 'wss'".into())),
+		};
+		let host = url.host_str().ok_or_else(|| WsHandshakeError::Url("No host in URL".into()))?.into();
+		let sockaddrs: Vec<SocketAddr> = url.socket_addrs(|| None).map_err(WsHandshakeError::ResolutionFailed)?;
```
niklasad1, line 282: switched to https://docs.rs/url/2.2.0/url/struct.Url.html#method.socket_addrs instead which provides the default port number if it's missing such that we don't have to do it manually anymore

ws-client/src/transport.rs
```diff
@@ -146,43 +159,8 @@ pub enum WsConnectError {
 
 /// Creates a new WebSocket connection based on [`WsConfig`](crate::WsConfig) represented as a Sender and Receiver pair.
 pub async fn websocket_connection(config: WsConfig<'_>) -> Result<(Sender, Receiver), WsHandshakeError> {
```
niklasad1, line 161: if you like the change i.e, provide `TryFrom<WsConfig> for WebSocketTransportClient { ... }` then I think we could remove this function.

insipx (approved):
LGTM.

tests/src/lib.rs
dvdplm: TIL!

ws-client/src/transport.rs
dvdplm: I think `String` is fine.

ws-client/src/transport.rs
dvdplm: Do you think we should add a comment stating this and what the default port is?

ws-client/src/transport.rs
```diff
@@ -277,6 +268,31 @@ impl<'a> WsTransportClientBuilder<'a> {
...
+
+	fn try_from(config: WsConfig<'a>) -> Result<Self, Self::Error> {
+		let url =
+			url::Url::parse(&config.url).map_err(|e| WsHandshakeError::Url(format!("Invalid URL: {}", e).into()))?;
+		let mode = match url.scheme() {
+			"ws" => Mode::Plain,
+			"wss" => Mode::Tls,
+			_ => return Err(WsHandshakeError::Url("URL scheme not supported, expects 'ws' or 'wss'".into())),
+		};
+		let host = url.host_str().ok_or_else(|| WsHandshakeError::Url("No host in URL".into()))?.into();
```
dvdplm, line 282: Isn't the error case taken care of by the `url::Url::parse()` call on line 276? Oh, I see, it is `None` for unix sockets... Ok, fine then.

The error is bit funny though, it's not really a handshaking error. Can we use a more appropriate error?

ws-client/src/transport.rs
```diff
@@ -277,6 +268,31 @@ impl<'a> WsTransportClientBuilder<'a> {
 	}
 }
 
+impl<'a> TryFrom<WsConfig<'a>> for WsTransportClientBuilder<'a> {
+	type Error = WsHandshakeError;
+
+	fn try_from(config: WsConfig<'a>) -> Result<Self, Self::Error> {
```
dvdplm, line 274: I quite like this pattern of converting a config into a builder.

dvdplm (approved):
A few minor nits.

ws-client/src/transport.rs
niklasad1: I think the best option is the document this on `WsConfig` i.e, default ports is `ws - 80, wss - 433` ^^

Here we could add a comment that `sock_addrs()` uses Https://docs.rs/url/2.2.0/url/struct.Url.html#method.port_or_known_default if the port is missing

ws-client/src/transport.rs
niklasad1: > The error is bit funny though, it's not really a handshaking error. Can we use a more appropriate error?

indeed, let's fix that in another PR

PR #215, opened 2021-02-17
prepare for client release

niklasad1:
The PR prepares for a alpha client release:
-  changes version: 0.1.0 -> 0.2.0-alpha (to avoid collisions, 0.1 is already released at crates.io)
- adds a changelog
- adds minimal documentation
- adds a few lints and fixes them


http-client/src/lib.rs
```diff
@@ -1,3 +1,25 @@
...
+//!
+//! It supports both [`tokio 1.0`](https://docs.rs/tokio/1.2.0/tokio/) and [`tokio 0.2`](https://docs.rs/tokio/0.2.25/tokio/index.html)
+//! via [Optional features](#optional-features).
+//!
+//! # Optional Features
+//!
+//! `jsonrpsee-http-client` uses the following [feature flags]:
+//!
+//! - `tokio1`: Enable to work with [`tokio 1.0`](https://docs.rs/tokio/1.2.0/tokio/) (mutual exclusive to `tokio02`)
+//! - `tokio0.2`: Enable to work with [`tokio 0.2`](https://docs.rs/tokio/0.2.25/tokio/index.html) (mutual exclusive to `tokio1`)
```
dvdplm, line 19:
```suggestion
//! - `tokio1`: Enable to use the library with [`tokio 1.0`](https://docs.rs/tokio/1.2.0/tokio/) (mutually exclusive with `tokio02`)
//! - `tokio0.2`: Enable to use the library with [`tokio 0.2`](https://docs.rs/tokio/0.2.25/tokio/index.html) (mutually exclusive with `tokio1`)
```

types/CHANGELOG.md
```diff
@@ -0,0 +1,7 @@
+# Changelog
```
dvdplm, line 1: I think we should have a single `CHANGELOG` or things will become messy and hard to read and keep up to date.

ws-client/src/lib.rs
```diff
@@ -1,3 +1,9 @@
+#![warn(missing_debug_implementations, missing_docs, rust_2018_idioms, unreachable_pub)]
```
dvdplm, line 1: I think we should agree on a standard here with @maciejhirsz too. I suggest we make `missing_docs` and `unreachable_pub` errors.

types/CHANGELOG.md
niklasad1: yeah that would be good but we don't have a wrapper/facade crate yet... :) 

I can remove the CHANGELOGs for now

ws-client/src/lib.rs
maciejhirsz: I'm happy with `missing_docs` and `unreachable_pub`, `rust_2018_idioms` seems overly pendantic with elided lifetimes (I'd rather have `Cow<str>` than `Cow<'_, str>`, etc.).

maciejhirsz (approved):
Just the one nitpick on lints that we just talked about with @niklasad1 :)

dvdplm (approved):

ws-client/src/lib.rs
niklasad1: I think we should keep it as `warn` because `deny/forbid` would force it stop compile on new toolchains as `rustc` gets smarter :)

[forbid(warnings) in substrate](https://github.com/paritytech/substrate/pull/2627/files) is an example of this :)

PR #216, opened 2021-02-17
client batch requests

### LINKED ISSUE #103: Batch Requests

niklasad1:
Closes #103

Needs to updated to work on WebSockets too before it's ready to go.

### DIFF
--- a/http-client/src/client.rs
+++ b/http-client/src/client.rs
@@ -1,13 +1,13 @@
 use crate::transport::HttpTransportClient;
+use alloc::collections::VecDeque;
+use core::convert::TryInto;
+use core::sync::atomic::{AtomicU64, Ordering};
 use jsonrpc::DeserializeOwned;
 use jsonrpsee_types::{
 	error::Error,
 	http::HttpConfig,
 	jsonrpc::{self, JsonValue},
 };
-
-use std::convert::TryInto;
-use std::sync::atomic::{AtomicU64, Ordering};
 
 /// JSON-RPC HTTP Client that provides functionality to perform method calls and notifications.
 ///
@@ -83,6 +83,57 @@ impl HttpClient {
 			}
 		}?;
 		jsonrpc::from_value(json_value).map_err(Error::ParseError)
+	}
+
+	/// Perform a batch request towards the server.
+	///
+	/// Returns `Ok` if all requests were answered succesfully.
+	/// Returns `Error` if any of the requests fails.
+	//
+	// TODO(niklasad1): maybe simplify generic `requests`, it's quite unreadable.
+	pub async fn batch_request<'a>(
+		&self,
+		requests: impl IntoIterator<Item = (impl Into<String>, impl Into<jsonrpc::Params>)>,
+	) -> Result<Vec<JsonValue>, Error> {
+		let mut calls = Vec::new();
+		let mut ids = VecDeque::new();
+
+		for (method, params) in requests.into_iter() {
+			let id = self.request_id.fetch_add(1, Ordering::SeqCst);
+			calls.push(jsonrpc::Call::MethodCall(jsonrpc::MethodCall {
+				jsonrpc: jsonrpc::Version::V2,
+				method: method.into(),
+				params: params.into(),
+				id: jsonrpc::Id::Num(id),
+			}));
+			ids.push_back(id);
+		}
+
+		let batch_request = jsonrpc::Request::Batch(calls);
+		let response = self
+			.transport
+			.send_request_and_wait_for_response(batch_request)
+			.await
+			.map_err(|e| Error::TransportError(Box::new(e)))?;
+
+		match response {
+			jsonrpc::Response::Single(_) => {
+				Err(Error::Custom("Server replied with single response to a batch request".to_string()))
+			}
+			jsonrpc::Response::Notif(_notif) => {
+				Err(Error::Custom("Server replied with notification to a a batch request".to_string()))
+			}
+			jsonrpc::Response::Batch(rps) => {
+				let mut responses = Vec::with_capacity(ids.len());
+				for rp in rps {
+					let next_id = ids
+						.pop_front()
+						.ok_or_else(|| Error::Custom("Batch request failed: ID not found".to_string()))?;
+					responses.push(Self::process_response(rp, next_id)?);
+				}
+				Ok(responses)
+			}
+		}
 	}
 
 	fn process_response(response: jsonrpc::Output, expected_id: u64) -> Result<JsonValue, Error> {
--- a/http-client/src/client.rs
+++ b/http-client/src/client.rs
@@ -1,5 +1,4 @@
 use crate::transport::HttpTransportClient;
-use alloc::collections::VecDeque;
 use core::convert::TryInto;
 use core::sync::atomic::{AtomicU64, Ordering};
 use jsonrpc::DeserializeOwned;
@@ -10,7 +9,6 @@ use jsonrpsee_types::{
 };
 // will not compile for WASM/no-std
 use std::collections::HashSet;
-
 
 /// JSON-RPC HTTP Client that provides functionality to perform method calls and notifications.
 ///
@@ -75,7 +73,14 @@ impl HttpClient {
 			.map_err(|e| Error::TransportError(Box::new(e)))?;
 
 		let json_value = match response {
-			jsonrpc::Response::Single(rp) => Self::process_response(rp, id),
+			jsonrpc::Response::Single(rp) => {
+				let (val, received_id) = Self::process_response(rp)?;
+				if id == received_id {
+					Ok(val)
+				} else {
+					Err(Error::InvalidRequestId)
+				}
+			}
 			// Server should not send batch response to a single request.
 			jsonrpc::Response::Batch(_rps) => {
 				Err(Error::Custom("Server replied with batch response to a single request".to_string()))
@@ -146,9 +151,9 @@ impl HttpClient {
 		}
 	}
 
-	fn process_response(response: jsonrpc::Output, expected_id: u64) -> Result<JsonValue, Error> {
-		match response.id() {
-			jsonrpc::Id::Num(n) if n == &expected_id => response.try_into().map_err(Error::Request),
+	fn process_response(response: jsonrpc::Output) -> Result<(JsonValue, u64), Error> {
+		match response.id().as_number().copied() {
+			Some(n) => Ok((response.try_into().map_err(Error::Request)?, n)),
 			_ => Err(Error::InvalidRequestId),
 		}
 	}
--- a/http-client/src/client.rs
+++ b/http-client/src/client.rs
@@ -99,10 +99,13 @@ impl HttpClient {
 	/// Returns `Error` if any of the requests fails.
 	//
 	// TODO(niklasad1): maybe simplify generic `requests`, it's quite unreadable.
-	pub async fn batch_request<'a>(
+	pub async fn batch_request<'a, T>(
 		&self,
 		requests: impl IntoIterator<Item = (impl Into<String>, impl Into<jsonrpc::Params>)>,
-	) -> Result<Vec<JsonValue>, Error> {
+	) -> Result<Vec<T>, Error>
+	where
+		T: DeserializeOwned,
+	{
 		let mut calls = Vec::new();
 		// NOTE(niklasad1): If more than `u64::MAX` requests are performed in the `batch` then duplicate IDs are used
 		// which we don't support because ID is used to uniquely identify a given request.
@@ -143,7 +146,8 @@ impl HttpClient {
 					if !ids.remove(&id) {
 						return Err(Error::InvalidRequestId);
 					}
-					let val: JsonValue = rp.try_into().map_err(Error::Request)?;
+					let json_val: JsonValue = rp.try_into().map_err(Error::Request)?;
+					let val = jsonrpc::from_value(json_val).map_err(Error::ParseError)?;
 					responses.push(val);
 				}
 				Ok(responses)
--- a/http-client/src/client.rs
+++ b/http-client/src/client.rs
@@ -8,6 +8,9 @@ use jsonrpsee_types::{
 	http::HttpConfig,
 	jsonrpc::{self, JsonValue},
 };
+// will not compile for WASM/no-std
+use std::collections::HashSet;
+
 
 /// JSON-RPC HTTP Client that provides functionality to perform method calls and notifications.
 ///
@@ -87,7 +90,7 @@ impl HttpClient {
 
 	/// Perform a batch request towards the server.
 	///
-	/// Returns `Ok` if all requests were answered succesfully.
+	/// Returns `Ok` if all requests were answered successfully.
 	/// Returns `Error` if any of the requests fails.
 	//
 	// TODO(niklasad1): maybe simplify generic `requests`, it's quite unreadable.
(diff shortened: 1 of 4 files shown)


http-client/src/client.rs
```diff
@@ -23,6 +22,71 @@ impl HttpClient {
...
 		Ok(Self { transport, request_id: AtomicU64::new(0) })
 	}
+
+	/// Perform a batch request towards the server.
+	///
+	/// Returns `Ok` if all requests were answered successfully.
+	/// Returns `Error` if any of the requests fails.
+	//
+	// TODO(niklasad1): maybe simplify generic `requests`, it's quite unreadable.
+	pub async fn batch_request<'a, T>(
```
niklasad1, line 32: move this to client trait?!

ws-client/src/manager.rs
```diff
@@ -53,6 +54,9 @@ pub struct RequestManager {
 	requests: FnvHashMap<RequestId, Kind>,
 	/// Reverse lookup, to find a request ID in constant time by `subscription ID` instead of looking through all requests.
 	subscriptions: HashMap<SubscriptionId, RequestId>,
+	/// Pending batch requests
+	// NOTE: BTreeSet to get sorted order of requestIDs.
+	batches: HashMap<BTreeSet<RequestId>, (Vec<RequestId>, PendingBatchOneshot)>,
```
maciejhirsz, line 59: `FnvHashMap<Vec<RequestId>, (Vec<RequestId>, PendingBatchOneshot)>,` and sort the Ids before accessing the map.

dvdplm:
> as the `server` is free to process the batch in any order such that the client has try any order of the request IDs.

The spec says "The Response objects being returned from a batch call MAY be returned in any order within the Array. The Client SHOULD match contexts between the set of Request objects and the resulting set of Response objects based on the id member within each Object." – but here you implement an ordered set of responses. Can you elaborate on this? What do they mean by "SHOULD match contexts"?

ws-client/src/client.rs
```diff
@@ -375,8 +420,50 @@
...
+
+				let mut ordered_responses = Vec::new();
+				for id in batch_ordered {
+					// NOTE(niklasad1): O(n)
+					let pos = match rps_unordered.iter().position(|(i, _)| *i == id) {
+						Some(pos) => pos,
+						None => unreachable!("All request ID's valid checked by RequestManager above; qed"),
+					};
+					// Dummy value should never be.
+					let (_, rp) = std::mem::replace(&mut rps_unordered[pos], (0, JsonValue::Number(0.into())));
```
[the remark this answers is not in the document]
dvdplm: Is there a guarantee that `Id` is non-zero? I took a glance and it seems that it boils down to a `u64` so using 0 as the tombstone is perhaps not 100% safe?

dvdplm (commented):
It seems like 90% of the complexity here comes from the requirement to return responses in the order they were submitted. As noted above I'm not 100% convinced this is what the spec actually says so before commenting further I'd like to be convinced I'm wrong.

niklasad1:
> as the `server` is free to process the batch in any order such that the client has try any order of the request IDs.

I meant that each "request/request object" in the batch may be returned by server in any order, clarified the PR description.
I think the tests I implemented show this quite clearly.

> The spec says "The Response objects being returned from a batch call MAY be returned in any order within the Array. The Client SHOULD match contexts between the set of Request objects and the resulting set of Response objects based on the id member within each Object." – but here you implement an ordered set of responses. Can you elaborate on this? What do they mean by "SHOULD match contexts"?

I interpreted it as each request ID should be matched with corresponding request ID in the response, or in their words "match contexts".

Technically, I implemented it in the same order as the requests are performed by the user.
When we receive a batch (internally a vec of requests), then first we copy to vec into a BTreeSet (which is ordered) and store orignal order in another Vec:

```rust
batches: HashMap<BTreeSet<RequestId>, (Vec<RequestId>, PendingBatchOneshot)>
```

This is just to avoid trying out all possible combinations of the request ID and the same could be achieved by sorting the `Vec<RequestId>` and put them in the HashMap as Maciej pointed out.

As I told Maciej I will try to provide some benchmarks on this

dvdplm:
@niklasad1 brainfart on my part; I was thinking that "client" here meant the user code that is sending the batch (which is completely wrong). Ignore me. :)

http-client/src/client.rs
```diff
@@ -23,6 +22,71 @@ impl HttpClient {
 		let transport = HttpTransportClient::new(target, config).map_err(|e| Error::TransportError(Box::new(e)))?;
 		Ok(Self { transport, request_id: AtomicU64::new(0) })
 	}
+
+	/// Perform a batch request towards the server.
+	///
+	/// Returns `Ok` if all requests were answered successfully.
+	/// Returns `Error` if any of the requests fails.
+	//
+	// TODO(niklasad1): maybe simplify generic `requests`, it's quite unreadable.
```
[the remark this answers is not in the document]
niklasad1: `requests: impl IntoIterator<Item = (impl Into<String>, impl Into<jsonrpc::Params>)>,`

Was what I referred to in the TODO must people will probably use a `Vec` and `IntoIterator` allows unordered collections such as HashMap which might be a footgun in the case.

So maybe accepting requests as a `slice` is better.

http-client/src/client.rs
```diff
@@ -23,6 +22,71 @@ impl HttpClient {
...
+
+		let batch_request = jsonrpc::Request::Batch(calls);
+		let response = self
+			.transport
+			.send_request_and_wait_for_response(batch_request)
+			.await
+			.map_err(|e| Error::TransportError(Box::new(e)))?;
+
+		match response {
+			jsonrpc::Response::Single(_) => {
```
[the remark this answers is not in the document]
niklasad1: This means that response was `deserialized` as single request object i.e, not batch (which has to contain an array of requests with a at least one request) which doesn't follow the spec.

So, we return a error for this instead of panicking.

http-client/src/client.rs
```diff
@@ -23,6 +22,71 @@ impl HttpClient {
...
+					// NOTE(niklasad1): O(n)
+					let pos = match ids.iter().position(|i| i == &id) {
+						Some(id) => id,
+						None => return Err(Error::InvalidRequestId),
+					};
+					let json_val: jsonrpc::JsonValue = rp.try_into().map_err(Error::Request)?;
+					json_responses[pos] = json_val;
+				}
+				let responses: Result<_, _> =
+					json_responses.into_iter().map(|val| jsonrpc::from_value(val).map_err(Error::ParseError)).collect();
```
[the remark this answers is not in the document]
niklasad1: yeah, I think it did like this because `jsonrpc::from_value` returns `T.

Then we either to add `T: Default` or put in an `Option` where the json_responses is declared.

ws-client/src/client.rs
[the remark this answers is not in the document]
niklasad1: This is just dummy value because each request IDs should be unique and read exactly once, ensured by the RequestManager if we don't have any bugs :)
 
So think of this as `std::mem::take` but I could instead clone the value if you like it better.

http-client/src/client.rs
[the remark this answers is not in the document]
dvdplm: `T: Default` sounds reasonable to me.

ws-client/src/client.rs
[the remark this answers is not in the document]
dvdplm: `std::mem::replace` is fine. My concern was more that we could end up with a valid ID of `0` and that might be bad? We are using `0` to signal "free" and the comment above, "Dummy value should never be", made me think `0` must never be used for an ID. If this is the case then we should use the type system to ensure IDs are non-zero.

http-client/src/client.rs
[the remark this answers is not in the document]
dvdplm: I like the idea of using a slice. @maciejhirsz thoughts?

dvdplm:
> Further, batch requests does consume a `request ID` for each request in the batch and it "might violate" the "max_concurrent_request_limit" depending on how we define that I suppose. For example if a user sets this to 10 and then perform a batch request of 10, then no more requests are allowed until the batch has been completed. So question is that if a batch request should be regarded as one request or the number of requests in the batch

I think a batch request containing `n` requests should count against the `max_concurrent_request_limit` as `n`. In my head the `max_concurrent_request_limit` setting should default to a fairly high value – 1000? hundreds? – and it should be thought of as a safety valve and not really as a throttling mechanism.

ws-client/src/client.rs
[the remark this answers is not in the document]
niklasad1: Switched to `Vec::swap_remove` it is `O(1)` but doesn't preserve order.

http-client/src/client.rs
```diff
@@ -79,4 +78,62 @@ impl Client for HttpClient {
...
+
+	async fn batch_request<T, M, P>(&self, batch: Vec<(M, P)>) -> Result<Vec<T>, Error>
+	where
+		T: DeserializeOwned + Default + Clone,
+		M: Into<String> + Send,
+		P: Into<jsonrpc::Params> + Send,
+	{
+		let mut calls = Vec::with_capacity(batch.len());
+		// NOTE(niklasad1): `ID` is not necessarily monotonically increasing.
+		let mut ordered_requests = Vec::with_capacity(batch.len());
```
niklasad1, line 90: I realized that we could have an additional request_set `HashSet<RequestID>`  here to do quick lookups to avoid O(n) lookups

http-client/src/client.rs
```diff
@@ -88,8 +89,9 @@ impl Client for HttpClient {
 		let mut calls = Vec::with_capacity(batch.len());
 		// NOTE(niklasad1): `ID` is not necessarily monotonically increasing.
 		let mut ordered_requests = Vec::with_capacity(batch.len());
+		let mut request_set = FnvHashMap::with_capacity_and_hasher(batch.len(), Default::default());
```
dvdplm, line 92: I wonder why they didn't keep `with_capacity()`.

niklasad1:
> I think a batch request containing n requests should count against the max_concurrent_request_limit as n. In my head the max_concurrent_request_limit setting should default to a fairly high value – 1000? hundreds? – and it should be thought of as a safety valve and not really as a throttling mechanism.

Indeed and this makes it more complicated as request ID and batch ID (request ID of request within the batch) needs to be distinguished. So, I'm not super excited about this and it's error prone to have return the request ID after it has been completed.

Maybe better just with a simple counter and guard the hashmap/other collections when inserting new requests.

ws-client/src/manager.rs
```diff
@@ -38,33 +38,63 @@ pub enum RequestStatus {
...
 type RequestId = u64;
+/// Wrapping counter for requests within a batch request.
+// TODO(niklasad1): new type for this
+type BatchId = u64;
+
+#[derive(Debug)]
+/// Batch state.
+pub struct BatchState {
+	/// Order that the request was performed in.
+	pub order: FnvHashMap<BatchId, usize>,
```
niklasad1, line 56: I added this to avoid the O(n) searches.

http-client/src/client.rs
niklasad1: :shrug:

ws-client/src/client.rs
```diff
@@ -374,8 +411,50 @@ async fn background_task(
...
+						Err(err) => {
+							let _ = front_error.send(err);
+							return;
+						}
+					};
+					digest.push(id);
+					rps_unordered.push((id, rp));
+				}
+
+				digest.sort();
```
dvdplm, line 439: Github Actions says "used sort instead of sort_unstable to sort primitive type `u64`" here. Do you see those warnings too or is it just me?

ws-client/src/jsonrpc_transport.rs
```diff
@@ -31,6 +31,45 @@ impl Sender {
...
+		let req_id = request_manager.next_request_id()?;
+		let mut calls = Vec::with_capacity(batch.requests.len());
+		let mut ids = Vec::with_capacity(batch.requests.len());
+
+		for (method, params) in batch.requests {
+			let batch_id = request_manager.next_batch_id();
+			ids.push(batch_id);
+			calls.push(jsonrpc::Call::MethodCall(jsonrpc::MethodCall {
+				jsonrpc: jsonrpc::Version::V2,
+				method: method.into(),
```
dvdplm, line 49: GH Actions says "useless conversion to the same type: `std::string::String`"

dvdplm (commented):
> So, I'm not super excited about this and it's error prone to have return the request ID after each request has been completed.

FWIW it's not something that should block this PR. Let's log an issue and move on for now.

niklasad1:
re benches:

I tried against `jsonrpc servers` because our `WebSocket server` doesn't support batches and funny enough I can't compile with `tokio 0.2` for tests because of our `test-util crate` uses `hyper 14` ^^

However, I did some manual wallclock benches

```
HTTP 10 requests in a batch took 0 ms
HTTP 100 requests in a batch took 0 ms
HTTP 1000 requests in a batch took 3 ms
HTTP 10000 requests in a batch took 35 ms
WebSocket 10 requests in a batch took 4 ms
WebSocket 100 requests in a batch took 0 ms
WebSocket 1000 requests in a batch took 3 ms
WebSocket 10000 requests in a batch took 34 ms
```

Looks alright, I think 48 concurrent HTTP requests took ~1ms on my machine

types/src/traits.rs
```diff
@@ -19,6 +19,18 @@ pub trait Client {
...
+
+	/// Send a [batch request](https://www.jsonrpc.org/specification#batch).
+	///
+	/// The response to batch are returned in the same order as it was inserted in the batch.
+	///
+	/// Returns `Ok` if all requests in the batch were answered successfully.
+	/// Returns `Error` if any of the requests in batch fails.
+	async fn batch_request<T, M, P>(&self, batch: Vec<(M, P)>) -> Result<Vec<T>, Error>
+	where
+		T: DeserializeOwned + Default + Clone,
```
niklasad1, line 31: `Default + Clone` is required to do `vec![T::default(); len]`

http-client/src/client.rs
```diff
@@ -79,4 +79,62 @@ impl Client for HttpClient {
 		}?;
 		jsonrpc::from_value(json_value).map_err(Error::ParseError)
 	}
+
+	async fn batch_request<T, M, P>(&self, batch: Vec<(M, P)>) -> Result<Vec<T>, Error>
```
niklasad1, line 83: I think it would be possible to have:

```rust
	async fn batch_request<T, M, P>(&self, batch: &[(M, P)]) -> Result<Vec<T>, Error>
	where
		T: DeserializeOwned + Default + Clone,
		M: Into<String> + Send,
		P: Into<Params> + Send + Sync,
		for<'a> &'a M: Into<String> + Send,
		for<'a> &'a P: Into<Params> + Send;
```

It becomes super complicated.

ws-client/src/jsonrpc_transport.rs
```diff
@@ -31,6 +31,45 @@ impl Sender {
...
+				jsonrpc: jsonrpc::Version::V2,
+				method,
+				params,
+				id: jsonrpc::Id::Num(batch_id),
+			}));
+		}
+
+		if let Err(send_back) = request_manager.insert_pending_batch(ids, batch.send_back, req_id) {
+			request_manager.reclaim_request_id(req_id);
+			let _ = send_back.send(Err(Error::InvalidRequestId));
```
niklasad1, line 57: maybe a new error batches?

types/src/error.rs
```diff
@@ -30,6 +30,9 @@ pub enum Error {
 	/// Frontend/backend channel error.
 	#[error("Frontend/backend channel error: {0}")]
 	Internal(#[source] futures::channel::mpsc::SendError),
+	/// Invalid response,
+	#[error("Invalid response detected: {0}")]
```
dvdplm, line 34:
```suggestion
	#[error("Invalid response: {0}")]
```

http-client/src/client.rs
dvdplm: Fair enough. Let's use a `Vec` then, keep it simple and opinionated.

dvdplm (approved):

Issue #212, opened 2021-02-13
[HTTP server]: ACK notifications / send empty response to notifications


niklasad1:
Basically, if a client does the following:

```rust
    let client = HttpClient::new("http://127.0.0.1:3030", HttpConfig::default())?;
    client
        .notification("say_hello", jsonrpc::Params::None)
        .await
        .unwrap();
```

It will wait forever because it's possible not to ACK notifications with the current notification design in the HTTP server.
It would be better just to return an empty response with status code 200 and simplify the use for client implementations. 

I think https://github.com/paritytech/jsonrpc does this too.

Tthus, it means that notifications from the `jsonrpsee HTTP client` will block forever/not work ^^

niklasad1:
I guess it would be fine in the client to do `send_now_or_never()` for now.

niklasad1:
We don't have exposed in API anymore

PR #223, opened 2021-03-02
[ws client]: register request ID for unsubscribe requests.

### LINKED ISSUE #220: [ws client]: register request ID for unsubscriptions
After #174 the server responds with the request ID for `unsubscriptions` those are currently not registered in the request manager and those responses will be regarded as `invalid requests` because the request ID is most likely not registered.

Example:
  1. The subscription capacity gets full because of too many unread notifications
  2. drops the subscription
  3. client send `unsubscribe_to_some_method, id: x`
  4. server ACKs `unsubscribe_to_some_method` the request with `id: x` 
  5. client doesn't know what `id: x`  is and closes the connection because of `Error::Internal(_)`.

I'm quite surprised that [subscription test with drop](https://github.com/paritytech/jsonrpsee/blob/master/tests/src/lib.rs#L97) passes but I think the subscriptions itself has to be dropped and a request or something has to be performed to detect the error because it happens in the background thread.

niklasad1:
Fixes #220

### DIFF
--- a/tests/src/lib.rs
+++ b/tests/src/lib.rs
@@ -106,9 +106,9 @@ async fn ws_subscription_several_clients_with_drop() {
 		config.max_subscription_capacity = u32::MAX as usize;
 
 		let client = WsClient::new(config).await.unwrap();
-		let hello_sub: WsSubscription<JsonValue> =
+		let hello_sub: WsSubscription<String> =
 			client.subscribe("subscribe_hello", Params::None, "unsubscribe_hello").await.unwrap();
-		let foo_sub: WsSubscription<JsonValue> =
+		let foo_sub: WsSubscription<u64> =
 			client.subscribe("subscribe_foo", Params::None, "unsubscribe_foo").await.unwrap();
 		clients.push((client, hello_sub, foo_sub))
 	}
@@ -117,13 +117,18 @@ async fn ws_subscription_several_clients_with_drop() {
 		for (_client, hello_sub, foo_sub) in &mut clients {
 			let hello = hello_sub.next().await.unwrap();
 			let foo = foo_sub.next().await.unwrap();
-			assert_eq!(hello, JsonValue::String("hello from subscription".to_owned()));
-			assert_eq!(foo, JsonValue::Number(1337_u64.into()));
+			assert_eq!(&hello, "hello from subscription");
+			assert_eq!(foo, 1337);
 		}
 	}
 
 	for i in 0..5 {
-		let (client, _, _) = clients.remove(i);
+		let (client, hello_sub, foo_sub) = clients.remove(i);
+		drop(hello_sub);
+		drop(foo_sub);
+		// Send this request to make sure the client's background thread hasn't
+		// been canceled.
+		let _r: String = client.request("say_hello", Params::None).await.unwrap();
 		drop(client);
 	}
 
@@ -134,8 +139,8 @@ async fn ws_subscription_several_clients_with_drop() {
 		for (_client, hello_sub, foo_sub) in &mut clients {
 			let hello = hello_sub.next().await.unwrap();
 			let foo = foo_sub.next().await.unwrap();
-			assert_eq!(hello, JsonValue::String("hello from subscription".to_owned()));
-			assert_eq!(foo, JsonValue::Number(1337_u64.into()));
+			assert_eq!(&hello, "hello from subscription");
+			assert_eq!(foo, 1337);
 		}
 	}
 }
--- a/ws-client/src/client.rs
+++ b/ws-client/src/client.rs
@@ -268,7 +268,7 @@ async fn background_task(
 					log::trace!("[backend]: client prepares to send request={:?}", method);
 					match sender.start_request(method, params).await {
 						Ok(id) => {
-							if let Err(send_back) = manager.insert_pending_call(id, send_back) {
+							if let Err(Some(send_back)) = manager.insert_pending_call(id, Some(send_back)) {
 								let _ = send_back.send(Err(Error::DuplicateRequestId));
 							}
 						}
@@ -306,7 +306,9 @@ async fn background_task(
 						if let Some((_sink, unsubscribe_method)) = manager.remove_subscription(request_id, sub_id.clone()) {
 							if let Ok(json_sub_id) = jsonrpc::to_value(sub_id) {
 								let params = jsonrpc::Params::Array(vec![json_sub_id]);
-								let _ = sender.start_request(unsubscribe_method, params).await;
+								if let Ok(id) = sender.start_request(unsubscribe_method, params).await {
+										let _ = manager.insert_pending_call(id, None);
+								}
 							}
 						}
 					}
@@ -320,8 +322,13 @@ async fn background_task(
 				Some(Ok(jsonrpc::Response::Single(response))) => {
 					match process_response(&mut manager, response, max_capacity_per_subscription) {
 						Ok(Some((unsubscribe, params))) => {
-							if let Err(e) = sender.start_request(unsubscribe, params).await {
-								log::error!("Failed to send unsubscription response: {:?}", e);
+							match sender.start_request(unsubscribe, params).await {
+								Ok(id) => {
+									if let Err(e) = manager.insert_pending_call(id, None) {
+										log::error!("Error: {:?}", e);
+									}
+								}
+								Err(e) => log::error!("Failed to send unsubscription request: {:?}", e),
 							}
 						}
 						Ok(None) => (),
@@ -331,22 +338,9 @@ async fn background_task(
 						}
 					}
 				}
-				Some(Ok(jsonrpc::Response::Batch(responses))) => {
-					// if any request fails, throw away entire batch.
-					for response in responses {
-						match process_response(&mut manager, response, max_capacity_per_subscription) {
-							Ok(Some((unsubscribe, params))) => {
-								if let Err(e) = sender.start_request(unsubscribe, params).await {
-									log::error!("Failed to send unsubscription response: {:?}", e);
-								}
-							}
-							Ok(None) => (),
-							Err(e) => {
-								log::error!("Error: {:?} terminating client", e);
-								return;
-							}
-						}
-					}
+				Some(Ok(jsonrpc::Response::Batch(_responses))) => {
+					log::error!("Received batch response but not supported");
+					return;
 				}
 				Some(Ok(jsonrpc::Response::Notif(notif))) => {
 					let sub_id = notif.params.subscription;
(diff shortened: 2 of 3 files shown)


ws-client/src/client.rs
```diff
@@ -394,10 +389,11 @@ fn process_response(
...
 			let response = response.try_into().map_err(Error::Request);
-			match send_back_oneshot.send(response) {
-				Err(Err(e)) => Err(e),
-				Err(Ok(_)) => Err(Error::Custom("Frontend channel closed".into())),
-				Ok(_) => Ok(None),
+			match send_back_oneshot.map(|tx| tx.send(response)) {
+				Some(Err(Err(e))) => Err(e),
+				Some(Err(Ok(_))) => Err(Error::Custom("Frontend channel closed".into())),
+				Some(Ok(_)) => Ok(None),
+				None => Ok(None),
```
niklasad1, line 396: I added that `send_back` can be `None` because subscriptions are different from ordinary requests and the actual `drop imp` doesn't care about the response.

ws-client/src/client.rs
```diff
@@ -320,8 +322,13 @@ async fn background_task(
 				Some(Ok(jsonrpc::Response::Single(response))) => {
 					match process_response(&mut manager, response, max_capacity_per_subscription) {
 						Ok(Some((unsubscribe, params))) => {
-							if let Err(e) = sender.start_request(unsubscribe, params).await {
-								log::error!("Failed to send unsubscription response: {:?}", e);
+							match sender.start_request(unsubscribe, params).await {
+								Ok(id) => {
+									if let Err(e) = manager.insert_pending_call(id, None) {
```
niklasad1, line 327: Ideally, move the request manager to the sender instead to avoid having to this manual inserting.

In addition, if the request ID is already occupied leaves the client in a weird state i.e, might map the request to a request id. Also an argument why to move the request manager to the `JSONRPC Sender`, i.e, might be better to check if the request ID is not in the request manager and to avoid sending another request with the same ID. However, that's very unlikely one would need to send more requests than u64::MAX for that to happen.

ws-client/src/tests.rs
```diff
@@ -98,6 +98,7 @@ async fn internal_error_works() {
 
 #[tokio::test]
 async fn subscription_works() {
+	let _ = env_logger::try_init();
```
niklasad1, line 101:
```suggestion
```

tests/src/lib.rs
```diff
@@ -117,13 +117,18 @@ async fn ws_subscription_several_clients_with_drop() {
...
+			assert_eq!(foo, 1337);
 		}
 	}
 
 	for i in 0..5 {
-		let (client, _, _) = clients.remove(i);
+		let (client, hello_sub, foo_sub) = clients.remove(i);
+		drop(hello_sub);
+		drop(foo_sub);
+		// Send this request to make sure the client's background thread hasn't
```
niklasad1, line 129:
```suggestion
		// Send this request to make sure that the client's background thread hasn't
```

ws-client/src/client.rs
```diff
@@ -251,153 +261,129 @@ async fn background_task(
 		let next_backend = backend_event.next();
 		futures::pin_mut!(next_frontend, next_backend);
 
-		futures::select! {
```
niklasad1, line 254: changed to `futures::select` to improve readability, reduces one level of indentation.

ws-client/src/jsonrpc_transport.rs
```diff
@@ -15,81 +20,97 @@ pub async fn websocket_connection(config: WsConfig<'_>) -> Result<(Sender, Recei
...
+		Self { transport }
 	}
 
-	/// Inner implementation for starting either a request or a subscription.
-	async fn start_impl(
+	/// Sends a request to the server but it doesn’t wait for a response.
+	/// Instead, you have keep the request ID and use the Receiver to get the response.
+	///
+	/// Returns Ok() if the request was successfully sent otherwise Err(_).
+	pub async fn start_request(
```
niklasad1, line 38: you may wonder why we don't do the same for the receiver but the reason is that we move the receiver into a `stream::unfold` which takes ownership of the stream.

so won't work without moving the `request_manager` to a Mutex which I prefer not to do even though it would make code in the background to be easier to read.

ws-client/src/manager.rs
```diff
@@ -34,32 +34,60 @@ pub enum RequestStatus {
...
 type UnsubscribeMethod = String;
-type RequestId = u64;
+type RequestId = u8;
 
-#[derive(Debug, Default)]
+#[derive(Debug)]
 /// Manages and monitors JSONRPC v2 method calls and subscriptions.
 pub struct RequestManager {
+	/// Free list, 0 indicates a free slot and 1 an occupied slot.
+	free_list: [u8; 256],
```
niklasad1, line 47: we could make this faster by reqister the last fetched request ID instead of starting from zero every time in `fn next_request_id`

gregdhill (approved):

ws-client/src/manager.rs
niklasad1: after some thinking, it's probably unnecessary to make assumptions about how many unanswered requests which covers all use-cases.

maciejhirsz, line 47: Suggestion 1: rename `free_list` to `request_id_slots`.

Suggestion 2: Change the type to `[bool; 256]` or `[Slot; 256]` where `Slot` is an enum with `Vacant` and `Occupied` or some such to make the state clearer.

Could also make it dense with `[u64; 4]` for 1/8th of memory, and using [`leading_zeroes`](https://doc.rust-lang.org/std/primitive.u64.html#method.leading_zeros) to make the iteration technically faster, at least on some implementations of x86.

maciejhirsz (approved):
Looks good to me, just some suggestions on how id recycling can be done.

ws-client/src/jsonrpc_transport.rs
```diff
@@ -15,81 +20,99 @@ pub async fn websocket_connection(config: WsConfig<'_>) -> Result<(Sender, Recei
...
-		self.transport.send_request(request).await?;
-
-		Ok(id)
+		match self.transport.send_request(req).await {
+			Ok(_) => {
+				request_manager.insert_pending_call(id, request.send_back).expect("ID unused checked above; qed");
+				Ok(())
+			}
+			Err(e) => {
+				let str_err = e.to_string();
```
niklasad1, line 63: these are `to_string` are kind of ugly the reason is that the actual error is sent over the channel to the frontend and the error string is returned to background thread for logging.

niklasad1: we could remove the `to_string` but for now let's keep them as this crate is still not super stable.

ws-client/src/manager.rs
```diff
@@ -34,32 +37,54 @@ pub enum RequestStatus {
...
-	// NOTE: FnvHashMap is used here because RequestId is not under the caller's control and is known to be a short key (u64).
+	// NOTE: FnvHashMap is used here because RequestId is not under the caller's control and is known to be a short key.
 	requests: FnvHashMap<RequestId, Kind>,
 	/// Reverse lookup, to find a request ID in constant time by `subscription ID` instead of looking through all requests.
 	subscriptions: HashMap<SubscriptionId, RequestId>,
 }
 
+impl Default for RequestManager {
+	fn default() -> Self {
+		Self { free_slots: (0..65535).collect(), requests: FnvHashMap::default(), subscriptions: HashMap::default() }
```
niklasad1, line 60: increased the to `u16::MAX` to be on the safe side

ws-client/src/manager.rs
niklasad1: I made it an `VecDeque` instead and made it possible configure the size, ok? However, I didn't implement it in the client though I couldn't think of a good name and the config is already complicated....

Feel free to push a fix to this branch if you have something better in mind :)

tests/src/lib.rs
```diff
@@ -185,13 +190,11 @@ async fn ws_more_request_than_buffer_should_not_deadlock() {
 
 	for _ in 0..6 {
 		let c = client.clone();
-		requests.push(tokio::spawn(async move {
-			let _: JsonValue = c.request("say_hello", Params::None).await.unwrap();
-		}));
+		requests.push(tokio::spawn(async move { c.request::<String, _, _>("say_hello", Params::None).await }));
```
niklasad1, line 193: after the request manager limit, it will reject the request direct because once it can't allocate a request ID it would propagated the users, kind of fixes https://github.com/paritytech/jsonrpsee/issues/225.

It will increase channel capacity but be dropped when it reaches the background thread.

dvdplm (approved):
lgtm. I like the reuse mechanism here. Left nitpicks.

ws-client/src/manager.rs
```diff
@@ -34,32 +37,57 @@ pub enum RequestStatus {
...
+		Self::new(REQUEST_MANAGER_DEFAULT_MAX_SIZE)
+	}
+}
+
 impl RequestManager {
-	/// Create an empty `RequestManager`.
-	pub fn new() -> Self {
-		Self::default()
+	/// Create a new `RequestManager` with specified capacity.
+	pub fn new(slot_capacity: RequestId) -> Self {
```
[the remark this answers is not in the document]
niklasad1: it was just because it was used internally as u64, I can make it an usize probably a more convenient for users.

ws-client/src/manager.rs
```diff
@@ -34,32 +37,57 @@ pub enum RequestStatus {
...
 	/// List of requests that are waiting for a response from the server.
-	// NOTE: FnvHashMap is used here because RequestId is not under the caller's control and is known to be a short key (u64).
+	// NOTE: FnvHashMap is used here because RequestId is not under the caller's control and is known to be a short key.
 	requests: FnvHashMap<RequestId, Kind>,
 	/// Reverse lookup, to find a request ID in constant time by `subscription ID` instead of looking through all requests.
 	subscriptions: HashMap<SubscriptionId, RequestId>,
 }
 
+impl Default for RequestManager {
+	fn default() -> Self {
```
niklasad1, line 62: this was super stupid because `new` calls `default::default` recursively and ends up with a stack overflow ^^

types/src/client.rs
```diff
@@ -16,41 +16,50 @@ pub struct Subscription<Notif> {
...
+	/// When we get a response from the server about that subscription, we send the result on
+	/// this channel. If the subscription succeeds, we return a [Receiver](futures::channel::mpsc::Receiver) that will receive
+	/// notifications.
+	pub send_back: oneshot::Sender<Result<(mpsc::Receiver<JsonValue>, SubscriptionId), Error>>,
+}
+
 /// Message that the Client can send to the background task.
 #[derive(Debug)]
 pub enum FrontToBack {
 	/// Send a one-shot notification to the server. The server doesn't give back any feedback.
```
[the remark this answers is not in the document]
niklasad1: I think we should stray the last comment because I think it's rather implementation specific and the `JSONRPC spec` does state anything about it:

> A Request object that is a Notification signifies the Client's lack of interest in the corresponding Response object, and as such no Response object needs to be returned to the client.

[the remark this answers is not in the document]
niklasad1: For WS it's fine to not respond but for HTTP `200 OK` seems reasonable for a server to respond with see #212

PR #224, opened 2021-03-03
RpcModule and RpcContextModule

maciejhirsz:
This PR adds module functionality to the ws-server (although it should be fairly easy to make it general use for either ws and http).

Modules are equivalent to `IoHandler` while `Context` in `RpcContextModule` is equivalent to `Metadata` in `jsonrpc-core`, with the major difference here is that `RpcModule` is a concrete type without any generics, and `RpcContextModule` can be converted to `RpcModule` once all methods that need the extra context are defined.

I expect to do changes as I run into unexpected issues when migrating RPC layer in Substrate, so I would consider the API to be wildly unstable at this point.

### DIFF
--- a/ws-server/src/server.rs
+++ b/ws-server/src/server.rs
@@ -41,7 +41,7 @@ use tokio::{
 use tokio_stream::{wrappers::TcpListenerStream, StreamExt};
 use tokio_util::compat::TokioAsyncReadCompatExt;
 
-use crate::types::{ConnectionId, RpcSender, RpcId, RpcMethod, Methods};
+use crate::types::{ConnectionId, Methods, RpcId, RpcMethod, RpcSender};
 use crate::types::{JsonRpcError, JsonRpcErrorParams};
 use crate::types::{JsonRpcInvalidRequest, JsonRpcRequest, JsonRpcResponse, TwoPointZero};
 use crate::types::{JsonRpcNotification, JsonRpcNotificationParams};
--- a/ws-server/src/server.rs
+++ b/ws-server/src/server.rs
@@ -48,7 +48,7 @@ use crate::types::{JsonRpcNotification, JsonRpcNotificationParams};
 
 mod module;
 
-pub use module::Module;
+pub use module::RpcModule;
 
 type SubscriptionId = u64;
 
@@ -164,7 +164,7 @@ fn send_error(id: RpcId, tx: RpcSender, code: i32, message: &str) {
 }
 
 pub struct Server {
-	root: Module,
+	root: RpcModule,
 	listener: TcpListener,
 }
 
@@ -173,7 +173,7 @@ impl Server {
 	pub async fn new(addr: impl ToSocketAddrs) -> anyhow::Result<Self> {
 		let listener = TcpListener::bind(addr).await?;
 
-		Ok(Server { listener, root: Module::new() })
+		Ok(Server { listener, root: RpcModule::new() })
 	}
 
 	/// Register a new RPC method, which responds with a given callback.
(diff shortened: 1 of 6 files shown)


ws-server/src/server.rs
```diff
@@ -211,67 +191,23 @@ impl Server {
...
-			match result {
-				Ok(()) => (),
-				Err(err) => {
-					// Remove subscribe method if unsubscribe has failed
-					self.methods.remove(subscribe_method_name);
-					return Err(err);
-				}
-			}
-		}
+		self.root.register_subscription(subscribe_method_name, unsubscribe_method_name)
```
maciejhirsz, line 194: Instead of doing this proxy, we can probably add `Deref` and `DerefMut` impls to `Server` that carries over the methods for the root `RpcModule` internally.

ws-server/src/server/module.rs
```diff
@@ -0,0 +1,161 @@
+use super::*;
+
+pub struct RpcModule {
```
niklasad1, line 3: maybe derive Default?

```suggestion
#[derive(Default)]
pub struct RpcModule {
```

ws-server/src/server/module.rs
```diff
@@ -0,0 +1,161 @@
...
+
+		for (name, callback) in other.methods {
+			self.methods.insert(name, callback);
+		}
+
+		Ok(())
+	}
+}
+
+pub struct RpcContextModule<Context> {
```
niklasad1, line 121: is this similar to `Meta` in `jsonrpc`?

niklasad1 (approved):
looks clean to me :+1: but it's been while since I looked at this code in substrate FWIW

ws-server/src/server/module.rs
```diff
@@ -0,0 +1,162 @@
...
+	}
+
+	/// Register a new RPC subscription, with subscribe and unsubscribe methods.
+	pub fn register_subscription(
+		&mut self,
+		subscribe_method_name: &'static str,
+		unsubscribe_method_name: &'static str,
+	) -> Result<SubscriptionSink, Error> {
+		if subscribe_method_name == unsubscribe_method_name {
+			return Err(Error::MethodAlreadyRegistered(subscribe_method_name.into()));
```
dvdplm, line 56: This should be some other error type I think. The problem isn't really that the method is registered but that the names of the (un-)subscribe methods are the same, `Error::InvalidSubscriptionName` perhaps?

ws-server/src/server/module.rs
```diff
@@ -0,0 +1,162 @@
...
+
+					Ok(())
+				}),
+			);
+		}
+
+		Ok(SubscriptionSink { method: subscribe_method_name, subscribers })
+	}
+
+	pub(crate) fn into_map(self) -> Methods {
```
dvdplm, line 105:
```suggestion
	pub(crate) fn into_methods(self) -> Methods {
```

…maybe?

ws-server/src/server/module.rs
```diff
@@ -0,0 +1,162 @@
...
+	}
+
+	pub(crate) fn merge(&mut self, other: RpcModule) -> Result<(), Error> {
+		for name in other.methods.keys() {
+			self.verify_method_name(name)?;
+		}
+
+		for (name, callback) in other.methods {
+			self.methods.insert(name, callback);
+		}
```
dvdplm, line 116: Why verify first and then insert?

dvdplm (commented):
Looks real nice. :)

ws-server/src/server/module.rs
maciejhirsz: So we don't have to roll-back inserts if there is a name conflict down the line, that is if registering the module fails, the state of the server will remain unchanged.

dvdplm (approved):

PR #234, opened 2021-03-09
[ws client]: background error reporting to frontend

### LINKED ISSUE #188: [ws client]: better error message when a request fails
For the `WebSocket client` there is a background thread that drives the request to completion if the request fails it's possible that just the background thread is terminated and the actual error is never propagated to the frontend. Then error message just says `Error::TransportError("background task closed")`, [an example of this](https://github.com/paritytech/jsonrpsee/blob/master/ws-client/src/tests.rs#L113-#L122)

Thus, it could be possible to introduce an additional channel just to send back the error cause when that happens.

niklasad1:
Closes #188

### DIFF
--- a/tests/src/proc_macros.rs
+++ b/tests/src/proc_macros.rs
@@ -1,3 +1,6 @@
+// Not all proc macros are used let's suppress it for now.
+#![allow(dead_code)]
+
 use crate::helpers;
 use jsonrpsee_http_client::*;
 use jsonrpsee_ws_client::*;
--- a/types/src/error.rs
+++ b/types/src/error.rs
@@ -30,6 +30,9 @@ pub enum Error {
 	/// Frontend/backend channel error.
 	#[error("Frontend/backend channel error: {0}")]
 	Internal(#[source] futures::channel::mpsc::SendError),
+	/// The background task has been terminated.
+	#[error("The background task been terminated because: {0}; restart required")]
+	RestartNeeded(String),
 	/// Failed to parse the data that the server sent back to us.
 	#[error("Parse error: {0}")]
 	ParseError(#[source] jsonrpc::ParseError),
--- a/ws-client/src/client.rs
+++ b/ws-client/src/client.rs
@@ -56,20 +56,22 @@ enum ErrorFromBack {
 }
 
 impl ErrorFromBack {
-	async fn step(self) -> (Self, String) {
+	async fn read_error(self) -> (Self, Error) {
 		match self {
 			Self::Unread(rx) => {
 				let msg = match rx.await {
 					Ok(msg) => msg.to_string(),
-					// This is a bug.
+					// This should never happen because the receiving end is still alive.
+					// Would be a bug in the logic of the background task.
 					Err(_) => {
 						log::error!("Could not found error from backend");
 						String::new()
 					}
 				};
-				(Self::Read(msg.clone()), msg)
+				let err = Error::RestartNeeded(msg.clone());
+				(Self::Read(msg), err)
 			}
-			Self::Read(msg) => (Self::Read(msg.clone()), msg),
+			Self::Read(msg) => (Self::Read(msg.clone()), Error::RestartNeeded(msg)),
 		}
 	}
 }
@@ -82,6 +84,7 @@ pub struct WsClient {
 	/// Channel to send requests to the background task.
 	to_back: mpsc::Sender<FrontToBack>,
 	/// If the background thread terminates the error is sent to this channel.
+	// NOTE(niklasad1): This is a Mutex because the `Client` is Clone.
 	error: Arc<Mutex<ErrorFromBack>>,
 	/// Request timeout
 	request_timeout: Option<Duration>,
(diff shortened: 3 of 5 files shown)


ws-client/src/client.rs
```diff
@@ -115,15 +150,33 @@ impl WsClient {
...
+		let (err_tx, err_rx) = oneshot::channel();
 
 		let (sender, receiver) = jsonrpc_transport::websocket_connection(config.clone())
 			.await
 			.map_err(|e| Error::TransportError(Box::new(e)))?;
 
 		async_std::task::spawn(async move {
-			background_task(sender, receiver, from_front, max_capacity_per_subscription, max_concurrent_requests).await;
+			background_task(
+				sender,
```
niklasad1, line 161: maybe introduce a new type for all this params ^^

ws-client/src/client.rs
```diff
@@ -40,17 +41,53 @@ use jsonrpsee_types::{
...
+			Self::Read(msg) => (Self::Read(msg.clone()), Error::RestartNeeded(msg)),
+		}
+	}
+}
+
+/// WebSocket client that works by maintaining a background task running in parallel.
 ///
-/// > **Note**: This struct is designed to be easy to use, but it works by maintaining a background task running in parallel.
-#[derive(Clone, Debug)]
+/// It's possible that background thread gets terminated which makes the client unusable.
```
niklasad1, line 81:
```suggestion
/// It's possible that the background thread gets terminated that makes the client unusable.
```

types/src/error.rs
```diff
@@ -30,6 +30,9 @@ pub enum Error {
 	/// Frontend/backend channel error.
 	#[error("Frontend/backend channel error: {0}")]
 	Internal(#[source] futures::channel::mpsc::SendError),
+	/// The background task has been terminated.
+	#[error("The background task been terminated because: {0}; restart required")]
+	RestartNeeded(String),
```
niklasad1, line 35: Could be  `Box<Error>` but since `Error` is not `Clone` it doesn't help much.

ws-client/src/client.rs
```diff
@@ -40,17 +41,53 @@ use jsonrpsee_types::{
...
+/// It's possible that the background thread gets terminated that makes the client unusable.
+/// An error `Error::RestartNeeded` is returned if that occurs and you have to manually
+/// handle that such as dropping client.
+#[derive(Debug)]
 pub struct WsClient {
 	/// Channel to send requests to the background task.
 	to_back: mpsc::Sender<FrontToBack>,
+	/// If the background thread terminates the error is sent to this channel.
+	// NOTE(niklasad1): This is a Mutex to circumvent that the async fns takes immutable references.
+	error: Arc<Mutex<ErrorFromBack>>,
```
maciejhirsz, line 90: Does it still need the `Arc` here?

niklasad1: no, nice catch :1st_place_medal:

maciejhirsz (approved):

tests/src/proc_macros.rs
```diff
@@ -1,3 +1,6 @@
+// Not all proc macros are used let's suppress it for now.
+#![allow(dead_code)]
```
dvdplm, line 2: Isn't there an `unused_macros` lint we can use instead?

ws-client/src/client.rs
```diff
@@ -40,17 +41,52 @@ use jsonrpsee_types::{
...
+
+impl ErrorFromBack {
+	async fn read_error(self) -> (Self, Error) {
+		match self {
+			Self::Unread(rx) => {
+				let msg = match rx.await {
+					Ok(msg) => msg.to_string(),
+					// This should never happen because the receiving end is still alive.
+					// Would be a bug in the logic of the background task.
+					Err(_) => "Error reason could not be found".to_string(),
```
dvdplm, line 68:
```suggestion
					Err(_) => "Error reason could not be found. This is a bug. Please open an issue.".to_string(),
```

ws-client/src/client.rs
```diff
@@ -40,17 +41,52 @@ use jsonrpsee_types::{
...
+			Self::Read(msg) => (Self::Read(msg.clone()), Error::RestartNeeded(msg)),
+		}
+	}
+}
+
+/// WebSocket client that works by maintaining a background task running in parallel.
 ///
-/// > **Note**: This struct is designed to be easy to use, but it works by maintaining a background task running in parallel.
-#[derive(Clone, Debug)]
+/// It's possible that the background thread gets terminated that makes the client unusable.
```
dvdplm, line 80:
```suggestion
/// It's possible that the background thread is terminated and this makes the client unusable.
```

ws-client/src/client.rs
```diff
@@ -40,17 +41,52 @@ use jsonrpsee_types::{
...
+/// An error `Error::RestartNeeded` is returned if that occurs and you have to manually
+/// handle that such as dropping client.
```
dvdplm, line 82:
```suggestion
/// An error `Error::RestartNeeded` is returned if this happens and users has to manually
/// handle dropping and restarting a new client.
```

ws-client/src/client.rs
dvdplm: Maybe, but it's not too bad yet. Let's hold off on that for now?

dvdplm (approved):

tests/src/proc_macros.rs
niklasad1: I think it only works for `declarative macors (macro_rules!)`, didn't work in this case

ws-client/src/client.rs
niklasad1: yeah, let's ignore for now :)

PR #244, opened 2021-03-15
[ws client]: unsubscribe direct when subscription is dropped.

niklasad1:
When a subscription is dropped because of `max_notifs_per_subscription` is exceeded
the client didn't send an `unsubscribe request` until the subscription was dropped which this commit fixes.

This causes the server to still send messages to the subscription and the client would get lots of
error messages: `InvalidRequestId` because it's not interested in the `subscription` anymore (removed from request manager)

### DIFF
--- a/ws-client/src/client.rs
+++ b/ws-client/src/client.rs
@@ -350,17 +350,11 @@ async fn background_task(
 				log::trace!("Closing subscription: {:?}", sub_id);
 				// NOTE: The subscription may have been closed earlier if
 				// the channel was full or disconnected.
-				if let Some((_, unsub_method)) = manager
+				if let Some(unsub_request) = manager
 					.get_request_id_by_subscription_id(&sub_id)
-					.and_then(|req_id| manager.remove_subscription(req_id, sub_id.clone()))
+					.and_then(|req_id| build_unsubscribe_request(&mut manager, req_id, sub_id))
 				{
-					let json_sub_id = jsonrpc::to_value(sub_id).expect("SubscriptionID to JSON is infallible; qed");
-					let request = RequestMessage {
-						method: unsub_method,
-						params: jsonrpc::Params::Array(vec![json_sub_id]),
-						send_back: None,
-					};
-					send_unsubscribe_request(&mut sender, &mut manager, request).await;
+					send_unsubscribe_request(&mut sender, &mut manager, unsub_request).await;
 				}
 			}
 			Either::Right((Some(Ok(jsonrpc::Response::Single(response))), _)) => {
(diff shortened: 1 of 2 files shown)


ws-client/src/client.rs
```diff
@@ -466,11 +459,8 @@ fn process_response(
 				match send_back_oneshot.send(Ok((subscribe_rx, sub_id.clone()))) {
```
dvdplm, line 459: Github says "redundant clone" on line 449 here.

ws-client/src/client.rs
```diff
@@ -488,6 +478,17 @@ async fn send_unsubscribe_request(
...
+
+fn build_unsubscribe_request(
+	manager: &mut RequestManager,
+	req_id: u64,
+	sub_id: SubscriptionId,
+) -> Option<RequestMessage> {
+	let (_, unsub) = manager.remove_subscription(req_id, sub_id.clone())?;
+	manager.reclaim_request_id(req_id);
+	let json_sub_id = jsonrpc::to_value(sub_id).expect("SubscriptionId to JSON is infallible; qed");
+	Some(RequestMessage { method: unsub, params: jsonrpc::Params::Array(vec![json_sub_id]), send_back: None })
```
dvdplm, line 493: Maybe we should have a `params!` macro to reduce the noise a bit:

`params![json_sub_id]` vs `jsonrpc::Params::Array(vec![json_sub_id])`

ws-client/src/client.rs
```diff
@@ -488,6 +478,17 @@ async fn send_unsubscribe_request(
...
+		log::error!("Send unsubscribe request failed: {:?}", e);
 	}
 }
+
+fn build_unsubscribe_request(
+	manager: &mut RequestManager,
+	req_id: u64,
+	sub_id: SubscriptionId,
+) -> Option<RequestMessage> {
+	let (_, unsub) = manager.remove_subscription(req_id, sub_id.clone())?;
```
dvdplm, line 490: This `sub_id.clone()` is a bit unfortunate.

ws-client/src/client.rs
```diff
@@ -488,6 +478,17 @@ async fn send_unsubscribe_request(
 	request: RequestMessage,
 ) {
 	if let Err(e) = sender.start_request(request, manager).await {
-		log::error!("send unsubscribe request failed: {:?}", e);
+		log::error!("Send unsubscribe request failed: {:?}", e);
 	}
 }
+
+fn build_unsubscribe_request(
```
dvdplm, line 485: "request" is an inflated term in this library (any networking library I guess) and I wonder if a better name would be `build_unsubscribe_message`? Like: use "request" for actual users making a call for data from some server and "message" for internal communication between components?

dvdplm (approved):
nits and questions

ws-client/src/client.rs
niklasad1: might go away with `maciejs serde raw value` design :)

ws-client/src/client.rs
niklasad1: Yeah, it's because the `RequestManager` is using the HashMap entry api internally.

Clone should be cheap though, `u64` or `String`

maciejhirsz (approved):
LGTM

PR #245, opened 2021-03-15
[http client]: request ID `Ordering::Relaxed`

niklasad1:
Request ID is just a counter and doesn't require any synchronization besides an atomic value.

Fixes https://github.com/paritytech/jsonrpsee/pull/199#discussion_r575219903

### DIFF
--- a/http-client/src/client.rs
+++ b/http-client/src/client.rs
@@ -49,7 +49,7 @@ impl Client for HttpClient {
 		P: Into<jsonrpc::Params> + Send,
 	{
 		// NOTE: `fetch_add` wraps on overflow which is intended.
-		let id = self.request_id.fetch_add(1, Ordering::SeqCst);
+		let id = self.request_id.fetch_add(1, Ordering::Relaxed);
 		let request = jsonrpc::Request::Single(jsonrpc::Call::MethodCall(jsonrpc::MethodCall {
 			jsonrpc: jsonrpc::Version::V2,
 			method: method.into(),


http-client/src/client.rs
```diff
@@ -49,7 +49,7 @@ impl Client for HttpClient {
 		P: Into<jsonrpc::Params> + Send,
 	{
 		// NOTE: `fetch_add` wraps on overflow which is intended.
-		let id = self.request_id.fetch_add(1, Ordering::SeqCst);
+		let id = self.request_id.fetch_add(1, Ordering::Relaxed);
```
dvdplm, line 52: This is ok because:
1. it is the only place the request_id is written in this method
2. even if another addition were to happen on a separate thread we would still be fine
3. even is both 1 and 2 turned out to be false it would not be a catastrophic failure because it would not lead to a crash or to wrong data being sent to the wrong user or some other **bad** badness; it would just mean that this request might get mixed up with another one.

Did I get that right-ish?

dvdplm (approved):

http-client/src/client.rs
niklasad1: Sort of :)

Basically, as long as `request ID` is used standalone it still atomic and is safe as long it's not used to reference memory or to do some other synchronization.

For example this would would be bad with `relaxed`

```rust
   let mut secret_data = [1, 2, 3, 4, 5];
   let lock = AtomicBool(false);
   
   while lock.compare_and_swap(false, true, Ordering::Relaxed) {
     // DON'T DO THIS: secret_data isn't ensured to be synchronized but the lock is still Atomic
     secret_data = write_some_secret();
   }
```

This on the other hand is safe but because all operations after Release/Acquire would be ordered
and that includes non-atomic variables too:

```rust
   let mut secret_data = [1, 2, 3, 4, 5];
   let lock = AtomicBool(false);
   
   while lock.compare_and_swap(false, true, Ordering::Acquire) {
     // OK: secret_data is synchronized after Release
     secret_data = write_some_secret();
   }
```

niklasad1: This would just be faster on `weakly ordered platforms` such as ARM so we could just ignore for now too.

PR #260, opened 2021-04-02
add `jsonrpsee` root crate

### LINKED ISSUE #214: [repo org]: add root/facade crate `jsonrpsee`
We should provide a wrapper crate that exports all internal crates such that users doesn't have manually import each indiviudal crate, https://github.com/paritytech/jsonrpsee/pull/177#pullrequestreview-563912962

niklasad1:
Closes #214


proc-macros/src/lib.rs
```diff
@@ -225,27 +225,27 @@ fn build_client_functions(api: &api_def::ApiDefinition) -> Result<Vec<proc_macro
 			params_to_json.push(quote_spanned!(pat_span=>
 				map.insert(
 					#rpc_param_name.to_string(),
-					jsonrpsee_types::jsonrpc::to_value(#generated_param_name.into()).map_err(|e| jsonrpsee_types::error::Error::Custom(format!("{:?}", e)))?
+					jsonrpsee::types::jsonrpc::to_value(#generated_param_name.into()).map_err(|e| jsonrpsee::types::error::Error::Custom(format!("{:?}", e)))?
```
niklasad1, line 228: note, this will only work when using it from `jsonrpsee` root crate which is the most likely use-case.

worth fixing this for when using the internal crate `jsonrpsee_proc_macros` individually?!

jsonrpsee/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "jsonrpsee"
+description = "JSON-RPC crate"
+version = "0.2.0-alpha.4"
+authors = ["Parity Technologies <admin@parity.io>", "Pierre Krieger <pierre.krieger1708@gmail.com>"]
+license = "MIT"
+edition = "2018"
+
+[dependencies]
+http-client = { path = "../http-client", version = "0.2.0-alpha.4", package = "jsonrpsee-http-client" }
```
niklasad1, line 10: make it possible to use tokio02?

jsonrpsee/src/lib.rs
```diff
@@ -0,0 +1,9 @@
+//! jsonrpsee wrapper crate.
+
+pub use http_client;
```
niklasad1, line 3: feature gate all internal crates?!

dvdplm: You mean like `ws-client`, `ws-server`, `http-client`,`http-server`? Hmm. I suspect most users will want either client or server, but many will want both transports?
Good topic for a conversation with @maciejhirsz too.

proc-macros/src/lib.rs
dvdplm: I think it's fine and even good to make the `jsonrpsee::types` crate only usable with the root crate. We don't really want to stick to semver for the `types` crate anyway; it's really more of an internal-use crate.

examples/Cargo.toml
```diff
@@ -4,17 +4,24 @@ version = "0.1.0"
...
+name = "http"
+path = "http.rs"
+
+[[example]]
+name = "ws"
+path = "ws.rs"
+
+[[example]]
+name = "ws_subscription"
+path = "ws_subscription.rs"
```
dvdplm, line 27: Isn't `path` enough?

jsonrpsee/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+http-server = { path = "../http-server", version = "0.2.0-alpha.4", package = "jsonrpsee-http-server" }
+ws-client = { path = "../ws-client", version = "0.2.0-alpha.4", package = "jsonrpsee-ws-client" }
+ws-server = { path = "../ws-server", version = "0.2.0-alpha.4", package = "jsonrpsee-ws-server" }
+types = { path = "../types", version = "0.2.0-alpha.4", package = "jsonrpsee-types" }
+utils = { path = "../utils", version = "0.2.0-alpha.4", package = "jsonrpsee-utils" }
+proc-macros = { path = "../proc-macros", version = "0.2.0-alpha.4", package = "jsonrpsee-proc-macros" }
```
dvdplm, line 16: Should this be renamed to `jsonrpsee-derive` perhaps?

jsonrpsee/src/lib.rs
niklasad1: yepp, my first intention was that `jsonrpsee` will include all and if people want only to use internal crates they can import them individually but it might be weird since `proc-macros` only works from `jsonrpsee crate`. Then the linker would just strip off all unused stuff but the crates will be downloaded still...

I think `server`, `client` and `macros` would be enough and for more fine-granularity then use individual crates.

examples/Cargo.toml
niklasad1: No unfortunately not, `name` is required and the standard path is the `/example`.

So for it work one would have to another examples folder as we had before sadly

jsonrpsee/Cargo.toml
niklasad1: I think the `proc-macros` is more accurate because my intuition is that `derive` implies `custom derive macro` but maybe it's just me ^^

I don't have strong opinions on it, //cc @maciejhirsz ^

proc-macros/src/lib.rs
niklasad1: Fixed that by using [Basti's proc macro crate](https://docs.rs/proc-macro-crate/1.0.0/proc_macro_crate/) instead

jsonrpsee/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+authors = ["Parity Technologies <admin@parity.io>", "Pierre Krieger <pierre.krieger1708@gmail.com>"]
+license = "MIT"
+edition = "2018"
+
+[dependencies]
+http-client = { path = "../http-client", version = "0.2.0-alpha.4", package = "jsonrpsee-http-client", optional = true }
+http-server = { path = "../http-server", version = "0.2.0-alpha.4", package = "jsonrpsee-http-server", optional = true }
+ws-client = { path = "../ws-client", version = "0.2.0-alpha.4", package = "jsonrpsee-ws-client", optional = true }
+ws-server = { path = "../ws-server", version = "0.2.0-alpha.4", package = "jsonrpsee-ws-server", optional = true }
+types = { path = "../types", version = "0.2.0-alpha.4", package = "jsonrpsee-types", optional = true }
```
niklasad1, line 14: we could remove `types` and re-export those from the clients instead.

Such as `traits` and the `types` used in the normal API/traits.

jsonrpsee/Cargo.toml
dvdplm: My general rule of thumb for naming things is that they should give a hint about what something **does**, rather than describe what it **is**. So a datastructure containing customer information implemented as a  linked list should be named "customer_info" and not "customer_linked_list". Here `proc-macros` tells the reader what it is, but gives no hint about what it does.

Let's yak shave this name and come up with something better before `0.2`. Not a blocker for this PR.

jsonrpsee/Cargo.toml
dvdplm: I agree, let's do that.

proc-macros/Cargo.toml
```diff
@@ -13,4 +13,5 @@ proc-macro = true
 Inflector = "0.11.4"
 proc-macro2 = "1.0"
 quote = "1.0"
-syn = { version = "1.0", features = ["full", "extra-traits", "parsing", "printing", "proc-macro", "visit"] }
\ No newline at end of file
+syn = { version = "1.0", features = ["full", "extra-traits", "parsing", "printing", "proc-macro", "visit"] }
```
dvdplm, line 16:
```suggestion
syn = { version = "1.0", features = ["full", "extra-traits", "visit"] }
```

"parsing", "printing" and  "proc-macro" are enabled by default I think. Alternatively we can be explicit and do:

`syn = { version = "1.0", features = ["full", "extra-traits", "parsing", "printing", "proc-macro", "visit"], default-features = false }` (but maybe we need `clone-impls` too then?

proc-macros/src/lib.rs
```diff
@@ -189,6 +191,12 @@ fn build_client_impl(api: &api_def::ApiDefinition) -> Result<proc_macro2::TokenS
 fn build_client_functions(api: &api_def::ApiDefinition) -> Result<Vec<proc_macro2::TokenStream>, syn::Error> {
 	let visibility = &api.visibility;
 
+	let _jsonrpsee_types = match crate_name("jsonrpsee-types") {
+		Ok(FoundCrate::Itself) => unreachable!(),
```
dvdplm, line 195:
```suggestion
		Ok(FoundCrate::Itself) => panic!("Deriving RPC methods in the `types` crate is not supported"),
```

dvdplm (approved):
Left a few nits.

proc-macros/Cargo.toml
niklasad1: cool, removed the unused ones.

proc-macros/src/lib.rs
```diff
@@ -189,6 +191,14 @@ fn build_client_impl(api: &api_def::ApiDefinition) -> Result<proc_macro2::TokenS
 fn build_client_functions(api: &api_def::ApiDefinition) -> Result<Vec<proc_macro2::TokenStream>, syn::Error> {
 	let visibility = &api.visibility;
 
+	let _crate = match (crate_name("jsonrpsee-http-client"), crate_name("jsonrpsee-ws-client")) {
```
niklasad1, line 194: this is a hack just to find a client crate to get the types from, not sure if there's a nicer way to do without explicitly import the types crate.

proc-macros/Cargo.toml
```diff
@@ -10,7 +10,9 @@ edition = "2018"
 proc-macro = true
 
 [dependencies]
+jsonrpsee-types = { version = "0.2.0-alpha.4", path = "../types" }
```
niklasad1, line 13:
```suggestion
```

proc-macros/src/lib.rs
```diff
@@ -189,6 +191,14 @@ fn build_client_impl(api: &api_def::ApiDefinition) -> Result<proc_macro2::TokenS
...
+		(Ok(FoundCrate::Name(name)), _) => syn::Ident::new(&name, Span::call_site()),
+		(_, Ok(FoundCrate::Name(name))) => syn::Ident::new(&name, Span::call_site()),
+		(_, Err(e)) => return Err(syn::Error::new(Span::call_site(), &e)),
```
niklasad1, line 197: right, I think I need to support `FoundCrate::Itself` now because it should be possible to just use the one of the client crates and proc_macros ^^

jsonrpsee/src/lib.rs
maciejhirsz: I don't know how many people would use both transports, my hunch is most people (for clients at least) would only need one, but I have no data to back it so I reckon the current granularity is fine. We can always come back and add separate feature flags if need be.

maciejhirsz (approved):
LGTM

proc-macros/src/lib.rs
niklasad1: nvm I was wrong

Issue #227, opened 2021-03-05
[http client]: switch to surf as backend


niklasad1:
We would like to switch to surf for a couple of reasons:
- provides compat layer between `tokio` and `async-std`
- comes with TLS via rustls
- makes it compile for `WASM`
- reduces the number of `conditional compilation` for jsonrpsee itself

Currently blocked on https://github.com/http-rs/surf/issues/295

niklasad1:
Closing this we already have `web transport` which is sufficient

PR #261, opened 2021-04-02
[types] Remove unnecessary deps

c410-f3r:


types/src/v2/error.rs
```diff
@@ -1,16 +1,33 @@
...
-	#[error("invalid params")]
 	InvalidParams,
 }
 
+impl fmt::Debug for RpcError {
+	#[inline]
+	fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+		match *self {
+			Self::Unknown => write!(f, "unknown rpc error"),
+			Self::InvalidParams => write!(f, "invalid paramsr"),
```
niklasad1, line 16:
```suggestion
			Self::InvalidParams => write!(f, "invalid params"),
```

niklasad1 (commented):
Thanks for your PR.

Looks overall good but I don't really understand the rationale behind making this crate `no-std` without the client traits then it's not super useful, such as the proc macros wouldn't work. I suppose making the `SubscriptionClient trait` std only (subscription type) and make `Client trait` no-std would do it.

Can you elaborate why making the `no-std` is useful?

c410-f3r:
`no_std` support is usually a complementary and natural step when removing unnecessary dependencies. Besides, isn't it something desirable in the long run?

If necessary, I can remove `no_std` stuff. No problem.

niklasad1:
> no_std support is usually a complementary and natural step when removing unnecessary dependencies. Besides, isn't it something desirable in the long run?

Sure, why not it introduces some boiler plate code for the `error stuff` but worth it if somebody wants to use the types in `no-std` but we have no plans making client/server to support for `no-std` but we have plans to move the [HTTP client to use surf as backend for WASM support](https://github.com/paritytech/jsonrpsee/issues/227)

niklasad1 (changes requested):
LGTM except the typo, see inline suggestion.

niklasad1 (approved):

c410-f3r:
IIRC, `WASM` language isn't aware of `std` stuff (filesystem, threads, timer, etc) and that is one of the reasons for the creation of `WASI`.

niklasad1:
> IIRC, WASM language isn't aware of std stuff (filesystem, threads, timer, etc) and that is one of the reasons for the creation of WASI.

Yes, but it's possible to do with `JavaScript bindings` too as [surf/http-client](https://github.com/http-rs/http-client) does.

c410-f3r:
Cool! Thanks for the link

types/src/error.rs
```diff
@@ -16,62 +18,110 @@ impl<T: fmt::Display> fmt::Display for Mismatch<T> {
...
 	/// Configured max number of request slots exceeded.
-	#[error("Configured max number of request slots exceeded")]
 	MaxSlotsExceeded,
 	/// Custom error.
-	#[error("Custom error: {0}")]
 	Custom(String),
 }
 
+impl fmt::Debug for Error {
+	#[inline]
```
dvdplm, line 56: Let's not try to outsmart the compiler; formatting errors is not on the hot path for anything so please remove the inlining hints.

types/src/error.rs
```diff
@@ -16,62 +18,110 @@ impl<T: fmt::Display> fmt::Display for Mismatch<T> {
...
+			}
+			Self::WsRequestTimeout => write!(f, "Websocket request timeout"),
+			Self::MaxSlotsExceeded => write!(f, "Configured max number of request slots exceeded"),
+			Self::Custom(ref elem) => write!(f, "Custom error: {}", elem),
+		}
+	}
+}
+
+impl fmt::Display for Error {
+	#[inline]
```
dvdplm, line 86: same

dvdplm (changes requested):
Reducing compile times is great, thank you so much for looking into this.

I agree with @niklasad1 that `no-std` is not a goal and adds more noise than it's worth imo. I also think that the added boilerplate around errors is questionable. I prefer it the way it was.

Any chance we can keep the compile time improvements without the inlining, no-std, and error re-implementations?

types/src/v2/traits.rs
```diff
@@ -11,5 +11,7 @@ impl<R, T> RpcMethod<R> for T where T: Fn(RpcParams) -> Result<R, RpcError> + Se
 
 /// RPC Call Result.
 trait RpcResult {
```
niklasad1, line 13: please merge with master, this was removed recently.

niklasad1:
@dvdplm please take another look at this PR ^^

dvdplm (approved):

PR #264, opened 2021-04-06
refactor utils crate

niklasad1:
Move the `access control` stuff the HTTP server since it's not shared and make all mods in `utils` optional to not bring in any extra deps.


http-server/src/lib.rs
```diff
@@ -24,11 +24,12 @@
 // IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 // DEALINGS IN THE SOFTWARE.
 
+mod access_control;
 mod module;
 mod response;
 mod server;
 
-pub use jsonrpsee_utils::http::access_control::{AccessControl, AccessControlBuilder};
+pub use access_control::{AccessControl, AccessControlBuilder, AllowHosts, Host};
```
niklasad1, line 32: these re-exports are needed to configure the AccessControl properly (i.e, non-default ^^)

http-server/src/access_control/hosts.rs
```diff
@@ -169,17 +168,6 @@ pub fn is_host_valid(host: Option<&str>, allow_hosts: &AllowHosts) -> bool {
 	}
 }
 
-/// Updates given list of hosts with the address.
-pub fn update(hosts: Option<Vec<Host>>, address: &SocketAddr) -> Option<Vec<Host>> {
-	hosts.map(|current_hosts| {
-		let mut new_hosts = current_hosts.into_iter().collect::<HashSet<_>>();
```
dvdplm, line 175: Clippy was really angry about this one. :)

dvdplm (approved):

Issue #254, opened 2021-03-28
hyper13 and hyper14 compatibility


akru:
Looks like the latest jsonrpsee has some compatibility issues with fresh hyper.

```
   Compiling jsonrpsee-utils v0.2.0-alpha.3                                                                                    
   Compiling hyper-rustls v0.20.0                                                                                              
   Compiling hyper-multipart-rfc7578 v0.4.0-rc                                                                                 
error: feature `hyper13` and `hyper14` are mutably exclusive                                                                   
 --> /home/akru/.cargo/registry/src/github.com-1ecc6299db9ec823/jsonrpsee-utils-0.2.0-alpha.3/src/lib.rs:2:1                   
  |                                                                                                                            
2 | compile_error!("feature `hyper13` and `hyper14` are mutably exclusive");   
```

dvdplm:
Can you show your full Cargo.toml? Could it be your project has multiple versions of `hyper` in it?

niklasad1:
Please share your `Cargo.toml`, my guess it that you forgot to disable the default-features if you want to use `hyper13`.

Otherwise, it's bug on our end but I would surprised as [we test it in the CI](https://github.com/paritytech/jsonrpsee/blob/master/.github/workflows/ci.yml##L73-L77)

akru:
I use jsonrpsee as a dependency of https://github.com/paritytech/substrate-subxt (crates package version `0.15`). I guess it conflicts with something from https://github.com/airalab/robonomics, maybe https://docs.rs/ipfs-api/0.11.0/ipfs_api/

niklasad1:
Ok, I see.

`subxt` uses tokio1.0/hyper14 and `robonoimcs (substrate)` uses tokio0.2/hyper13 thus you will end-up with both `hyper13 and hyper14` features which causes it not to compile.

I suppose `subxt` can make it configurable to work with `tokio 0.2 or tokio 1.0` until the entire ecosystem has upgraded to tokio 1.0

niklasad1:
Closing this as this works as intended i.e, `tokio0.2` and `tokio1` is mutually exclusive and only one should be enabled.

Please file an issue in https://github.com/paritytech/substrate-subxt instead to make it configurable with `tokio 0.2`

PR #265, opened 2021-04-08
Derive `Clone` for `WsClient`

HCastano:
In the [`parity-bridges-common`](https://github.com/paritytech/parity-bridges-common) we currently rely on `WsClient` being `Clone`. This was removed 
in `jsonrpsee-ws-client` `v0.2.0-alpha.3`, which makes upgrading from `alpha.2` a breaking
change for us.

This PR makes `WsClient` `Clone`-able again.

### DIFF
--- a/ws-client/src/client.rs
+++ b/ws-client/src/client.rs
@@ -43,6 +43,7 @@ use jsonrpsee_types::{
 	traits::{Client, SubscriptionClient},
 };
 use std::marker::PhantomData;
+use std::sync::Arc;
 use std::time::Duration;
 use std::{borrow::Cow, convert::TryInto};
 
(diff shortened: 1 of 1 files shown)


niklasad1 (commented):
Hey Hernando,

We don't respect semver in the alpha releases,

> A pre-release version indicates that the version is unstable and might not satisfy the intended compatibility requirements as denoted by its associated normal version`

The client is not `Clone` anymore see https://github.com/paritytech/jsonrpsee/pull/243#discussion_r594153898. Please convince me why it should be `Clone`

Can't you pin it to `=alpha.2` in your `Cargo.toml`?

HCastano:
> Hey Hernando,
> 
> > which makes upgrading from alpha.2 a breaking
> > change for us.
> 
> [pre-releases may have breaking changes according to semver](https://semver.org/#spec-item-9) however it's not a great user-experience I guess but I thought `x.x.x-alpha` was sufficient to indicate that it may contain breaking changes.

That's a fair point. I guess I'm just used to expecting stability between releases in general.

> The client is not `Clone` anymore see [#243 (comment)](https://github.com/paritytech/jsonrpsee/pull/243#discussion_r594153898). Please convince me why it should be `Clone`

I don't have a good answer other than keeping the status quo. I'll give a go to `Arc`-ifying stuff on our end and close this PR if that works.

> Can't you pin it to `=alpha.2` in your `Cargo.toml`?

No. This is because we want to be integrated with the Polkadot repo, which is already using `alpha.3` (which comes from Substrate).

niklasad1:
> No. This is because we want to be integrated with the Polkadot repo, which is already using alpha.3 (which comes from Substrate).

Ok, I see watch out for https://github.com/paritytech/jsonrpsee/issues/254 then because `substrate` uses the feature `tokio02` on the http client.

HCastano:
Closing in favour of https://github.com/paritytech/parity-bridges-common/pull/892.

PR #268, opened 2021-04-12
Apease Clippy and some renames

dvdplm:


types/src/v2/traits.rs
```diff
@@ -11,5 +11,5 @@ impl<R, T> RpcMethod<R> for T where T: Fn(RpcParams) -> Result<R, RpcError> + Se
 
 /// RPC Call Result.
 trait RpcResult {
-	fn to_json(self, id: Option<&RawValue>) -> anyhow::Result<String>;
+	fn into_json(self, id: Option<&RawValue>) -> anyhow::Result<String>;
```
dvdplm, line 14: There are no impls of this trait anywhere, which is a bit odd. Do we need this trait @niklasad1 ?

niklasad1: No, I removed it in another pending PR.

Just a nit when moving those types from `ws server`

http-server/src/access_control/cors.rs
```diff
@@ -49,22 +49,22 @@ pub enum OriginProtocol {
 pub struct Origin {
 	protocol: OriginProtocol,
 	host: Host,
-	as_string: String,
+	host_with_proto: String,
```
niklasad1, line 52: almost like we should refactor this to a separate type ^^ :)

I think this entire `access_control module` needs to checked a little bit more carefully, I don't remember but this matcher stuff could maybe be replaced by the `url crate` where they have functions for parsing url related stuff.

niklasad1 (approved):
:+1:

Issue #249, opened 2021-03-18
[ws client]: better request ID and batch ID handling


niklasad1:
> I think a batch request containing n requests should count against the max_concurrent_request_limit as n. In my head the max_concurrent_request_limit setting should default to a fairly high value – 1000? hundreds? – and it should be thought of as a safety valve and not really as a throttling mechanism.

Indeed and this makes it more complicated as request ID and batch ID (request ID of request within the batch) needs to be distinguished. So, I'm not super excited about this and it's error prone to have return the request ID after each request has been completed.

Maybe better just with a simple counter and guard the hashmap/other collections when inserting new requests.
Also as additional safety check is to introduce new types for `request ID` and `batch ID` which wraps the actual ID such that it can only be generated the `RequestManager` and that `request ID` and `batch ID` doesn't get mixed unintentionally.

_Originally posted by @niklasad1 in https://github.com/paritytech/jsonrpsee/issues/216#issuecomment-801302950_

niklasad1:
Closed by #269

Issue #225, opened 2021-03-04
[ws client]: `max_concurrent_requests` doesn't really work


niklasad1:
As the all methods on the `client` for doing notifications/requests/subscriptions takes an immutable borrow to the client itself to work across threads (to be Send + Sync) forces us to clone the sender where the capacity is `num_senders + capacity`, it is documented [here](https://github.com/paritytech/jsonrpsee/blob/master/ws-client/src/client.rs#L82-#L85). Thus, it increases on every request and is not bounded in practice.

So I suggest to introduce semaphore/AtomicInt as bound instead of using `futures::mpsc::Sender` for that.

niklasad1:
Closed by #269

Issue #219, opened 2021-02-23
Homogenize WS and HTTP servers and clients


maciejhirsz:
Follow up to #174, the new ws-server introduced a set of types to speed up serialization and deserialization of the JSON, it would be nice if those can be re-used, and having the API somewhat similar.

- [x] Make http and ws servers use the same `JsonRpcX` structs.
- [x] Make the servers and clients use the same `JsonRpcX` structs.
- [x] Make the API between http and ws servers same-ish.

niklasad1:
Closed by #269

PR #269, opened 2021-04-12
[client] use types v2 (less alloc)

niklasad1:
Migrate to `jsonrpsee_types::v2` which causes lots of related changes.

The major change is the client traits and related `JsonRpcParam` type, 

```rust
enum JsonRpcParams<'a> {
    NoParams,
    Array(Vec<JsonValue>]),
    ArrayRef(&'a [JsonValue]),
    Map(BTreeMap<&'a str, JsonValue>),
}

fn request<'a>(method: &'str, params: JsonRpcParams<'a>) { .... } 

```

The changes were easy to port to `http client` but with `websocket client` I had to major refactoring and decided to make the transport abstraction clean i.e, doesn't know anything about `jsonrpc` it just sends messages. Thus, the clients are responsible for constructing the messages.

- [x] re-org types::v2 module perhaps request/response or serialize/deserial or something.- 
 
Follow up:
- macro for constructing `JsonRpcParams` (#275)
- smallvec instead of Vec for batch requests.
- write/port the serialize/deserialize tests for the types (#276)

Closing #249 #219 #225


benches/benches/bench.rs
```diff
@@ -1,15 +1,51 @@
...
+			}));
+			v1_serialize(request);
+		})
+	});
+}
+
+pub fn jsonrpsee_types_v2(crit: &mut Criterion) {
+	crit.bench_function("jsonrpsee_types_v2", |b| {
+		b.iter(|| {
+			let params: JsonRpcParams<_> = vec![&1, &2].into();
```
niklasad1, line 42: provide a macro for this:... 

params_vec![1, 2, 3], params_map!["key => val]

http-client/src/transport.rs
```diff
@@ -65,24 +68,33 @@ impl HttpTransportClient {
...
 
 	/// Send request and wait for response.
-	pub async fn send_request_and_wait_for_response(
+	pub async fn send_request_and_wait_for_response<'a, T, R>(
 		&self,
-		request: jsonrpc::Request,
-	) -> Result<jsonrpc::Response, Error> {
-		let response = self.send_request(request).await?;
+		request: impl Into<JsonRpcRequest<'a, T>>,
+	) -> Result<JsonRpcResponse<R>, Error>
```
niklasad1, line 84: maybe just return `Vec<u8>` I suppose I could use `serde::RawValue` then..

http-client/src/client.rs
```diff
@@ -48,115 +45,96 @@ pub struct HttpClient {
...
-	async fn batch_request<T, M, P>(&self, batch: Vec<(M, P)>) -> Result<Vec<T>, Error>
+	async fn batch_request<'a, R>(&self, batch: Vec<(&'a str, JsonRpcParams<'a>)>) -> Result<Vec<R>, Error>
 	where
-		T: DeserializeOwned + Default + Clone,
-		M: Into<String> + Send,
-		P: Into<jsonrpc::Params> + Send,
+		R: DeserializeOwned + Default + Clone,
 	{
-		let mut calls = Vec::with_capacity(batch.len());
+		let mut batch_request = Vec::with_capacity(batch.len());
```
niklasad1, line 92: todo: SmallVec?

jsonrpsee/Cargo.toml
```diff
@@ -12,6 +12,7 @@ http-server = { path = "../http-server", version = "0.2.0-alpha.4", package = "j
 ws-client = { path = "../ws-client", version = "0.2.0-alpha.4", package = "jsonrpsee-ws-client", optional = true }
 ws-server = { path = "../ws-server", version = "0.2.0-alpha.4", package = "jsonrpsee-ws-server", optional = true }
 proc-macros = { path = "../proc-macros", version = "0.2.0-alpha.4", package = "jsonrpsee-proc-macros", optional = true }
+types = { path = "../types", version = "0.2.0-alpha.4", package = "jsonrpsee-types", optional = true }
```
niklasad1, line 15: I had to include this again, if not  any of the `client crates` is in the users `Cargo.toml` it won't work.

That's the reason why I moved the tests to separate crate, is just worked before because it was in the same crate.

proc-macros/Cargo.toml
```diff
@@ -13,5 +13,5 @@ proc-macro = true
 Inflector = "0.11.4"
 proc-macro2 = "1.0"
 quote = "1.0"
-syn = { version = "1.0", default-features = false, features = ["extra-traits"] }
+syn = { version = "1.0", default-features = false, features = ["extra-traits", "full"] }
```
niklasad1, line 16: faulty feature probably leaked and works in the root.

types/src/v2/error.rs
```diff
@@ -1,14 +1,24 @@
...
-	#[error("invalid params")]
-	InvalidParams,
+/// [JSON-RPC Error object](https://www.jsonrpc.org/specification#error_object)
+#[derive(Error, Debug, PartialEq, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct JsonRpcErrorObject {
+	/// Error code
+	pub code: ErrorCode,
+	/// Message
+	pub message: String,
```
niklasad1, line 13: This message is totally useless, the message is deduced from the error code AFAIU

However, it needs to be sent sent/received according the spec. Maybe include in ErrorCode deser/ser impls.

types/src/v2/mod.rs
```diff
@@ -160,3 +176,126 @@ impl<'a> RpcParams<'a> {
...
+/// construct the parameters. Because `serde_json::to_value` serializes the type which
+/// allocates whereas `Into<JsonValue>` doesn't in most cases.
+#[derive(Serialize, Debug)]
+#[serde(untagged)]
+pub enum JsonRpcParams<'a> {
+	/// No params.
+	NoParams,
+	/// Positional params (heap allocated)
+	Array(Vec<JsonValue>),
+	/// Positional params (slices)
```
niklasad1, line 192:
```suggestion
	/// Positional params (slice)
```

types/src/v2/mod.rs
```diff
@@ -160,3 +176,126 @@ impl<'a> RpcParams<'a> {
...
+	pub jsonrpc: TwoPointZero,
+	/// Params.
+	pub params: JsonRpcNotificationParamsAlloc<T>,
+}
+
+/// Id of a subscription, communicated by the server.
+#[derive(Debug, PartialEq, Clone, Hash, Eq, Deserialize, Serialize)]
+#[serde(deny_unknown_fields)]
+#[serde(untagged)]
+pub enum SubscriptionId {
```
niklasad1, line 276: I think we could kill this type and just use JsonValue as the subscription ID, alternatively reduce it to u64 as the servers does.

ws-client/Cargo.toml
```diff
@@ -8,12 +8,13 @@ license = "MIT"
 
 [dependencies]
 async-trait = "0.1"
-async-std = { version = "1.9", default-features = false, features = ["std"] }
+async-std = "1.9"
```
niklasad1, line 11: sorry I made I mistake here, didn't realize that when default-features are disabled, the `net` module is not enabled and can't be enabled by any feature flag https://github.com/async-rs/async-std/blob/master/src/lib.rs#L314-#L318

niklasad1: the feature is probably leaked from somewhere else and works in the workspace

ws-client/src/client.rs
```diff
@@ -90,6 +96,62 @@ pub struct WsClient {
...
+	fn next_request_id(&self) -> Result<u64, Error> {
+		self.get_slot()?;
+		let id = self.current_id.fetch_add(1, Ordering::Relaxed);
+		Ok(id)
+	}
+
+	/// Attempts to get the `n` number next IDs that only counts as one request.
+	///
+	/// Fails if request limit has been exceeded.
+	fn next_request_ids(&self, len: usize) -> Result<Vec<u64>, Error> {
```
niklasad1, line 140: todo: SmallVec?!

ws-client/src/client.rs
```diff
@@ -400,34 +465,56 @@ async fn background_task(
...
+
 			// User called `subscribe` on the front-end.
-			Either::Left((Some(FrontToBack::Subscribe(subscribe)), _)) => {
-				log::trace!("[backend]: client prepares to start subscription: {:?}", subscribe);
-				if let Err(e) = sender.start_subscription(subscribe, &mut manager).await {
+			Either::Left((Some(FrontToBack::Subscribe(sub)), _)) => match sender.send(sub.raw).await {
+				Ok(_) => manager
+					.insert_pending_subscription(
+						sub.subscribe_id,
+						sub.unsubscribe_id,
```
niklasad1, line 508: to be on the safe-side we should probably register the unsubscribe ID in request manager, such that it can't be used by another request concurrently such that the unsubscription request gets thrown away.

technically, that's very unlikely to happen but still the counter can wrap around re-use request IDs,

ws-client/src/client.rs
dvdplm: imo, not before we have benchmarks proving it's worth it.

proc-macros/src/lib.rs
```diff
@@ -310,3 +299,24 @@ fn rpc_param_name(pat: &syn::Pat, _attrs: &[syn::Attribute]) -> syn::parse::Resu
...
+
+/// Search for `jsonrpsee` in `Cargo.toml`.
+fn find_jsonrpsee_crate() -> Result<proc_macro2::TokenStream, syn::Error> {
+	match crate_name("jsonrpsee") {
+		Ok(FoundCrate::Name(name)) => {
+			let ident = syn::Ident::new(&name, Span::call_site());
+			Ok(quote!(#ident::types))
+		}
+		Ok(FoundCrate::Itself) => panic!("Deriving RPC methods in any of the `jsonrpsee crates` is not supported"),
+		Err(_) => match (crate_name("jsonrpsee-http-client"), crate_name("jsonrpsee-ws-client")) {
```
niklasad1, line 311: This was a bug before we searched only for the clients crates in `Cargo.toml`.

So now:
1. Search for `jsonrpsee crate`
2. if not found search for `client crates`

types/src/v2/error.rs
```diff
@@ -32,3 +46,94 @@ pub const INVALID_PARAMS_MSG: &str = "Invalid params";
...
 /// Method not found error message.
 pub const METHOD_NOT_FOUND_MSG: &str = "Method not found";
+/// Reserved for implementation-defined server-errors.
+pub const SERVER_ERROR_MSG: &str = "Server error";
+/// Application defined error which is not in the reserved space (-32000..=-32768)
+pub const APPLICATION_ERROR_MSG: &str = "Application error";
+
+/// JSONRPC error code
+#[derive(Error, Debug, PartialEq, Copy, Clone)]
+pub enum ErrorCode {
```
niklasad1, line 56: copy-pasted from `old jsonrpc` types with some minor tweaks.

types/src/v2/mod.rs
```diff
@@ -160,3 +176,126 @@ impl<'a> RpcParams<'a> {
...
+	/// Numeric id
+	Num(u64),
+	/// String id
+	Str(String),
+}
+
+impl From<SubscriptionId> for JsonValue {
```
niklasad1, line 283: avoid to call `serde_json::to_value` and it's infallible

dvdplm (approved):

types/src/v2/error.rs
```diff
@@ -32,3 +58,147 @@ pub const INVALID_PARAMS_MSG: &str = "Invalid params";
...
+			INVALID_PARAMS_CODE => ErrorCode::InvalidParams,
+			INTERNAL_ERROR_CODE => ErrorCode::InternalError,
+			SERVER_ERROR_CODE_RANGE_START..=SERVER_ERROR_CODE_RANGE_END => ErrorCode::ServerError(code),
+			code => ErrorCode::ApplicationError(code),
+		}
+	}
+}
+
+impl<'a> serde::Deserialize<'a> for ErrorCode {
+	fn deserialize<D>(deserializer: D) -> Result<ErrorCode, D::Error>
```
niklasad1, line 135: these are ser/deser is a little bit weird (not symmetrical) but essentially the error message can be deduced from the error code so no need to store the message as allocated string.

ws-client/src/client.rs
niklasad1: yepp, we can add an issue for that this PR is already too big :(

types/src/lib.rs
```diff
@@ -5,17 +5,20 @@
...
 /// Shared client types.
-pub mod client;
+mod client;
 
 /// Traits
 pub mod traits;
+
+pub use beef::Cow;
+pub use client::*;
+pub use error::Error;
```
niklasad1, line 22: we could revisit this error type but it's quite hard to understand the different error types.

Technically we could kill the huge enum and have each client/server to have their own error type instead.

types/src/v2/error.rs
```diff
@@ -1,14 +1,36 @@
...
-	/// Invalid params in the RPC call.
-	#[error("invalid params")]
-	InvalidParams,
+/// [Failed JSON-RPC response object](https://www.jsonrpc.org/specification#response_object).
+#[derive(Serialize, Debug)]
+pub struct JsonRpcError<'a> {
+	/// JSON-RPC version.
+	pub jsonrpc: TwoPointZero,
+	/// Error.
+	pub error: ErrorCode,
```
niklasad1, line 15: @maciejhirsz I removed your type https://github.com/paritytech/jsonrpsee/blob/master/types/src/v2/mod.rs#L85-#L90. ErrorCode should be as cheap I think and more readable

types/src/v2/params.rs
```diff
@@ -0,0 +1,193 @@
...
+			SubscriptionId::Str(s) => s.into(),
+		}
+	}
+}
+
+/// Request Id
+#[derive(Debug, PartialEq, Clone, Hash, Eq, Deserialize, Serialize)]
+#[serde(deny_unknown_fields)]
+#[serde(untagged)]
+pub enum Id {
```
niklasad1, line 160: decided to bring this back, currently we just use `Number` is the clients and the servers is using RawValue

but probably worth to comply with the spec

maciejhirsz (approved):
LGTM

ws-client/src/client.rs
```diff
@@ -90,6 +97,62 @@ pub struct WsClient {
...
+		for _ in 0..len {
+			batch.push(self.current_id.fetch_add(1, Ordering::Relaxed));
+		}
+		Ok(batch)
+	}
+
+	fn reclaim_request_id(&self) {
+		let curr = self.current_pending.load(Ordering::Relaxed);
+		if curr > 0 {
+			self.current_pending.store(curr - 1, Ordering::Relaxed);
```
tomusdrw, line 153: There is a pretty nasty race condition here :)

PR #273, opened 2021-04-19
Fix clippy warning and enforce no warning in the CI

chevdor:
This PR fixes a (the single)  clippy warning and enforces in the CI that warnings are unwanted.

This PR is small/short but should not be merged lightly as it would require a major version bump.

It is also not critical to have immediately so it would be a good candidate to pack with other PRs that would change the public interface.


http-client/src/transport.rs
```diff
@@ -112,7 +112,7 @@ pub(crate) enum Error {
 
 	/// Failed to parse the JSON returned by the server into a JSON-RPC response.
 	#[error("Error while parsing the response body")]
-	ParseError(#[source] serde_json::error::Error),
+	Parsing(#[source] serde_json::error::Error),
```
niklasad1, line 115: we will remove this in #269 anyway

.github/workflows/ci.yml
```diff
@@ -40,7 +40,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: -- -A clippy::mutable_key_type
+          args: -- -A clippy::mutable_key_type -D warnings
```
niklasad1, line 43: I understand your point but we want don't want to fail the build just because some rather opinionated lint by clippy is triggered, so warnings is sufficient IMHO.

//cc @maciejhirsz @dvdplm 

However, I think you can remove `-A clippy::mutable_key_type`

http-client/src/transport.rs
niklasad1: which lint trigger this BTW?

.github/workflows/ci.yml
dvdplm: I looked at fixing this clippy warning and was scratching my head tbh: it wasn't clear to me that the lint is valid (and I don't think it is in our case).

I'm fine keeping this as a warning in CI for the time being; github makes the clippy warnings fairly visible in the UI. And wee're still far away from a 1.0 release.

dvdplm:
@chevdor I took the liberty to merge in master.

proc-macros/src/api_def.rs
```diff
@@ -154,7 +154,9 @@ impl ApiMethodAttrs {
 	/// Tries to merge another `ApiMethodAttrs` within this one. Returns an error if there is an
 	/// overlap in the attributes.
 	// TODO: span
-	fn try_merge(&mut self, other: ApiMethodAttrs) -> syn::parse::Result<()> {
+	fn try_merge(&mut self, other: ApiMethodAttrs)
+	//  -> syn::parse::Result<()>
+	{
 		if let Some(method) = other.method {
 			if self.method.is_some() {
 				// TODO: return Err(())
```
chevdor, line 162: @niklasad1 is there something here that is blocking?
I "fixed" the clippy issue but this is NOT a good option.

I understand that you planned on returning an Error but clippy is currently right in the sense that we don't need to return an Error if we don't actually return somewhere.

So the real fix here would be to implement the TODO and bring back the Error checking.

niklasad1: Nothing really,

I think you could just return `return Err(syn::Error::new(method.span(), "Duplicate method attribute found"))` but I think it's already checked in another method before this.

FWIW, not really an expert on this code it was written by Pierre for a while ago :)

chevdor: Ok, thanks for the input @niklasad1.
I definitely don't want to just remove a `Result<>` and miss this error, although we don#t have it today 🤫
So returning this Err now, we take a small risk that it breaks downstream.
@tomaka can you advise?

tomaka: I don't think we should listen to clippy if it's suggesting stupid things such as removing the `Result`.

niklasad1: I agree but I mean't brint back the `Result` as it was before and just replace `todo` and return `return Err(syn::Error::new(method.span(), "Duplicate method attribute found"))`

chevdor: I guess clippy does not read the comments so its conclusion was not off but I agree we should not remove the Result<>.

proc-macros/src/api_def.rs
```diff
@@ -153,11 +153,10 @@ impl syn::parse::Parse for ApiMethod {
 impl ApiMethodAttrs {
 	/// Tries to merge another `ApiMethodAttrs` within this one. Returns an error if there is an
 	/// overlap in the attributes.
-	// TODO: span
 	fn try_merge(&mut self, other: ApiMethodAttrs) -> syn::parse::Result<()> {
 		if let Some(method) = other.method {
-			if self.method.is_some() {
-				// TODO: return Err(())
+			if self.method.as_deref() == Some(&method) {
```
chevdor, line 158: @niklasad1 if I got the description of this function right, a part of the check was missing.
I added some tests on the way. Can you confirm?

.github/workflows/ci.yml
chevdor: At that point, there is not a single warning left. So it is now possible to be stricter, keep `-D warnings` and reject any new warnings. That is, with the option to allow them again later if some non sense lints shows up.

Do you still want to remove `-D warnings` nonetheless ?
Let me know and I will fix accordingly and remove `-A clippy::mutable_key_type` on the way.

proc-macros/src/api_def.rs
niklasad1: Looks good thanks :+1:

.github/workflows/ci.yml
niklasad1: yeah, please remove `-D warnings` the warnings is still shown at every PR which is sufficient for now.

still great that you took care of this

niklasad1 (approved):
LGTM modulo `-D warnings` in the CI

.github/workflows/ci.yml
chevdor: ok, will do 👍

niklasad1:
> dq: is this a breaking change because we now return an error in a logic branch that previously would have not error'ed? Since the return type doesn't change on try_merge could it be considered just a patch?

I would regard it as patch/bug fix but for now we don't care about about breaking changes because we are still at `pre release version`

chevdor:
@emostov in theory yes, with the last change I made, a fn has been changed.
It was returning a `Result` and so potentially an `Err` in theory but nothing in the code was ever returning an  `Err`. Now, in some conditions, it may.

But if you treated already the `Result` properly before (= assuming it could be an `Err`), you should not have a suprise.

emostov (approved):
Thanks @chevdor & @niklasad1 for the quick answer. From my PoV this looks good

proc-macros/src/api_def.rs
```diff
@@ -218,3 +217,29 @@ impl syn::parse::Parse for ApiMethods {
 		Ok(out)
 	}
 }
+
+#[cfg(test)]
+mod tests {
+	use super::*;
+
+	#[test]
+	fn returns_ok_if_attribue_does_not_exists() {
```
dvdplm, line 226:
```suggestion
	fn returns_ok_if_attribute_does_not_exists() {
```

proc-macros/src/api_def.rs
```diff
@@ -218,3 +217,29 @@ impl syn::parse::Parse for ApiMethods {
...
+		let mut a = ApiMethodAttrs { method: None, positional_params: true };
+		let b = ApiMethodAttrs { method: Some("method".to_string()), positional_params: true };
+		assert!(ApiMethodAttrs::try_merge(&mut a, b).is_ok());
+	}
+
+	#[test]
+	fn returns_an_error_if_attribue_already_exists() {
```
dvdplm, line 233:
```suggestion
	fn returns_an_error_if_attribute_already_exists() {
```

proc-macros/src/api_def.rs
```diff
@@ -218,3 +217,29 @@ impl syn::parse::Parse for ApiMethods {
...
+		let mut a = ApiMethodAttrs { method: Some("method1".to_string()), positional_params: true };
+		let b = ApiMethodAttrs { method: Some("method2".to_string()), positional_params: true };
+		assert!(ApiMethodAttrs::try_merge(&mut a, b).is_ok());
+	}
+
+	#[test]
+	fn returns_an_error_if_attribue_exists() {
```
dvdplm, line 240:
```suggestion
	fn returns_an_error_if_attribute_exists() {
```

proc-macros/src/api_def.rs
dvdplm: The test name speaks of returning an error, but the assertion does not check for that. Which is it?

dvdplm (changes requested):

proc-macros/src/api_def.rs
chevdor: 🤦 good catch (+ the damn typos).

dvdplm (approved):

dvdplm:
@chevdor feel free to merge when CI clears, and ty! :)

PR #274, opened 2021-04-19
Reclaim request id for unsubscribe requests

svyatonik:
When unsubscribe request is sent, the `send_back` is `None` (see [here](https://github.com/paritytech/jsonrpsee/blob/3f2af26fb85b9843506505560d7831a4ecb3e373/ws-client/src/client.rs#L618)). When response to this request is received, the request id is not reclaimed by the request manager (see [this](https://github.com/paritytech/jsonrpsee/blob/3f2af26fb85b9843506505560d7831a4ecb3e373/ws-client/src/client.rs#L554) and [this](https://github.com/paritytech/jsonrpsee/blob/3f2af26fb85b9843506505560d7831a4ecb3e373/ws-client/src/client.rs#L558)) => if app starts and drops subscriptions, it'll end up with [MaxSlotsExceeded](https://github.com/paritytech/jsonrpsee/blob/3f2af26fb85b9843506505560d7831a4ecb3e373/ws-client/src/manager.rs#L105) error. Like it already happened in https://github.com/paritytech/parity-bridges-common/issues/909 .

Unfortunately, I'm unable to compile master branch of `jsonrpsee` with latest stable rust => can't run tests and see if this patch breaks something. But at least applying this patch to the alpha.4 solves our issue.

### DIFF
--- a/jsonrpsee/tests/integration_tests.rs
+++ b/jsonrpsee/tests/integration_tests.rs
@@ -188,6 +188,7 @@ async fn ws_more_request_than_buffer_should_not_deadlock() {
 }
 
 #[tokio::test]
+#[ignore]
 async fn https_works() {
 	let client = HttpClientBuilder::default().build("https://kusama-rpc.polkadot.io").unwrap();
 	let response: String = client.request("system_chain", Params::None).await.unwrap();
@@ -195,6 +196,7 @@ async fn https_works() {
 }
 
 #[tokio::test]
+#[ignore]
 async fn wss_works() {
 	let client = WsClientBuilder::default().build("wss://kusama-rpc.polkadot.io").await.unwrap();
 	let response: String = client.request("system_chain", Params::None).await.unwrap();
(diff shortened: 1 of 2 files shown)


niklasad1 (approved):
Thanks, nice catch

However, when #269 when/if merged this logic will be simplified

niklasad1:
> Unfortunately, I'm unable to compile master branch of jsonrpsee with latest stable rust => can't run tests and see if this patch breaks something. But at least applying this patch to the alpha.4 solves our issue.

Hmm, really?

svyatonik:
Yeah - I'm getting [this](https://gist.github.com/svyatonik/2a848e7b010346fc0aa9570e5a0bfe40) when trying to run `cargo test --all`. Removing target also doesn't help - I'm not sure what else I could do :)

jsonrpsee/tests/integration_tests.rs
```diff
@@ -188,13 +188,15 @@ async fn ws_more_request_than_buffer_should_not_deadlock() {
 }
 
 #[tokio::test]
+#[ignore]
```
dvdplm, line 191: What is this about, are these tests failing even with this fix?

niklasad1: the nodes have been down so ignored those for now

PR #282, opened 2021-04-21
Use a constant to express max request body size and consistently use u32 in all crates

### LINKED ISSUE #257: [all crates]: make `10 MB default` a const shared by the crates.
Maybe make 10 * 1024 * 1024 a const for consistency; i've seen it a few times but making a const would make it easier to change the buffer_max_size in the future if we choose to do so, and also makes it a bit more readable

I think it's readable as it is, but would be nice QoL, not sure if it's actually used enough to be worth it though

_Originally posted by @insipx in https://github.com/paritytech/jsonrpsee/pull/253#discussion_r605046025_

dvdplm:
Closes #257 

`soketto` uses `usize` for the `max_message_size` so if we want to remain consistent we should do the same. I think a `u32` is ok though?

### DIFF
--- a/types/src/lib.rs
+++ b/types/src/lib.rs
@@ -6,7 +6,7 @@
 extern crate alloc;
 
 /// Ten megabytes.
-pub const TEN_MB: u32 = 10 * 1024 * 1024;
+pub const TEN_MB_SIZE_BYTES: u32 = 10 * 1024 * 1024;
 
 /// JSON-RPC 2.0 specification related types v2.
 pub mod v2;
--- a/types/src/lib.rs
+++ b/types/src/lib.rs
@@ -5,6 +5,9 @@
 
 extern crate alloc;
 
+/// Ten megabytes.
+pub const TEN_MB: u32 = 10 * 1024 * 1024;
+
 /// JSON-RPC 2.0 specification related types v2.
 pub mod v2;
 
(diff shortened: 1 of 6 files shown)


types/src/lib.rs
```diff
@@ -5,6 +5,9 @@
 
 extern crate alloc;
 
+/// Ten megabytes.
+pub const TEN_MB: u32 = 10 * 1024 * 1024;
```
niklasad1, line 9: I would prefer if we name it `TEN_MB_IN_BYTES` or change documentation to be explicit :)

niklasad1 (approved):

niklasad1:
> soketto uses usize for the max_message_size so if we want to remain consistent we should do the same. I think a u32 is ok though?

The reason I choose `u32` for HTTP was to be explicit and to avoid platform related issues. Most browsers support up to 4GB payloads for HTTP and that seemed reasonable.

WebSocket message frame size is `2^63` but 4GB seems reasonable for that too. I think soketto limits it to `usize`?

dvdplm:
> I think soketto limits it to `usize`?

Yes, it's `usize` in `soketto` (see for instance [here](https://github.com/paritytech/soketto/blob/develop/src/connection.rs#L19) and [here](https://github.com/paritytech/soketto/blob/develop/src/connection.rs#L98))

PR #285, opened 2021-04-23
Fix race conditions.

tomusdrw:
We see a lot of "Invalid request ID" errors after updating to `alpha.5`. I believe this might because of possible race condition in `reclaim_request_id`, although haven't tested this yet.

Also I'm not entirely sure if `Ordering::Relaxed` is the right choice in the entire `RequestIdGuard`, but I'm really not an expert in this. Honestly would rather be on the safe side and use `SeqCst` unless you are sure it's correct. :)

### DIFF
--- a/ws-client/src/client.rs
+++ b/ws-client/src/client.rs
@@ -119,7 +119,7 @@ impl RequestIdGuard {
 
 	fn get_slot(&self) -> Result<(), Error> {
 		self.current_pending
-			.fetch_updated(Ordering::Relaxed, Ordering::Relaxed, |val| {
+			.fetch_update(Ordering::Relaxed, Ordering::Relaxed, |val| {
 				if val >= self.max_concurrent_requests {
 					None
 				} else {
--- a/ws-client/src/client.rs
+++ b/ws-client/src/client.rs
@@ -118,13 +118,16 @@ impl RequestIdGuard {
 	}
 
 	fn get_slot(&self) -> Result<(), Error> {
-		if self.current_pending.load(Ordering::Relaxed) >= self.max_concurrent_requests {
-			Err(Error::MaxSlotsExceeded)
-		} else {
-			// NOTE: `fetch_add` wraps on overflow but that can't occur because `current_pending` is checked above.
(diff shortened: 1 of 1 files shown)


niklasad1:
ok, technically `Relaxed` should be sufficient as we don't do any synchronization on the counter's but it's inserted into RequestManager I realize now so it might be races there so better to make it `SeqCst` for now I suppose.

Do you have logs on this that I can take look at? Would be interesting to see how this occurs.

ws-client/src/client.rs
```diff
@@ -118,13 +118,16 @@ impl RequestIdGuard {
...
 	fn get_slot(&self) -> Result<(), Error> {
-		if self.current_pending.load(Ordering::Relaxed) >= self.max_concurrent_requests {
-			Err(Error::MaxSlotsExceeded)
-		} else {
-			// NOTE: `fetch_add` wraps on overflow but that can't occur because `current_pending` is checked above.
-			self.current_pending.fetch_add(1, Ordering::Relaxed);
-			Ok(())
-		}
+		self.current_pending
+			.fetch_update(Ordering::Relaxed, Ordering::Relaxed, |val| {
```
niklasad1, line 122: make this SeqCst

niklasad1:
Still it doesn't explain `Invalid RequestID`, the counter should still atomic and it would just make the `MaxSlots` not to work properly AFAIU.

ws-client/src/client.rs
```diff
@@ -149,10 +152,14 @@ impl RequestIdGuard {
 	}
 
 	fn reclaim_request_id(&self) {
-		let curr = self.current_pending.load(Ordering::Relaxed);
-		if curr > 0 {
-			self.current_pending.store(curr - 1, Ordering::Relaxed);
-		}
+		// NOTE we ignore the error here, since we are simply saturating at 0
+		let _ = self.current_pending.fetch_update(Ordering::Relaxed, Ordering::Relaxed, |val| {
```
niklasad1, line 156: make this SeqCst or Aquire/Release

niklasad1 (approved):
Thank you

PR #286, opened 2021-04-24
[types]: allow `data` field in jsonrpc error obj.

niklasad1:
Follow up on #285 

Basically when a error response with a `data` field is included it was regarded as bad response which this fixes

For example I tested this with `subxt` and the following was regarded as bad response before this PR:

```
[2021-04-24T10:44:15Z DEBUG jsonrpsee_ws_client::client] [backend]: recv unparseable message: Ok(Object({"error": Object({"code": Number(1014), "data": String("The transaction has too low priority to replace another transaction already in the pool."), "message": String("Priority is too low: (12497117750671931 vs 12497117750671931)")}), "id": Number(12), "jsonrpc": String("2.0")}))

```

//cc @tomusdrw I think this fixes your `InvalidRequestId` messages.

### DIFF
--- a/http-server/src/server.rs
+++ b/http-server/src/server.rs
@@ -38,10 +38,7 @@ use hyper::{
 };
 use jsonrpsee_types::error::{Error, GenericTransportError, RpcError};
 use jsonrpsee_types::v2::request::{JsonRpcInvalidRequest, JsonRpcRequest};
-use jsonrpsee_types::v2::{
-	error::{JsonRpcErrorCode, JsonRpcErrorObject},
-	params::RpcParams,
-};
+use jsonrpsee_types::v2::{error::JsonRpcErrorCode, params::RpcParams};
 use jsonrpsee_utils::{hyper_helpers::read_response_to_body, server::send_error};
 use serde::Serialize;
 use socket2::{Domain, Socket, Type};
@@ -189,12 +186,7 @@ impl Server {
 										log::error!("method_call: {} failed: {:?}", req.method, err);
 									}
 								} else {
-									let code = JsonRpcErrorCode::MethodNotFound;
-									send_error(
-										req.id,
-										&tx,
-										JsonRpcErrorObject { code, message: code.message(), data: None },
-									);
+									send_error(req.id, &tx, JsonRpcErrorCode::MethodNotFound.into());
 								}
 							}
 							Err(_e) => {
@@ -202,7 +194,7 @@ impl Server {
 									Ok(req) => (req.id, JsonRpcErrorCode::InvalidRequest),
 									Err(_) => (None, JsonRpcErrorCode::ParseError),
 								};
-								send_error(id, &tx, JsonRpcErrorObject { code, message: code.message(), data: None });
+								send_error(id, &tx, code.into());
 							}
 						};
 
--- a/http-server/src/server.rs
+++ b/http-server/src/server.rs
@@ -37,7 +37,10 @@ use hyper::{
 };
 use jsonrpsee_types::error::{Error, GenericTransportError, RpcError};
 use jsonrpsee_types::v2::request::{JsonRpcInvalidRequest, JsonRpcRequest};
-use jsonrpsee_types::v2::{error::ErrorCode, params::RpcParams};
+use jsonrpsee_types::v2::{
+	error::{JsonRpcErrorCode, JsonRpcErrorObject},
+	params::RpcParams,
+};
 use jsonrpsee_utils::{hyper_helpers::read_response_to_body, server::send_error};
 use serde::Serialize;
 use socket2::{Domain, Socket, Type};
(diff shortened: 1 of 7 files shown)


types/src/v2/error.rs
```diff
@@ -160,23 +163,33 @@ impl<'de> Visitor<'de> for ErrorCodeVisitor {
...
+
+		loop {
+			match access.next_entry::<&str, i32>() {
+				Ok(Some((key, val))) if key == ERROR_CODE_KEY && res.is_none() => {
+					res = Some(Ok(val.into()));
+				}
+				Ok(Some((key, _))) if key == ERROR_CODE_KEY => {
+					res = Some(Err(DeserializeError::duplicate_field(ERROR_CODE_KEY)));
+				}
+				Ok(None) => break,
```
niklasad1, line 183: deny unknown fields?!

http-server/src/server.rs
```diff
@@ -186,15 +189,20 @@ impl Server {
...
 							Err(_e) => {
-								let (id, err) = match serde_json::from_slice::<JsonRpcInvalidRequest>(&body) {
-									Ok(req) => (req.id, ErrorCode::InvalidRequest),
-									Err(_) => (None, ErrorCode::ParseError),
+								let (id, code) = match serde_json::from_slice::<JsonRpcInvalidRequest>(&body) {
+									Ok(req) => (req.id, JsonRpcErrorCode::InvalidRequest),
+									Err(_) => (None, JsonRpcErrorCode::ParseError),
 								};
-								send_error(id, &tx, err);
+								send_error(id, &tx, JsonRpcErrorObject { code, message: code.message(), data: None });
```
dvdplm, line 205: Maybe we can have a `JsonRpcErrorCode::into_error_obj()` for this? `send_error(id, &tx, code.into_error_obj());`?

types/src/v2/error.rs
```diff
@@ -21,18 +21,50 @@ pub struct JsonRpcError<'a> {
...
 impl fmt::Display for JsonRpcErrorAlloc {
 	fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-		write!(f, "{:?}: {}: {:?}", self.jsonrpc, self.error, self.id)
+		write!(f, "{:?}: {:?}: {:?}", self.jsonrpc, self.error, self.id)
 	}
 }
 
+/// JSON-RPC error object.
+#[derive(Debug, PartialEq, Clone, Deserialize)]
+#[serde(deny_unknown_fields)]
```
dvdplm, line 38: Do we really need this?

types/src/v2/error.rs
```diff
@@ -21,18 +21,50 @@ pub struct JsonRpcError<'a> {
...
+	/// Message
+	pub message: String,
+	/// Optional data
+	#[serde(skip_serializing_if = "Option::is_none")]
+	pub data: Option<JsonValue>,
+}
+
+impl From<JsonRpcErrorCode> for JsonRpcErrorObjectAlloc {
+	fn from(code: JsonRpcErrorCode) -> Self {
+		Self { message: code.message().to_owned(), code, data: None }
```
dvdplm, line 51:
```suggestion
		Self { code, message: code.message().to_owned(), data: None }
```
…to keep clippy happy?

types/src/v2/error.rs
```diff
@@ -21,18 +21,50 @@ pub struct JsonRpcError<'a> {
...
+}
+
+/// JSON-RPC error object with no extra allocations.
+#[derive(Debug, Serialize)]
+#[serde(deny_unknown_fields)]
+pub struct JsonRpcErrorObject<'a> {
+	/// Code
+	pub code: JsonRpcErrorCode,
+	/// Message
+	pub message: &'a str,
```
dvdplm, line 62: Could this be a `Cow` you think?

dvdplm (approved):
lgtm modulo nits/suggestions

types/src/v2/error.rs
niklasad1: The reason is the huge error type which requires ownership and we return it the user.

The actual body is read and dropped there maybe we could have like `to_owned` or something for  `JsonRpcErrorObject`, it's quite annoying of to have two different types but perhaps we could tackle that in another PR

types/src/v2/error.rs
niklasad1: the intention is that this type should only be used for serialize the object and `Cow` should not be needed then unless we merge to two different types somehow ^^

types/src/v2/error.rs
```diff
@@ -121,7 +126,7 @@ impl JsonRpcErrorCode {
 	}
 
 	/// Returns the message for the given error code.
-	pub const fn message(&self) -> &str {
+	pub const fn message(&self) -> &'static str {
```
dvdplm, line 129: I guess a `&str` returned from a `const fn` is `'static` by definition, but I don't mind being explicit. 👍

niklasad1: It didn't work in the `From impl` so I had put `'static` there.

```
impl<'a> From<JsonRpcErrorCode> for JsonRpcErrorObject<'a> {
	fn from(code: JsonRpcErrorCode) -> Self {
		Self { code, message: code.message(), data: None }
	}
}
```

Not sure if there's a better work around for that

PR #289, opened 2021-04-26
Use SeqCst everywhere

dvdplm:
See https://github.com/paritytech/jsonrpsee/pull/285

I went over all the atomic operations and realised I could not convince myself that all uses of `Relaxed` were ok, so better safe than sorry: let's switch to `SeqCst`.

### DIFF
--- a/http-client/src/client.rs
+++ b/http-client/src/client.rs
@@ -64,7 +64,7 @@ impl Client for HttpClient {
 		R: DeserializeOwned,
 	{
 		// NOTE: `fetch_add` wraps on overflow which is intended.
-		let id = self.request_id.fetch_add(1, Ordering::Relaxed);
+		let id = self.request_id.fetch_add(1, Ordering::SeqCst);
 		let request = JsonRpcCallSer::new(Id::Number(id), method, params);
 
 		let body = self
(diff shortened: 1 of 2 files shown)


niklasad1 (approved):

niklasad1:
It is better to safe than sorry but it's just used a simple counter which should be safe, as we don't care about ordering just that it's incremented but I got it so wrong before so I should probably be quiet. 

I read some `C++` guidelines on this and the approach for `Ordering::Relaxed` is to wrap it in `new type` such makes only possible to do `add` and nothing else. Then the rest of the code doesn't have to care about it.

dvdplm:
> I read some `C++ guidelines` on this and they recommend for `Ordering::Relaxed` is to wrap it in `new type` such makes only possible to do `add` and nothing else. Then the rest of the code doesn't have to care about it.

That sounds like a pretty good pattern. Do you want me to make changes to this PR in that sense?

niklasad1:
> That sounds like a pretty good pattern. Do you want me to make changes to this PR in that sense?

It's better to handle it another PR. 
I think we could change the internal request ID guard to something like this for better readability/technical debt:

```
#[derive(Debug)]
struct RequestIdGuard {
	// Ongoing concurrent request.
	current_pending: RwLock<usize>,
	/// Max concurrent pending requests allowed.
	max_concurrent_requests: usize,
	/// Get the next request ID.
	current_id: AtomicCounterU64,
}
```

There's https://crates.io/crates/atomic-counter but it only works for `usize`

Issue #276, opened 2021-04-20
[types]: jsonrpc_v2 types write serialize/deserialize tests.


niklasad1:
We should have some fine-grained tests for the types to comply with the spec.

For example using [serde_json::RawValue as JSON-RPC Params](https://github.com/paritytech/jsonrpsee/blob/master/types/src/v2/request.rs#L20) would not ensure that it is an valid `JSON-RPC param`...

Ideally we would unify the types and don't have to test so many different types. For example is many cases we have one type for serializing and one for deserialization, see JsonRpcParams and RpcParams.

PR #292, opened 2021-04-26
[http server] Batch requests

dvdplm:
Implement http server batch requests.

[Spec](https://www.jsonrpc.org/specification#batch).

We diverge from the spec in a few places:
- sending a batch of notifications should receive the empty string, `""`, in response. Instead we return `{"jsonrpc":"2.0","result":"","id":null}` for each notification
- sending an invalid batch with numbers, e.g. `[123]` we should return an `Invalid Request` error (which we do) with the `id` set to `null` (which we don't; somehow serde parses the `123` into the `id` and I don't think it's worth the hassle to fix that)
- sending a batch of several invalid requests, e.g. `[1, 2, 3]`, should return one `Invalid Request` for each, but to us that's a `Parse Error`. Again I think that it's not really worth the hassle to fix.

One thing that annoys me with this PR is that if we decide to support notifications properly, the number of `serde_json::from_slice` calls will be even bigger than it is now (3 for the worst case), so the lack of support for `untagged` enums with `RawValue` becomes really problematic. We'd have to get pretty creative at that point. :/

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -51,7 +51,7 @@ pub fn batched_http_requests(crit: &mut Criterion) {
 	let rt = TokioRuntime::new().unwrap();
 	let url = rt.block_on(helpers::http_server());
 	let client = Arc::new(HttpClientBuilder::default().build(&url).unwrap());
-	run_round_trip_with_batch(&rt, crit, client.clone(), "batched_http_round_trip", 10);
+	run_round_trip_with_batch(&rt, crit, client.clone(), "http batch requests");
 }
 
 pub fn websocket_requests(crit: &mut Criterion) {
@@ -73,21 +73,26 @@ fn run_round_trip(rt: &TokioRuntime, crit: &mut Criterion, client: Arc<impl Clie
 	});
 }
 
+/// Benchmark http batch requests over batch sizes of 2, 5, 10, 50 and 100 RPCs in each batch.
 fn run_round_trip_with_batch(
 	rt: &TokioRuntime,
 	crit: &mut Criterion,
 	client: Arc<impl Client>,
 	name: &str,
-	batch_size: usize,
 ) {
-	let batch = vec![("say_hello", JsonRpcParams::NoParams); batch_size];
-	crit.bench_function(name, |b| {
-		b.iter(|| {
-			rt.block_on(async {
-				black_box(client.batch_request::<String>(batch.clone()).await.unwrap());
+	let mut group = crit.benchmark_group(name);
+	for batch_size in [2, 5, 10, 50, 100usize].iter() {
+		let batch = vec![("say_hello", JsonRpcParams::NoParams); *batch_size];
+		group.throughput(Throughput::Elements(*batch_size as u64));
+		group.bench_with_input(BenchmarkId::from_parameter(batch_size), batch_size, |b, _| {
+			b.iter(|| {
+				rt.block_on( async {
+					client.batch_request::<String>(batch.clone()).await.unwrap()
+				})
 			})
-		})
-	});
+		});
+	}
+	group.finish();
 }
 
 fn run_concurrent_round_trip<C: 'static + Client + Send + Sync>(
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -74,7 +74,13 @@ fn run_round_trip(rt: &TokioRuntime, crit: &mut Criterion, client: Arc<impl Clie
 	});
 }
 
-fn run_round_trip_with_batch(rt: &TokioRuntime, crit: &mut Criterion, client: Arc<impl Client>, name: &str, batch_size: usize) {
+fn run_round_trip_with_batch(
+	rt: &TokioRuntime,
+	crit: &mut Criterion,
+	client: Arc<impl Client>,
+	name: &str,
+	batch_size: usize,
+) {
 	let batch = vec![("say_hello", JsonRpcParams::NoParams); batch_size];
 	crit.bench_function(name, |b| {
 		b.iter(|| {
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -74,22 +74,13 @@ fn run_round_trip(rt: &TokioRuntime, crit: &mut Criterion, client: Arc<impl Clie
 }
 
 /// Benchmark http batch requests over batch sizes of 2, 5, 10, 50 and 100 RPCs in each batch.
-fn run_round_trip_with_batch(
-	rt: &TokioRuntime,
-	crit: &mut Criterion,
-	client: Arc<impl Client>,
-	name: &str,
-) {
+fn run_round_trip_with_batch(rt: &TokioRuntime, crit: &mut Criterion, client: Arc<impl Client>, name: &str) {
 	let mut group = crit.benchmark_group(name);
 	for batch_size in [2, 5, 10, 50, 100usize].iter() {
 		let batch = vec![("say_hello", JsonRpcParams::NoParams); *batch_size];
 		group.throughput(Throughput::Elements(*batch_size as u64));
 		group.bench_with_input(BenchmarkId::from_parameter(batch_size), batch_size, |b, _| {
-			b.iter(|| {
-				rt.block_on( async {
-					client.batch_request::<String>(batch.clone()).await.unwrap()
-				})
-			})
+			b.iter(|| rt.block_on(async { client.batch_request::<String>(batch.clone()).await.unwrap() }))
 		});
 	}
 	group.finish();
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -9,7 +9,6 @@ use jsonrpsee::{
 	ws_client::WsClientBuilder,
 };
 use std::sync::Arc;
-use std::time::Instant;
 use tokio::runtime::Runtime as TokioRuntime;
 
 mod helpers;
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -9,11 +9,12 @@ use jsonrpsee::{
 	ws_client::WsClientBuilder,
 };
 use std::sync::Arc;
+use std::time::Instant;
 use tokio::runtime::Runtime as TokioRuntime;
 
 mod helpers;
 
-criterion_group!(benches, http_requests, websocket_requests, jsonrpsee_types_v2);
+criterion_group!(benches, http_requests, batched_http_requests, websocket_requests, jsonrpsee_types_v2);
 criterion_main!(benches);
 
 fn v2_serialize<'a>(req: JsonRpcCallSer<'a>) -> String {
@@ -47,6 +48,13 @@ pub fn http_requests(crit: &mut Criterion) {
 	run_concurrent_round_trip(&rt, crit, client.clone(), "http_concurrent_round_trip");
 }
 
+pub fn batched_http_requests(crit: &mut Criterion) {
+	let rt = TokioRuntime::new().unwrap();
+	let url = rt.block_on(helpers::http_server());
+	let client = Arc::new(HttpClientBuilder::default().build(&url).unwrap());
+	run_round_trip_with_batch(&rt, crit, client.clone(), "batched_http_round_trip", 10);
+}
+
 pub fn websocket_requests(crit: &mut Criterion) {
 	let rt = TokioRuntime::new().unwrap();
 	let url = rt.block_on(helpers::ws_server());
@@ -61,6 +69,17 @@ fn run_round_trip(rt: &TokioRuntime, crit: &mut Criterion, client: Arc<impl Clie
 		b.iter(|| {
 			rt.block_on(async {
 				black_box(client.request::<String>("say_hello", JsonRpcParams::NoParams).await.unwrap());
+			})
+		})
+	});
+}
+
+fn run_round_trip_with_batch(rt: &TokioRuntime, crit: &mut Criterion, client: Arc<impl Client>, name: &str, batch_size: usize) {
+	let batch = vec![("say_hello", JsonRpcParams::NoParams); batch_size];
+	crit.bench_function(name, |b| {
+		b.iter(|| {
+			rt.block_on(async {
+				black_box(client.batch_request::<String>(batch.clone()).await.unwrap());
 			})
 		})
 	});
--- a/http-server/src/server.rs
+++ b/http-server/src/server.rs
@@ -169,7 +169,7 @@ impl Server {
 								// NOTE(niklasad1): connection ID is unused thus hardcoded to `0`.
 								if let Err(err) = (method)(id, params, &tx, 0) {
 									log::error!(
-										"execution of method call {} failed: {:?}, request id={:?}",
+										"execution of method call '{}' failed: {:?}, request id={:?}",
 										method_name,
 										err,
 										id
--- a/http-server/src/server.rs
+++ b/http-server/src/server.rs
@@ -197,13 +197,14 @@ impl Server {
 
 						// NOTE(niklasad1): it's a channel because it's needed for batch requests.
 						let (tx, mut rx) = mpsc::unbounded();
+						// Is this a single request or a batch (or error)?
+						let mut single = true;
 
 						// For [technical reasons](https://github.com/serde-rs/json/issues/497), `RawValue` can't be
 						// used with untagged enums at the moment. This means we can't use an `SingleOrBatch` untagged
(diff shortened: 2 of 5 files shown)


http-server/src/server.rs
```diff
@@ -175,31 +197,53 @@ impl Server {
...
+						// used with untagged enums at the moment. This means we can't use an `SingleOrBatch` untagged
+						// enum here and have to try each case individually: first the single request case, then the
+						// batch case and lastly the error. For the worst case – unparseable input – we make three calls
+						// to [`serde_json::from_slice`] which is pretty annoying.
+						if let Ok(JsonRpcRequest { id, method: method_name, params, .. }) =
+							serde_json::from_slice::<JsonRpcRequest>(&body)
+						{
+							execute(id, &tx, &method_name, params);
+						} else if let Ok(batch) = serde_json::from_slice::<Vec<JsonRpcRequest>>(&body) {
+							if batch.len() > 0 {
```
insipx, line 213:
```suggestion
							if !batch.is_empty() {
```

http-server/src/server.rs
```diff
@@ -175,31 +197,53 @@ impl Server {
...
-								log::debug!("recv: {:?}", req);
-								let params = RpcParams::new(req.params.map(|params| params.get()));
-								if let Some(method) = methods.get(&*req.method) {
-									// NOTE(niklasad1): connection ID is unused thus hardcoded to `0`.
-									if let Err(err) = (method)(req.id, params, &tx, 0) {
-										log::error!("method_call: {} failed: {:?}", req.method, err);
-									}
-								} else {
-									send_error(req.id, &tx, JsonRpcErrorCode::MethodNotFound.into());
+						// For reasons outlined [here](https://github.com/serde-rs/json/issues/497), `RawValue` can't be
```
insipx, line 203: let's make this an issue to track it easier

benches/bench.rs
```diff
@@ -9,11 +9,12 @@ use jsonrpsee::{
 	ws_client::WsClientBuilder,
 };
 use std::sync::Arc;
+use std::time::Instant;
```
insipx, line 12:
```suggestion
```

insipx (commented):
It does worry me that we don't return exactly to spec, but at the same time I don't think any of these things being slightly off will lead to system-critical errors so I'm pretty sure it's fine too

niklasad1:
> sending a batch of notifications should receive the empty string, "", in response. Instead we return {"jsonrpc":"2.0","result":"","id":null} for each notification

Technically, we don't support notifications so those are registered as ordinary method calls and treated as that. So we would need have `register_notification` method that handles them differently to work according to the spec.

> sending an invalid batch with numbers, e.g. [123] we should return an Invalid Request error (which we do) with the id set to null (which we don't; somehow serde parses the 123 into the id and I don't think it's worth the hassle to fix that)

Can you elaborate what you mean by that? Is the `"id":[123]`? That would be parsed a number if you call [RpcParams::one](https://github.com/paritytech/jsonrpsee/blob/master/types/src/v2/params.rs#L92) on it would work yes. Maybe we should have typed `Id` similar to what we have in the client.

> sending a batch of several invalid requests, e.g. [1, 2, 3], should return one Invalid Request for each, but to us that's a Parse Error. Again I think that it's not really worth the hassle to fix.

Right, that's sounds almost impossible to properly with an nested enum. Maybe enough to motivation to switch back to huge output enum ^^

http-server/src/tests.rs
```diff
@@ -50,6 +62,81 @@ async fn single_method_call_with_params() {
...
+#[tokio::test]
+async fn batched_notifications() {
+	let _ = env_logger::try_init();
+
+	let addr = server().await;
+	let uri = to_http_uri(addr);
+
+	let req = r#"[
+        {"jsonrpc": "2.0", "method": "notif", "params": [1,2,4]},
+        {"jsonrpc": "2.0", "method": "notif", "params": [7]}
```
niklasad1, line 95: this probably because `Id` is regarded a `None` not sure, so a typed Id could fix this

niklasad1 (commented):
I'm favor of ditching `serde_json::value::RawValue` this sounds to hard to do properly without untagged enums. We could open a tracking issue to that `serde issue` and migrate when it's fixed but I guess it's not anytime soon.

//cc @maciejhirsz any ideas or opinions ^

http-server/src/server.rs
```diff
@@ -153,6 +157,24 @@ impl Server {
 				Ok::<_, HyperError>(service_fn(move |request| {
 					let methods = methods.clone();
 					let access_control = access_control.clone();
+
+					let execute =
```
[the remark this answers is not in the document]
dvdplm: I know, and I agree, and I tried but couldn't make it work; I'd have to look up the `method` from the `methods` hash map in `start()` and pass it along so it didn't really shorten up the code much.

Lately I've started to shift my views on when it's right to refactor for briefness. The rule of thumb is still "terse is good", but for some cases I've started to think it's alright to keep the code long when it's "the main loop", whatever that means for each specific case, i.e. the most important thing that a given program does. Splitting things up too much can force the reader to jump around more than is ideal and even if it's long it is sometimes more readable to keep it all in one place. I know this is hand wavey (and for this case here I wanted to do exactly what you suggest), but yeah, just wanted to share those thoughts.

http-server/src/server.rs
dvdplm: https://github.com/paritytech/jsonrpsee/issues/296

dvdplm:
> > sending a batch of notifications should receive the empty string, "", in response. Instead we return {"jsonrpc":"2.0","result":"","id":null} for each notification
> 
> Technically, we don't support notifications so those are registered as ordinary method calls and treated as that. So we would need have `register_notification` method that handles them differently to work according to the spec but the response looks wrong anyway.

Yeah, I think it's fine to be non-spec compliant for notifications. We explicitly do not support them, so as long as we're not doing anything completely bonkers I think it's ok. I don't know what the purpose of notifications is but if I were to guess it's sort of a performance optimization? Either way: are you ok keeping the tests and comments around as support for any future dev that works on adding notification support?

> > sending an invalid batch with numbers, e.g. [123] we should return an Invalid Request error (which we do) with the id set to null (which we don't; somehow serde parses the 123 into the id and I don't think it's worth the hassle to fix that)
> 
> Can you elaborate what you mean by that? Is it `"id":[123]`? 

No! It's `… "id": 123}`, so serde actually parses `[123]`. 

> Maybe we should have typed `Id` similar to what we have in the client.

The `id` is typed in the server, `Id::Num(…)`/`Id::Null`/`Id::Str(…)` – or what do you mean?

> Right, that's sounds very hard to do properly without an nested enum. Maybe enough to motivation to switch back to huge output enum ^^

I'd need to understand the tradeoffs better tbh. Not in this PR though!

niklasad1:
> The id is typed in the server, Id::Num(…)/Id::Null/Id::Str(…) – or what do you mean?

I mean in the [v2::response::JsonRpcResponse](https://github.com/paritytech/jsonrpsee/blob/master/types/src/v2/response.rs#L14), either to change to have the concrete type `Id` or manually check it/parse

http-server/src/server.rs
```diff
@@ -175,31 +197,54 @@ impl Server {
...
 							}
-							Err(_e) => {
-								let (id, code) = match serde_json::from_slice::<JsonRpcInvalidRequest>(&body) {
-									Ok(req) => (req.id, JsonRpcErrorCode::InvalidRequest),
-									Err(_) => (None, JsonRpcErrorCode::ParseError),
-								};
-								send_error(id, &tx, code.into());
-							}
+						} else {
+							log::error!("[service_fn], Cannot parse request body={:?}", String::from_utf8_lossy(&body));
```
niklasad1, line 223: how readable is this, I suspect it just raw bytes?

can't you try to parse it as `JsonValue` instead?

Anyway, users could do try to parse it as JSON offline I guess

dvdplm: > how readable is this, I suspect it just raw bytes?

It prints the payload as text, it's "human readable". Should I remove it you think (a huge incoming and malformed payload could put strain on the logger perhaps)?

niklasad1: ok, maybe I confused this with `serde_json::to_string let's keep it

http-server/src/server.rs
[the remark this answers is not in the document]
niklasad1: It's mainly because of the closure that is passed to hyper but yeah this changes made the code really quite hard to read, it wasn't great before either.

Maybe we split it the response handling to helper functions, basically you have to read the bytes here to take ownership over it which used to later to build a message to send to the background task.

benches/bench.rs
```diff
@@ -47,6 +47,13 @@ pub fn http_requests(crit: &mut Criterion) {
 	run_concurrent_round_trip(&rt, crit, client.clone(), "http_concurrent_round_trip");
 }
 
+pub fn batched_http_requests(crit: &mut Criterion) {
+	let rt = TokioRuntime::new().unwrap();
+	let url = rt.block_on(helpers::http_server());
+	let client = Arc::new(HttpClientBuilder::default().build(&url).unwrap());
+	run_round_trip_with_batch(&rt, crit, client.clone(), "batched_http_round_trip", 10);
```
niklasad1, line 54: would be nice to with different batch sizes just to see how it performs with bigger input ^^

http-server/src/server.rs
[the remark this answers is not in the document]
niklasad1: Personally, I found [this](https://github.com/paritytech/jsonrpsee/blob/master/ws-client/src/client.rs#L554-#L599) more readable but it's still quite long ^^

(maybe because I wrote it lol)

http-server/src/server.rs
```diff
@@ -175,31 +197,54 @@ impl Server {
...
+							let (id, code) = match serde_json::from_slice::<JsonRpcInvalidRequest>(&body) {
+								Ok(req) => (req.id, JsonRpcErrorCode::InvalidRequest),
+								Err(_) => (None, JsonRpcErrorCode::ParseError),
+							};
+							send_error(id, &tx, code.into());
+						}
+						rx.close();
```
niklasad1, line 230: I suppose this is performed to mark to that we are done sending any further messages?!
Such that we the receiver will receive eventually `Ok(None)` when all messages has been sent?

I think it deserves a comment because it was not straightforward to me at least

http-server/src/server.rs
```diff
@@ -175,31 +197,54 @@ impl Server {
...
+						let response = if single {
+							rx.next().await.expect("Sender is still alive managed by us above; qed")
+						} else {
+							let mut buf = String::with_capacity(2048);
```
niklasad1, line 234: Why bother estimate `with_capacity`?

The number of requests in the batch is known but the size of the response is unknown?

http-server/src/server.rs
```diff
@@ -175,31 +197,54 @@ impl Server {
...
+							buf.push('[');
+							let mut buf = rx
+								.fold(buf, move |mut acc, response| async move {
+									acc = [acc, response].concat();
+									acc.push(',');
+									acc
+								})
+								.await;
+							buf.pop();
```
niklasad1, line 243: this is used to remove the trailing comma of the last entry?

niklasad1 (commented):
Looks alright, most of the issues that appeared in this PR is related https://github.com/paritytech/jsonrpsee/issues/276

but I have some additional questions ^^

http-server/src/server.rs
dvdplm: Yes exactly, I read this in the docs:

> Closes the receiving half of a channel without dropping it.
>This prevents any further messages from being sent on the channel while still enabling the receiver to drain messages that are buffered.

I don't know if it's necessary to close the channel in this case but thought maybe it's a "best practice" to do so and if we ever decide to execute batch requests on separate tasks (and threads) with deadlines it's good to make sure the channel can't be written to. I'll add a comment.

http-server/src/server.rs
dvdplm: Exactly

http-server/src/server.rs
dvdplm: Yes. I'm just guessing the value here ofc, 2kb felt like it should cater for 80% of RPC calls. Premature optimization? Should I remove it?

niklasad1: it's fine :)

http-server/src/server.rs
niklasad1: In this case, I think so because otherwise you have to drop the sender(s) to close the channel. Because we have control of over both the sender and receiver (the sender is just borrowed by the call closure)

niklasad1 (approved):

http-server/src/server.rs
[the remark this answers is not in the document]
dvdplm: K, I tried another approach, moving `execute` to be a method on `Server` so I can access `self.root` (i.e. the function pointers we call to run the request), but it doesn't work because `self` is moved into the closure when calling `make_service_fn()`.
On the client side it's a bit easier because we don't have to call anything. Or I'm too limited to see how to do it! Pointers welcome!

http-server/src/server.rs
```diff
@@ -153,6 +158,30 @@ impl Server {
...
+								let params = RpcParams::new(params.map(|params| params.get()));
+								// NOTE(niklasad1): connection ID is unused thus hardcoded to `0`.
+								if let Err(err) = (method)(id, params, &tx, 0) {
+									log::error!(
+										"execution of method call {} failed: {:?}, request id={:?}",
+										method_name,
+										err,
+										id
+									);
+								}
```
dvdplm, line 177: @niklasad1 What does the connected client see in this case, i.e. when the method they're calling fails? Shouldn't we be sending back the error here in addition to logging it?

dvdplm: Is that what you're doing in https://github.com/paritytech/jsonrpsee/pull/295 perhaps?

niklasad1: yes, #295 should fix that but it doesn't log it FWIW

http-server/src/server.rs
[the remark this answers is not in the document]
insipx: We could just add another parameter, `&Methods` to the execute function so execute becomes:

```rust
fn execute(methods: &Methods, id: Option<&RawValue>, tx: RpcSender, method_name: &str, params: Option<&RawValue>) {
	if let Some(method) = methods.get(method_name) {
		let params = RpcParams::new(params.map(|params| params.get()));
		// NOTE(niklasad1): connection ID is unused thus hardcoded to `0`.
		if let Err(err) = (method)(id, params, &tx, 0) {
			log::error!("execution of method call {} failed: {:?}, request id={:?}", method_name, err, id);
		}
	} else {
		send_error(id, tx, JsonRpcErrorCode::MethodNotFound.into());
	}
}
```

and then call it like `execute(&methods, id, &tx, &method_name, params);`

I guess the tradeoff is having a long function signature for `execute`, though.

I think I agree with your analysis of `main_loops` though. Often it can be hard to make them shorter and would result in more confusing code than just keeping it all together. In this case I think `execute` fn is self-explanatory enough in that it just executes the right function for the rpc call.

insipx (approved):

insipx:
> I'm favor of ditching serde_json::value::RawValue this sounds to hard to do properly without untagged enums. We could open a tracking issue to that serde issue and migrate when it's fixed but I guess it's not anytime soon.

Yeah, I find it interesting that the serde attempt to solve this is sortof blocked on Associated type defaults being included in rust, but there's no knowing if that will be done soon. Maybe worth revisiting anyway since it's been ~2 years?

http-server/src/server.rs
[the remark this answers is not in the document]
dvdplm: @insipx I could have sworn I tried that; your version works too. I can go either way here, @niklasad1 thoughts?

dvdplm:
> blocked on Associated type defaults being included in rust, but there's no knowing if that will be done soon. Maybe worth revisiting anyway since it's been ~2 years?

TIL [this](https://github.com/rust-lang/rust/pull/61812) was merged in March, so maybe there's hope?

benches/bench.rs
dvdplm: Done in 78f3b25c3a02277153f2c85e9f6e1e712296faeb

http-server/src/server.rs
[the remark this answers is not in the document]
dvdplm: FWIW the performance of @insipx's version is identical.

PR #294, opened 2021-04-27
unify error types in the servers.

niklasad1:
It was really annoying to deal with when trying to integrate in substrate.

Maybe we should try to split up the error in with different types instead to reduce the scope instead of depending on the big `Error enum`


dvdplm (approved):
I think I ran into a similar frustration the other day. This seems like a good first step.

PR #295, opened 2021-04-29
[servers] return error if context or params fails

niklasad1:
This PR changes the servers behavior when they get a response with params or user provided context that fails. 
Instead of just terminating the connection we now return back a `JSON-RPC error`.

I changed the `RpcMethod trait` to take an generic Error instead such that we can configure it statically instead of matching of the entire enum when the is just one or two possible errors.

### DIFF
--- a/http-server/src/module.rs
+++ b/http-server/src/module.rs
@@ -1,4 +1,4 @@
-use jsonrpsee_types::v2::error::{JsonRpcErrorCode, JsonRpcErrorObject};
+use jsonrpsee_types::v2::error::{JsonRpcErrorCode, JsonRpcErrorObject, CONTEXT_EXECUTION_FAILED_CODE};
 use jsonrpsee_types::{
 	error::{Error, InvalidParams, ServerCallError},
 	traits::RpcMethod,
@@ -103,7 +103,7 @@ impl<Context> RpcContextModule<Context> {
 						send_error(id, tx, JsonRpcErrorCode::InvalidParams.into())
 					}
 					Err(ServerCallError::ContextFailed(err)) => {
-						let err = JsonRpcErrorObject { code: 1.into(), message: &err.to_string(), data: None };
+						let err = JsonRpcErrorObject { code: JsonRpcErrorCode::ServerError(CONTEXT_EXECUTION_FAILED_CODE), message: &err.to_string(), data: None };
 						send_error(id, tx, err)
 					}
 				};
--- a/http-server/src/module.rs
+++ b/http-server/src/module.rs
@@ -1,4 +1,4 @@
-use jsonrpsee_types::v2::error::{JsonRpcErrorCode, JsonRpcErrorObject, CONTEXT_EXECUTION_FAILED_CODE};
+use jsonrpsee_types::v2::error::{JsonRpcErrorCode, JsonRpcErrorObject, CALL_EXECUTION_FAILED_CODE};
 use jsonrpsee_types::{
 	error::{CallError, Error, InvalidParams},
 	traits::RpcMethod,
@@ -102,7 +102,7 @@ impl<Context> RpcContextModule<Context> {
 					Err(CallError::InvalidParams(_)) => send_error(id, tx, JsonRpcErrorCode::InvalidParams.into()),
 					Err(CallError::Failed(err)) => {
 						let err = JsonRpcErrorObject {
-							code: JsonRpcErrorCode::ServerError(CONTEXT_EXECUTION_FAILED_CODE),
+							code: JsonRpcErrorCode::ServerError(CALL_EXECUTION_FAILED_CODE),
 							message: &err.to_string(),
 							data: None,
 						};
--- a/http-server/src/module.rs
+++ b/http-server/src/module.rs
@@ -1,6 +1,6 @@
 use jsonrpsee_types::v2::error::{JsonRpcErrorCode, JsonRpcErrorObject, CONTEXT_EXECUTION_FAILED_CODE};
 use jsonrpsee_types::{
-	error::{Error, InvalidParams, ServerCallError},
+	error::{CallError, Error, InvalidParams},
 	traits::RpcMethod,
 	v2::params::RpcParams,
 };
@@ -88,7 +88,7 @@ impl<Context> RpcContextModule<Context> {
 	where
 		Context: Send + Sync + 'static,
 		R: Serialize,
-		F: Fn(RpcParams, &Context) -> Result<R, ServerCallError> + Send + Sync + 'static,
+		F: Fn(RpcParams, &Context) -> Result<R, CallError> + Send + Sync + 'static,
 	{
 		self.module.verify_method_name(method_name)?;
 
@@ -99,10 +99,8 @@ impl<Context> RpcContextModule<Context> {
 			Box::new(move |id, params, tx, _| {
 				match callback(params, &*ctx) {
 					Ok(res) => send_response(id, tx, res),
-					Err(ServerCallError::InvalidParams(_)) => {
-						send_error(id, tx, JsonRpcErrorCode::InvalidParams.into())
-					}
-					Err(ServerCallError::ContextFailed(err)) => {
+					Err(CallError::InvalidParams(_)) => send_error(id, tx, JsonRpcErrorCode::InvalidParams.into()),
+					Err(CallError::Failed(err)) => {
 						let err = JsonRpcErrorObject {
 							code: JsonRpcErrorCode::ServerError(CONTEXT_EXECUTION_FAILED_CODE),
 							message: &err.to_string(),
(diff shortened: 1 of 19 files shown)


http-server/src/module.rs
```diff
@@ -91,10 +97,16 @@ impl<Context> RpcContextModule<Context> {
...
-
-				send_response(id, tx, result);
-
+				match callback(params, &*ctx) {
+					Ok(res) => send_response(id, tx, res),
+					Err(ServerCallError::InvalidParams(_)) => {
+						send_error(id, tx, JsonRpcErrorCode::InvalidParams.into())
+					}
+					Err(ServerCallError::ContextFailed(err)) => {
+						let err = JsonRpcErrorObject { code: 1.into(), message: &err.to_string(), data: None };
```
niklasad1, line 106: hardcoded to `1` :cold_face:

emostov (approved):

http-server/src/module.rs
dvdplm: I think it ends up [here](https://github.com/paritytech/jsonrpsee/blob/c106710a2f7c14c2deacd549bcd5d7beccf8d121/types/src/v2/error.rs#L136) and that the number doesn't matter?
Still, not sure I follow why not use `JsonRpcErrorCode::ServerError(1)` explicitly here? And what the plan is for the `ServerError` code is.

types/src/error.rs
```diff
@@ -16,12 +16,37 @@ impl<T: fmt::Display> fmt::Display for Mismatch<T> {
...
+	fn from(params: InvalidParams) -> Self {
+		Self::InvalidParams(params)
+	}
+}
+
 /// Error type.
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
+	/// Error that may occur when a server fails to execute provided closure/callback.
+	/// Note, will most likely imply that the JSON-RPC request was valid.
```
dvdplm, line 44: Isn't it guaranteed that the request was valid if execution gets to a point where a `ServerCallError` can happen?

types/src/error.rs
```diff
@@ -16,12 +16,37 @@ impl<T: fmt::Display> fmt::Display for Mismatch<T> {
...
+impl From<InvalidParams> for ServerCallError {
+	fn from(params: InvalidParams) -> Self {
+		Self::InvalidParams(params)
+	}
+}
+
 /// Error type.
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
+	/// Error that may occur when a server fails to execute provided closure/callback.
```
dvdplm, line 43: The wording here seems to imply that these are errors that can happen when a closure/callback was provided, implicitly suggesting that there could be other ways to use the library. I think it'd be less easy to misunderstand if we just had "Error that may occur when a server fails to execute a call.".

ws-server/src/tests.rs
```diff
@@ -58,6 +58,19 @@ async fn single_method_call_with_params_works() {
...
 
+#[tokio::test]
+async fn single_method_call_with_faulty_params_returns_err() {
+	let _ = env_logger::try_init();
+	let (server_started_tx, server_started_rx) = oneshot::channel::<SocketAddr>();
+	tokio::spawn(server(server_started_tx));
+	let server_addr = server_started_rx.await.unwrap();
+	let mut client = WebSocketTestClient::new(server_addr).await.unwrap();
+
+	let req = r#"{"jsonrpc":"2.0","method":"add", "params":["Invalid"],"id":1}"#;
```
dvdplm, line 69: Would be good to cover a few more odd cases, e.g. `{"jsonrpc":"2.0","method":"add", "params":["Invalid"],"id":"hey ho"}`, `{"jsonrpc":"2.0","method":"add", "params": "Invalid", "id":1}`, `{"id":1}` etc

http-server/src/module.rs
```diff
@@ -91,10 +97,16 @@ impl<Context> RpcContextModule<Context> {
...
-				let result = callback(params, &*ctx)?;
-
-				send_response(id, tx, result);
-
+				match callback(params, &*ctx) {
+					Ok(res) => send_response(id, tx, res),
+					Err(ServerCallError::InvalidParams(_)) => {
+						send_error(id, tx, JsonRpcErrorCode::InvalidParams.into())
+					}
+					Err(ServerCallError::ContextFailed(err)) => {
```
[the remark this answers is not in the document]
niklasad1: It means that the user code/context/metadata wrapped in the closure failed nothing really related to jsonrpc spec.

The users of this API is supposed to provide their own errors wrapped in `ContextFailed` not to have use `Error::Custom` or similar.

http-server/src/module.rs
niklasad1: yepp, I think we could dedicate/document that an error code for this but `-32000 to -32099 | Server error | Reserved for implementation-defined server-errors` seems better for this

types/src/error.rs
niklasad1: yes, currently but that could change so I was vague here

ws-server/src/tests.rs
niklasad1: those would simplify be invalid `JSON-RPC requests`, seems unrelated we already got tests for this, except params  is not array or object.

http-server/src/tests.rs
```diff
@@ -42,14 +42,23 @@ async fn single_method_call_with_params() {
...
+async fn single_method_call_with_faulty_params_returns_err() {
+	let addr = server().await;
+	let uri = to_http_uri(addr);
+
+	let req = r#"{"jsonrpc":"2.0","method":"add", "params":["Invalid"],"id":1}"#;
+	let response = http_request(req.into(), uri).await.unwrap();
+	assert_eq!(response.status, StatusCode::OK);
+	assert_eq!(response.body, invalid_params(Id::Num(1)));
+}
+
```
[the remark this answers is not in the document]
niklasad1: sure, good idea.

ws-server/src/tests.rs
niklasad1: I get where the motivation for those tests but more related to #276 than this PR, will not address that here.

ws-server/src/server/module.rs
```diff
@@ -35,16 +38,17 @@ impl RpcModule {
 	pub fn register_method<F, R>(&mut self, method_name: &'static str, callback: F) -> Result<(), Error>
 	where
 		R: Serialize,
-		F: RpcMethod<R>,
+		F: RpcMethod<R, InvalidParams>,
```
maciejhirsz, line 41: Should we do something like `dyn Error` here instead? Or implement our own error trait that gives us the JSON-RPC error code along with the message?

niklasad1: > Or implement our own error trait that gives us the JSON-RPC error code along with the message?

That sounds neat and I like but it might be annoying users to explicitly implement that trait for the error types that they "plug-in" in `Error::ContextFailed(user err)`? However, it would be a way for the users to explicitly add there own error codes for this

types/src/error.rs
```diff
@@ -16,12 +16,36 @@ impl<T: fmt::Display> fmt::Display for Mismatch<T> {
...
+/// Invalid params.
+#[derive(Debug)]
+pub struct InvalidParams;
+
+/// Error that may occur when a server fails to execute a call.
+#[derive(Debug, thiserror::Error)]
+pub enum ServerCallError {
+	#[error("Invalid params in the RPC call")]
+	/// InvalidParams,
+	InvalidParams(InvalidParams),
```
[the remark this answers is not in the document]
niklasad1: I like `CallErrror` better too

types/src/error.rs
```diff
@@ -34,7 +58,7 @@ pub enum Error {
 	/// The background task has been terminated.
 	#[error("The background task been terminated because: {0}; restart required")]
 	RestartNeeded(String),
```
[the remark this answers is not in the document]
niklasad1: No, this is purely client related.

So, we could wrap in an `Client` or something to more specific. The WS server might close a connection though but no error internally is propogated

types/src/error.rs
```diff
@@ -16,12 +16,36 @@ impl<T: fmt::Display> fmt::Display for Mismatch<T> {
 	}
 }
 
+/// Invalid params.
+#[derive(Debug)]
+pub struct InvalidParams;
```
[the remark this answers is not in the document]
niklasad1: It was just that `register_method_without_context` can only fail when the params couldn't be parsed so it was introduced to avoid matching on unreachable enum variants

dvdplm (approved):
Let's get this in, just fix up the naming. Still a bit unsure about how the errors will work out in practice but this is better than what we had so let's merge it.

types/src/error.rs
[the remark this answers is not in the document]
niklasad1: I think it could be removed, lemme check

ws-server/src/server/module.rs
niklasad1: Let's create an issue for this and address it another PR

types/src/error.rs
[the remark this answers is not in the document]
niklasad1: I think `CallError::InvalidParams` is clear enough and `CallError::BadCallParams` looks redundant to me.

Maybe you are referring to ZST `InvalidParams`?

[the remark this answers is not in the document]
dvdplm: > I think `CallError::InvalidParams` is clear enough and `CallError::BadCallParams` looks redundant to me.

You're right, leave it as-is.

Issue #299, opened 2021-05-04
[server]: error types when register methods


niklasad1:
Should we do something like `dyn Error` here instead? Or implement our own error trait that gives us the JSON-RPC error code along with the message?

_Originally posted by @maciejhirsz in https://github.com/paritytech/jsonrpsee/pull/295#r624936990_

PR #300, opened 2021-05-04
[ws-server] Batch support

dvdplm:
Add batch request support for the websocket server. Continuation of #292.

Closes https://github.com/paritytech/jsonrpsee/issues/277

TODO:

- [x] better tests
- [x] sort out error, switch from `InvalidParams` to `CallError::Failed`
- [x] benchmarks

### DIFF
--- a/http-server/src/module.rs
+++ b/http-server/src/module.rs
@@ -1,6 +1,6 @@
 use jsonrpsee_types::v2::error::{JsonRpcErrorCode, JsonRpcErrorObject, CALL_EXECUTION_FAILED_CODE};
 use jsonrpsee_types::{
-	error::{CallError, Error, InvalidParams},
+	error::{CallError, Error},
 	traits::RpcMethod,
 	v2::params::RpcParams,
 };
@@ -45,12 +45,9 @@ impl RpcModule {
 			Box::new(move |id, params, tx, _| {
 				match callback(params) {
 					Ok(res) => send_response(id, tx, res),
-					// TODO: this looks wonky...
-					Err(CallError::InvalidParams(InvalidParams)) => {
-						send_error(id, tx, JsonRpcErrorCode::InvalidParams.into())
-					}
+					Err(CallError::InvalidParams) => send_error(id, tx, JsonRpcErrorCode::InvalidParams.into()),
 					Err(CallError::Failed(e)) => {
-						// TODO: do something smart(-er) with the error?
+						// TODO: Return the error message (`e`) to clients, see https://github.com/paritytech/jsonrpsee/issues/299
 						log::error!("Call failed with: {}", e);
 						send_error(id, tx, JsonRpcErrorCode::ServerError(CALL_EXECUTION_FAILED_CODE).into())
 					}
@@ -107,7 +104,7 @@ impl<Context> RpcContextModule<Context> {
 			Box::new(move |id, params, tx, _| {
 				match callback(params, &*ctx) {
 					Ok(res) => send_response(id, tx, res),
-					Err(CallError::InvalidParams(_)) => send_error(id, tx, JsonRpcErrorCode::InvalidParams.into()),
+					Err(CallError::InvalidParams) => send_error(id, tx, JsonRpcErrorCode::InvalidParams.into()),
 					Err(CallError::Failed(err)) => {
 						let err = JsonRpcErrorObject {
 							code: JsonRpcErrorCode::ServerError(CALL_EXECUTION_FAILED_CODE),
--- a/http-server/src/module.rs
+++ b/http-server/src/module.rs
@@ -46,10 +46,14 @@ impl RpcModule {
 				match callback(params) {
 					Ok(res) => send_response(id, tx, res),
 					Err(CallError::InvalidParams) => send_error(id, tx, JsonRpcErrorCode::InvalidParams.into()),
-					Err(CallError::Failed(e)) => {
-						// TODO: Return the error message (`e`) to clients, see https://github.com/paritytech/jsonrpsee/issues/299
-						log::error!("Call failed with: {}", e);
-						send_error(id, tx, JsonRpcErrorCode::ServerError(CALL_EXECUTION_FAILED_CODE).into())
+					Err(CallError::Failed(err)) => {
+						log::error!("Call failed with: {}", err);
+						let err = JsonRpcErrorObject {
+							code: JsonRpcErrorCode::ServerError(CALL_EXECUTION_FAILED_CODE),
+							message: &err.to_string(),
+							data: None,
+						};
+						send_error(id, tx, err)
 					}
 				};
 
--- a/http-server/src/module.rs
+++ b/http-server/src/module.rs
@@ -36,7 +36,7 @@ impl RpcModule {
 	pub fn register_method<F, R>(&mut self, method_name: &'static str, callback: F) -> Result<(), Error>
 	where
 		R: Serialize,
-		F: RpcMethod<R, InvalidParams>,
+		F: RpcMethod<R, CallError>,
 	{
 		self.verify_method_name(method_name)?;
 
@@ -45,7 +45,15 @@ impl RpcModule {
 			Box::new(move |id, params, tx, _| {
 				match callback(params) {
 					Ok(res) => send_response(id, tx, res),
-					Err(InvalidParams) => send_error(id, tx, JsonRpcErrorCode::InvalidParams.into()),
+					// TODO: this looks wonky...
+					Err(CallError::InvalidParams(InvalidParams)) => {
+						send_error(id, tx, JsonRpcErrorCode::InvalidParams.into())
+					}
+					Err(CallError::Failed(e)) => {
+						// TODO: do something smart(-er) with the error?
+						log::error!("Call failed with: {}", e);
+						send_error(id, tx, JsonRpcErrorCode::ServerError(CALL_EXECUTION_FAILED_CODE).into())
+					}
 				};
 
 				Ok(())
--- a/http-server/src/module.rs
+++ b/http-server/src/module.rs
@@ -1,5 +1,10 @@
-use jsonrpsee_types::{traits::RpcMethod, v2::params::RpcParams, Error};
-use jsonrpsee_utils::server::{send_response, Methods};
+use jsonrpsee_types::v2::error::{JsonRpcErrorCode, JsonRpcErrorObject, CALL_EXECUTION_FAILED_CODE};
+use jsonrpsee_types::{
+	error::{CallError, Error, InvalidParams},
+	traits::RpcMethod,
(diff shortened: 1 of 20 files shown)


http-server/src/module.rs
```diff
@@ -45,7 +45,15 @@ impl RpcModule {
...
 					Ok(res) => send_response(id, tx, res),
-					Err(InvalidParams) => send_error(id, tx, JsonRpcErrorCode::InvalidParams.into()),
+					// TODO: this looks wonky...
+					Err(CallError::InvalidParams(InvalidParams)) => {
+						send_error(id, tx, JsonRpcErrorCode::InvalidParams.into())
+					}
+					Err(CallError::Failed(e)) => {
+						// TODO: do something smart(-er) with the error?
+						log::error!("Call failed with: {}", e);
+						send_error(id, tx, JsonRpcErrorCode::ServerError(CALL_EXECUTION_FAILED_CODE).into())
```
niklasad1, line 55: you could include the error message in the error too if you'd like but requires some more boiler plate

dvdplm: Yeah I was wondering if we had a plan for that. Do you know? Users will be sad about seeing only an error code.

utils/src/server.rs
```diff
@@ -50,3 +51,21 @@ pub fn send_error(id: RpcId, tx: RpcSender, error: JsonRpcErrorObject) {
 		log::error!("Error sending response to the client: {:?}", err)
 	}
 }
+
+/// Read all the results of all method calls in a batch request from the ['Stream']. Format the result into a single
+/// `String` appropriately wrapped in `[`/`]`.
+pub async fn collect_batch_responses(rx: mpsc::UnboundedReceiver<String>) -> String {
```
niklasad1, line 57:
```suggestion
pub async fn collect_batch_response(rx: mpsc::UnboundedReceiver<String>) -> String {
```

http-server/src/module.rs
```diff
@@ -45,7 +45,15 @@ impl RpcModule {
 			Box::new(move |id, params, tx, _| {
 				match callback(params) {
 					Ok(res) => send_response(id, tx, res),
-					Err(InvalidParams) => send_error(id, tx, JsonRpcErrorCode::InvalidParams.into()),
+					// TODO: this looks wonky...
+					Err(CallError::InvalidParams(InvalidParams)) => {
+						send_error(id, tx, JsonRpcErrorCode::InvalidParams.into())
+					}
```
dvdplm, line 51: @niklasad1 what about this one, I am still not sure why `CallError::InvalidParams` has an `InvalidParam` field. It looks redundant here.

ws-server/src/server.rs
```diff
@@ -26,6 +26,7 @@
 
 use futures_channel::mpsc;
 use futures_util::io::{BufReader, BufWriter};
+use futures_util::stream::StreamExt;
```
dvdplm, line 29: @niklasad1 I realized that the ws server was using `tokio_stream::StreamExt` and the http server uses `futures_util::stream::StreamExt`. I don't know which is better or if they were using different versions on purpose. This here seems to work though.

ws-server/src/server.rs
```diff
@@ -166,38 +171,69 @@ async fn background_task(socket: tokio::net::TcpStream, methods: Arc<Methods>, i
...
+		// to [`serde_json::from_slice`] which is pretty annoying.
+		// Our [issue](https://github.com/paritytech/jsonrpsee/issues/296).
+		if let Ok(req) = serde_json::from_slice::<JsonRpcRequest>(&data) {
+			execute(&tx, req);
+		} else if let Ok(batch) = serde_json::from_slice::<Vec<JsonRpcRequest>>(&data) {
+			if !batch.is_empty() {
+				// Batch responses must be sent back as a single message so we the results from each request in the
+				// batch and read the results off of a new channel, `rx2`, and then send the complete batch response
+				// back to the client over `tx`.
+				let (tx2, mut rx2) = mpsc::unbounded::<String>();
```
niklasad1, line 217: ok, so create a channel for each batch request we receive then?

perhaps rename `tx2` and `rx2` to batch_tx and and batch_rx or something it little more descriptive.

ws-server/src/server.rs
```diff
@@ -166,38 +171,69 @@ async fn background_task(socket: tokio::net::TcpStream, methods: Arc<Methods>, i
...
+				for req in batch {
+					execute(&tx2, req);
+				}
+				// TODO: add a test with a slow method call to prove this is correct.
+				rx2.close();
```
niklasad1, line 221: It's really hard to test this without extracting this in a separate function or something.

But I'm almost certain that this is correct otherwise the receiving end wouldn't get `Ok(None)` and the receiver would never "complete", as we use fold in `collect_batch_response`

ws-server/src/server/module.rs
```diff
@@ -47,7 +47,15 @@ impl RpcModule {
 			Box::new(move |id, params, tx, _| {
 				match callback(params) {
 					Ok(res) => send_response(id, tx, res),
-					Err(InvalidParams) => send_error(id, tx, JsonRpcErrorCode::InvalidParams.into()),
+					// TODO: this looks wonky...
+					Err(CallError::InvalidParams(InvalidParams)) => {
```
niklasad1, line 51: I think we can remove this `InvalidParams` type and just use to `InvalidParams` variant,

perhaps link to https://github.com/paritytech/jsonrpsee/issues/299

niklasad1: thus, if CallError would implement a trait or something `JsonRpcErrorCode` then you could just to `err.to_error_code()` or something similar

ws-server/src/server/module.rs
```diff
@@ -47,7 +47,15 @@ impl RpcModule {
...
+						send_error(id, tx, JsonRpcErrorCode::InvalidParams.into())
+					}
+					Err(CallError::Failed(e)) => {
```
niklasad1, line 54: I'm still curious to understand how this could occur here

ws-server/src/tests.rs
```diff
@@ -202,3 +232,24 @@ async fn invalid_request_should_not_close_connection() {
...
+async fn valid_request_that_fails_to_execute_should_not_close_connection() {
+	let addr = server().await;
+	let mut client = WebSocketTestClient::new(addr).await.unwrap();
+
+	// Good request, executes fine
+	let request = r#"{"jsonrpc":"2.0","method":"say_hello","id":33}"#;
+	let response = client.send_request_text(request).await.unwrap();
+	assert_eq!(response, ok_response(JsonValue::String("hello".to_owned()), Id::Num(33)));
+
+	// Good request, but causes error.
```
niklasad1, line 284:
```suggestion
	// Good request with unexpected params
```

http-server/src/module.rs
niklasad1: You could remove it, it's redudant

http-server/src/module.rs
niklasad1: we already uses it for the context module because I assumed that only `InvalidParams` could happen here it was not considered.

so you could copy-paste that code from [here](https://github.com/paritytech/jsonrpsee/blob/master/ws-server/src/server/module.rs#L160-#L168) or extract to helper function

ws-server/src/server.rs
niklasad1: I think they are different not sure how but but uses the underlying `Stream trait` from futures.

IIRC, it's tokio here because `TcpListenerStream` is used from `tokio_stream` and avoided to another dep on `futures_util`

niklasad1 (commented):
Looks overall good, not exactly sure of the overhead w.r.t spawning an unbounded channel per batch request. 

Should be faster than reusing the existing channel and we should limit incoming requests based on payload size anyway so should be fine w.r.t OOM related issues.

ws-server/src/server.rs
dvdplm: Do you want me to switch back?

niklasad1:
An additional bonus to this PR, would be to add integration tests for batch requests (both HTTP and WS)

Could address it in another PR also.

ws-server/src/server.rs
dvdplm: The doubt I had when looking at it again today was: when the docs say no further sends can happen, do they talk about __time__ (as in: the instant `close` is called, can no further sends happen from any thread?) or to they mean __program order__ (as in: no `.send()` calls can be made after this line of code?).
I added a test where one of the method calls sleeps for 1500ms and it seems to work fine.

ws-server/src/server.rs
niklasad1: no, it's fine :)

http-server/src/module.rs
dvdplm: Ended up copy-pasting.

ws-server/src/tests.rs
dvdplm: No, the point here is to show that a **valid** request object that still fails to execute correctly comes back with the expected (app-level) error. And that this doesn't take down the server ofc.

niklasad1: ah, I see missed that it was an error,

niklasad1 (approved):

ws-server/src/server.rs
```diff
@@ -166,38 +171,69 @@ async fn background_task(socket: tokio::net::TcpStream, methods: Arc<Methods>, i
...
+		// For reasons outlined [here](https://github.com/serde-rs/json/issues/497), `RawValue` can't be used with
+		// untagged enums at the moment. This means we can't use an `SingleOrBatch` untagged enum here and have to try
+		// each case individually: first the single request case, then the batch case and lastly the error. For the
+		// worst case – unparseable input – we make three calls to [`serde_json::from_slice`] which is pretty annoying.
+		// Our [issue](https://github.com/paritytech/jsonrpsee/issues/296).
+		if let Ok(req) = serde_json::from_slice::<JsonRpcRequest>(&data) {
+			execute(&tx, req);
+		} else if let Ok(batch) = serde_json::from_slice::<Vec<JsonRpcRequest>>(&data) {
+			if !batch.is_empty() {
+				// Batch responses must be sent back as a single message so we the results from each request in the
```
insipx, line 213:
```suggestion
				// Batch responses must be sent back as a single message so we read the results from each request in the
```

not sure if read or get here, I think read makes sense though

ws-server/src/server.rs
```diff
@@ -166,38 +171,69 @@ async fn background_task(socket: tokio::net::TcpStream, methods: Arc<Methods>, i
...
+				// batch and read the results off of a new channel, `rx2`, and then send the complete batch response
```
insipx, line 214:
```suggestion
				// batch and read the results off of a new channel, `rx_batch`, and then send the complete batch response
```

insipx (approved):
Looks good other than small nits

PR #303, opened 2021-05-06
[ws client] RegisterNotification support

### LINKED ISSUE #301: [ws client] How to handle incoming notifications?
I'm using jsonrpsee to implement a websocket rpc client, and I need to handle some incoming notifications.  I don't see a way to register a handler for incoming notifications and curious if I'm missing something in the docs / examples.  Is this not supported?

Thanks!

billylindeman:
This adds the ability to register a NotificationHandler\<T\> for a method name, allowing jsonrpsee ws clients to support server originated notifications!  #301 

Usage:

```
let mut offers: NotificationHandler<SessionDescription> = client.register_notification("offer").await?;
```

This is my first PR on jsonrpsee so let me know what you think :). I think the error cases could use a second glance by someone more experienced on the project but it does work ATM.

@niklasad1 @dvdplm 

fixes #301

### DIFF
--- a/test-utils/src/types.rs
+++ b/test-utils/src/types.rs
@@ -1,4 +1,5 @@
 use futures_channel::mpsc::{self, Receiver, Sender};
+use futures_channel::oneshot;
 use futures_util::{
 	future::FutureExt,
 	io::{BufReader, BufWriter},
@@ -118,10 +119,19 @@ impl WebSocketTestServer {
 
 	// Spawns a dummy `JSONRPC v2` WebSocket server that sends out a pre-configured `hardcoded notification` for every connection.
 	pub async fn with_hardcoded_notification(sockaddr: SocketAddr, notification: String) -> Self {
-		let listener = async_std::net::TcpListener::bind(sockaddr).await.unwrap();
-		let local_addr = listener.local_addr().unwrap();
-		let (tx, rx) = mpsc::channel::<()>(4);
-		tokio::spawn(server_backend(listener, rx, ServerMode::Notification(notification)));
+		let (tx, rx) = mpsc::channel::<()>(1);
+		let (addr_tx, addr_rx) = oneshot::channel();
+
+		std::thread::spawn(move || {
+			let rt = tokio::runtime::Runtime::new().unwrap();
+			let listener = rt.block_on(async_std::net::TcpListener::bind(sockaddr)).unwrap();
+			let local_addr = listener.local_addr().unwrap();
+
+			addr_tx.send(local_addr).unwrap();
+			rt.block_on(server_backend(listener, rx, ServerMode::Notification(notification)));
+		});
+
+		let local_addr = addr_rx.await.unwrap();
 
 		Self { local_addr, exit: tx }
 	}
@@ -209,7 +219,8 @@ async fn connection_task(socket: async_std::net::TcpStream, mode: ServerMode, mu
 	loop {
 		let next_ws = ws_stream.next().fuse();
 		let next_exit = exit.next().fuse();
-		let time_out = tokio::time::sleep(Duration::from_secs(1)).fuse();
+		let time_out = tokio::time::sleep(Duration::from_millis(200)).fuse();
+
 		pin_mut!(time_out, next_exit, next_ws);
 
 		select! {
--- a/test-utils/src/types.rs
+++ b/test-utils/src/types.rs
@@ -95,6 +95,8 @@ pub enum ServerMode {
 	Response(String),
 	// Send out a subscription ID on a request and continuously send out data on the subscription.
 	Subscription { subscription_id: String, subscription_response: String },
+	// Send out a notification after timeout
+	Notification(String),
 }
 
 /// JSONRPC v2 dummy WebSocket server that sends a hardcoded response.
@@ -110,6 +112,16 @@ impl WebSocketTestServer {
 		let local_addr = listener.local_addr().unwrap();
 		let (tx, rx) = mpsc::channel::<()>(4);
 		tokio::spawn(server_backend(listener, rx, ServerMode::Response(response)));
+
+		Self { local_addr, exit: tx }
+	}
+
+	// Spawns a dummy `JSONRPC v2` WebSocket server that sends out a pre-configured `hardcoded notification` for every connection.
+	pub async fn with_hardcoded_notification(sockaddr: SocketAddr, notification: String) -> Self {
+		let listener = async_std::net::TcpListener::bind(sockaddr).await.unwrap();
+		let local_addr = listener.local_addr().unwrap();
+		let (tx, rx) = mpsc::channel::<()>(4);
+		tokio::spawn(server_backend(listener, rx, ServerMode::Notification(notification)));
 
 		Self { local_addr, exit: tx }
 	}
@@ -202,10 +214,18 @@ async fn connection_task(socket: async_std::net::TcpStream, mode: ServerMode, mu
 
 		select! {
 			_ = time_out => {
-				if let ServerMode::Subscription { subscription_response, .. } = &mode {
-					if let Err(e) = sender.send_text(&subscription_response).await {
-						log::warn!("send response to subscription: {:?}", e);
-					}
+				 match &mode {
+					ServerMode::Subscription { subscription_response, .. } => {
+						if let Err(e) = sender.send_text(&subscription_response).await {
+							log::warn!("send response to subscription: {:?}", e);
+						}
+					},
+					ServerMode::Notification(n) => {
+						if let Err(e) = sender.send_text(&n).await {
+							log::warn!("send notification: {:?}", e);
+						}
+					},
+					_ => {}
 				}
 			}
 			ws = next_ws => {
@@ -217,12 +237,13 @@ async fn connection_task(socket: async_std::net::TcpStream, mode: ServerMode, mu
 							if let Err(e) = sender.send_text(&r).await {
 								log::warn!("send response to request error: {:?}", e);
 							}
-						}
+						},
 						ServerMode::Subscription { subscription_id, .. } => {
 							if let Err(e) = sender.send_text(&subscription_id).await {
 								log::warn!("send subscription id error: {:?}", e);
 							}
-						}
+						},
+						_ => {}
 					}
 				}
 			}
--- a/types/src/client.rs
+++ b/types/src/client.rs
@@ -124,6 +124,28 @@ where
 	}
 }
 
+impl<Notif> NotificationHandler<Notif>
+where
+	Notif: DeserializeOwned,
+{
+	/// Returns the next notification from the stream
+	/// This may return `None` if the method has been unregistered,
+	/// may happen if the channel becomes full or is dropped.
+	///
+	/// Ignores any malformed packet.
+	pub async fn next(&mut self) -> Option<Notif> {
+		loop {
+			match self.notifs_rx.next().await {
+				Some(n) => match serde_json::from_value(n) {
+					Ok(parsed) => return Some(parsed),
+					Err(e) => log::debug!("NotificationHandler response error: {:?}", e),
+				},
+				None => return None,
+			}
+		}
+	}
+}
+
 impl<Notif> Drop for Subscription<Notif> {
 	fn drop(&mut self) {
 		// We can't actually guarantee that this goes through. If the background task is busy, then
--- a/types/src/client.rs
+++ b/types/src/client.rs
@@ -92,6 +92,8 @@ pub enum FrontToBack {
 	Subscribe(SubscriptionMessage),
 	/// Register a notification handler
 	RegisterNotification(RegisterNotificationMessage),
+	/// Unregister a notification handler
+	UnregisterNotification(String),
 	/// When a subscription channel is closed, we send this message to the background
 	/// task to mark it ready for garbage collection.
 	// NOTE: It is not possible to cancel pending subscriptions or pending requests.
@@ -130,5 +132,15 @@ impl<Notif> Drop for Subscription<Notif> {
 		// to the `Subscription` has been closed, and will perform the unsubscribe.
 		let id = std::mem::replace(&mut self.id, SubscriptionId::Num(0));
 		let _ = self.to_back.send(FrontToBack::SubscriptionClosed(id)).now_or_never();
+	}
+}
+
+impl<Notif> Drop for NotificationHandler<Notif> {
+	fn drop(&mut self) {
+		// We can't actually guarantee that this goes through. If the background task is busy, then
+		// the channel's buffer will be full, and our unsubscription request will never make it.
+		// However, when a notification arrives, the background task will realize that the channel
+		// to the `Subscription` has been closed, and will perform the unsubscribe.
+		let _ = self.to_back.send(FrontToBack::UnregisterNotification((&self.method).to_owned())).now_or_never();
 	}
 }
--- a/types/src/client.rs
+++ b/types/src/client.rs
@@ -56,6 +56,18 @@ pub struct SubscriptionMessage {
 	pub send_back: oneshot::Sender<Result<(mpsc::Receiver<JsonValue>, SubscriptionId), Error>>,
 }
 
+/// OnNotification message.
+#[derive(Debug)]
+pub struct OnNotificationMessage {
+	/// Request ID of the subscribe message.
+	pub sub_id: SubscriptionId,
+	pub req_id: u64,
+	/// We return a [`mpsc::Receiver`] that will receive notifications.
+	/// When we get a response from the server about that subscription, we send the result over
+	/// this channel.
+	pub send_back: oneshot::Sender<Result<(mpsc::Receiver<JsonValue>, SubscriptionId), Error>>,
+}
+
 /// Message that the Client can send to the background task.
 #[derive(Debug)]
 pub enum FrontToBack {
@@ -67,6 +79,8 @@ pub enum FrontToBack {
 	Request(RequestMessage),
 	/// Send a subscription request to the server.
 	Subscribe(SubscriptionMessage),
+	/// Create a notification handler subscription
+	OnNotification(OnNotificationMessage),
 	/// When a subscription channel is closed, we send this message to the background
 	/// task to mark it ready for garbage collection.
 	// NOTE: It is not possible to cancel pending subscriptions or pending requests.
--- a/types/src/client.rs
+++ b/types/src/client.rs
@@ -60,8 +60,9 @@ pub struct SubscriptionMessage {
 #[derive(Debug)]
 pub struct OnNotificationMessage {
 	/// Request ID of the subscribe message.
-	pub sub_id: SubscriptionId,
 	pub req_id: u64,
+	/// SubscriptionId the method name this notification handler is attached to
+	pub sub_id: SubscriptionId,
 	/// We return a [`mpsc::Receiver`] that will receive notifications.
 	/// When we get a response from the server about that subscription, we send the result over
 	/// this channel.
--- a/types/src/client.rs
+++ b/types/src/client.rs
@@ -23,7 +23,7 @@ pub struct NotificationHandler<Notif> {
 	pub to_back: mpsc::Sender<FrontToBack>,
 	/// Channel from which we receive notifications from the server, as encoded `JsonValue`s.
 	pub notifs_rx: mpsc::Receiver<JsonValue>,
-	/// Method,
+	/// Method Name
 	pub method: String,
 	/// Marker in order to pin the `Notif` parameter.
 	pub marker: PhantomData<Notif>,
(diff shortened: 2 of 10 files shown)


ws-client/src/helpers.rs
```diff
@@ -72,6 +72,33 @@ pub fn process_subscription_response(
...
+		Some(send_back_sink) => match send_back_sink.try_send(notif.params) {
+			Ok(()) => Ok(()),
+			Err(err) => {
+				log::error!("Error sending notification subscription {:?} error: {:?}", sub_id, err);
+				Err(Error::Internal(err.into_send_error()))
+			}
+		},
+		None => {
+			log::error!("Subscription ID: {:?} not an active subscription", sub_id);
+			Err(Error::InvalidSubscriptionId)
```
billylindeman, line 97: This error is a bit confusing on the frontend so maybe we should have a specific error for unexpected notification

niklasad1 (commented):
Looks overall good, thanks.

However, I think the code could be simplified/more readable by just using the method name and the Request Manager but it might needs some additional boiler plate code I suppose.

types/src/client.rs
```diff
@@ -56,6 +56,19 @@ pub struct SubscriptionMessage {
 	pub send_back: oneshot::Sender<Result<(mpsc::Receiver<JsonValue>, SubscriptionId), Error>>,
 }
 
+/// OnNotification message.
+#[derive(Debug)]
+pub struct OnNotificationMessage {
+	/// Request ID of the subscribe message.
+	pub req_id: u64,
+	/// SubscriptionId the method name this notification handler is attached to
+	pub sub_id: SubscriptionId,
```
niklasad1, line 65:
```suggestion
	pub method_name: SubscriptionId,
```

types/src/client.rs
```diff
@@ -56,6 +56,19 @@ pub struct SubscriptionMessage {
 	pub send_back: oneshot::Sender<Result<(mpsc::Receiver<JsonValue>, SubscriptionId), Error>>,
 }
 
+/// OnNotification message.
+#[derive(Debug)]
+pub struct OnNotificationMessage {
+	/// Request ID of the subscribe message.
+	pub req_id: u64,
```
niklasad1, line 63: This seems needless given that we only care about that `method_name`

I assume this is a workaround to work to work how current subscriptions are handled in the `RequestManager`.

types/src/traits.rs
```diff
@@ -44,6 +44,11 @@ pub trait SubscriptionClient: Client {
 	) -> Result<Subscription<Notif>, Error>
 	where
 		Notif: DeserializeOwned;
+
+	/// Register a Subscription<Notif> that will listen for incoming JSON-RPC notifications
+	async fn on_notification<'a, Notif>(&self, method: &'a str) -> Result<Subscription<Notif>, Error>
```
niklasad1, line 49: `register_notification` seems like a more appropriate name for this?

ws-client/src/client.rs
```diff
@@ -550,6 +587,20 @@ async fn background_task(
...
 				}
 			}
+
+			// User called `on_notification` on the front-end.
+			Either::Left((Some(FrontToBack::OnNotification(sub)), _)) => {
+				log::trace!("[backend] registering notification handler: {:?}", sub.sub_id);
+				let (subscribe_tx, subscribe_rx) = mpsc::channel(max_notifs_per_subscription);
+
+				if manager.insert_notification_handler(sub.req_id, sub.sub_id.clone(), subscribe_tx).is_ok() {
+					sub.send_back
```
niklasad1, line 597: This expect should be removed because if the user drops the notification handle to this will panic,

so just ignore the error here by `let _ = send(...)`

ws-client/src/helpers.rs
niklasad1: Yeah, just add a new error `InvalidNotification`, `UnregisteredNotification` or similar.

ws-client/src/client.rs
```diff
@@ -455,6 +456,42 @@ impl SubscriptionClient for WsClient {
 		};
 		Ok(Subscription { to_back: self.to_back.clone(), notifs_rx, marker: PhantomData, id })
 	}
+
+	/// Register a notification handler for async messages from the server.
+	///
+	async fn on_notification<'a, N>(&self, method: &'a str) -> Result<Subscription<N>, Error>
```
niklasad1, line 462: I think we need a new type for `Subscription`, the drop implementation will not work I think.

billylindeman: Yeah agreed actually this would let me clean things up quite a bit. I'll work on that today

types/src/client.rs
billylindeman: Yep, I will refactor out

ws-client/src/client.rs
billylindeman: Are you sure about this one?  This insert should happen in the code path for register_notification() and that async is blocked on this send (or it should fail to register).

types/src/client.rs
```diff
@@ -132,3 +134,13 @@ impl<Notif> Drop for Subscription<Notif> {
 		let _ = self.to_back.send(FrontToBack::SubscriptionClosed(id)).now_or_never();
 	}
 }
+
+impl<Notif> Drop for NotificationHandler<Notif> {
```
billylindeman, line 138: @niklasad1 Can you take a look at my drop implementation / follow this code path?

niklasad1: it looks good

types/src/client.rs
```diff
@@ -132,3 +134,13 @@ impl<Notif> Drop for Subscription<Notif> {
...
+	fn drop(&mut self) {
+		// We can't actually guarantee that this goes through. If the background task is busy, then
```
niklasad1, line 140: can you update this comment? little outdated now.

types/src/client.rs
```diff
@@ -132,3 +134,13 @@ impl<Notif> Drop for Subscription<Notif> {
...
+		// the channel's buffer will be full, and our unsubscription request will never make it.
+		// However, when a notification arrives, the background task will realize that the channel
+		// to the `Subscription` has been closed, and will perform the unsubscribe.
+		let _ = self.to_back.send(FrontToBack::UnregisterNotification((&self.method).to_owned())).now_or_never();
```
niklasad1, line 144:
```suggestion
        let notif_method = std::mem::replace(self.method);
		let _ = self.to_back.send(FrontToBack::UnregisterNotification(notif_method)).now_or_never();
```

ws-client/src/client.rs
```diff
@@ -602,6 +602,12 @@ async fn background_task(
 					let _ = reg.send_back.send(Err(Error::MethodAlreadyRegistered(reg.method)));
 				}
 			}
+
+			// User called `on_notification` on the front-end.
+			Either::Left((Some(FrontToBack::UnregisterNotification(method)), _)) => {
+				log::trace!("[backend] unregistering notification handler: {:?}", method);
+				let _ = manager.remove_notification_handler(&method);
```
niklasad1, line 609: I think you can pass this by value instead then you don't have to clone/allocate in `remove_notification`

ws-client/src/client.rs
niklasad1: That's a good comment but the reason is that the user may drop the `register_notifiction future` which is possible and then this will panic AFAIU ^^

Probably quite unlikely but better be safe than sorry I guess

ws-client/src/client.rs
```diff
@@ -455,6 +456,43 @@ impl SubscriptionClient for WsClient {
...
+			Ok(Ok(val)) => val,
+			Ok(Err(err)) => return Err(err),
+			Err(_) => return Err(self.read_error_from_backend().await),
+		};
+
+		Ok(NotificationHandler {
+			to_back: self.to_back.clone(),
+			notifs_rx,
+			marker: PhantomData,
+			method: method.to_owned(),
```
niklasad1, line 493:
```suggestion
			method,
```

ws-client/src/manager.rs
```diff
@@ -240,6 +261,17 @@ impl RequestManager {
 		}
 	}
 
+	/// Get a mutable reference to underlying `Sink` in order to send incmoing notifications to the subscription.
+	///
+	/// Returns `Some` if the `request_id` was registered as a NotificationHandler otherwise `None`.
+	pub fn as_notification_handler_mut(&mut self, method: &str) -> Option<&mut SubscriptionSink> {
+		if let Some(sink) = self.notification_handlers.get_mut(&method.to_owned()) {
```
niklasad1, line 268:
```suggestion
		self.notification_handlers.get_mut(method)
```

ws-client/src/client.rs
billylindeman: 👍  ah that does make sense

ws-client/src/helpers.rs
```diff
@@ -72,6 +72,27 @@ pub fn process_subscription_response(
...
+///
+/// Returns Ok() if the response was successfully handled
+/// Returns Err() if there was no handler for the method
+
+pub fn process_notification(manager: &mut RequestManager, notif: JsonRpcNotifResponse<JsonValue>) -> Result<(), Error> {
+	match manager.as_notification_handler_mut(&notif.method.to_owned()) {
+		Some(send_back_sink) => match send_back_sink.try_send(notif.params) {
+			Ok(()) => Ok(()),
+			Err(err) => {
+				log::error!("Error sending notification subscription {:?} error: {:?}", notif.method, err);
```
niklasad1, line 85: An error here will be that the channel is full which means that the user couldn't keep with the notifications.

We should drop the notification to indicate this to the user, we do that for subscriptions.

billylindeman:
Thanks for all the feed back @niklasad1 I should be able to address all these tomorrow and update the PR

types/src/client.rs
```diff
@@ -17,6 +17,18 @@ pub struct Subscription<Notif> {
 	pub marker: PhantomData<Notif>,
 }
 
+/// Active NotificationHandler on a Client.
+pub struct NotificationHandler<Notif> {
+	/// Channel to send requests to the background task.
+	pub to_back: mpsc::Sender<FrontToBack>,
+	/// Channel from which we receive notifications from the server, as encoded `JsonValue`s.
+	pub notifs_rx: mpsc::Receiver<JsonValue>,
+	/// Method,
```
dvdplm, line 26:
```suggestion
	/// Method name
```
Maybe it'll be clearer once I read through, but why is this needed?

types/src/client.rs
```diff
@@ -17,6 +17,18 @@ pub struct Subscription<Notif> {
...
+	pub method: String,
+	/// Marker in order to pin the `Notif` parameter.
+	pub marker: PhantomData<Notif>,
```
dvdplm, line 29: Does this need to be `pub`?

types/src/error.rs
```diff
@@ -67,6 +67,9 @@ pub enum Error {
 	/// Invalid request ID.
 	#[error("Invalid request ID")]
 	InvalidRequestId,
+	/// Unregistered notification method
```
dvdplm, line 70: Can we extend the docs here a little bit, it's not immediately clear (to me) what "notification" means here, it being a word used for a few different things in the library. Maybe just a note about when this can happen?

ws-client/src/client.rs
```diff
@@ -455,6 +456,43 @@ impl SubscriptionClient for WsClient {
...
+
+		let res = send_back_rx.await;
+		let (notifs_rx, method) = match res {
+			Ok(Ok(val)) => val,
+			Ok(Err(err)) => return Err(err),
+			Err(_) => return Err(self.read_error_from_backend().await),
+		};
+
+		Ok(NotificationHandler {
+			to_back: self.to_back.clone(),
```
dvdplm, line 490: Do you really need to clone `to_back` both here and on line 470 above?

ws-client/src/client.rs
```diff
@@ -550,6 +588,26 @@ async fn background_task(
...
+				if manager.insert_notification_handler(&reg.method, subscribe_tx).is_ok() {
+					reg.send_back
+						.send(Ok((subscribe_rx, reg.method)))
+						.expect("error sending response for notification handler");
+				} else {
+					let _ = reg.send_back.send(Err(Error::MethodAlreadyRegistered(reg.method)));
+				}
+			}
+
+			// User called `on_notification` on the front-end.
```
dvdplm, line 606: Is this comment correct? Why does calling `on_notification` cause de-registration?

ws-client/src/client.rs
```diff
@@ -566,12 +624,23 @@ async fn background_task(
...
 					log::debug!("[backend]: recv subscription {:?}", notif);
 					if let Err(Some(unsub)) = process_subscription_response(&mut manager, notif) {
 						let _ = stop_subscription(&mut sender, &mut manager, unsub).await;
 					}
 				}
+				// Incoming Notification
+				else if let Ok(notif) = serde_json::from_slice::<JsonRpcNotifResponse<_>>(&raw) {
+					log::debug!("[backend]: recv notification {:?}", notif);
+					match process_notification(&mut manager, notif) {
+						Ok(_) => return,
```
dvdplm, line 637: @niklasad1 is it correct to exit the `loop` here?

dvdplm (commented):
Good progress here, tyvm.

I for one could really use a proper example of how to use this, it would help clarify what the API looks like. Do you mind adding one?

types/src/client.rs
```diff
@@ -17,6 +17,18 @@ pub struct Subscription<Notif> {
 	pub marker: PhantomData<Notif>,
 }
 
+/// Active NotificationHandler on a Client.
+pub struct NotificationHandler<Notif> {
```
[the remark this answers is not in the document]
niklasad1: No, the issue is the drop implementation but I guess we could have a flag or something internally.

This is fine for me though :)

ws-client/src/manager.rs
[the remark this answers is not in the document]
niklasad1: `get_mut` should be enough ^

types/src/client.rs
niklasad1: It's need because the user declares that wants to "subscribe" for all notifications for `method` then this is used to filter out notifications that the user it interested in.

It's in essence similar to `subscriptions` but then the subscription ID is used for that instead.

ws-client/src/client.rs
niklasad1, line 606:
```suggestion
			// User called `register_notification` on the front-end.
```

ws-client/src/client.rs
niklasad1: nice catch, it shouldn't

types/src/client.rs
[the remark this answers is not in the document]
dvdplm: I tried this and it compiles but I realized there are no tests so my code is likely wrong then. 

My drop impl just became `let _ = self.inner.to_back.send(FrontToBack::UnregisterNotification((&self.method).to_owned())).now_or_never();`

dvdplm (changes requested):
Some good tests would be good.

types/src/client.rs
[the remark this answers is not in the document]
niklasad1: yeah, that should work too.

types/src/client.rs
billylindeman: Not sure.. @niklasad1?  It is pub in the Subscription<T>

niklasad1: It needs to be pub unless we provide a `fn new` ^

ws-client/src/client.rs
```diff
@@ -595,18 +590,16 @@ async fn background_task(
...
-						.send(Ok((subscribe_rx, reg.method)))
-						.expect("error sending response for notification handler");
+					let _ = reg.send_back.send(Ok((subscribe_rx, reg.method)));
 				} else {
 					let _ = reg.send_back.send(Err(Error::MethodAlreadyRegistered(reg.method)));
 				}
 			}
 
-			// User called `on_notification` on the front-end.
+			// User dopped the notificationHandler for this method
```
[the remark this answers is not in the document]
billylindeman: yes  /sigh

ws-client/src/client.rs
```diff
@@ -634,10 +627,10 @@ async fn background_task(
 				else if let Ok(notif) = serde_json::from_slice::<JsonRpcNotifResponse<_>>(&raw) {
 					log::debug!("[backend]: recv notification {:?}", notif);
 					match process_notification(&mut manager, notif) {
-						Ok(_) => return,
+						Ok(_) => {}
 						Err(err) => {
 							let _ = front_error.send(err);
-							return;
+							break;
```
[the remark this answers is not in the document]
niklasad1: It doesn't matter in this context, so I was just inconsistent I guess ^

ws-client/src/manager.rs
```diff
@@ -240,6 +261,13 @@ impl RequestManager {
 		}
 	}
 
+	/// Get a mutable reference to underlying `Sink` in order to send incmoing notifications to the subscription.
+	///
+	/// Returns `Some` if the `request_id` was registered as a NotificationHandler otherwise `None`.
+	pub fn as_notification_handler_mut(&mut self, method: &str) -> Option<&mut SubscriptionSink> {
+		self.notification_handlers.get_mut(&method.to_owned())
```
niklasad1, line 268:
```suggestion
		self.notification_handlers.get_mut(method)
```

ws-client/src/helpers.rs
niklasad1: Then the user has re-register the method again to use it again or increase the size of buffer ^

types/src/client.rs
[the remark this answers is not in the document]
billylindeman: > Some good tests would be good.

@dvdplm I added a basic test case for the happy path.  Open to suggestions for more tests

billylindeman:
@dvdplm I added notification support to the mock server and a test for it.  Any other test suggestions?

ws-client/src/helpers.rs
```diff
@@ -72,6 +72,27 @@ pub fn process_subscription_response(
...
+///
+/// Returns Ok() if the response was successfully handled
+/// Returns Err() if there was no handler for the method
+pub fn process_notification(manager: &mut RequestManager, notif: JsonRpcNotifResponse<JsonValue>) -> Result<(), Error> {
+	match manager.as_notification_handler_mut(notif.method.to_owned()) {
+		Some(send_back_sink) => match send_back_sink.try_send(notif.params) {
+			Ok(()) => Ok(()),
+			Err(err) => {
+				log::error!("Error sending notification, dropping handler for {:?} error: {:?}", notif.method, err);
+				let _ = manager.remove_notification_handler(notif.method.to_owned());
```
billylindeman, line 85: @niklasad1 So we should remove the handler her / drop it in this scope?

niklasad1: yepp, that should be it.

if you could just port this [test](https://github.com/paritytech/jsonrpsee/blob/master/tests/tests/integration_tests.rs#L139-#L166) to the register_notification that would be great, then it should be good

feel free to split it into two tests if you want.

test-utils/src/types.rs
```diff
@@ -114,6 +116,16 @@ impl WebSocketTestServer {
 		Self { local_addr, exit: tx }
 	}
 
+	// Spawns a dummy `JSONRPC v2` WebSocket server that sends out a pre-configured `hardcoded notification` for every connection.
+	pub async fn with_hardcoded_notification(sockaddr: SocketAddr, notification: String) -> Self {
+		let listener = async_std::net::TcpListener::bind(sockaddr).await.unwrap();
```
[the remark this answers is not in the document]
niklasad1: It's `async-std` when I wrote it for a long time ago to avoid using `compat` stuff but we could get rid of that the make the dependency tree a little smaller

niklasad1 (commented):
Thanks grinding this out,

One last thing:
Can you could just port this [test](https://github.com/paritytech/jsonrpsee/blob/master/tests/tests/integration_tests.rs#L139-#L166) to register_notification that would be great, then it should be good

ws-client/src/helpers.rs
billylindeman: done

test-utils/src/types.rs
```diff
@@ -114,6 +117,25 @@ impl WebSocketTestServer {
 		Self { local_addr, exit: tx }
 	}
 
+	// Spawns a dummy `JSONRPC v2` WebSocket server that sends out a pre-configured `hardcoded notification` for every connection.
+	pub async fn with_hardcoded_notification(sockaddr: SocketAddr, notification: String) -> Self {
+		let (tx, rx) = mpsc::channel::<()>(1);
+		let (addr_tx, addr_rx) = oneshot::channel();
+
+		std::thread::spawn(move || {
```
niklasad1, line 125: why spawning a thread for this instead `tokio spawn`?

PR #312, opened 2021-05-11
Change order of type params to register_method

dvdplm:
I think it reads a bit better with the return type being the first type param and the second anonymous, i.e. `register_method::<MyReturnType, _>(… …)` than `register_method::<_, MyReturnType>(… …)`. ﻿

### DIFF
--- a/ws-server/src/tests.rs
+++ b/ws-server/src/tests.rs
@@ -5,6 +5,7 @@ use jsonrpsee_test_utils::helpers::*;
 use jsonrpsee_test_utils::types::{Id, TestContext, WebSocketTestClient};
 use jsonrpsee_types::error::{CallError, Error};
 use serde_json::Value as JsonValue;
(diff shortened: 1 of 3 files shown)


insipx (approved):
nice, yeah second as anonymous is better imo

EDIT: I approved this but it seems like CI doesn't like the style, what's the point of B::Two?

niklasad1 (approved):

ws-server/src/tests.rs
```diff
@@ -4,6 +4,7 @@ use crate::{RpcContextModule, WsServer};
 use jsonrpsee_test_utils::helpers::*;
 use jsonrpsee_test_utils::types::{Id, TestContext, WebSocketTestClient};
 use jsonrpsee_types::error::{CallError, Error};
+use serde::Serialize;
```
niklasad1, line 7:
```suggestion
```

PR #316, opened 2021-05-12
[ws server]: SubscriptionSink: different API for register subscription with or without params.

niklasad1:
Example design to address to that subscriptions on the server-side needs to expose an API to get params in "subscription request"

This design provides different APIs to register subscriptions with or without params.
Such that if the incorrect param type is received in the request, nothing will be transmitted to the subscriber.

Warning, very boiler-plateyyy

### DIFF
--- a/ws-server/src/server.rs
+++ b/ws-server/src/server.rs
@@ -29,11 +29,10 @@ use futures_util::io::{BufReader, BufWriter};
 use futures_util::stream::StreamExt;
 use parking_lot::Mutex;
 use rustc_hash::FxHashMap;
-use serde::Serialize;
+use serde::{Serialize, de::DeserializeOwned};
 use serde_json::value::to_raw_value;
 use soketto::handshake::{server::Response, Server as SokettoServer};
-use std::net::SocketAddr;
-use std::sync::Arc;
+use std::{net::SocketAddr, sync::Arc};
 use tokio::net::{TcpListener, ToSocketAddrs};
 use tokio_stream::wrappers::TcpListenerStream;
 use tokio_util::compat::TokioAsyncReadCompatExt;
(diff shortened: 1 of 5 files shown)


ws-server/src/server.rs
```diff
@@ -120,6 +162,15 @@ impl Server {
 		self.root.register_subscription(subscribe_method_name, unsubscribe_method_name)
 	}
 
+	/// Register a new RPC subscription with the possibility to get the params in subscription request.
+	pub fn register_subscription_with_params<P: DeserializeOwned + Send + Sync + 'static>(
```
dvdplm, line 166: Shouldn't this be the default? I don't see the point of having both `register_subscription` and `register_subscription_with_params` and force users to decide which they want to use. (But maybe it's like this just to experiment?)

ws-server/src/server/module.rs
```diff
@@ -122,6 +122,66 @@ impl RpcModule {
...
+		self.verify_method_name(subscribe_method_name)?;
+		self.verify_method_name(unsubscribe_method_name)?;
+
+		let subscribers = Arc::new(Mutex::new(FxHashMap::default()));
+
+		{
+			let subscribers = subscribers.clone();
+			self.methods.insert(
+				subscribe_method_name,
+				Box::new(move |id, params, tx, _| {
```
dvdplm, line 144: How come the `conn` isn't important now?

ws-server/src/server.rs
niklasad1: it's much cleaner otherwise we must have been a bunch of runtime checks not to send out responses on the wrong subscription when using send.

Like the user must still make the assumption whether the use `send_all` or `send based on this input to a specific subscription`, so why not have diffrent types/API for it? Then we can enforce more stuff at compile-time than at runtime

but it's still an experiment :)

ws-server/src/server/module.rs
```diff
@@ -122,6 +122,66 @@ impl RpcModule {
 		Ok(SubscriptionSink { method: subscribe_method_name, subscribers })
 	}
 
+	/// Register a new RPC subscription, with subscribe and unsubscribe methods.
+	pub fn register_subscription_with_params<P: DeserializeOwned + Send + Sync + 'static>(
```
niklasad1, line 126: todo: remove boiler plate when we decided whether this is ok or not

maciejhirsz: Do we need two methods for this? Wouldn't it be enough to have a default `()` here?

niklasad1: yeah, good point completely useless

niklasad1:
Opened #318 to show a different solution, personally I like this better.

niklasad1:
//cc @maciejhirsz @dvdplm can you take a look at this and #318?

I think `design #1` will be a little more annoying to handle with `proc macros` or they need different annotations.

However, I need this to properly implement subscriptions in substrate FWIW

niklasad1:
Going forward with #318

PR #318, opened 2021-05-12
[servers]: subscriptionSink support params.

niklasad1:
Currently, the params within a subscription request is kept inside the server thus not possible to access the params.

This PR changes the `SubscriptionSink<Param>` to take a type parameter that must correspond to the param in the subscription request otherwise the request will be denied (not send back a subscription ID)

It adds two additional APIs:

- `call_with_params`: a way to apply a closure to the params and send out the response on that specific subscription
- `to_sinks`: consumes the current subscriptions, to expose the inner subscription sinks.


ws-server/src/server.rs
```diff
@@ -87,6 +92,85 @@ impl SubscriptionSink {
...
 	}
+
+	/// Extract subscriptions with input.
+	/// Usually, you want process the input before sending back
+	/// data on that subscription.
+	pub fn extract_with_input(&self) -> Vec<InnerSinkWithParams<P>> {
+		let mut subs = self.subscribers.lock();
+
+		let mut input = Vec::new();
+		*subs = std::mem::replace(&mut *subs, FxHashMap::default())
```
dvdplm, line 103:
```suggestion
		*subs = std::mem::take(&mut *subs)
```

ws-server/src/server.rs
```diff
@@ -87,6 +92,85 @@ impl SubscriptionSink {
 
 		Ok(())
 	}
+
+	/// Extract subscriptions with input.
+	/// Usually, you want process the input before sending back
+	/// data on that subscription.
+	pub fn extract_with_input(&self) -> Vec<InnerSinkWithParams<P>> {
```
dvdplm, line 99: Still struggling to grok the impact of this. It looks like it could be a pretty big performance hit to loop over all the subscriptions to find the ones using parameters?

niklasad1: That's a good comment, I was mostly thinking that the `Mutex` is the bottleneck in this context.

But again, this code is very "draftisch" and I haven't really put in much effort into mostly to show a point.
I guess we could put `with params` and `params` in different containers to avoid `O(n)` lookups or something

niklasad1: Best way is ofc to benchmark this :)

maciejhirsz (commented):
I had to have a bigger think about this and #316. I think I prefer this PR since it's simpler, and the `InnerSink` makes sense to me.

niklasad1:
> I had to have a bigger think about this and #316. I think I prefer this PR since it's simpler, and the InnerSink makes sense to me.

Ok, let's then polish this and create some benchmarks it on after @dvdplm grumbles ^^

tests/tests/integration_tests.rs
```diff
@@ -54,6 +54,20 @@ async fn ws_subscription_works() {
...
 
+#[tokio::test]
+async fn ws_subscription_with_input_works() {
+	let server_addr = websocket_server_with_subscription().await;
+	let server_url = format!("ws://{}", server_addr);
+	let client = WsClientBuilder::default().build(&server_url).await.unwrap();
+	let mut add_one: Subscription<u64> =
+		client.subscribe("subscribe_add_one", vec![1.into()].into(), "unsubscribe_add_one").await.unwrap();
+
+	for _ in 0..10 {
```
niklasad1, line 65: just show a point that the actual param doesn't get mutated.

utils/src/server/rpc_module.rs
```diff
@@ -242,8 +246,45 @@ impl SubscriptionSink {
...
+
+		Ok(())
+	}
+
+	/// Send a message to all subscriptions that could parse `P` as input.
+	///
+	/// F: is closure that you need to provide to apply on the input P.
+	pub fn send_all_with_params<T, F>(&self, f: F) -> anyhow::Result<()>
+	where
+		F: Fn(&P) -> T,
```
niklasad1, line 268: do we need `F: Fn(&mut P) -> T` too?

i,e, to mutate the param?

examples/ws_subscription.rs
```diff
@@ -54,10 +54,10 @@ async fn main() -> anyhow::Result<()> {
 
 async fn run_server() -> anyhow::Result<SocketAddr> {
 	let mut server = WsServer::new("127.0.0.1:0").await?;
-	let mut subscription = server.register_subscription("subscribe_hello", "unsubscribe_hello").unwrap();
+	let subscription = server.register_subscription::<String>("subscribe_hello", "unsubscribe_hello").unwrap();
```
dvdplm, line 57:
```suggestion
	let subscription = server.register_subscription::<()>("subscribe_hello", "unsubscribe_hello").unwrap();
```

To make it clear that the `P` type parameter is the parameters sent to the subscription (and there are none here).

examples/ws_subscription.rs
```diff
@@ -54,10 +54,10 @@ async fn main() -> anyhow::Result<()> {
...
 
 	std::thread::spawn(move || loop {
-		subscription.send(&"hello my friend").unwrap();
+		subscription.send_all(&"hello my friend").unwrap();
```
dvdplm, line 60: What do you think of renaming `send_all` to `broadcast`?

tests/tests/integration_tests.rs
dvdplm: maybe 5 iterations is enough? lol

utils/src/server/rpc_module.rs
```diff
@@ -27,7 +27,7 @@ pub type SubscriptionId = u64;
 /// Sink that is used to send back the result to the server for a specific method.
 pub type MethodSink = mpsc::UnboundedSender<String>;
 
-type Subscribers = Arc<Mutex<FxHashMap<(ConnectionId, SubscriptionId), MethodSink>>>;
+type Subscribers<P> = Arc<Mutex<FxHashMap<(ConnectionId, SubscriptionId), InnerSink<P>>>>;
```
dvdplm, line 30: Should this be `pub`? I can't think of a good reason right now, but it feels like maybe it should?

dvdplm, line 30:
```suggestion
/// Map of subscribers keyed by the connection and subscription ids to an [`InnerSink`] that contains the parameters
/// they used to subscribe and the tx side of a channel used to convey results&errors back.
type Subscribers<P> = Arc<Mutex<FxHashMap<(ConnectionId, SubscriptionId), InnerSink<P>>>>;
```

utils/src/server/rpc_module.rs
```diff
@@ -27,7 +27,7 @@ pub type SubscriptionId = u64;
...
 
 /// Sets of JSON-RPC methods can be organized into a "module" that are in turn registered on server or,
 /// alternatively, merged with other modules to construct a cohesive API.
```
dvdplm, line 33:
```suggestion
/// Sets of JSON-RPC methods can be organized into a "module"s that are in turn registered on the server or,
/// alternatively, merged with other modules to construct a cohesive API.
```

utils/src/server/rpc_module.rs
```diff
@@ -88,11 +88,11 @@ impl RpcModule {
 	}
 
 	/// Register a new RPC subscription, with subscribe and unsubscribe methods.
-	pub fn register_subscription(
+	pub fn register_subscription<P: DeserializeOwned + Send + Sync + 'static>(
```
dvdplm, line 91: What do you think about using `Params` instead of just `P`?

utils/src/server/rpc_module.rs
```diff
@@ -88,11 +88,11 @@ impl RpcModule {
 	}
 
 	/// Register a new RPC subscription, with subscribe and unsubscribe methods.
```
dvdplm, line 90:
```suggestion
	/// Register a new RPC subscription, with subscribe and unsubscribe methods. Returns a [`SubscriptionSink`]. If a
	/// method with the same name is already registered, an [`Error::MethodAlreadyRegistered`] is returned.
    /// If the subscription does not take any parameters, set `P` to `()`.
```

utils/src/server/rpc_module.rs
```diff
@@ -217,16 +219,18 @@ impl<Context> RpcContextModule<Context> {
...
+	subscribers: Subscribers<P>,
 }
 
-impl SubscriptionSink {
-	/// Send data back to subscribers.
-	/// If a send fails (likely a broken connection) the subscriber is removed from the sink.
-	/// O(n) in the number of subscribers.
-	pub fn send<T>(&mut self, result: &T) -> anyhow::Result<()>
+impl<P> SubscriptionSink<P> {
+	/// Send a message on all the subscriptions
```
dvdplm, line 228:
```suggestion
	/// Send a message on all the subscribers.
```

utils/src/server/rpc_module.rs
```diff
@@ -217,16 +219,18 @@ impl<Context> RpcContextModule<Context> {
...
+	///
+	/// If you have subscriptions with params/input you should most likely
+	/// call `call_with_params` to the process the input/params and send out
+	/// the result on each subscription individually instead.
```
dvdplm, line 232:
```suggestion
	/// Don't use this if there are subscriptions with params. Use [`send_with_params`] instead to process the params
	/// and send the result on each subscription individually instead.
```

utils/src/server/rpc_module.rs
```diff
@@ -242,8 +246,45 @@ impl SubscriptionSink {
 				params: JsonRpcNotificationParams { subscription: *sub_id, result: &*result },
 			})?;
 
-			// Track broken connections
-			if sender.unbounded_send(msg).is_err() {
+			// Log broken connections
```
dvdplm, line 249:
```suggestion
			// Track broken connections
```

We're not logging anything I think?

utils/src/server/rpc_module.rs
```diff
@@ -242,8 +246,45 @@ impl SubscriptionSink {
...
+		for entry in errored {
+			subs.remove(&entry);
+		}
+
+		Ok(())
+	}
+
+	/// Send a message to all subscriptions that could parse `P` as input.
+	///
+	/// F: is closure that you need to provide to apply on the input P.
```
dvdplm, line 265:
```suggestion
	/// F: is a closure that you need to provide to apply on the input P.
```

utils/src/server/rpc_module.rs
dvdplm: Hmm. Maybe, but until we know we do I say we wait.

utils/src/server/rpc_module.rs
```diff
@@ -242,8 +246,45 @@ impl SubscriptionSink {
...
+
+		// Remove broken connections
+		for entry in errored {
+			subs.remove(&entry);
+		}
+
+		Ok(())
+	}
+
+	/// Send a message to all subscriptions that could parse `P` as input.
```
dvdplm, line 263:
```suggestion
	/// Send a message to all subscribers one by one, parsing the params they sent with the provided closure. If the
	/// closure `F` fails to parse the params the message is not sent.
```

utils/src/server/rpc_module.rs
```diff
@@ -242,8 +246,45 @@ impl SubscriptionSink {
...
+	///
+	/// F: is closure that you need to provide to apply on the input P.
+	pub fn send_all_with_params<T, F>(&self, f: F) -> anyhow::Result<()>
```
dvdplm, line 266: How about `send_each`?

utils/src/server/rpc_module.rs
```diff
@@ -252,6 +293,14 @@ impl SubscriptionSink {
 		for entry in errored {
 			subs.remove(&entry);
```
dvdplm, line 294: Log here?

dvdplm (approved):
lgtm, modulo a bunch of nits and some naming.

utils/src/server/rpc_module.rs
niklasad1: No, it's just an internal type so no need to make it `pub` yet.

examples/ws_subscription.rs
niklasad1: sounds good.

utils/src/server/rpc_module.rs
niklasad1: it looks like I need it for my `author_WatchExtrinsic` branch, I will try it out now

Thus, changed to `F: Fn(&mut Params) -> T`

utils/src/server/rpc_module.rs
```diff
@@ -219,18 +223,18 @@ impl<Context> RpcContextModule<Context> {
...
+pub struct SubscriptionSink<Params> {
 	method: &'static str,
-	subscribers: Subscribers<P>,
+	subscribers: Subscribers<Params>,
 }
 
-impl<P> SubscriptionSink<P> {
-	/// Send a message on all the subscriptions
+impl<Params> SubscriptionSink<Params> {
+	/// Send a message on the all the subscribers.
```
dvdplm, line 232:
```suggestion
	/// Send a message to all subscribers.
```

utils/src/server/rpc_module.rs
```diff
@@ -217,16 +223,18 @@ impl<Context> RpcContextModule<Context> {
 
 /// Used by the server to send data back to subscribers.
 #[derive(Clone)]
-pub struct SubscriptionSink {
+pub struct SubscriptionSink<Params> {
```
maciejhirsz, line 226: Maybe adding a default here would prevent us from having turbofish `register_subscription::<()>`: 

```suggestion
pub struct SubscriptionSink<Params = ()> {
```

Not 100% that compiler will figure it out, but might be worth a try.

utils/src/server/rpc_module.rs
```diff
@@ -252,6 +260,51 @@ impl SubscriptionSink {
...
+
 		Ok(())
 	}
+
+	/// Send a message to all subscriptions that could parse `P` as input.
+	///
+	/// F: is a closure that you need to provide to apply on the input P.
+	pub fn send_each<T, F>(&self, f: F) -> anyhow::Result<()>
+	where
+		F: Fn(&mut Params) -> T,
```
maciejhirsz, line 273: `Params` being `&mut` look like a foot gun to me.

I also think the closure should support error handling in user land:

`F: Fn(&Params) -> anyhow::Result<Option<T>>`, where:

+ `Ok(Some(T))` - no errors, send `T` to the subscriber.
+ `Ok(None)` - no errors, don't send anything to the subscriber.
+ `Err(_)` - error, log it (not sure if we need to send anything back?).

utils/src/server/rpc_module.rs
```diff
@@ -252,6 +260,51 @@ impl SubscriptionSink {
...
+		T: Serialize,
+	{
+		let mut subs = self.subscribers.lock();
+		let mut errored = Vec::new();
+
+		for ((conn_id, sub_id), sender) in subs.iter_mut() {
+			let result = match sender.params.as_mut().map(|p| to_raw_value(&f(p))) {
```
maciejhirsz, line 280:
```suggestion
			let result = match sender.params.as_ref().map(f) {
```
Closure return type is known, and it's ok if it's transient since it's going to be serialized in the same block, so there is no reason to use `RawValue` here.

utils/src/server/rpc_module.rs
```diff
@@ -252,6 +260,51 @@ impl SubscriptionSink {
...
+				Some(Ok(res)) => res,
+				_ => continue,
```
maciejhirsz, line 282: I'm still unsure if / why we need to do this filtering here. The subscription method should either require params, or not. If the user wants to do optional params, they can do `register_subscription::<Option<T>>`.

maciejhirsz (commented):
Code looks good to me, but I still have some opinionated suggestions to the API, feel free to disagree with anything :).

utils/src/server/rpc_module.rs
dvdplm: Seems like it doesn't. I still like the suggestion of providing a default though, maybe the compiler will get smart enough one day.

utils/src/server/rpc_module.rs
niklasad1: ofc, I found an edge-case where this doesn't work,

old code:

```rust
loop {
	let handle = match self.xt_sink.next() {
		Some(h) => h,
		None => continue,
	};
	
	let xt = handle.params();
	let best_block_hash = self.client.info().best_hash;
	let dxt = TransactionFor::<P>::decode(&mut &xt[..]).unwrap();
	let stream = match self.pool.submit_and_watch(&generic::BlockId::hash(best_block_hash), TX_SOURCE, dxt).await {
		Ok(stream) => stream,
		// just drop the subscription.
		Err(_) => continue,
	};

	// TODO: ugly blocks the executor from accepting other subs.
	stream.for_each(|item| {
		let _ = handle.send_all(&item);
		futures::future::ready(())
	}).await;
}
```

Basically, I found no way of porting this design because firstly, the param(s) is fetched to create a stream and that stream is polled and when a new element is received that is send by the sink. Thus, I don't know how to do it without having an API for fetch params and send it out only on that subscriber.

I suppose it could be possible to wrap the stream in the closure to seems quite messy to poll in and so on with having some other API to do it.

utils/src/server/rpc_module.rs
niklasad1: I think it will if we move the `register_subscription` to a trait with `P: Send + Sync + DeserializeOwned = ()`

utils/src/server/rpc_module.rs
niklasad1: The reason was that [this](https://github.com/paritytech/jsonrpsee/blob/master/types/src/v2/params.rs#L17) requires RawValue

utils/src/server/rpc_module.rs
dvdplm: It would have to be `F: Fn(&Option<Params>) -> anyhow::Result<Option<T>>` I think?

This __seems__ to work:

```rust
	pub fn send_each<T, F>(&self, f: F) -> anyhow::Result<()>
	where
		F: Fn(&Option<Params>) -> anyhow::Result<Option<T>>,
		T: Serialize,
	{
		let mut subs = self.subscribers.lock();
		let mut errored = Vec::new();

		for ((conn_id, sub_id), sender) in subs.iter_mut() {
			match f(&sender.params) {
				Ok(Some(res)) => {
					let msg = serde_json::to_string(&JsonRpcSubscriptionResponse {
						jsonrpc: TwoPointZero,
						method: self.method,
						params: JsonRpcNotificationParams { subscription: *sub_id, result: &to_raw_value(&res).expect("TODO what to do here?") },
					})?;

					// Mark broken connections, to be removed.
					if sender.sink.unbounded_send(msg).is_err() {
						errored.push((*conn_id, *sub_id));
					}
				},
				Ok(None) => log::warn!("Calling `send_each` for subscriptions that don't use params is… bad?"),
				Err(e) => log::error!("err: {:?}", e)
			}
		}

		// Remove broken connections
		for entry in errored {
			subs.remove(&entry);
		}

		Ok(())
	}
```

utils/src/server/rpc_module.rs
maciejhirsz: You can just make the struct generic, no? :)

utils/src/server/rpc_module.rs
dvdplm: hmm, do you have a branch?

utils/src/server/rpc_module.rs
maciejhirsz: Params being `Option<Params>` instead of just `Params` is a separate issue, see [the other comment](https://github.com/paritytech/jsonrpsee/pull/318#discussion_r635927172).

utils/src/server/rpc_module.rs
niklasad1: https://github.com/paritytech/substrate/tree/na-jsonrpsee-params-subscribe haven't pushed the stuff on it yet

utils/src/server/rpc_module.rs
niklasad1: yeah, that's possible but I'd prefer to do it another PR :)

utils/src/server/rpc_module.rs
niklasad1: yeah, that's neat.

types/src/v2/params.rs
```diff
@@ -83,10 +83,8 @@ impl<'a> RpcParams<'a> {
 	where
 		T: Deserialize<'a>,
 	{
-		match self.0 {
-			None => Err(CallError::InvalidParams),
-			Some(params) => serde_json::from_str(params).map_err(|_| CallError::InvalidParams),
-		}
+		let params = self.0.unwrap_or("null");
```
niklasad1, line 86: cover the case if the P is Option or () but maybe it's weird because technically not according to the spec.

utils/src/server/rpc_module.rs
```diff
@@ -27,9 +27,11 @@ pub type SubscriptionId = u64;
 /// Sink that is used to send back the result to the server for a specific method.
 pub type MethodSink = mpsc::UnboundedSender<String>;
 
-type Subscribers = Arc<Mutex<FxHashMap<(ConnectionId, SubscriptionId), MethodSink>>>;
+/// Map of subscribers keyed by the connection and subscription ids to an [`InnerSink`] that contains the parameters
+/// they used to subscribe and the tx side of a channel used to convey results and errors back.
+type Subscribers<P> = Arc<Mutex<FxHashMap<(ConnectionId, SubscriptionId), InnerSink<P>>>>;
```
TarikGul, line 32: 👍

examples/ws_sub_with_params.rs
```diff
@@ -0,0 +1,73 @@
...
+// DEALINGS IN THE SOFTWARE.
+
+use jsonrpsee::{
+	ws_client::{traits::SubscriptionClient, v2::params::JsonRpcParams, WsClientBuilder},
+	ws_server::WsServer,
+};
+use std::net::SocketAddr;
+
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
niklasad1: yeah indeed, it's a procedural macro that generates code that initializes the  `tokio runtime/threadpool` with along `fn async main` as you said.

TarikGul (approved):

niklasad1:
Superseeded by #336

Issue #255, opened 2021-03-30
Error: TransportError(NoAddressFound)


songtianyi:
jsonrpsee new client report error on vaild wss address
To reproduce the error
```rust
use jsonrpsee_types::{
	jsonrpc::{JsonValue, Params},
	traits::Client,
};
use jsonrpsee_ws_client::{WsClient, WsConfig};
const SERVER_URI: &str = "wss://pc2-rpc.darwinia.network";


#[tokio::main]
async fn main() -> Result<(), Box<dyn std::error::Error>> {

	let client = WsClient::new(WsConfig::with_url(SERVER_URI)).await?;
        println!("OK");
	Ok(())
}
```
```toml
[package]
name = "ws-rpc-test"
version = "0.1.0"
authors = ["songtianyi"]
edition = "2018"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
jsonrpsee-ws-client = "0.2.0-alpha.2"
jsonrpsee-types = "0.2.0-alpha.3"
[dependencies.tokio]
package = "tokio"
version = "0.2.22"
features = ["full"]
```

niklasad1:
Ok, I'm quite surprised by this I think https://github.com/paritytech/jsonrpsee/pull/209 fixed this.

Which version did you get this on?
AFAIK, all `0.2.0-alpha.x` releases should work...

songtianyi:
> Ok, I'm quite surprised by this I think #209 fixed this.
> 
> Which version did you get this on?
> AFAIK, all `0.2.0-alpha.x` releases should work...

jsonrpsee-ws-client = "0.2.0-alpha.2" still does

niklasad1:
I can't reproduce this, are you sure, can you double check `Cargo.lock` please?

However, don't matter that much when `alpha.3` works

usbalbin:
I got the same error when trying to connect to a server with a self signed cert. Could this be the same issue? Any way to get it to work with self signed certs?

*Running alpha.6*

```rust
let client = WsClientBuilder::default().build("wss://some-ip:1234").await.unwrap();
```

```
wss://some-ip:1234
Failed to connect to sockaddr: wss://some-ip:1234 with err: Io(Custom { kind: InvalidData, error: WebPKIError(CAUsedAsEndEntity) })
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: TransportError(NoAddressFound)', src/main.rs:26:91
```

[...]

PR #321, opened 2021-05-15
Add certificate_store() to WsClientBuilder

usbalbin:
Tries to solve my problems with self signed certs, see [part of issue 255](https://github.com/paritytech/jsonrpsee/issues/255#issuecomment-833015649).

I still can not get this to work with self signed certs(same error). However I happened to have a valid lets encrypt cert laying around that I tested. I still get the same error when not specifying `custom_certificate` however when I do, everything works :)

### DIFF
--- a/ws-client/src/client.rs
+++ b/ws-client/src/client.rs
@@ -39,7 +39,7 @@ use crate::{
 	manager::RequestManager, BatchMessage, Error, FrontToBack, NotificationHandler, RegisterNotificationMessage,
 	RequestMessage, Subscription, SubscriptionMessage,
 };
-use async_std::{path::PathBuf, sync::Mutex};
+use async_std::sync::Mutex;
 use async_trait::async_trait;
 use futures::{
 	channel::{mpsc, oneshot},
@@ -168,7 +168,7 @@ impl RequestIdGuard {
 /// Configuration.
 #[derive(Clone, Debug)]
 pub struct WsClientBuilder<'a> {
-	custom_certificate: Option<PathBuf>,
+	use_system_certificates: bool,
 	max_request_body_size: u32,
 	request_timeout: Option<Duration>,
 	connection_timeout: Duration,
@@ -181,7 +181,7 @@ pub struct WsClientBuilder<'a> {
 impl<'a> Default for WsClientBuilder<'a> {
 	fn default() -> Self {
 		Self {
-			custom_certificate: None,
+			use_system_certificates: false,
 			max_request_body_size: TEN_MB_SIZE_BYTES,
 			request_timeout: None,
 			connection_timeout: Duration::from_secs(10),
@@ -194,9 +194,9 @@ impl<'a> Default for WsClientBuilder<'a> {
 }
 
 impl<'a> WsClientBuilder<'a> {
-	/// Specify custom cert file
-	pub fn custom_certificate(mut self, path: impl Into<PathBuf>) -> Self {
-		self.custom_certificate = Some(path.into());
+	/// Set wheather to use system certificates
+	pub fn use_system_certificates(mut self, use_system_certificates: bool) -> Self {
+		self.use_system_certificates = use_system_certificates;
 		self
 	}
 
@@ -257,7 +257,7 @@ impl<'a> WsClientBuilder<'a> {
 	///
 	/// `wss://host` - port 443 is used
 	pub async fn build(self, url: &'a str) -> Result<WsClient, Error> {
-		let custom_certificate = self.custom_certificate;
+		let use_system_certificates = self.use_system_certificates;
 		let max_capacity_per_subscription = self.max_notifs_per_subscription;
 		let max_concurrent_requests = self.max_concurrent_requests;
 		let request_timeout = self.request_timeout;
@@ -267,7 +267,7 @@ impl<'a> WsClientBuilder<'a> {
 		let (sockaddrs, host, mode) = parse_url(url).map_err(|e| Error::Transport(Box::new(e)))?;
 
 		let builder = WsTransportClientBuilder {
-			custom_certificate,
+			use_system_certificates,
 			sockaddrs,
 			mode,
 			host,
--- a/ws-client/src/client.rs
+++ b/ws-client/src/client.rs
@@ -181,7 +181,7 @@ pub struct WsClientBuilder<'a> {
 impl<'a> Default for WsClientBuilder<'a> {
 	fn default() -> Self {
 		Self {
-			use_system_certificates: false,
+			use_system_certificates: true,
 			max_request_body_size: TEN_MB_SIZE_BYTES,
 			request_timeout: None,
 			connection_timeout: Duration::from_secs(10),
--- a/ws-client/src/client.rs
+++ b/ws-client/src/client.rs
@@ -24,10 +24,6 @@
 // IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 // DEALINGS IN THE SOFTWARE.
 
-use crate::{helpers::{
-	build_unsubscribe_message, process_batch_response, process_error_response, process_notification,
-	process_single_response, process_subscription_response, stop_subscription,
-}, transport::CertificateStore};
 use crate::traits::{Client, SubscriptionClient};
 use crate::transport::{parse_url, Receiver as WsReceiver, Sender as WsSender, WsTransportClientBuilder};
 use crate::v2::error::JsonRpcErrorAlloc;
@@ -35,6 +31,13 @@ use crate::v2::params::{Id, JsonRpcParams};
 use crate::v2::request::{JsonRpcCallSer, JsonRpcNotificationSer};
 use crate::v2::response::{JsonRpcNotifResponse, JsonRpcResponse, JsonRpcSubscriptionResponse};
 use crate::TEN_MB_SIZE_BYTES;
+use crate::{
+	helpers::{
+		build_unsubscribe_message, process_batch_response, process_error_response, process_notification,
+		process_single_response, process_subscription_response, stop_subscription,
+	},
+	transport::CertificateStore,
+};
 use crate::{
 	manager::RequestManager, BatchMessage, Error, FrontToBack, NotificationHandler, RegisterNotificationMessage,
 	RequestMessage, Subscription, SubscriptionMessage,
--- a/ws-client/src/client.rs
+++ b/ws-client/src/client.rs
@@ -24,10 +24,10 @@
 // IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 // DEALINGS IN THE SOFTWARE.
 
-use crate::helpers::{
+use crate::{helpers::{
 	build_unsubscribe_message, process_batch_response, process_error_response, process_notification,
 	process_single_response, process_subscription_response, stop_subscription,
-};
+}, transport::CertificateStore};
 use crate::traits::{Client, SubscriptionClient};
 use crate::transport::{parse_url, Receiver as WsReceiver, Sender as WsSender, WsTransportClientBuilder};
 use crate::v2::error::JsonRpcErrorAlloc;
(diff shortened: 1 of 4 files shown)


ws-client/src/transport.rs
```diff
@@ -227,7 +233,24 @@ impl<'a> WsTransportClientBuilder<'a> {
 					match self.mode {
 						Mode::Plain => TlsOrPlain::Plain(socket),
 						Mode::Tls => {
-							let connector = async_tls::TlsConnector::default();
+							let mut client_config = rustls::ClientConfig::default();
+							if let Some(path) = &self.custom_certificate {
+								// TODO: How to give use async for file io and still give add_pem_file() a std::io::BufRead?
+								let f = std::fs::File::open(&path)?;
+								let mut reader = io::BufReader::new(f);
+								let (added, failed) = client_config
```
usbalbin, line 241: Do we want to expose or handle `added` and `failed` in any way differently?

For example, perhaps the contents of the `if let` should be in `custom_certificate()` which would then the result of `add_pem_file()` (plus any errors)

ws-client/src/transport.rs
```diff
@@ -227,7 +233,24 @@ impl<'a> WsTransportClientBuilder<'a> {
 					match self.mode {
 						Mode::Plain => TlsOrPlain::Plain(socket),
 						Mode::Tls => {
-							let connector = async_tls::TlsConnector::default();
+							let mut client_config = rustls::ClientConfig::default();
+							if let Some(path) = &self.custom_certificate {
+								// TODO: How to give use async for file io and still give add_pem_file() a std::io::BufRead?
+								let f = std::fs::File::open(&path)?;
```
usbalbin, line 239: Is there any way to solve this async:ier?

[`add_pem_file`](https://docs.rs/rustls/0.19.1/rustls/struct.RootCertStore.html#method.add_pem_file) wants an `std::io::BufRead`.

niklasad1: async_std::fs::read(path).await?;

ws-client/src/transport.rs
niklasad1: No, I think it's fine to log it and return an error if the certificate could not be read from the Path.

ws-client/src/transport.rs
```diff
@@ -227,7 +233,24 @@ impl<'a> WsTransportClientBuilder<'a> {
...
+								let mut reader = io::BufReader::new(f);
+								let (added, failed) = client_config
+									.root_store
+									.add_pem_file(&mut reader)
```
niklasad1, line 243: I realized that another solution is to use `rustls_native_certificates` instead such as [hyper-rustls does](https://github.com/ctz/hyper-rustls/blob/master/src/connector.rs#L34-#L42)

niklasad1:
> I still can not get this to work with self signed certs(same error).

Ok, turns out that that [webpki doesn't support self-signed certificates](https://github.com/briansmith/webpki/blob/d4cff28a541343c17c464145d3406949da6d9b77/src/cert.rs#L99-L116)

We might want to use `rustls-native-certs` instead




hmm, it might be better to go the another route and to use `rustls_native_certs::load_native_certs()` instead.

ws-client/src/transport.rs
```diff
@@ -227,7 +229,12 @@ impl<'a> WsTransportClientBuilder<'a> {
 					match self.mode {
 						Mode::Plain => TlsOrPlain::Plain(socket),
 						Mode::Tls => {
-							let connector = async_tls::TlsConnector::default();
+							let mut client_config = rustls::ClientConfig::default();
+							if self.use_system_certificates {
+								client_config.root_store = rustls_native_certs::load_native_certs().map_err(|(_, e)| e)?;
```
usbalbin, line 234: `load_native_certs` is blocking, do we want to `spawn_blocking` or something similar?

niklasad1, line 234: it's fine but you could move this to `fn build` to avoid doing it several times.

ws-client/src/client.rs
```diff
@@ -179,6 +181,7 @@ pub struct WsClientBuilder<'a> {
 impl<'a> Default for WsClientBuilder<'a> {
 	fn default() -> Self {
 		Self {
+			use_system_certificates: false,
```
niklasad1, line 184: I won't mind changing `WsClient` to use the `native certificate store` by default, we use it in the `HttpClient` for example

ws-client/src/transport.rs
usbalbin: That would cause the certs to be loaded even when Plain is used. Would something like this 6261d94ed be better?

ws-client/src/client.rs
usbalbin: Ok :)

Would you also prefer i called `use_system_certificates` something like `use_native_certificate_store`?

ws-client/src/transport.rs
usbalbin: Oh nevermind

ws-client/src/transport.rs
```diff
@@ -201,8 +204,9 @@ impl<'a> WsTransportClientBuilder<'a> {
 
 	/// Try establish the connection.
 	pub async fn build(self) -> Result<(Sender, Receiver), WsHandshakeError> {
+		let mut client_config = None;
```
niklasad1, line 207: what I meant is that you could do:

```rust
let client_config = match self.client_config {
   Some(client_config) => client_config,
   None => {
      let mut config = rustls::ClientConfig::default();
	  if self.use_system_certificates {
	    config.root_store = rustls_native_certs::load_native_certs().map_err(|(_, e)| e)?;
	}
	client_config.get_or_insert(config)
	}
};
let connector: async_tls::TlsConnector = client_config.into();
```

here instead and just pass by the connector by reference to `try_connect`?!

usbalbin: Yes, i did not realize we have access to self.mode in `fn build`. Sorry :)

ws-client/src/client.rs
niklasad1: Personally, I would prefer to have an enum configure the certificate store, boolean for that seems it little bit hard to understand.

such as:

```
enum CertificateStore {
  Native,
  WebPki,
}

/// Set certificate store to use
///
/// Default: native certificate store
pub fn certificate_store(mut self, cert: CertificateStore) -> Self {
		self.certificate_store = cert;
		self
	}
```

But you could just remove the `use_system_certificates option and hardcode it to native then fine for me, `webpiki` seems more secure but I think it's too complicated to use for most users anyway.

usbalbin: Do you think exposing [RootCertStore](https://docs.rs/rustls/0.19.1/rustls/struct.RootCertStore.html) would be to take it too far?

Then the user could do what ever they wanted. We could still have `rustls_native_certs::load_native_certs()` as default.

ws-client/src/manager.rs
```diff
@@ -160,10 +160,10 @@ impl RequestManager {
 
 	/// Removes a notification handler
 	pub fn remove_notification_handler(&mut self, method: String) -> Result<(), Error> {
-		if let Some(_) = self.notification_handlers.remove(&method) {
+		if self.notification_handlers.remove(&method).is_some() {
 			Ok(())
 		} else {
-			Err(Error::UnregisteredNotification(method.to_owned()))
```
niklasad1, line 166: :+1: thanks

ws-client/src/transport.rs
```diff
@@ -201,8 +216,20 @@ impl<'a> WsTransportClientBuilder<'a> {
 
 	/// Try establish the connection.
 	pub async fn build(self) -> Result<(Sender, Receiver), WsHandshakeError> {
+		let client_config = match self.mode {
```
niklasad1, line 219:
```suggestion
		let connector = match self.mode {
```

ws-client/src/transport.rs
```diff
@@ -201,8 +216,20 @@ impl<'a> WsTransportClientBuilder<'a> {
...
+						rustls_native_certs::load_native_certs().map_err(|(_, e)| WsHandshakeError::NativeCert(e))?;
+				}
+				Some(client_config.into())
+			}
+			Mode::Plain => None,
+		};
+
 		for sockaddr in &self.sockaddrs {
-			match self.try_connect(*sockaddr).await {
+			match self.try_connect(*sockaddr, &client_config).await {
```
niklasad1, line 232:
```suggestion
			match self.try_connect(*sockaddr, &connector).await {
```

niklasad1 (approved):
Good work, LGTM

ws-client/src/client.rs
niklasad1: let's ignore for it now, I think most users wants us to control that but perhaps add an issue for it.

usbalbin:
> > I still can not get this to work with self signed certs(same error).
> 
> Ok, turns out that that [webpki doesn't support self-signed certificates](https://github.com/briansmith/webpki/blob/d4cff28a541343c17c464145d3406949da6d9b77/src/cert.rs#L99-L116)
> 
> We might want to use `rustls-native-certs` instead.

Maybe I am doing something wrong but I still can not get a self signed cert to work.

(Running ubuntu 20.04)

I have run
```bash
openssl req -x509 -newkey rsa:4096 -keyout server.key -out server.crt -days 365 -sha256 -nodes --subj '/CN=localhost/'
# Move both files to correct place for server program...

cp server.crt /server.crt /usr/share/ca-certificates/xo.crt
dpkg-reconfigure ca-certificates
# Select `Ask` and tick the checkbox for my cert
reboot now
# start server program
# start client program, try to connect to server
# Boom:
[2021-05-16T16:23:41Z DEBUG rustls::anchors] add_pem_file processed 130 valid and 0 invalid certs
[2021-05-16T16:23:41Z DEBUG rustls::client::hs] No cached session for DNSNameRef("localhost")
[2021-05-16T16:23:41Z DEBUG rustls::client::hs] Not resuming any session
[2021-05-16T16:23:41Z DEBUG rustls::client::hs] Using ciphersuite TLS13_AES_256_GCM_SHA384
[2021-05-16T16:23:41Z DEBUG rustls::client::tls13] Not resuming
[2021-05-16T16:23:41Z DEBUG rustls::client::tls13] TLS1.3 encrypted extensions: []
[2021-05-16T16:23:41Z DEBUG rustls::client::hs] ALPN protocol is None
[2021-05-16T16:23:41Z WARN  rustls::session] Sending fatal alert BadCertificate
[2021-05-16T16:23:41Z DEBUG jsonrpsee_ws_client::transport] Failed to connect to sockaddr: [::1]:8080 with err: Io(Custom { kind: InvalidData, error: WebPKIError(CAUsedAsEndEntity) })
[2021-05-16T16:23:41Z DEBUG rustls::client::hs] No cached session for DNSNameRef("localhost")
[2021-05-16T16:23:41Z DEBUG rustls::client::hs] Not resuming any session
[2021-05-16T16:23:41Z DEBUG rustls::client::hs] Using ciphersuite TLS13_AES_256_GCM_SHA384
[2021-05-16T16:23:41Z DEBUG rustls::client::tls13] Not resuming
[2021-05-16T16:23:41Z DEBUG rustls::client::tls13] TLS1.3 encrypted extensions: []
[2021-05-16T16:23:41Z DEBUG rustls::client::hs] ALPN protocol is None
[2021-05-16T16:23:41Z WARN  rustls::session] Sending fatal alert BadCertificate
[2021-05-16T16:23:41Z DEBUG jsonrpsee_ws_client::transport] Failed to connect to sockaddr: 127.0.0.1:8080 with err: Io(Custom { kind: InvalidData, error: WebPKIError(CAUsedAsEndEntity) })
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Transport(NoAddressFound)', /home/albin/Rock-Algot/tools/src/xo.rs:94:18
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Do you think there is something wrong on my end or anything more to be fixed in the code? :)

niklasad1:
> Do you think there is something wrong on my end or anything more to be fixed in the code? :)

hmm, I'm no expert in adding certificates to `certificate store` but maybe worth a try testing with [mkcert](https://github.com/FiloSottile/mkcert).

Thus:

`$ mkcert -install`
`$ mkcert -key-file key.pem -cert-file cert.pem localhost`

"Works" or I'm not getting bad certificates at least ^^, we should probably set it in the CI but so complicated so we have tested against some remote `wss` and `https` targets.

usbalbin:
Yes that seems to work fine. Thanks a lot :)

dvdplm (approved):
tyvm!

usbalbin:
Thank you :)

PR #325, opened 2021-05-18
[types]: ID type instead of serde_json::RawValue

niklasad1:
Similar should be done on `Params` and `JsonRpcNotifParams` to work properly but this PR is large enough as it is.

Further info see https://github.com/paritytech/jsonrpsee/pull/292#issuecomment-829874041

### DIFF
--- a/http-client/src/client.rs
+++ b/http-client/src/client.rs
@@ -2,11 +2,11 @@ use crate::traits::Client;
 use crate::transport::HttpTransportClient;
 use crate::v2::request::{JsonRpcCallSer, JsonRpcNotificationSer};
 use crate::v2::{
-	error::JsonRpcErrorAlloc,
+	error::JsonRpcError,
 	params::{Id, JsonRpcParams},
 	response::JsonRpcResponse,
 };
-use crate::{Error, JsonRawValue, TEN_MB_SIZE_BYTES};
+use crate::{Error, TEN_MB_SIZE_BYTES};
 use async_trait::async_trait;
 use fnv::FnvHashMap;
 use serde::de::DeserializeOwned;
@@ -76,12 +76,12 @@ impl Client for HttpClient {
 		let response: JsonRpcResponse<_> = match serde_json::from_slice(&body) {
 			Ok(response) => response,
 			Err(_) => {
-				let err: JsonRpcErrorAlloc = serde_json::from_slice(&body).map_err(Error::ParseError)?;
-				return Err(Error::Request(err));
+				let err: JsonRpcError = serde_json::from_slice(&body).map_err(Error::ParseError)?;
+				return Err(Error::Request(err.to_string()));
 			}
 		};
 
-		let response_id = parse_request_id(response.id)?;
+		let response_id = response.id.as_number().copied().ok_or(Error::InvalidRequestId)?;
 
 		if response_id == id {
 			Ok(response.result)
@@ -115,15 +115,15 @@ impl Client for HttpClient {
 		let rps: Vec<JsonRpcResponse<_>> = match serde_json::from_slice(&body) {
 			Ok(response) => response,
 			Err(_) => {
-				let err: JsonRpcErrorAlloc = serde_json::from_slice(&body).map_err(Error::ParseError)?;
-				return Err(Error::Request(err));
+				let err: JsonRpcError = serde_json::from_slice(&body).map_err(Error::ParseError)?;
+				return Err(Error::Request(err.to_string()));
 			}
 		};
 
 		// NOTE: `R::default` is placeholder and will be replaced in loop below.
 		let mut responses = vec![R::default(); ordered_requests.len()];
 		for rp in rps {
-			let response_id = parse_request_id(rp.id)?;
+			let response_id = rp.id.as_number().copied().ok_or(Error::InvalidRequestId)?;
 			let pos = match request_set.get(&response_id) {
 				Some(pos) => *pos,
 				None => return Err(Error::InvalidRequestId),
@@ -131,15 +131,5 @@ impl Client for HttpClient {
 			responses[pos] = rp.result
 		}
 		Ok(responses)
-	}
-}
-
-fn parse_request_id(raw: Option<&JsonRawValue>) -> Result<u64, Error> {
-	match raw {
-		None => Err(Error::InvalidRequestId),
-		Some(id) => {
-			let id = serde_json::from_str(id.get()).map_err(Error::ParseError)?;
-			Ok(id)
-		}
 	}
 }
--- a/http-server/src/server.rs
+++ b/http-server/src/server.rs
@@ -36,7 +36,7 @@ use hyper::{
 use jsonrpsee_types::error::{CallError, Error, GenericTransportError};
 use jsonrpsee_types::v2::error::JsonRpcErrorCode;
 use jsonrpsee_types::v2::params::{Id, RpcParams};
-use jsonrpsee_types::v2::request::{JsonRpcInvalidRequest, JsonRpcRequest, JsonRpcNotification};
+use jsonrpsee_types::v2::request::{JsonRpcInvalidRequest, JsonRpcNotification, JsonRpcRequest};
 use jsonrpsee_utils::hyper_helpers::read_response_to_body;
 use jsonrpsee_utils::server::helpers::{collect_batch_response, send_error};
 use jsonrpsee_utils::server::rpc_module::{MethodSink, RpcModule};
--- a/http-server/src/server.rs
+++ b/http-server/src/server.rs
@@ -34,8 +34,9 @@ use hyper::{
 	Error as HyperError,
 };
 use jsonrpsee_types::error::{CallError, Error, GenericTransportError};
+use jsonrpsee_types::v2::error::JsonRpcErrorCode;
+use jsonrpsee_types::v2::params::{Id, RpcParams};
 use jsonrpsee_types::v2::request::{JsonRpcInvalidRequest, JsonRpcRequest};
-use jsonrpsee_types::v2::{error::JsonRpcErrorCode, params::RpcParams};
 use jsonrpsee_utils::hyper_helpers::read_response_to_body;
 use jsonrpsee_utils::server::helpers::{collect_batch_response, send_error};
 use jsonrpsee_utils::server::rpc_module::{MethodSink, RpcModule};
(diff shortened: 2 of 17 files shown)


types/src/v2/params.rs
```diff
@@ -1,9 +1,11 @@
 use crate::error::InvalidParams;
 use alloc::collections::BTreeMap;
+use std::borrow::Cow;
```
niklasad1, line 3: couldn't use `beef::Cow` here some problem with serialize bounds.

types/src/v2/params.rs
```diff
@@ -190,8 +193,49 @@ impl Id {
 			_ => None,
 		}
 	}
+
+	/// Creates owned data from borrowed data, allocates only for Strings.
+	pub fn to_owned(&self) -> Id<'static> {
```
niklasad1, line 198: I think this can be removed.

types/src/v2/params.rs
dvdplm: Hmm. @maciejhirsz thoughts?

ws-server/src/server.rs
```diff
@@ -141,12 +141,12 @@ async fn background_task(
 	let execute = move |tx: &MethodSink, req: JsonRpcRequest| {
 		if let Some(method) = methods.get(&*req.method) {
 			let params = RpcParams::new(req.params.map(|params| params.get()));
-			if let Err(err) = (method)(req.id, params, &tx, conn_id) {
+			if let Err(err) = (method)(req.id.clone(), params, &tx, conn_id) {
```
dvdplm, line 144: So these clones are cheap then you say?

dvdplm (approved):
Good cleanup here, and good riddance of those `*Alloc` types. 👍

ws-server/src/server.rs
niklasad1: yes :pray:

http-client/src/client.rs
```diff
@@ -76,12 +76,12 @@ impl Client for HttpClient {
 		let response: JsonRpcResponse<_> = match serde_json::from_slice(&body) {
 			Ok(response) => response,
 			Err(_) => {
-				let err: JsonRpcErrorAlloc = serde_json::from_slice(&body).map_err(Error::ParseError)?;
-				return Err(Error::Request(err));
+				let err: JsonRpcError = serde_json::from_slice(&body).map_err(Error::ParseError)?;
+				return Err(Error::Request(err.to_string()));
```
niklasad1, line 80: Note: I changed the error type to a String in `Error::Request` to get rid of the alloc type.

http-server/src/server.rs
```diff
@@ -211,23 +212,27 @@ impl Server {
 						// Our [issue](https://github.com/paritytech/jsonrpsee/issues/296).
 						if let Ok(req) = serde_json::from_slice::<JsonRpcRequest>(&body) {
 							execute(&tx, req);
+						} else if let Ok(_req) = serde_json::from_slice::<JsonRpcNotification>(&body) {
+							return Ok::<_, HyperError>(response::ok_response("".into()));
```
niklasad1, line 216: send empty response to notification,

currently we don't care if the method is registered or not (that API is not exposed)

http-server/src/server.rs
```diff
@@ -211,23 +212,27 @@ impl Server {
...
 								single = false;
 								for req in batch {
 									execute(&tx, req);
 								}
 							} else {
-								send_error(None, &tx, JsonRpcErrorCode::InvalidRequest.into());
+								send_error(Id::Null, &tx, JsonRpcErrorCode::InvalidRequest.into());
 							}
+						} else if let Ok(_batch) = serde_json::from_slice::<Vec<JsonRpcNotification>>(&body) {
+							return Ok::<_, HyperError>(response::ok_response("".into()));
```
niklasad1, line 227: send empty response to batch notification,

currently we don't care if the method is registered or not (that API is not exposed)

types/src/v2/error.rs
```diff
@@ -2,57 +2,30 @@ use crate::v2::params::{Id, TwoPointZero};
...
-impl From<JsonRpcErrorCode> for JsonRpcErrorObjectAlloc {
-	fn from(code: JsonRpcErrorCode) -> Self {
-		Self { code, message: code.message().to_owned(), data: None }
+impl<'a> fmt::Display for JsonRpcError<'a> {
+	fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+		write!(f, "{}", serde_json::to_string(&self).expect("infallible; qed"))
 	}
 }
 
 /// JSON-RPC error object with no extra allocations.
```
niklasad1, line 27: fix docs
```suggestion
/// JSON-RPC error object.
```

types/src/v2/params.rs
```diff
@@ -4,6 +4,8 @@ use serde::de::{self, Deserializer, Unexpected, Visitor};
 use serde::ser::Serializer;
 use serde::{Deserialize, Serialize};
 use serde_json::{value::RawValue, Value as JsonValue};
+use std::borrow::Cow;
+
```
niklasad1, line 8:
```suggestion
```

types/src/v2/response.rs
```diff
@@ -1,30 +1,45 @@
...
+
+/// JSON-RPC subscription response.
+#[derive(Serialize)]
+pub struct JsonRpcSubscriptionResponse<'a> {
+	/// JSON-RPC version.
+	pub jsonrpc: TwoPointZero,
+	/// Method
+	pub method: &'a str,
+	/// Params.
+	pub params: JsonRpcNotificationParams<'a>,
```
niklasad1, line 25: In another PR; merge `JsonRpcNotificationParams` and `JsonRpcNotificationParamsNotification`

types/src/v2/request.rs
```diff
@@ -21,22 +21,24 @@ pub struct JsonRpcRequest<'a> {
...
+#[serde(deny_unknown_fields)]
 pub struct JsonRpcNotification<'a> {
 	/// JSON-RPC version.
 	pub jsonrpc: TwoPointZero,
 	/// Name of the method to be invoked.
 	pub method: &'a str,
 	/// Parameter values of the request.
-	pub params: JsonRpcNotificationParams<'a>,
+	#[serde(borrow)]
+	pub params: Option<&'a RawValue>,
```
niklasad1, line 41: In another PR, hopefully merge JsonRpcParams/RpcParams and replace RawValue here,

That's the reason why some de-serialize tests still fail

http-client/src/client.rs
dvdplm: Yeah I noticed that. I don't think there much else we can do right?

types/src/v2/error.rs
```diff
@@ -180,47 +154,15 @@ impl serde::Serialize for JsonRpcErrorCode {
 
 #[cfg(test)]
 mod tests {
-	use super::{
-		Id, JsonRpcError, JsonRpcErrorAlloc, JsonRpcErrorCode, JsonRpcErrorObject, JsonRpcErrorObjectAlloc,
-		TwoPointZero,
-	};
-
-	#[test]
-	fn deserialize_works() {
```
niklasad1, line 189: TODO: add this test back.

types/src/v2/response.rs
```diff
@@ -1,30 +1,45 @@
...
+	/// Method
+	pub method: &'a str,
 	/// Params.
 	pub params: JsonRpcNotificationParamsAlloc<T>,
 }
 
 /// JSON-RPC notification response.
-#[derive(Deserialize, Debug)]
+#[derive(Deserialize, Serialize, Debug)]
+#[serde(deny_unknown_fields)]
```
niklasad1, line 42: BTW, we have almost two identical types for `notifications` and `subscriptions`, currently the subscription type requires params and the notification does not.

Thus, I think it's that a notification with params is not propagated to the client even if it's registered.

Let's merge them in another PR and check them manually in RequestManager instead.

types/src/v2/params.rs
niklasad1:
```
error[E0277]: the trait bound `beef::generic::Cow<'_, str, beef::wide::internal::Wide>: Serialize` is not satisfied
   --> types/src/v2/params.rs:168:6
    |
168 |     Str(Cow<'a, str>),
    |         ^^^^^^^^^^^^ the trait `Serialize` is not implemented for `beef::generic::Cow<'_, str, beef::wide::internal::Wide>`
    |
   ::: ./.cargo/registry/src/github.com-1ecc6299db9ec823/serde-1.0.126/src/ser/mod.rs:247:18
    |
247 |     fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
    |                  - required by this bound in `serialize`
```

http-server/src/tests.rs
```diff
@@ -169,14 +169,11 @@ async fn batched_notifications() {
...
-	let req = r#"[
-        {"jsonrpc": "2.0", "method": "notif", "params": [1,2,4]},
-        {"jsonrpc": "2.0", "method": "notif", "params": [7]}
-	]"#;
+	let req = r#"[{"jsonrpc": "2.0", "method": "notif", "params": [1,2,4]},{"jsonrpc": "2.0", "method": "notif", "params": [7]}]"#;
 	let response = http_request(req.into(), uri).await.unwrap();
 	assert_eq!(response.status, StatusCode::OK);
-	// Note: this is *not* according to spec. Response should be the empty string, `""`.
-	assert_eq!(response.body, r#"[{"jsonrpc":"2.0","result":"","id":null},{"jsonrpc":"2.0","result":"","id":null}]"#);
+	// Note: on HTTP we ack the notification with an empty response.
```
TarikGul, line 175: What does `ack` mean here?

niklasad1: ACK - acknowledge, maybe it's a bad term ^^

TarikGul: Ahh makes sense, yea I wouldn't have been able to deduce that. But now that I know what it means I like it. For the general user, it would probably make sense to write out the whole word though in this case.

niklasad1, line 175:
```suggestion
	// Note: on HTTP acknowledge the notification with an empty response.
```

TarikGul (approved):
LGTM, After a brief explanation from @niklasad1 I was able to follow along the type changes, and understand the direction of the PR. 👍

Issue #291, opened 2021-04-26
[rpc context module]: register async methods


niklasad1:
It would be nice if we could make it possible for the servers to allow to register `async methods`.

Something like:

```rust
   rpc_module.register_method_async("chain_getHeader", |params, chain| async {
     let hash = params.one()?;     
     chain.header(Some(hash)).await
   })?;
```

niklasad1:
I did some naive attempts on this but couldn't get it work because `async closures` is thing in stable Rust AFAIU

//cc @maciejhirsz

niklasad1:
Closed by #327

PR #327, opened 2021-05-20
Basic support for async methods

popzxc:
Related to #291

This PR introduces support for `register_async_method` and it, *uhm*, even seems to be working.

I'm not quite fond of introduced `Owned*` types, but not sure that my `Pin` kung fu is strong enough to deal with all these lifetimes.

### DIFF
--- a/types/src/v2/request.rs
+++ b/types/src/v2/request.rs
@@ -43,7 +43,12 @@ impl OwnedJsonRpcRequest {
 			jsonrpc: self.jsonrpc,
 			id: self.id.borrowed(),
 			method: Cow::borrowed(self.method.as_ref()),
-			params: self.params.as_ref().map(|s| serde_json::from_str(&s).unwrap()),
+			params: self.params.as_ref().map(|s| {
+				// Note: while this object *may* be created not from the `JsonRpcRequest` object,
+				// using an invalid field to construct it would be a logical invariant break.
+				serde_json::from_str(&s)
+					.expect("OwnedJsonRpcRequest is only created from JsonRpcRequest, so this conversion must be safe")
+			}),
 		}
 	}
 }
--- a/types/src/v2/request.rs
+++ b/types/src/v2/request.rs
@@ -3,6 +3,8 @@ use beef::Cow;
 use serde::{Deserialize, Serialize};
 use serde_json::value::RawValue;
 
+use super::params::OwnedId;
+
 /// [JSON-RPC request object](https://www.jsonrpc.org/specification#request-object)
 #[derive(Deserialize, Debug)]
 #[serde(deny_unknown_fields)]
@@ -18,6 +20,43 @@ pub struct JsonRpcRequest<'a> {
 	/// Parameter values of the request.
 	#[serde(borrow)]
 	pub params: Option<&'a RawValue>,
+}
+
+/// Owned version of [`JsonRpcRequest`].
+#[derive(Deserialize, Debug, Clone)]
+#[serde(deny_unknown_fields)]
+pub struct OwnedJsonRpcRequest {
+	/// JSON-RPC version.
+	pub jsonrpc: TwoPointZero,
+	/// Request ID
+	pub id: OwnedId,
+	/// Name of the method to be invoked.
+	pub method: String,
+	/// Parameter values of the request.
+	pub params: Option<String>,
+}
+
+impl OwnedJsonRpcRequest {
+	/// Converts `OwnedJsonRpcRequest` into borrowed `JsonRpcRequest`.
+	pub fn borrowed(&self) -> JsonRpcRequest<'_> {
+		JsonRpcRequest {
+			jsonrpc: self.jsonrpc,
+			id: self.id.borrowed(),
+			method: Cow::borrowed(self.method.as_ref()),
+			params: self.params.as_ref().map(|s| serde_json::from_str(&s).unwrap()),
+		}
+	}
+}
+
+impl<'a> From<JsonRpcRequest<'a>> for OwnedJsonRpcRequest {
+	fn from(borrowed: JsonRpcRequest<'a>) -> Self {
+		Self {
+			jsonrpc: borrowed.jsonrpc,
+			id: borrowed.id.into(),
+			method: borrowed.method.as_ref().to_owned(),
+			params: borrowed.params.map(|p| p.get().to_owned()),
+		}
+	}
 }
 
 /// Invalid request with known request ID.
--- a/utils/src/server/rpc_module.rs
+++ b/utils/src/server/rpc_module.rs
@@ -1,9 +1,10 @@
 use crate::server::helpers::{send_error, send_response};
 use futures_channel::{mpsc, oneshot};
+use futures_util::{future::BoxFuture, FutureExt};
 use jsonrpsee_types::error::{CallError, Error};
-use jsonrpsee_types::traits::RpcMethod;
+use jsonrpsee_types::traits::{AsyncRpcMethod, RpcMethod};
 use jsonrpsee_types::v2::error::{JsonRpcErrorCode, JsonRpcErrorObject, CALL_EXECUTION_FAILED_CODE};
-use jsonrpsee_types::v2::params::{Id, JsonRpcNotificationParams, RpcParams, TwoPointZero};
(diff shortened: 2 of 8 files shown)


niklasad1 (commented):
Cool,

We have been thinking a little bit about this and thought about similar to what you have done but instead `spawn` the future in leaf `async fn register_method`

Example:

```rust
	/// Register a new async RPC method, which responds with a given callback.
	pub fn register_method_async<F, R>(&mut self, method_name: &'static str, callback: F) -> Result<(), Error>
	where
		R: Serialize + 'static + Send + Sync,
                // TODO: should have RpcParams or RpcParamsOwned
		F: Fn(()) -> BoxFuture<'static, Result<R, CallError>> + Send + Sync + 'static + Copy
	{
		self.verify_method_name(method_name)?;

		self.methods.insert(
			method_name,
			Box::new(move |id, params, tx, _| {
				let tx_clone = tx.clone();
				let fut = async move {
                                        // TODO: add real code here
					let res = callback(()).boxed().await.unwrap();
                                        send_response(Id::Null, &tx_clone, res);
				};

				let _ = tokio::spawn(fut);
				Ok(())
			}),
		);
		Ok(())
	}
```

popzxc:
Regarding the lifetime issue: setting the lifetime in `BoxFuture` won't help, as it's not the data lifetime, but rather the type lifetime. Additionally, the thing I said about `Pin` will only be valid in case we process all the requests sequentially. Otherwise, we will *have* to use owned types, because the future may be scheduled in another thread *after* the initial request object got destroyed in the handler loop.

Regarding using the `tokio::spawn` inside of the future: I initially thought about it too but decided that it's not the right way to do things.
The reasons for that are:

1. Currently, `types` and `utils` are free from the `tokio` dependencies. This means that one may implement an `async-std` (or `actix-rt`, or whatever) based server using the `RpcModule` module abstraction. It's a nice level of abstraction and I think it's a good thing to keep it in that way.
2. Using `tokio::spawn` *inside* of the future gives less flexibility for the server implementation. To elaborate: in the current implementation, sync and async methods are handled sequentially. It's not the most efficient way, but it has the right to be implemented. Alternative options are:
    - Handle sync methods sequentially, and handle async methods via `tokio::spawn`.
    - Handle both sync and async methods via `tokio::spawn`. This approach is actually interesting, because sync methods may be distributed among several runtime threads and will be executed in parallel despite being sync.

Not bounding on `tokio` on the `utils` crate level is a better choice IMHO. Moreover, the cases listed above may be implemented atop of this PR relatively easily, but not vise versa.

However, I'm not going to insist. If you are adamant that it's the way you see the implementation, I'll be happy to adapt the codebase accordingly.

niklasad1:
Ok, thanks for the explanation that makes sense to me.

Overall the implementation looks clean, nice.

utils/src/server/rpc_module.rs
```diff
@@ -83,6 +116,45 @@ impl RpcModule {
...
+	where
+		R: Serialize + Send + Sync + 'static,
+		F: AsyncRpcMethod<R, CallError> + Copy + Send + Sync + 'static,
+	{
+		self.verify_method_name(method_name)?;
+
+		self.async_methods.insert(
+			method_name,
+			Box::new(move |id, params, tx, _| {
+				let tx = tx.clone();
```
niklasad1, line 135: remove clone?!

or is it possible to borrow it in `AsyncMethods` instead?

types/src/v2/request.rs
```diff
@@ -20,6 +22,43 @@ pub struct JsonRpcRequest<'a> {
...
+}
+
+impl OwnedJsonRpcRequest {
+	/// Converts `OwnedJsonRpcRequest` into borrowed `JsonRpcRequest`.
+	pub fn borrowed(&self) -> JsonRpcRequest<'_> {
+		JsonRpcRequest {
+			jsonrpc: self.jsonrpc,
+			id: self.id.borrowed(),
+			method: Cow::borrowed(self.method.as_ref()),
+			params: self.params.as_ref().map(|s| serde_json::from_str(&s).unwrap()),
```
niklasad1, line 46: please use `expect` with some explanation why it can't panic

utils/src/server/rpc_module.rs
```diff
@@ -206,6 +282,48 @@ impl<Context> RpcContextModule<Context> {
...
+		Context: Sync + Send + 'static,
+	{
+		self.module.verify_method_name(method_name)?;
+		let ctx = self.ctx.clone();
+
+		self.module.method_types.insert(method_name, MethodType::Async);
+		self.module.async_methods.insert(
+			method_name,
+			Box::new(move |id, params, tx, _| {
+				let tx = tx.clone();
```
niklasad1, line 303: same here; redundant clone

utils/src/server/rpc_module.rs
```diff
@@ -143,20 +217,21 @@ impl RpcModule {
 		Ok(SubscriptionSink { method: subscribe_method_name, subscribers })
 	}
 
-	/// Convert a module into methods.
```
niklasad1, line 146: right, this doesn't make sense anymore because `Methods` and `AsyncMethods` are two different types, but why not `-> (Methods, AsyncMethods)`?

popzxc: I thought that keeping the resolving logic in `RpcModule` can lead to a bit more simple code. Like, the lookup of method existence in both sync and methods collections is unified under the `method_type`.

All the methods that should not be called in the context where this removed method would've been called are taking `&mut self` reference, thus we won't be able to call them as long as we keep the object immutable.

Or is there a reasoning that I don't see?

popzxc:
It's weird that the tests job got stuck and runs for over 4 hours already. Locally it passes almost immediately.

Did that ever happen before? It may be a signal of some `await`ed future never getting finished, and I wonder whether it could've been caused by my changes.

niklasad1:
> Did that ever happen before? It may be a signal of some awaited future never getting finished, and I wonder whether it could've been caused by my changes.

It's a GitHub CI issue that occurs "occasionally" but I haven't been able to reproduce it locally, but nothing related to this PR.

dvdplm:
> I hesitate a bit and don't want to be annoying, but is there an estimate when I can wait for the review?

Not annoying! This was next on our list after we sort out subscriptions. I think we'll get to this on Monday so might be smart to wait a bit to rebase.

maciejhirsz (approved):

maciejhirsz:
Code looks good to me if this is the API we decide on.

utils/src/server/rpc_module.rs
niklasad1: I think it might be issues with #350 that's all I can think of.

niklasad1 (approved):
I'm down for merging this and maybe improve down the road.

> Handle both sync and async methods via tokio::spawn. This approach is actually interesting, because sync methods may be distributed among several runtime threads and will be executed in parallel despite being sync.

Another benefit might be that we don't need to distinguish between async and sync calls with removes some boiler plate I guess. However, regarding the concurrency I think the major improvement here is for batch requests and maybe requests for the WebSocket server but it would interesting to explore IMHO. FYI, HTTP requests are spawned on separate task by hyper.

utils/src/server/rpc_module.rs
popzxc: I've created a `MethodsHolder` structure which contains both `Methods` and `AsyncMethods` and has a convenient interface as a midground.

maciejhirsz:
@popzxc Hey, we're just talking about merging this and wanted to ask you if you can fix the CI, which you did :). Do you want to commit any more changes here or are we go?

utils/src/server/rpc_module.rs
```diff
@@ -28,41 +35,104 @@ pub type MethodSink = mpsc::UnboundedSender<String>;
...
+/// Identifier of the method type.
+#[derive(Debug, Copy, Clone)]
+pub enum MethodType {
+	/// Synchronous method handler.
+	Sync,
+	/// Asynchronous method handler.
+	Async,
+}
+
+/// Collection of syncrhonous and asynchronous methods.
```
niklasad1, line 47:
```suggestion
/// Collection of synchronous and asynchronous methods.
```

popzxc:
> Do you want to commit any more changes here or are we go?

Nah, I think it can be merged.

PR #332, opened 2021-05-22
Add timeouts to the futures in tests

popzxc:
I was a bit worried by the fact that CI jobs sometimes get stuck on their own.
I think it *may* be a signal of some hard-to-get but yet dangerous issue (or some kind of misconfiguration).

This PR adds ~~5~~60-second timeout to the futures spawned in tests, so it will be easier to get enough information and probably find the common pattern in these failures. Also, with it jobs will fail fast, without having to wait for the timeout.

### DIFF
--- a/http-client/src/tests.rs
+++ b/http-client/src/tests.rs
@@ -9,19 +9,22 @@ use jsonrpsee_test_utils::TimeoutFutureExt;
 
 #[tokio::test]
 async fn method_call_works() {
-	let result =
-		run_request_with_response(ok_response("hello".into(), Id::Num(0))).with_timeout().await.unwrap().unwrap();
+	let result = run_request_with_response(ok_response("hello".into(), Id::Num(0)))
+		.with_default_timeout()
+		.await
+		.unwrap()
+		.unwrap();
 	assert_eq!(JsonValue::String("hello".into()), result);
 }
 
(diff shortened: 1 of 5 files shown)


niklasad1 (commented):
It seems rather opinionated to assume that `test-futures` will complete in 5 seconds but it would certainly make the CI more responsive and help find the cause but maybe increase the timeout a bit?

So why not but I think `fn with_timeout(self)` is too implicit it should be renamed something more descriptive such as `fn timeout_five_seconds` or take the actual timeout period as parameter.

niklasad1 (approved):

dvdplm:
@niklasad1 What do you think of making all requests time out by default? So that all calls are wrapped in a generous default timeout (1min seems good) and users can use `with_timeout()` to change that if they want?

niklasad1:
> @niklasad1 What do you think of making all requests time out by default? So that all calls are wrapped in a generous default timeout (1min seems good) and users can use with_timeout() to change that if they want?

We kind of already have this for the [WsClient](https://github.com/paritytech/jsonrpsee/blob/master/ws-client/src/client.rs#L212-#L216) but it's not enabled by default.

I don't know whether having it as default is good but most requests would probably be answered in 1 min :)
It needs to be added to HTTP client too FWIW.

PR #333, opened 2021-05-24
Switch from anyhow::Result to std Result

dvdplm:
Don't return `anyhow::Result`﻿and instead just use `thiserror` and standard `Result<T, Error>` where `Error` is the `jsonrpsee_types::error::Error` type.

The downside is that with this PR `jsonrpsee_types` depends on both `soketto` and `hyper`. I think making `jsonrpsee_types` as dependency-free as possible was a big part of an external crate in the first place. OTOH all users will depend on at least one of them.

### DIFF
--- a/examples/http.rs
+++ b/examples/http.rs
@@ -49,7 +49,7 @@ async fn main() -> anyhow::Result<()> {
 async fn run_server() -> anyhow::Result<SocketAddr> {
 	let mut server = HttpServerBuilder::default().build("127.0.0.1:0".parse()?)?;
 	server.register_method("say_hello", |_| Ok("lo"))?;
-	let addr = server.local_addr();
+	let addr = server.local_addr()?;
 	tokio::spawn(async move { server.start().await });
-	addr
+	Ok(addr)
 }
--- a/types/Cargo.toml
+++ b/types/Cargo.toml
@@ -18,7 +18,5 @@ log = { version = "0.4", default-features = false }
 serde = { version = "1", default-features = false, features = ["derive"] }
 serde_json = { version = "1", default-features = false, features = ["alloc", "raw_value", "std"] }
 thiserror = "1.0"
-# TODO: a bit sad to have to pull in soketto to handle errors
 soketto = "0.4"
-# TODO: a bit sad to have to pull in soketto to handle errors
 hyper = "0.14"
--- a/types/Cargo.toml
+++ b/types/Cargo.toml
@@ -18,3 +18,7 @@ log = { version = "0.4", default-features = false }
 serde = { version = "1", default-features = false, features = ["derive"] }
 serde_json = { version = "1", default-features = false, features = ["alloc", "raw_value", "std"] }
 thiserror = "1.0"
+# TODO: a bit sad to have to pull in soketto to handle errors
+soketto = "0.4"
+# TODO: a bit sad to have to pull in soketto to handle errors
+hyper = "0.14"
--- a/types/src/error.rs
+++ b/types/src/error.rs
@@ -31,7 +31,7 @@ pub enum CallError {
 pub enum Error {
 	/// Error that occurs when a call failed.
 	#[error("Server call failed: {0}")]
-	Call(CallError),
+	Call(#[from] CallError),
 	/// Networking error or error on the low-level protocol layer.
 	#[error("Networking or low-level protocol error: {0}")]
 	Transport(#[source] Box<dyn std::error::Error + Send + Sync>),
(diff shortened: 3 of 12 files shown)


examples/http.rs
```diff
@@ -49,7 +49,7 @@ async fn main() -> anyhow::Result<()> {
 async fn run_server() -> anyhow::Result<SocketAddr> {
 	let mut server = HttpServerBuilder::default().build("127.0.0.1:0".parse()?)?;
 	server.register_method("say_hello", |_| Ok("lo"))?;
-	let addr = server.local_addr();
+	let addr = server.local_addr()?;
 	tokio::spawn(async move { server.start().await });
-	addr
+	Ok(addr)
```
maciejhirsz, line 54: Good catch.

types/src/error.rs
```diff
@@ -88,3 +89,39 @@ pub enum GenericTransportError<T: std::error::Error + Send + Sync> {
 	#[error("Transport error: {0}")]
 	Inner(T),
 }
+
+impl From<CallError> for Error {
```
maciejhirsz, line 93: Can't `thiserror` derive these?

types/Cargo.toml
```diff
@@ -18,3 +18,7 @@ log = { version = "0.4", default-features = false }
 serde = { version = "1", default-features = false, features = ["derive"] }
 serde_json = { version = "1", default-features = false, features = ["alloc", "raw_value", "std"] }
 thiserror = "1.0"
+# TODO: a bit sad to have to pull in soketto to handle errors
+soketto = "0.4"
```
niklasad1, line 22: hmm, can't you use [this](https://github.com/paritytech/jsonrpsee/blob/master/types/src/error.rs#L36) instead?

or do we want concrete error types for every error?

types/src/error.rs
dvdplm: TIL!

types/Cargo.toml
niklasad1, line 22: yeah, I don't know how handle those without removing the shared error type.

but maybe a good idea to kill the big error type and that each crates has to define their own error type?

dvdplm: Won't that make it tricky to share code between the servers?

niklasad1: Yeah, unless you come up with super nice generic error type that both can use but then you need some extra boiler-plate.

Like `map.err(|e| Error::Transport(Box::new(e))`

dvdplm: I can't think of a better way other than adding features, which I'm not sure is such a good idea.

types/Cargo.toml
```diff
@@ -18,3 +18,7 @@ log = { version = "0.4", default-features = false }
...
+# TODO: a bit sad to have to pull in soketto to handle errors
```
niklasad1, line 23:
```suggestion
# TODO: a bit sad to have to pull in hyper to handle errors
```

niklasad1 (approved):

maciejhirsz (approved):

types/Cargo.toml
niklasad1: ok, you mean making `soketto` and `hyper` optional dependencies?

That's ok but why not let each crate implement implement the `From<TransportError> for Error`?
Then we shouldn't bring in any extra dependencies but a little bit of "redundant code", like for client/server as long they use the same transport library

utils/src/server/rpc_module.rs
```diff
@@ -130,7 +130,8 @@ impl RpcModule {
 			self.methods.insert(
 				unsubscribe_method_name,
 				Box::new(move |id, params, tx, conn| {
-					let sub_id = params.one().map_err(|e| anyhow::anyhow!("{:?}", e))?;
+					// let sub_id = params.one().map_err(|e| anyhow::anyhow!("{:?}", e))?;
```
niklasad1, line 133:
```suggestion
```

types/Cargo.toml
dvdplm: > hat's ok but why not let each crate implement implement the `From<TransportError> for Error`?
> Then we shouldn't bring in any extra dependencies but a little bit of "redundant code", like for client/server as long they use the same transport library

https://github.com/paritytech/jsonrpsee/pull/334

utils/src/server/rpc_module.rs
dvdplm: My bad. Fixed this one in https://github.com/paritytech/jsonrpsee/pull/334

PR #334, opened 2021-05-24
[servers] Mmove transport specific errors to respective crates

dvdplm:
Move transport-specific error conversion to http/ws crates respectively.

This lets us avoid pulling in hyper and soketto as dependencies to the `jsonrpsee_types` crate.

### DIFF
--- a/http-server/src/lib.rs
+++ b/http-server/src/lib.rs
@@ -28,6 +28,8 @@ mod access_control;
 mod response;
 mod server;
 
+use std::fmt::{self, Debug, Display};
+
 pub use access_control::{AccessControl, AccessControlBuilder, AllowHosts, Host};
 pub use jsonrpsee_types::{Error, TEN_MB_SIZE_BYTES};
 pub use jsonrpsee_utils::server::rpc_module::{Methods, RpcContextModule, RpcModule};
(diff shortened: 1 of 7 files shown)


ws-server/src/server.rs
```diff
@@ -107,7 +109,7 @@ async fn background_task(
 	socket: tokio::net::TcpStream,
 	methods: Arc<Methods>,
 	conn_id: ConnectionId,
-) -> Result<(), Error> {
+) -> Result<(), WsError> {
```
niklasad1, line 112: shouldn't this be `Error` for a uniform API for the servers?

http-server/src/lib.rs
```diff
@@ -28,10 +28,40 @@ mod access_control;
...
+	}
+}
+
+impl From<hyper::Error> for HttpError {
+	fn from(hyper_err: hyper::Error) -> HttpError {
+		HttpError(Error::Transport(Box::new(hyper_err)))
+	}
+}
+
+impl From<std::io::Error> for HttpError {
```
niklasad1, line 57: is not `From<std::io::Error> for Error` in jsonrpsee_types sufficient?

dvdplm: It is, but there's an `io::Error` raised in `ws-server` as well. Maybe I can move the conversion from `jsonrpsee_types` to `ws-server` (but it has to be in two places either way).

The downside of this PR is that it becomes less obvious where errors are defined/handled. I'm not sure if I like it tbh.

ws-server/src/server.rs
dvdplm: For posterity: you are correct, the API should be consistent and return the same error type, i.e. `jsonrpsee_types::Error`. That would require a bunch of yucky `map_err` which would need to be annotated, e.g. `.map_err::<WsError, _>(Into::into)`.

dvdplm:
Given how ugly this is we prefer to have `jsonrpsee_types` depend on `hyper`/`soketto` instead.

Issue #306, opened 2021-05-10
[ws client]: Handshake failed connecting to valid URL


chevdor:
Considering the following example (not in the repo):
```
use jsonrpsee::ws_client::WsClientBuilder;
use std::env;

fn get_ws_node() -> String {
	env::var("POLKADOT_WS").unwrap_or("ws://localhost:9944".to_string())
}

#[tokio::main]
async fn main() -> anyhow::Result<()> {
	env_logger::init();

	let url = String::from(get_ws_node());
	println!("Connecting to {:?}", &url);
	let client = WsClientBuilder::default().build(&url).await?;
	assert!(client.is_connected());
        println!("connected");
	Ok(())
}
```

This will succeed:
```
POLKADOT_WS=wss://rpc.polkadot.io c run --example ws_polkadot  
```

But this will fail despite the endpoint to be valid and up:
```
POLKADOT_WS=wss://polkadot.api.onfinality.io/public-ws c run --example ws_polkadot
    Finished dev [unoptimized + debuginfo] target(s) in 0.13s
     Running `target/debug/examples/ws_polkadot`
Connecting to "wss://polkadot.api.onfinality.io/public-ws"
Error: Networking or low-level protocol error: Couldn't find any IP address for this hostname

Caused by:
    Couldn't find any IP address for this hostname
```

niklasad1:
Looks like the server responds with `HTTP redirect`, https://github.com/paritytech/jsonrpsee/blob/master/ws-client/src/transport.rs#L251

We should fix that

niklasad1:
@chevdor we switched to `native certificate storage` which seemed to fix this, can you try to reproduce this again?

Works for me now.

niklasad1:
Closed by #335

PR #335, opened 2021-05-24
[ws client]: parse `path` from the URL

niklasad1:
Closing #306 

This commit changes that the path/HTTP resource path is parsed from the URL and removes that option
from WsClientBuilder.

### DIFF
--- a/ws-client/src/client.rs
+++ b/ws-client/src/client.rs
@@ -25,7 +25,7 @@
 // DEALINGS IN THE SOFTWARE.
 
 use crate::traits::{Client, SubscriptionClient};
-use crate::transport::{parse_url, Receiver as WsReceiver, Sender as WsSender, WsTransportClientBuilder};
+use crate::transport::{Receiver as WsReceiver, RemoteWsTarget, Sender as WsSender, WsTransportClientBuilder};
 use crate::v2::error::JsonRpcError;
 use crate::v2::params::{Id, JsonRpcParams};
 use crate::v2::request::{JsonRpcCallSer, JsonRpcNotificationSer};
@@ -259,14 +259,9 @@ impl<'a> WsClientBuilder<'a> {
 		let (to_back, from_front) = mpsc::channel(self.max_concurrent_requests);
 		let (err_tx, err_rx) = oneshot::channel();
 
-		let (sockaddrs, host, mode, path) = parse_url(url).map_err(|e| Error::Transport(Box::new(e)))?;
-
 		let builder = WsTransportClientBuilder {
 			certificate_store,
-			sockaddrs,
-			mode,
-			host,
-			path,
+			target: RemoteWsTarget::parse(url).map_err(|e| Error::Transport(Box::new(e)))?,
 			timeout: self.connection_timeout,
 			origin_header: self.origin_header,
 			max_request_body_size: self.max_request_body_size,
--- a/ws-client/src/client.rs
+++ b/ws-client/src/client.rs
@@ -25,7 +25,7 @@
 // DEALINGS IN THE SOFTWARE.
 
 use crate::traits::{Client, SubscriptionClient};
-use crate::transport::{Receiver as WsReceiver, Target, Sender as WsSender, WsTransportClientBuilder};
+use crate::transport::{Receiver as WsReceiver, Sender as WsSender, Target, WsTransportClientBuilder};
 use crate::v2::error::JsonRpcError;
 use crate::v2::params::{Id, JsonRpcParams};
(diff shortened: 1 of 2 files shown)


ws-client/src/transport.rs
```diff
@@ -191,30 +202,6 @@ impl Receiver {
 }
 
 impl<'a> WsTransportClientBuilder<'a> {
-	/// Sets the URL to pass during the HTTP handshake.
-	///
-	/// The default URL is `/`.
-	pub fn with_handshake_url(mut self, url: impl Into<Cow<'a, str>>) -> Self {
```
niklasad1, line 197: unsed, thus removed.

ws-client/src/client.rs
```diff
@@ -175,8 +175,7 @@ pub struct WsClientBuilder<'a> {
 	max_request_body_size: u32,
 	request_timeout: Option<Duration>,
 	connection_timeout: Duration,
-	origin: Option<Cow<'a, str>>,
-	handshake_url: Cow<'a, str>,
```
niklasad1, line 179: removed this and parses it from the URL instead

ws-client/src/client.rs
```diff
@@ -210,8 +208,8 @@ impl<'a> WsClientBuilder<'a> {
 	}
 
 	/// Set request timeout.
-	pub fn request_timeout(mut self, timeout: Option<Duration>) -> Self {
```
niklasad1, line 213: annoying builder method; unrelated to this PR

ws-client/src/client.rs
```diff
@@ -210,8 +208,8 @@ impl<'a> WsClientBuilder<'a> {
...
-		self.request_timeout = timeout;
+	pub fn request_timeout(mut self, timeout: Duration) -> Self {
```
niklasad1, line 211: annoying builder method; unrelated to this PR

chevdor:
I tested the fix and that's a PASS for me.

ws-client/src/client.rs
```diff
@@ -267,16 +259,16 @@ impl<'a> WsClientBuilder<'a> {
 		let (to_back, from_front) = mpsc::channel(self.max_concurrent_requests);
 		let (err_tx, err_rx) = oneshot::channel();
 
-		let (sockaddrs, host, mode) = parse_url(url).map_err(|e| Error::Transport(Box::new(e)))?;
+		let (sockaddrs, host, mode, path) = parse_url(url).map_err(|e| Error::Transport(Box::new(e)))?;
```
dvdplm, line 262: 👍

ws-client/src/transport.rs
```diff
@@ -323,7 +310,7 @@ impl From<soketto::connection::Error> for WsConnectError {
 }
 
 /// Helper to parse an URL to a WebSocket address.
-pub fn parse_url(url: impl AsRef<str>) -> Result<(Vec<SocketAddr>, Host, Mode), WsHandshakeError> {
+pub fn parse_url(url: impl AsRef<str>) -> Result<(Vec<SocketAddr>, Host, Mode, Path), WsHandshakeError> {
```
dvdplm, line 313: 4 is about the limit for tuples imo. It's fine here but if there are ever more items added here I think we might want a `PeerConfig` struct or something.

ws-client/src/transport.rs
```diff
@@ -46,6 +46,17 @@ impl Host {
...
+/// String representation of the path of the URL.
+#[derive(Clone, Debug)]
+pub struct Path(String);
+
+impl Path {
+	/// Extracts a string slice from the inner String.
+	pub fn as_str(&self) -> &str {
+		self.0.as_str()
+	}
+}
```
dvdplm, line 58: Not sure if this type is pulling its weight, we could probably do without it yes?

dvdplm (approved):
Good sleuthing here. Left a few suggestions, up to you if you want to implement them. :)

ws-client/src/transport.rs
niklasad1: let's do `PeerInfo` or `ParsedWsAddress` and group all in type instead.

I just don't like have a bunch of Strings in tuple, it's error-prone.

TarikGul (approved):
I like the change. 👍  LGTM

ws-client/src/transport.rs
dvdplm: Maybe "peer" is a poor choice, this is not really a p2p context. How about "Remote": `RemoteInfo`/`RemoteAddress`?

niklasad1: RemoteWsTarget ok?

Issue #337, opened 2021-05-25
NoAddressFound on some endpoints


chevdor:
This issue could be related to #320.

I am using `WsClientBuilder::default().build(&url)` to connect to various endpoints.
Most endpoints work, that includes:
- wss://kusama.elara.patract.io
- wss://westend.elara.patract.io

The following however, fail with `Transport(NoAddressFound)`:
- wss://rococo.elara.patract.io
- wss://47.241.106.95 (manually resolved IP of the above)

niklasad1:
Are you using latest master or running on `0.2.0-alpha.6?

On master(using the native certificate store) I get the following:

```
[2021-05-25T19:47:25Z DEBUG rustls::anchors] add_pem_file processed 128 valid and 0 invalid certs
[2021-05-25T19:47:25Z DEBUG rustls::client::hs] No cached session for DNSNameRef("rococo.elara.patract.io")
[2021-05-25T19:47:25Z DEBUG rustls::client::hs] Not resuming any session
[2021-05-25T19:47:26Z DEBUG rustls::client::hs] Using ciphersuite TLS13_AES_128_GCM_SHA256
[2021-05-25T19:47:26Z DEBUG rustls::client::tls13] Not resuming
[2021-05-25T19:47:26Z DEBUG rustls::client::tls13] TLS1.3 encrypted extensions: []
[2021-05-25T19:47:26Z DEBUG rustls::client::hs] ALPN protocol is None
[2021-05-25T19:47:26Z DEBUG rustls::client::tls13] Ticket saved
[2021-05-25T19:47:26Z DEBUG jsonrpsee_ws_client::transport] Failed to connect to sockaddr: 47.241.106.95:443 with err: Rejected { status_code: 405 }
Error: Networking or low-level protocol error: Couldn't find any IP address for this hostname
```

Thus, the `dns name` was resolved the IP address but the server refused the request, status code 405.
The error message could be improved though.

chevdor:
I am using:
```
jsonrpsee = {git = "https://github.com/paritytech/jsonrpsee", features = ["full"]}
```

and that should be rather "fresh":
```
git+https://github.com/paritytech/jsonrpsee#a60896e16c192b2fb4f3efea166ebcb4d9d1ba3b
```

niklasad1:
Ok, right,

Can you run it with `RUST_LOG="debug" cargo run --example ws` and provide the logs?

[...]

PR #338, opened 2021-05-25
[ws client]: return last seen error in connect

niklasad1:
The connect error is ignored and always returns `NoAddressFound` which this commit fixes.

Related to #337

### DIFF
--- a/ws-client/src/transport.rs
+++ b/ws-client/src/transport.rs
@@ -185,9 +185,9 @@ impl<'a> WsTransportClientBuilder<'a> {
 				}
 			}
 		}
-		// NOTE(niklasad1): this is most likely unreachable because `Url::socket_addrs` doesn't
-		// return an empty `Vec` if no socket addresses were found for this host.
-		err.unwrap_or_else(|| Err(WsHandshakeError::NoAddressFound(self.target.host)))
+		// NOTE(niklasad1): this is most likely unreachable because [`Url::socket_addrs`] doesn't
+		// return an empty `Vec` if no socket address was found for the host name.
+		err.unwrap_or(Err(WsHandshakeError::NoAddressFound(self.target.host)))
 	}
 
 	async fn try_connect(
--- a/ws-client/src/transport.rs
+++ b/ws-client/src/transport.rs
@@ -188,15 +188,17 @@ impl<'a> WsTransportClientBuilder<'a> {
 			Mode::Plain => None,
 		};
 
+		let mut err = Err(WsHandshakeError::NoAddressFound);
 		for sockaddr in &self.target.sockaddrs {
 			match self.try_connect(*sockaddr, &connector).await {
 				Ok(res) => return Ok(res),
 				Err(e) => {
 					log::debug!("Failed to connect to sockaddr: {:?} with err: {:?}", sockaddr, e);
+					err = Err(e.into());
 				}
 			}
 		}
-		Err(WsHandshakeError::NoAddressFound)
+		err
 	}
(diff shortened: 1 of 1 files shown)


ws-client/src/transport.rs
```diff
@@ -188,15 +188,17 @@ impl<'a> WsTransportClientBuilder<'a> {
...
 			match self.try_connect(*sockaddr, &connector).await {
 				Ok(res) => return Ok(res),
 				Err(e) => {
 					log::debug!("Failed to connect to sockaddr: {:?} with err: {:?}", sockaddr, e);
+					err = Err(e.into());
 				}
 			}
 		}
-		Err(WsHandshakeError::NoAddressFound)
+		err
```
dvdplm, line 201: It's a bit of a weird behaviour imo, but given `std::net::TcpStream::connect` does it, I guess it's fine we do it too. :/

An alternative could be to return immediately, at the first error, **unless** the error is `NoAddressFound`?

Either way, we need to adjust the docs for `WsHandshakeError`.

dvdplm (approved):
(but please adjust the docs [here](https://github.com/paritytech/jsonrpsee/blob/ec3e039c322a368941376c5e486e8cbbbe818415/ws-client/src/transport.rs#L151))

ws-client/src/transport.rs
niklasad1: It's documented [here](https://github.com/paritytech/jsonrpsee/blob/5aeda9384fa65fc1aafa14a96b5e5eb114524c60/ws-client/src/transport.rs#L123-#L128), not sufficient?

TarikGul (approved):
LGTM

ws-client/src/transport.rs
niklasad1: I think we can remove `ResolutionFailed` because it's not possible anymore because the IP addresses are "resolved" before this but the sockaddrs could still be empty.

dvdplm: > It's documented [here](https://github.com/paritytech/jsonrpsee/blob/5aeda9384fa65fc1aafa14a96b5e5eb114524c60/ws-client/src/transport.rs#L123-#L128), not sufficient?

I read that as if it applied only to the `WsHandshakeError::Connect` variant, but it applies to all variants actually, like if I give 3 IPs, first fails with `CertificateStore`, second with `NoAddressFound` and the third `Connect`, the caller would only see `Connect`, correct? So maybe the docs should move to the enum itself? Not sure.

niklasad1: fine, I merged to `WsNewError` with `WsHandshakeError` and added docs for the entire enum `WsHandshakeError`

it's technically not only `WsHandshakeError` in there such as `bad url` but better at least.

also improved the error messages a bit, I think this is good enough for now?

ws-client/src/transport.rs
```diff
@@ -82,81 +82,68 @@ pub enum CertificateStore {
...
+pub enum WsError {
 	/// Error in the WebSocket connection.
-	#[error("error in the WebSocket connection")]
-	Ws(#[source] soketto::connection::Error),
+	#[error("WebSocket connection error: {}", 0)]
+	Connection(#[source] soketto::connection::Error),
 
-	/// Failed to parse the JSON returned by the server into a JSON-RPC response.
-	#[error("error while parsing the response body")]
+	/// Failed to parse message in JSON.
```
niklasad1, line 138:
```suggestion
	/// Failed to parse the message in JSON.
```

PR #342, opened 2021-05-27
chore(deps): update futures 0.3.14

niklasad1:
The reason is that `Receiver::next` won't panic polling after `Ok(None)` has been received.


dvdplm (approved):
👍

PR #345, opened 2021-05-27
Subscription with context example

dvdplm:


types/src/client.rs
```diff
@@ -116,7 +116,10 @@ where
 			match self.notifs_rx.next().await {
 				Some(n) => match serde_json::from_value(n) {
 					Ok(parsed) => return Some(parsed),
-					Err(e) => log::debug!("Subscription response error: {:?}", e),
+					Err(e) => {
+						log::error!("Subscription response error: {:?}", e);
+						return None
+					}
```
dvdplm, line 122: This is the only code change here.

niklasad1 (approved):

examples/weather.rs
```diff
@@ -0,0 +1,134 @@
...
+				} else {
+					log::trace!(target: "server", "Same weather as before")
+				}
+				std::thread::sleep(std::time::Duration::from_millis(500));
+			});
+			Ok(())
+		})
+		.unwrap();
+
+	server.register_module(module.into_module()).unwrap();
```
dvdplm, line 129: @niklasad1 I think this is a bit awkward, having to cast my context module into a module to register it with the server. Would dynamic dispatch be possible here you think, so that `register_module` takes a `dyn RpcModuleTraitObject` instead?

niklasad1: I think we should merge `RpcModule` and `RpcContextModule` with `NoopContext` and just have a type def:

```
pub struct RpcContextModule<Context> {
...
}

pub type RpcModule: RpcContextModule<()>;

```

I think that would fix it without having do trait objects

examples/weather.rs
```diff
@@ -0,0 +1,134 @@
...
+	api_client: restson::RestClient,
+	last_weather: Weather,
+}
+
+async fn run_server() -> anyhow::Result<SocketAddr> {
+	let mut server = WsServer::new("127.0.0.1:0").await?;
+
+	let api_client = restson::RestClient::new("http://api.openweathermap.org").unwrap();
+	let last_weather = Weather::default();
+	let cx = Mutex::new(WeatherApiCx { api_client, last_weather });
```
niklasad1, line 107: I suppose you could move this into closure but just an example so cool.

dvdplm: That doesn't actually work, because the `restson` library is crap and requires a mutable ref to do `get()`.

examples/weather.rs
dvdplm: That's a good idea, I'll explore that.

PR #350, opened 2021-05-28
Unify RpcModule types

dvdplm:
Instead of to duplicate code between `RpcModule` and `RpcContextModule`, use a single module type. Remove the `register_*` methods for `Server` and instead force users to always create an `RpcModule`.


types/src/traits.rs
```diff
@@ -50,8 +50,3 @@ pub trait SubscriptionClient: Client {
 	where
 		Notif: DeserializeOwned;
 }
-
-/// JSON-RPC server interface for managing method calls.
-pub trait RpcMethod<R, E>: Fn(RpcParams) -> Result<R, E> + Send + Sync + 'static {}
```
dvdplm, line 55: I removed this because I had to use the explicit type anyway [here](https://github.com/paritytech/jsonrpsee/pull/350/files#diff-7cf339924bf9f55b7b2a32c481d69d8e70114e91a0c074b7228bf7f80f964f8aR59) because rustc could not infer the types anyway. Not sure how big of a deal this is though?

utils/src/server/rpc_module.rs
```diff
@@ -147,7 +146,8 @@ impl RpcModule {
 						sub_id,
 						is_online: online_tx,
 					};
-					callback(params, sink)
+					// TODO: maybe params, context, sink is a more natural order?
```
dvdplm, line 149: Any strong opinions on this?

http-server/src/server.rs
```diff
@@ -113,26 +112,19 @@ pub struct Server {
...
+	methods: Methods,
 	/// Max request body size.
 	max_request_body_size: u32,
 	/// Access control
 	access_control: AccessControl,
 }
 
 impl Server {
-	/// Register a new RPC method, which responds with a given callback.
-	pub fn register_method<F, R>(&mut self, method_name: &'static str, callback: F) -> Result<(), Error>
```
niklasad1, line 125: random idea until we have a proper builder for this; is keep this method but internally create a RpcModule<()> and append it the the methods?

ws-server/src/server.rs
```diff
@@ -51,34 +50,14 @@ impl Server {
...
-	pub fn register_method<R, F>(&mut self, method_name: &'static str, callback: F) -> Result<(), Error>
-	where
-		R: Serialize,
-		F: Fn(RpcParams) -> Result<R, CallError> + Send + Sync + 'static,
-	{
-		self.root.register_method(method_name, callback)
-	}
-
-	/// Register a new RPC subscription, with subscribe and unsubscribe methods.
-	pub fn register_subscription<F>(
```
niklasad1, line 67: same question here; if we could keep this for now but create a  `temp RpcModule` internally here and append it the methods?

tests/tests/helpers.rs
```diff
@@ -25,7 +25,12 @@
 // DEALINGS IN THE SOFTWARE.
 
 use futures_channel::oneshot;
-use jsonrpsee::{http_server::HttpServerBuilder, ws_server::WsServer};
+// TODO: a bit annoying to import ws_server::RpcModule when in truth it's available from both http and ws servers.
+// Export it directly from the façade crate?
```
niklasad1, line 29: :+1:

utils/src/server/rpc_module.rs
niklasad1: no, //cc @maciejhirsz

http-server/src/server.rs
```diff
@@ -113,26 +112,19 @@ pub struct Server {
...
-	{
-		self.root.register_method(method_name, callback)
-	}
-
 	/// Register all methods from a module on this server.
-	pub fn register_module(&mut self, module: RpcModule) -> Result<(), Error> {
-		self.root.merge(module)
+	pub fn register_module<Context>(&mut self, module: RpcModule<Context>) -> Result<(), Error> {
+		// TODO: must check for duplicate method names
+		self.methods.extend(module.into_methods());
```
niklasad1, line 126: I suggest having a separate type `Methods` and have a method `fn merge`

http-server/src/tests.rs
```diff
@@ -37,23 +39,22 @@ pub async fn server_with_context() -> SocketAddr {
 	let mut server = HttpServerBuilder::default().build("127.0.0.1:0".parse().unwrap()).unwrap();
 
 	let ctx = TestContext;
-	let mut rpc_ctx = RpcContextModule::new(ctx);
+	let mut rpc_module = RpcModule::new(ctx);
```
niklasad1, line 42: maybe have to different `constructors` on RpcModule such as `with_context` and `default`

http-server/src/server.rs
niklasad1: I think this will be annoying to implement nvm

http-server/src/tests.rs
dvdplm: What does `without_context` use as the context? `()`?

niklasad1: yeah, maybe overkill

utils/src/server/rpc_module.rs
```diff
@@ -390,26 +235,28 @@ impl SubscriptionSink {
 mod tests {
 	use super::*;
 	#[test]
-	fn rpc_context_modules_can_merge_with_rpc_module() {
-		// Prove that we can merge an RpcContextModule with a RpcModule.
+	// TODO: does it make sense to merge two RpcModules with the same context? Previously this worked because an
+	// RpcContextModule was cast to an RpcModule and then merged (so the `Context` type param was tossed away). Now that
+	// they are the same they can't be cast. But their methods can still have different contexts. Maybe the RpcModule
+	// should not have a type param at all, and instead only the register_method/register_subscription functions?
```
niklasad1, line 241: hmm, how would a RpcModule access its context in method calls and subscriptions then?

utils/src/server/rpc_module.rs
```diff
@@ -390,26 +235,28 @@ impl SubscriptionSink {
 mod tests {
 	use super::*;
 	#[test]
-	fn rpc_context_modules_can_merge_with_rpc_module() {
-		// Prove that we can merge an RpcContextModule with a RpcModule.
+	// TODO: does it make sense to merge two RpcModules with the same context? Previously this worked because an
```
niklasad1, line 238: sounds reasonable to me to allow merging to modules with the same context but not with different context.

the proper way is to add the module(s) to a server I suppose.

niklasad1 (commented):
Looks reasonable to me but would nice a to a simple way to register methods and subscriptions on the server too.

It may be handled by a builder instead and ignored in this PR. 
Very nice to get rid of redundant code :1st_place_medal:

utils/src/server/rpc_module.rs
dvdplm: Yeah, this is an unfinished thought. My train of thought was:
two modules can have different contexts
when we register them on the server we actually just take the methods
…so coming from two different modules (with different contexts) is not a problem
…but that sort of hints at the context actually being a property of the method, not the module
=> and so, I thought, maybe the context could be passed into the `register_*` methods and "hidden" in the closure and we wouldn't need the `Context` type parameter at all on the module.

dvdplm: I tried out this idea and it works, for some definition of "works". It requires the `register_*` methods to pass in the context explicitly. Overall the API gets pretty clunky.

```
pub struct RpcModule {
	methods: Methods,
	subscribers: Subscribers,
}
impl RpcModule {
        … …
	pub fn register_subscription<F, Context>(
		&mut self,
		ctx: Arc<Context>,
		callback: F,
	) -> Result<(), Error>
	where
		F: Fn(RpcParams, SubscriptionSink, Arc<Context>) -> Result<(), Error> + Send + Sync + 'static,
	{ … … }
	pub fn register_method<R, F, Context>(&mut self, method_name: &'static str, ctx: Arc<Context>, callback: F) -> Result<(), Error>
	where
		R: Serialize,
		F: Fn(RpcParams, &Context) -> Result<R, CallError> + Send + Sync + 'static,
	{ … … }
}
```

dvdplm: PoC branch [here](https://github.com/paritytech/jsonrpsee/compare/dp-merge-rpc-module-types-and-get-rid-of-context-type-param-on-module?expand=1)

utils/src/server/rpc_module.rs
```diff
@@ -31,22 +29,18 @@ pub type MethodSink = mpsc::UnboundedSender<String>;
...
-	pub fn new() -> Self {
-		RpcModule { methods: Methods::default(), subscribers: Subscribers::default() }
-	}
-
-	/// Add context for this module, turning it into an `RpcContextModule`.
-	pub fn with_context<Context>(self, ctx: Context) -> RpcContextModule<Context> {
-		RpcContextModule { ctx: Arc::new(ctx), module: self, subscribers: Subscribers::default() }
+impl<Context> RpcModule<Context> {
+	/// Create a new module with a given shared `Context`.
+	pub fn new(ctx: Context) -> Self {
```
dvdplm, line 42: @niklasad1 Having to add a context everywhere is what got me thinking about why the *module* needs a context when in reality it's actually an attribute of a *call*.

http-server/src/server.rs
dvdplm: `Methods`, being a type alias, can't have `impl`s. :/

niklasad1: yeah, I meant `pub struct Methods(FxHashMap<&'static str, Method>)`

utils/src/server/rpc_module.rs
niklasad1: Ok, I see.

It's a bit annoying that have `Arc<Context>` in register_method, for example in substrate it would force us to `clone` lots of Arc's instead of one. So not sure about this one.

utils/src/server/rpc_module.rs
niklasad1: yes, that's a good observation I think it's by-design to avoid having to clone the `Context` on each call.

Otherwise, you either have to borrow (sounds tricky) or move (take by value) context in the each register_method.

For register_subscription it's assumed that the `context` will be moved into a thread or task and need to be Arc:ed or impl Clone.  However, we could let the user decide that an just require `Context` there but might be hard understand the compiler errors...

examples/proc_macro.rs
```diff
@@ -24,7 +24,10 @@
 // IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 // DEALINGS IN THE SOFTWARE.
 
-use jsonrpsee::{http_client::HttpClientBuilder, http_server::HttpServerBuilder};
+use jsonrpsee::{
+	http_client::HttpClientBuilder,
+	http_server::{HttpServerBuilder, RpcModule},
```
niklasad1, line 29: use `utils::RpcModule` here?

http-server/src/server.rs
```diff
@@ -113,26 +112,31 @@ pub struct Server {
...
-	pub fn register_method<F, R>(&mut self, method_name: &'static str, callback: F) -> Result<(), Error>
-	where
-		R: Serialize,
-		F: Fn(RpcParams) -> Result<R, CallError> + Send + Sync + 'static,
-	{
-		self.root.register_method(method_name, callback)
+	/// Register all [`Methods`] from an [`RpcModule`] on this server. In case a method already is registered with the same name,
+	/// the method is not added. Returns the number of methods registered on the server after the new module's methods
+	/// were added.
+	pub fn register_module<Context>(&mut self, module: RpcModule<Context>) -> usize {
```
niklasad1, line 126: I would prefer to return `Error` here as it was before when duplicate methods are detected to make explicit for the user the handle the error.

http-server/src/server.rs
```diff
@@ -113,26 +112,31 @@ pub struct Server {
...
+				log::warn!("Can't add method `{}`: already exists. Skipping", name);
+			}
+		}
+		self.methods.len()
 	}
 
-	/// Register all methods from a module on this server.
-	pub fn register_module(&mut self, module: RpcModule) -> Result<(), Error> {
-		self.root.merge(module)
+	/// Returns a reference to the [`Methods`] registered on the server.
```
niklasad1, line 137: hmm, when would a user need to access `Methods`?

isn't better just to return `method names` here? but I like the idea

tests/tests/helpers.rs
```diff
@@ -25,7 +25,7 @@
 // DEALINGS IN THE SOFTWARE.
 
 use futures_channel::oneshot;
-use jsonrpsee::{http_server::HttpServerBuilder, ws_server::WsServer};
+use jsonrpsee::{http_server::HttpServerBuilder, ws_server::WsServer, RpcModule};
```
niklasad1, line 28: :+1:

utils/src/server/rpc_module.rs
niklasad1: are planning to fix this in the current PR? Otherwise, open an issue and link to it

http-server/src/server.rs
dvdplm: I'll remove this actually, good catch, was part of a previous version. We can always add new methods, removing them is harder.

utils/src/server/rpc_module.rs
dvdplm: We can revisit if it becomes annoying in practise, no need to decide now.

http-server/src/server.rs
dvdplm: …actually no. I am using it tests atm and it I think it's a bit odd that the `Server` offers no introspection at all atm. :/ 
I had this on the `RpcModule` for a while and got confused, sorry for the noise.

http-server/src/server.rs
```diff
@@ -136,7 +136,7 @@ impl Server {
 
 	/// Returns a `Vec` with all the method names registered on this server.
 	pub fn method_names(&self) -> Vec<String> {
```
niklasad1, line 138: we could do `impl IntoIterator <Item = &'static str>` here instead to avoid allocations

niklasad1 (approved):

utils/src/server/rpc_module.rs
maciejhirsz: I reckon being able to put some state into your callbacks is a pretty common use case. It's not strictly necessary, but without it you'd have to provide your own state (likely by doing what we are doing internally with `Arc<Context>`), which could be very boiler plate'y.

That said, if we can swipe all that boiler plate under the rug with proc macro instead, we could potentially remove the context altogether. I do like the `RpcModule` not having generics for ease of use too.

maciejhirsz (approved):
This looks fine to me, it's definitely simpler, and if the added generics aren't a problem in substrate I think it's a win.

niklasad1:
> This looks fine to me, it's definitely simpler, and if the added generics aren't a problem in substrate I think it's a win.

Not what I'm aware of and if you want other context you could just add another module with different context and then merge or add it on the server.

PR #354, opened 2021-05-31
chore(scripts): publish script

niklasad1:


dvdplm:
@tomusdrw 's [script](https://github.com/paritytech/jsonrpc/blob/master/_automate/publish.sh) seems to do a bit more than this and has been working well over there, just saying. :)

niklasad1:
yeah, it's inspired by that ofcourse but I want to actually check all crates could be published before creating the tag

but we could use `https://github.com/paritytech/cargo-unleash` perhaps instead.

dvdplm:
> but we could use `https://github.com/paritytech/cargo-unleash` perhaps instead.

oh, I like that idea.

tomusdrw:
>  but I want to actually check all crates could be published before creating the tag

You can still do that with my script just passing `--dry-run` to parameters (note `$@` near `cargo publish`).

However it only works if the published crates can use older (already published) versions. I.e. it won't work if you do a major release and have dependent crates for instance.

niklasad1:
> You can still do that with my script just passing --dry-run to parameters (note $@ near cargo publish).

Right, my bash reading skillz are poor nowdays.

> However it only works if the published crates can use older (already published) versions. I.e. it won't work if you do a major release and have dependent crates for instance.

yeah good point

dvdplm (approved):

PR #355, opened 2021-06-01
[rpc module]: server-side close subscription

niklasad1:
If the subscription request is answered by the server and then the server decides to close the subscription in a background task. Then it's not possible to send a response on some request to cancel that subscription.

Thus, this PR adds functionality to send additional close message when the `SubscriptionSink` when it's dropped.

In the client, the changes are made accordingly to redirect this to the user when the subscription is polled.

### DIFF
--- a/tests/tests/helpers.rs
+++ b/tests/tests/helpers.rs
@@ -71,7 +71,15 @@ pub async fn websocket_server_with_subscription() -> SocketAddr {
 			})
 			.unwrap();
 
-		module.register_subscription("subscribe_noop", "unsubscribe_noop", |_, _, _| Ok(())).unwrap();
+		module
+			.register_subscription("subscribe_noop", "unsubscribe_noop", |_, sink, _| {
+				std::thread::spawn(move || {
+					std::thread::sleep(Duration::from_secs(5));
+					drop(sink);
+				});
+				Ok(())
+			})
+			.unwrap();
 
 		server.register_module(module).unwrap();
 		rt.block_on(async move {
--- a/tests/tests/helpers.rs
+++ b/tests/tests/helpers.rs
@@ -25,7 +25,7 @@
 // DEALINGS IN THE SOFTWARE.
 
 use futures_channel::oneshot;
-use jsonrpsee::{http_server::HttpServerBuilder, ws_server::WsServer, RpcModule};
+use jsonrpsee::{http_server::HttpServerBuilder, ws_server::WsServerBuilder, RpcModule};
 use std::net::SocketAddr;
 use std::time::Duration;
 
@@ -35,7 +35,8 @@ pub async fn websocket_server_with_subscription() -> SocketAddr {
 	std::thread::spawn(move || {
 		let rt = tokio::runtime::Runtime::new().unwrap();
 
-		let mut server = rt.block_on(WsServer::new("127.0.0.1:0")).unwrap();
+		let mut server = rt.block_on(WsServerBuilder::default().build("127.0.0.1:0")).unwrap();
+
 		let mut module = RpcModule::new(());
 		module.register_method("say_hello", |_, _| Ok("hello")).unwrap();
 
@@ -97,8 +98,7 @@ pub async fn websocket_server() -> SocketAddr {
 
 	std::thread::spawn(move || {
 		let rt = tokio::runtime::Runtime::new().unwrap();
-
-		let mut server = rt.block_on(WsServer::new("127.0.0.1:0")).unwrap();
+		let mut server = rt.block_on(WsServerBuilder::default().build("127.0.0.1:0")).unwrap();
 		let mut module = RpcModule::new(());
 		module.register_method("say_hello", |_, _| Ok("hello")).unwrap();
 		server.register_module(module).unwrap();
--- a/tests/tests/helpers.rs
+++ b/tests/tests/helpers.rs
@@ -36,22 +36,45 @@ pub async fn websocket_server_with_subscription() -> SocketAddr {
 		let rt = tokio::runtime::Runtime::new().unwrap();
 
 		let mut server = rt.block_on(WsServer::new("127.0.0.1:0")).unwrap();
-		let mut sub_hello = server.register_subscription("subscribe_hello", "unsubscribe_hello").unwrap();
-		let mut sub_foo = server.register_subscription("subscribe_foo", "unsubscribe_foo").unwrap();
 
 		server.register_method("say_hello", |_| Ok("hello")).unwrap();
 
-		server_started_tx.send(server.local_addr().unwrap()).unwrap();
+		server
+			.register_subscription("subscribe_hello", "unsubscribe_hello", |_, sink| {
+				std::thread::spawn(move || loop {
+					let _ = sink.send(&"hello from subscription");
+					std::thread::sleep(Duration::from_millis(50));
+				});
+				Ok(())
+			})
+			.unwrap();
 
-		rt.spawn(server.start());
+		server
+			.register_subscription("subscribe_foo", "unsubscribe_foo", |_, sink| {
+				std::thread::spawn(move || loop {
+					let _ = sink.send(&1337);
+					std::thread::sleep(Duration::from_millis(100));
+				});
+				Ok(())
+			})
+			.unwrap();
+
+		server
+			.register_subscription("subscribe_add_one", "unsubscribe_add_one", |params, sink| {
+				let mut count: usize = params.one()?;
+				std::thread::spawn(move || loop {
+					count = count.wrapping_add(1);
+					let _ = sink.send(&count);
+					std::thread::sleep(Duration::from_millis(100));
+				});
+				Ok(())
+			})
+			.unwrap();
 
 		rt.block_on(async move {
-			loop {
-				tokio::time::sleep(Duration::from_millis(100)).await;
+			server_started_tx.send(server.local_addr().unwrap()).unwrap();
 
-				sub_hello.send(&"hello from subscription").unwrap();
-				sub_foo.send(&1337_u64).unwrap();
-			}
+			server.start().await
 		});
 	});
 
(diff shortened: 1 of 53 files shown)


tests/tests/integration_tests.rs
```diff
@@ -33,47 +33,57 @@ use jsonrpsee::{
...
-		let next = add_one.next().await.unwrap();
+		let next = add_one.next().with_default_timeout().await.unwrap().unwrap();
 		assert_eq!(next, i);
 	}
 }
 
 #[tokio::test]
 async fn ws_method_call_works() {
-	let server_addr = websocket_server().await;
+	let server_addr = websocket_server().with_default_timeout().await.unwrap();
```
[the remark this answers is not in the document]
niklasad1: sorry, I reverted this will make another PR unrelated this PR.

tests/tests/helpers.rs
```diff
@@ -71,6 +71,16 @@ pub async fn websocket_server_with_subscription() -> SocketAddr {
 			})
 			.unwrap();
 
+		module
+			.register_subscription("subscribe_noop", "unsubscribe_noop", |_, sink, _| {
```
niklasad1, line 75: This in example to show an edge-case.

If the subscription request is answered by the server and then the server decides to close the subscription. Then it's not possible to send a response on some request to cancel the subscription.

I don't think it's a clean solution this case. AFAIU, `jsonrpc` has it too.

Thus, I suggest an ugly way of handling this; just send a `notif response` with some weird type as result then rely on the assumption that the client will not add this type on the `Subscription<T>` then it should fail the client calls `Subscription::next`

utils/src/server/rpc_module.rs
```diff
@@ -258,21 +259,26 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
...
-			let subscribers = self.subscribers.clone();
 			self.methods.methods.insert(
 				unsubscribe_method_name,
-				Box::new(move |id, params, tx, conn| {
+				Box::new(move |id, params, _, conn_id| {
 					let sub_id = params.one()?;
-					subscribers.lock().remove(&(conn, sub_id));
-					send_response(id, tx, "Unsubscribed");
+					if let Some(sink) = subscribers.lock().remove(&(conn_id, sub_id)) {
+						send_response(id, &sink.0, "Unsubscribe");
```
niklasad1, line 280: revert

tests/tests/integration_tests.rs
```diff
@@ -240,3 +241,19 @@ async fn ws_unsubscribe_releases_request_slots() {
 	let _: Subscription<JsonValue> =
 		client.subscribe("subscribe_hello", JsonRpcParams::NoParams, "unsubscribe_hello").await.unwrap();
 }
+
+#[tokio::test]
+async fn server_should_be_able_to_close_subscriptions() {
```
niklasad1, line 246: this will fails; I just added this to show a point

utils/src/server/rpc_module.rs
```diff
@@ -343,16 +348,76 @@ impl SubscriptionSink {
 		self.inner_send(msg).map_err(Into::into)
 	}
 
-	fn inner_send(&self, msg: String) -> Result<(), Error> {
-		if self.is_online() {
-			self.inner.unbounded_send(msg).map_err(|e| Error::Internal(e.into_send_error()))
+	fn inner_send(&mut self, msg: String) -> Result<(), Error> {
+		let res = if let Some(online) = self.keep_alive.as_ref() {
+			if online.is_canceled() {
+				return Err(Error::SubscriptionClosed(None));
```
niklasad1, line 354: this needs to improved.

tests/tests/integration_tests.rs
```diff
@@ -241,3 +242,22 @@ async fn ws_unsubscribe_releases_request_slots() {
...
+	let server_addr = websocket_server_with_subscription().await;
+	let server_url = format!("ws://{}", server_addr);
+
+	let client = WsClientBuilder::default().max_concurrent_requests(1).build(&server_url).await.unwrap();
```
niklasad1, line 251: bad copy-paste

niklasad1, line 251:
```suggestion
	let client = WsClientBuilder::default().build(&server_url).await.unwrap();
```

utils/src/server/rpc_module.rs
```diff
@@ -275,21 +276,25 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
...
-			let subscribers = self.subscribers.clone();
 			self.methods.callbacks.insert(
 				unsubscribe_method_name,
-				MethodCallback::Sync(Box::new(move |id, params, tx, conn| {
+				MethodCallback::Sync(Box::new(move |id, params, tx, conn_id| {
 					let sub_id = params.one()?;
-					subscribers.lock().remove(&(conn, sub_id));
-					send_response(id, tx, "Unsubscribed");
+					subscribers.lock().remove(&(conn_id, sub_id));
+					send_response(id, &tx, "Unsubscribe");
```
niklasad1, line 297:
```suggestion
					send_response(id, &tx, "Unsubscribed");
```

utils/src/server/rpc_module.rs
```diff
@@ -343,19 +348,85 @@ impl SubscriptionSink {
...
 	}
 
-	fn inner_send(&self, msg: String) -> Result<(), Error> {
-		if self.is_online() {
-			self.inner.unbounded_send(msg).map_err(|e| Error::Internal(e.into_send_error()))
+	fn inner_send(&mut self, msg: String) -> Result<(), Error> {
+		let res = if let Some(online) = self.keep_alive.as_ref() {
+			if online.is_canceled() {
+				return Err(subscription_closed_by_client());
+			}
```
dvdplm, line 355:
```suggestion
		let res = if let Some(keep_alive) = self.keep_alive.as_ref() {
			if keep_alive.is_canceled() {
				return Err(subscription_closed_by_client());
			}
```

utils/src/server/rpc_module.rs
```diff
@@ -343,19 +348,85 @@ impl SubscriptionSink {
 		self.inner_send(msg).map_err(Into::into)
 	}
 
-	fn inner_send(&self, msg: String) -> Result<(), Error> {
-		if self.is_online() {
-			self.inner.unbounded_send(msg).map_err(|e| Error::Internal(e.into_send_error()))
+	fn inner_send(&mut self, msg: String) -> Result<(), Error> {
+		let res = if let Some(online) = self.keep_alive.as_ref() {
+			if online.is_canceled() {
+				return Err(subscription_closed_by_client());
```
dvdplm, line 354: This early return here, it means we don't run lines 365-367 to close the `KeepAlive`, why is that?

utils/src/server/rpc_module.rs
```diff
@@ -343,19 +348,83 @@ impl SubscriptionSink {
...
 	}
 
-	fn inner_send(&self, msg: String) -> Result<(), Error> {
-		if self.is_online() {
-			self.inner.unbounded_send(msg).map_err(|e| Error::Internal(e.into_send_error()))
+	fn inner_send(&mut self, msg: String) -> Result<(), Error> {
+		let res = if let Some(keep_alive) = self.keep_alive.as_ref() {
+			if !keep_alive.is_canceled() {
+				// unbounded send only fails if the receiver has been dropped.
+				self.inner.unbounded_send(msg).map_err(|_| subscription_closed_by_client())
```
dvdplm, line 355: 👍

utils/src/server/rpc_module.rs
```diff
@@ -275,21 +276,25 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 						inner: method_sink.clone(),
 						method: subscribe_method_name,
 						sub_id,
-						is_online: online_tx,
+						keep_alive: Some(KeepAlive {
```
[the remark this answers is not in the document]
niklasad1: I think these is a sign of weird abstraction, I will remove this KeepAlive.

utils/src/server/rpc_module.rs
```diff
@@ -33,7 +33,7 @@ pub type MethodSink = mpsc::UnboundedSender<String>;
 
 type Subscribers = Arc<Mutex<FxHashMap<UniqueSubscriptionEntry, (MethodSink, oneshot::Receiver<()>)>>>;
 
-/// Represent a unique subscription entry based on [`SubscriptionID`] and [`ConnectionID`].
+/// Represent a unique subscription entry based on SubscriptionID and ConnectionID.
```
dvdplm, line 36: The rustdoc complaint was caused by a misspelled type name I think, `ID` vs `Id`. So think this should work:

```
Represent a unique subscription entry based on [`SubscriptionId`] and [`ConnectionId`].
```

utils/src/server/rpc_module.rs
```diff
@@ -326,9 +330,13 @@ pub struct SubscriptionSink {
 	/// Method.
 	method: &'static str,
 	/// Subscription ID.
-	sub_id: SubscriptionId,
-	/// Whether the subscription is still connected or not.
-	keep_alive: Option<KeepAlive>,
```
dvdplm: Looks cleaner without this for sure. 👍

utils/src/server/rpc_module.rs
```diff
@@ -31,7 +31,14 @@ pub type SubscriptionId = u64;
 /// Sink that is used to send back the result to the server for a specific method.
 pub type MethodSink = mpsc::UnboundedSender<String>;
 
-type Subscribers = Arc<Mutex<FxHashMap<(ConnectionId, SubscriptionId), (MethodSink, oneshot::Receiver<()>)>>>;
+type Subscribers = Arc<Mutex<FxHashMap<UniqueSubscriptionEntry, (MethodSink, oneshot::Receiver<()>)>>>;
+
+/// Represent a unique subscription entry based on [`SubscriptionID`] and [`ConnectionID`].
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+struct UniqueSubscriptionEntry {
```
dvdplm, line 38: `SubscriptionKey`?

dvdplm (approved):

maciejhirsz (approved):

utils/src/server/rpc_module.rs
niklasad1: thanks changed it

PR #356, opened 2021-06-01
[ws client]: fix propogate error in subscription response

### LINKED ISSUE #349: [types client]: Subscription::next() propogate error when parsing the response fails.
When the wrong type is assigned in the `Subscription<T>`, then it's never propagated to the user as any malformed responses are ignored, [see docs here](https://docs.rs/jsonrpsee-types/0.2.0-alpha.7/jsonrpsee_types/struct.Subscription.html#method.next)

That should be changed to propagate when parsing the response fails, as it's not easy detect why no subscription responses are received unless the logs are read.

niklasad1:
Fixes #349 and removes some boiler plate to unify `Subscription` and `NotificationHandler`.

Renamed it `Notification`, technically it as handle for "subscribing" to either a subscription ID or method name.

### DIFF
--- a/tests/tests/integration_tests.rs
+++ b/tests/tests/integration_tests.rs
@@ -31,7 +31,7 @@ mod helpers;
 use helpers::{http_server, websocket_server, websocket_server_with_subscription};
 use jsonrpsee::{
 	http_client::{traits::Client, Error, HttpClientBuilder},
-	ws_client::{traits::SubscriptionClient, v2::params::JsonRpcParams, JsonValue, Subscription, WsClientBuilder},
+	ws_client::{traits::SubscriptionClient, v2::params::JsonRpcParams, JsonValue, Notification, WsClientBuilder},
 };
 use std::sync::Arc;
 use std::time::Duration;
@@ -41,14 +41,14 @@ async fn ws_subscription_works() {
 	let server_addr = websocket_server_with_subscription().await;
 	let server_url = format!("ws://{}", server_addr);
 	let client = WsClientBuilder::default().build(&server_url).await.unwrap();
-	let mut hello_sub: Subscription<String> =
+	let mut hello_sub: Notification<String> =
 		client.subscribe("subscribe_hello", JsonRpcParams::NoParams, "unsubscribe_hello").await.unwrap();
-	let mut foo_sub: Subscription<u64> =
+	let mut foo_sub: Notification<u64> =
 		client.subscribe("subscribe_foo", JsonRpcParams::NoParams, "unsubscribe_foo").await.unwrap();
 
 	for _ in 0..10 {
-		let hello = hello_sub.next().await.unwrap();
-		let foo = foo_sub.next().await.unwrap();
+		let hello = hello_sub.next().await.unwrap().unwrap();
+		let foo = foo_sub.next().await.unwrap().unwrap();
 		assert_eq!(&hello, "hello from subscription");
 		assert_eq!(foo, 1337);
 	}
@@ -59,11 +59,11 @@ async fn ws_subscription_with_input_works() {
 	let server_addr = websocket_server_with_subscription().await;
 	let server_url = format!("ws://{}", server_addr);
 	let client = WsClientBuilder::default().build(&server_url).await.unwrap();
-	let mut add_one: Subscription<u64> =
+	let mut add_one: Notification<u64> =
 		client.subscribe("subscribe_add_one", vec![1.into()].into(), "unsubscribe_add_one").await.unwrap();
 
 	for i in 2..4 {
-		let next = add_one.next().await.unwrap();
+		let next = add_one.next().await.unwrap().unwrap();
 		assert_eq!(next, i);
 	}
 }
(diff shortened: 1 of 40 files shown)


types/src/traits.rs
```diff
@@ -45,8 +48,10 @@ pub trait SubscriptionClient: Client {
 	where
 		Notif: DeserializeOwned;
 
-	/// Register a NotificationHandler<Notif> that will listen for incoming JSON-RPC notifications
-	async fn register_notification<'a, Notif>(&self, method: &'a str) -> Result<NotificationHandler<Notif>, Error>
+	/// Register a method subscription, the method name is used to only propogate
```
niklasad1, line 51:
```suggestion
	/// Register a method subscription, this is used to filter only server notifications that a user is interested in.
```

tests/tests/integration_tests.rs
```diff
@@ -48,8 +48,8 @@ async fn ws_subscription_works() {
 		client.subscribe("subscribe_foo", JsonRpcParams::NoParams, "unsubscribe_foo").await.unwrap();
 
 	for _ in 0..10 {
-		let hello = hello_sub.next().await.unwrap();
-		let foo = foo_sub.next().await.unwrap();
+		let hello = hello_sub.next().await.unwrap().unwrap();
+		let foo = foo_sub.next().await.unwrap().unwrap();
 		assert_eq!(&hello, "hello from subscription");
 		assert_eq!(foo, 1337);
```
dvdplm, line 54:
```suggestion
		let hello = hello_sub.next().await.unwrap();
		let foo = foo_sub.next().await.unwrap();
		assert_eq!(hello, Some("hello from subscription".into()));
		assert_eq!(foo, Some(1337));
```

Maybe it's easier to read rather than the double `.unwrap()`?

types/src/client.rs
```diff
@@ -109,63 +117,31 @@ where
 	Notif: DeserializeOwned,
 {
 	/// Returns the next notification from the stream
```
dvdplm, line 119:
```suggestion
	/// Returns the next notification from the stream.
```

types/src/client.rs
```diff
@@ -109,63 +117,31 @@ where
...
-}
-
-impl<Notif> NotificationHandler<Notif>
-where
-	Notif: DeserializeOwned,
-{
-	/// Returns the next notification from the stream
-	/// This may return `None` if the method has been unregistered,
+	/// This may return `Ok(None)` if the subscription has been terminated,
 	/// may happen if the channel becomes full or is dropped.
```
dvdplm, line 121:
```suggestion
	/// Returns `Ok(None)` if the subscription has been terminated on the server side, e.g. the server dropped the
	/// connection, crashed, or dropped the connection for whatever reason.
```

I am not sure I understand how the channel being full would cause `None`. Can you help me understand? Wouldn't that mean that the server can't send messages on it but the client could still read from it?

types/src/traits.rs
```diff
@@ -45,8 +48,10 @@ pub trait SubscriptionClient: Client {
 	where
 		Notif: DeserializeOwned;
 
-	/// Register a NotificationHandler<Notif> that will listen for incoming JSON-RPC notifications
-	async fn register_notification<'a, Notif>(&self, method: &'a str) -> Result<NotificationHandler<Notif>, Error>
+	/// Register a method subscription, this is used to filter only server notifications that a user is interested in.
+	///
+	/// The `Notif` param is a generic type to receive generic subscriptions, see [`Subscription`](crate::client::Subscription) for further documentation.
```
dvdplm, line 53:
```suggestion
	/// The `Notif` param is a generic type to receive generic subscriptions, see [`Subscription`] for further documentation.
```

types/src/traits.rs
```diff
@@ -28,11 +28,14 @@ pub trait Client {
...
+	/// a `Subscription ID` that is used to fetch messages on that subscription,
 	///
 	/// The `subscribe_method` and `params` are used to ask for the subscription towards the
 	/// server.
 	///
+	/// The params may be used as input for the subscription for the server to process.
+	///
 	/// The `unsubscribe_method` is used to close the subscription.
 	///
 	/// The `Notif` param is a generic type to receive generic subscriptions, see [`Subscription`](crate::client::Subscription) for further documentation.
```
dvdplm, line 41:
```suggestion
	/// The `Notif` param is a generic type to receive generic subscriptions, see [`Subscription`] for further documentation.
```

dvdplm (approved):

ws-client/src/client.rs
```diff
@@ -455,12 +455,11 @@ impl SubscriptionClient for WsClient {
...
 		};
-		Ok(Subscription { to_back: self.to_back.clone(), notifs_rx, marker: PhantomData, id })
+		Ok(Subscription::new(self.to_back.clone(), notifs_rx, SubscriptionKind::Subscription(id)))
 	}
 
-	/// Register a notification handler for async messages from the server.
-	///
-	async fn register_notification<'a, N>(&self, method: &'a str) -> Result<NotificationHandler<N>, Error>
+	/// Subscribe to a specific method.
+	async fn subscribe_method<'a, N>(&self, method: &'a str) -> Result<Subscription<N>, Error>
```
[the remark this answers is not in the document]
niklasad1: I like `subscribe_to_method`

dvdplm (approved):

PR #357, opened 2021-06-01
Fix clippy lints

popzxc:
While it's nothing critical, it's kinda annoying to see these warnings all the time (I use `cargo clippy` instead of `cargo check`).


niklasad1 (approved):
thanks

dvdplm (commented):
❤️

PR #359, opened 2021-06-01
Misc async methods tinkerings

dvdplm:
I was late to review #327 so here's a PR with suggestions I wanted to make there.﻿

It's a bit noisy mainly because of a bunch of renames:

`Methods` –> `SyncMethods`
`MethodsHandler` –> `Methods`

…and all the fallout from that.

### DIFF
--- a/types/src/v2/params.rs
+++ b/types/src/v2/params.rs
@@ -209,7 +209,7 @@ impl<'a> Id<'a> {
 	}
 }
 
-/// Owned version of [`Id`] that always allocates memory for its members.
+/// Owned version of [`Id`] that allocates memory for the `Number` and `Str` variants.
 #[derive(Debug, PartialEq, Clone, Hash, Eq, Deserialize, Serialize)]
 #[serde(deny_unknown_fields)]
 #[serde(untagged)]
@@ -228,7 +228,7 @@ impl OwnedId {
 		match self {
 			Self::Null => Id::Null,
 			Self::Number(num) => Id::Number(*num),
-			Self::Str(str) => Id::Str(Cow::borrowed(str.as_ref())),
+			Self::Str(str) => Id::Str(Cow::borrowed(str)),
 		}
 	}
 }
--- a/types/src/v2/params.rs
+++ b/types/src/v2/params.rs
@@ -209,7 +209,7 @@ impl<'a> Id<'a> {
 	}
 }
 
-/// Owned version of [`Id`] that allocates memory for the `Number` and `Str` variants.
+/// Owned version of [`Id`] that allocates memory for the `Str` variant.
 #[derive(Debug, PartialEq, Clone, Hash, Eq, Deserialize, Serialize)]
 #[serde(deny_unknown_fields)]
 #[serde(untagged)]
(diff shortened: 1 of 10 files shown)


ws-server/src/server.rs
```diff
@@ -124,40 +124,49 @@ async fn background_task(
...
+	// Note: This handler expects method existence to be checked prior to starting the server and will panic if the
 	// method does not exist.
 	let execute_async = |tx: MethodSink, req: OwnedJsonRpcRequest| {
 		let async_methods = methods.clone();
 		async move {
 			let req = req.borrowed();
-			let method = async_methods.async_method(&*req.method).unwrap();
+			let method = async_methods
+				.async_method(&*req.method)
+				.expect(&format!("async method '{}' is not registered on the server – this is a bug", req.method));
```
popzxc, line 161: It's generally better to use `unwrap_or_else(|| panic!(..))` in such cases.

popzxc (approved):
Totally makes sense.

types/src/v2/params.rs
```diff
@@ -209,7 +209,7 @@ impl<'a> Id<'a> {
 	}
 }
 
-/// Owned version of [`Id`] that always allocates memory for its members.
+/// Owned version of [`Id`] that allocates memory for the `Number` and `Str` variants.
```
niklasad1, line 212: should only allocate for `Str` FWIW

utils/src/server/rpc_module.rs
```diff
@@ -283,11 +283,11 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 	}
 
 	/// Convert a module into methods. Consumes self.
-	pub fn into_methods(self) -> MethodsHolder {
+	pub fn into_methods(self) -> Methods {
 		self.methods
 	}
 
-	/// Merge two [`RpcModule`]'s by adding all [`Method`]s from `other` into `self`.
+	/// Merge two [`RpcModule`]'s by adding all [`SyncMethod`]s from `other` into `self`.
```
niklasad1, line 290:
```suggestion
	/// Merge two [`RpcModule`]'s by adding all [`Method`]s from `other` into `self`.
```

ws-server/src/server.rs
```diff
@@ -124,40 +124,49 @@ async fn background_task(
...
 	//
-	// Note: This handler expects method existence to be checked prior to the call and will panic if
+	// Note: This handler expects method existence to be checked prior to starting the server and will panic if the
 	// method does not exist.
 	let sync_methods = methods.clone();
 	let execute_sync = move |tx: &MethodSink, req: JsonRpcRequest| {
-		let method = sync_methods.method(&*req.method).unwrap();
+		let method = sync_methods
+			.sync_method(&*req.method)
+			.expect(&format!("sync method '{}' is not registered on the server – this is a bug", req.method));
```
niklasad1, line 140: unwrap_or_else(|| panic!("sync method '{}' is not registered on the server – this is a bug", req.method));

types/src/v2/params.rs
niklasad1, line 212:
```suggestion
/// Owned version of [`Id`] that allocates memory for the `Str` variant.
```

niklasad1 (changes requested):
great, some minor things to clean up before merging

see inline comments

maciejhirsz (approved):

niklasad1 (approved):

PR #360, opened 2021-06-02
Enable docs/debug/pub warnings for server and types crates

popzxc:
This PR enables `#![warn(missing_debug_implementations, missing_docs, unreachable_pub)]` for server and types crates similarly to the client ones, and fixes issues uncovered by them.

### DIFF
--- a/types/src/client.rs
+++ b/types/src/client.rs
@@ -6,7 +6,8 @@ use serde::de::DeserializeOwned;
 use serde_json::Value as JsonValue;
 
 /// Active subscription on a Client.
-pub struct Subscription<Notif> {
+#[derive(Debug)]
+pub struct Subscription<Notif: std::fmt::Debug> {
 	/// Channel to send requests to the background task.
 	pub to_back: mpsc::Sender<FrontToBack>,
 	/// Channel from which we receive notifications from the server, as encoded `JsonValue`s.
@@ -18,6 +19,7 @@ pub struct Subscription<Notif> {
 }
 
 /// Active NotificationHandler on a Client.
+#[derive(Debug)]
 pub struct NotificationHandler<Notif> {
 	/// Channel to send requests to the background task.
 	pub to_back: mpsc::Sender<FrontToBack>,
@@ -104,7 +106,7 @@ pub enum FrontToBack {
 
 impl<Notif> Subscription<Notif>
 where
-	Notif: DeserializeOwned,
+	Notif: DeserializeOwned + std::fmt::Debug,
 {
 	/// Returns the next notification from the stream
 	/// This may return `None` if the subscription has been terminated,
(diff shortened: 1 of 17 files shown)


niklasad1 (approved):
:+1:

types/src/client.rs
```diff
@@ -6,7 +6,8 @@ use serde::de::DeserializeOwned;
 use serde_json::Value as JsonValue;
 
 /// Active subscription on a Client.
-pub struct Subscription<Notif> {
+#[derive(Debug)]
+pub struct Subscription<Notif: std::fmt::Debug> {
```
maciejhirsz, line 10: I've mixed feelings about this, I like having strict lints, but this looks really noisy to me and I'm not sure there is much value to it.

Also it doesn't look like the bound is really necessary since it's only applied to `PhantomData`?

niklasad1: > (Sidenote: why is PhantomData pub?)

it because we have no constructor for it, I have fixed it in #356

niklasad1: we could do a manual debug implementation for this instead.

popzxc: Makes sense, replaced it with the manual implementation.

types/src/client.rs
```diff
@@ -17,7 +17,18 @@ pub struct Subscription<Notif> {
...
+		f.debug_struct("Subscription")
+			.field("to_back", &self.to_back)
+			.field("notifs_rx", &self.notifs_rx)
+			.field("id", &self.id)
+			.finish()
+	}
+}
+
 /// Active NotificationHandler on a Client.
+#[derive(Debug)]
```
maciejhirsz, line 31: This one has generic `PhantomData` too. I reckon we should be consistent about either having manual impls for both skipping the marker, or derives on both.

dvdplm (approved):
I don't have strong opinions about mandatory `Debug` impls, I can go either way.

popzxc:
> I don't have strong opinions about mandatory Debug impls

Honestly, throughout the years I've found myself multiple times print-debugging weird things and getting frustrated because some object doesn't implement `Debug`. I think for a library, it's better to always have `Debug`.

utils/src/server/rpc_module.rs
```diff
@@ -52,6 +52,14 @@ pub struct Methods {
 	async_methods: AsyncMethods,
 }
 
+impl std::fmt::Debug for Methods {
+	fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+		// Including types only is sufficient, as it will provide information
+		// about all the stored method alongside with their types.
```
dvdplm, line 58:
```suggestion
		// Including types only is sufficient, as it contains information
		// about the registered methods and if they are sync or async.
```

PR #362, opened 2021-06-02
Use a single map for methods

maciejhirsz:
Instead of having one map with an enum for type, and then a map each for sync and async, we can just use a single map with newtype variants.

This removes one map lookup and atomic clone of the map for sync methods ~(in principle we should be able to get rid of the map lookup for async methods too, but it might require a bit more of a unique container than std `HashMap`)~.

Update: Changed `AsyncMethod` to use `Arc` instead of `Box`, so now we can avoid the extra map lookup on each async call too (still need to clone an `Arc`, but this time it's the method itself instead of the whole map.

Also changed `method_names` to return `Vec<&'static str>`.

### DIFF
--- a/utils/src/server/rpc_module.rs
+++ b/utils/src/server/rpc_module.rs
@@ -35,21 +35,18 @@ pub type MethodSink = mpsc::UnboundedSender<String>;
 
 type Subscribers = Arc<Mutex<FxHashMap<(ConnectionId, SubscriptionId), (MethodSink, oneshot::Receiver<()>)>>>;
 
-/// Identifier of the method type.
-#[derive(Debug, Copy, Clone)]
-pub enum MethodType {
+/// Callback wrapper that can be either sync or async.
+pub enum MethodCallback {
 	/// Synchronous method handler.
-	Sync,
+	Sync(SyncMethod),
 	/// Asynchronous method handler.
-	Async,
+	Async(AsyncMethod),
 }
 
 /// Collection of synchronous and asynchronous methods.
 #[derive(Default)]
 pub struct Methods {
-	method_types: FxHashMap<&'static str, MethodType>,
-	sync_methods: SyncMethods,
-	async_methods: AsyncMethods,
+	callbacks: FxHashMap<&'static str, MethodCallback>,
 }
 
 impl Methods {
@@ -59,7 +56,7 @@ impl Methods {
 	}
 
 	fn verify_method_name(&mut self, name: &str) -> Result<(), Error> {
-		if self.sync_methods.contains_key(name) || self.async_methods.contains_key(name) {
+		if self.callbacks.contains_key(name) {
 			return Err(Error::MethodAlreadyRegistered(name.into()));
 		}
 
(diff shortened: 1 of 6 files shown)


utils/src/server/rpc_module.rs
```diff
@@ -35,21 +37,44 @@ pub type MethodSink = mpsc::UnboundedSender<String>;
...
+				let tx = tx.clone();
+				let params = OwnedRpcParams::from(RpcParams::new(req.params.map(|params| params.get())));
+				let id = OwnedId::from(req.id);
+
+				(callback)(id, params, tx, conn_id).await
+			}
+		};
+
+		if let Err(err) = result {
+			log::error!("execution of sync method call '{}' failed: {:?}, request id={:?}", req.method, err, id);
```
niklasad1, line 68:
```suggestion
			log::error!("execution of method call '{}' failed: {:?}, request id={:?}", req.method, err, id);
```

utils/src/server/rpc_module.rs
```diff
@@ -69,41 +94,25 @@ impl Methods {
...
+	/// Returns the method callback.
+	pub fn method(&self, method_name: &str) -> Option<&MethodCallback> {
+		self.callbacks.get(method_name)
 	}
 
 	/// Returns a `Vec` with all the method names registered on this server.
-	pub fn method_names(&self) -> Vec<String> {
-		self.method_types.keys().map(|name| name.to_string()).collect()
+	pub fn method_names(&self) -> Vec<&'static str> {
+		self.callbacks.keys().map(|name| *name).collect()
```
niklasad1, line 115: is this `deref` needed? does it work with `self.callbacks.keys().copied().collect()?

alternatively we can do do `impl IntoIterator<Item = &'static str>`

niklasad1 (approved):
:1st_place_medal:

utils/src/server/rpc_module.rs
```diff
@@ -35,21 +37,44 @@ pub type MethodSink = mpsc::UnboundedSender<String>;
...
+}
+
+impl MethodCallback {
+	/// Execute the callback, sending the resulting JSON to the specified sink.
+	pub async fn execute(&self, tx: &MethodSink, req: JsonRpcRequest<'_>, conn_id: ConnectionId) {
+		let id = req.id.clone();
+
+		let result = match self {
+			MethodCallback::Sync(callback) => {
+				let params = RpcParams::new(req.params.map(|params| params.get()));
```
dvdplm, line 50: You could hoist this outside the match and then do `let params = OwnedRpcParams::from(params);` on line 60.

utils/src/server/rpc_module.rs
dvdplm:
```suggestion
		self.callbacks.keys().copied().collect()
```
…says clippy.

maciejhirsz: Yeah, was thinking about that, or just `impl Iterator<Item = &'static str>`, I don't know what the lifetime requirements for this are in substrate.

dvdplm (approved):
Good stuff. Left two nits.

PR #364, opened 2021-06-02
Stick code to a single runtime

popzxc:
First, I noticed that in some places, both `tokio` and `async-std` mixed together in tests.

It's generally not a good idea to mix globally accessible runtimes, because it may result in weird things happening.
It may have been the cause for weirdly behaving tests and not working timeouts (e.g. `tokio::timeout` may not work if it will be executed by `async-std` runtime).

Then I noticed that majority of code is using `tokio`, but `ws-client` is based on `async-std`. In my opinion it's weird, because if someone will want to use both `http-client` and `ws-client`, they will have to use two runtimes.

As a result, I removed `async-std` from everywhere, leaving only `tokio`.

Hopefully, it will help with both tests behavior and consistency.

### DIFF
--- a/ws-client/Cargo.toml
+++ b/ws-client/Cargo.toml
@@ -30,4 +30,4 @@ rustls-native-certs = "0.5.0"
 [dev-dependencies]
 env_logger = "0.8"
 jsonrpsee-test-utils = { path = "../test-utils" }
-tokio = { version = "1", features = ["full"] }
+async-std = { version = "1", features = ["attributes"] }
--- a/ws-client/Cargo.toml
+++ b/ws-client/Cargo.toml
@@ -11,8 +11,9 @@ documentation = "https://docs.rs/jsonrpsee-ws-client"
 
 [dependencies]
 async-trait = "0.1"
-async-std = "1"
-async-tls = "0.11"
+tokio = { version = "1", features = ["net", "time"] }
+tokio-rustls = "0.22"
+tokio-util = { version = "0.6", features = ["compat"] }
 fnv = "1"
 futures = { version = "0.3.14", default-features = false, features = ["std"] }
 jsonrpsee-types = { path = "../types", version = "=0.2.0-alpha.7" }
@@ -23,11 +24,9 @@ soketto = "0.5"
 pin-project = "1"
 thiserror = "1"
 url = "2"
-webpki = { version = "0.22", features = ["std"] }
 rustls = "0.19.1"
 rustls-native-certs = "0.5.0"
 
 [dev-dependencies]
 env_logger = "0.8"
 jsonrpsee-test-utils = { path = "../test-utils" }
-async-std = { version = "1", features = ["attributes"] }
--- a/ws-client/Cargo.toml
+++ b/ws-client/Cargo.toml
@@ -11,14 +11,14 @@ documentation = "https://docs.rs/jsonrpsee-ws-client"
 
 [dependencies]
 # Tokio v1 deps
-tokio1 = { package="tokio", version = "1", features = ["net", "time"], optional=true }
-tokio1-rustls = { package="tokio-rustls", version = "0.22", optional=true }
-tokio1-util = { package="tokio-util", version = "0.6", features = ["compat"], optional=true }
+tokioV1 = { package="tokio", version = "1", features = ["net", "time"], optional=true }
+tokioV1-rustls = { package="tokio-rustls", version = "0.22", optional=true }
+tokioV1-util = { package="tokio-util", version = "0.6", features = ["compat"], optional=true }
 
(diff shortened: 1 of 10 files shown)


ws-client/src/client.rs
```diff
@@ -269,7 +269,7 @@ impl<'a> WsClientBuilder<'a> {
 
 		let (sender, receiver) = builder.build().await.map_err(|e| Error::Transport(Box::new(e)))?;
 
-		async_std::task::spawn(async move {
+		tokio::spawn(async move {
```
niklasad1, line 272: the only downside with this is that `tokio` must be initialized for this to work,

niklasad1 (approved):
nice, I have been thinking about fixing this for a while.... so great.

this is mainly because initially this codebase was written in `async-std`  but has been migrated to tokio.

ws-client/src/client.rs
popzxc: It's an `async` function, thus it's unlikely that it will be executed without runtime.

niklasad1:
> Then I noticed that majority of code is using tokio, but ws-client is based on async-std. In my opinion it's weird, because if someone will want to use both http-client and ws-client, they will have to use two runtimes.

good point, but the major thing why this is still async-std is that we have some legacy that is still using tokio 0.1/tokio 0.2 then this worked without having the tokio compat hacks that we have in `http client`. Basically, to support tokio0.2 or tokio1 via feature flags.


Not sure, if we should include this in the 0.2 release.

/cc @dvdplm @maciejhirsz

ws-client/src/client.rs
niklasad1: true, we could add some docs that it must be spawned on `tokio`

dvdplm: Given it changes the current behaviour I think we should document this.

dvdplm (approved):
Not my strong suit to review this sort of code, but I do like the concept and the consistency. I think we need to do this if we want to go the "all method calls are async"-route anyway.
It'd be good to properly document that `tokio` is now the chosen runtime for `jsonrpsee`.

popzxc:
I have added a compatibility layer to support tokio v0.2.
However, integration tests are currently failing with `tokioV02` feature enabled, since the runtime created there is tokio v1 runtime. I'm not sure whether it worth fixing, since v02 *unit* tests do pass, and v1 is the one that is going to be supported long-term.

niklasad1:
> I'm not sure whether it worth fixing, since v02 unit tests do pass, and v1 is the one that is going to be supported long-term.

Indeed, I don't think it's worth fixing it should be deprecated really soon "hopefully"

ws-client/Cargo.toml
```diff
@@ -23,11 +31,14 @@ soketto = "0.5"
...
 rustls-native-certs = "0.5.0"
 
 [dev-dependencies]
 env_logger = "0.8"
 jsonrpsee-test-utils = { path = "../test-utils" }
-tokio = { version = "1", features = ["full"] }
+
+[features]
+default = ["tokioV1"]
+tokioV1 = ["tokio1", "tokio1-rustls", "tokio1-util"]
```
niklasad1, line 43: I don't really have any opinions if the `tokioV1` or `tokio1` is better for naming the features but we should stick to one naming scheme.

Either please change the `http client` feature flags to `tokioVx` or rename these to `tokioX`

ws-client/src/tokio.rs
```diff
@@ -0,0 +1,45 @@
+//! Compatibility layer for supporting both tokio v0.2 and v1.
```
niklasad1, line 1: this was neat.

ws-client/src/tokio.rs
```diff
@@ -0,0 +1,45 @@
...
+
+pub(crate) use tokio_impl::*;
+
+#[cfg(feature = "tokioV1")]
```
niklasad1, line 5: I think we want an explicit compile-error if both features are enabled:

```
#[cfg(all(feature = "tokio1", feature = "tokio02"))]
compile_error!("feature `tokio1` and `tokio02` are mutably exclusive");
```

ws-client/src/tokio.rs
```diff
@@ -1,45 +1,53 @@
 //! Compatibility layer for supporting both tokio v0.2 and v1.
 
+// Check that either v1 or v0.2 feature is enabled.
+#[cfg(not(any(feature = "tokio1", feature = "tokio02")))]
+compile_error!("Either `tokiov1` or `tokiov02` feature must be enabled");
```
niklasad1, line 5:
```suggestion
compile_error!("Either `tokio1` or `tokio02` feature must be enabled");
```

maciejhirsz (approved):

Issue #361, opened 2021-06-02
[ws server]: add builder 


niklasad1:
The `WebSocket server` should have a similar builder pattern [as the HTTP server](https://docs.rs/jsonrpsee-http-server/0.2.0-alpha.7/jsonrpsee_http_server/struct.HttpServerBuilder.html)

Then provide functionality to configure it properly such as `CORS`, `max body size` and so on.

dvdplm:
There is some duplicated code in the ws/http servers atm. I will not address that as part of this ticket, but something to ponder.

niklasad1:
Closed by #365

PR #365, opened 2021-06-02
Add a builder to the ws server

dvdplm:
Add minimal builder. Partially addresses https://github.com/paritytech/jsonrpsee/issues/361 but there are plenty of config options we could add (see e.g. what `parity-ws` [offers](https://github.com/paritytech/ws-rs/blob/master/src/lib.rs#L131)).

In particular this PR does not address the need to handle allowlists for hosts and origins (used in `substrate` e.g. [here](https://github.com/paritytech/substrate/blob/4652f9e00f0e3079b9ed40ff806829f17fd1ddcf/client/rpc-servers/src/lib.rs#L129)), which looks like it needs a change in soketto. When a client handshakes with a `soketto` server we get a [`ClientRequest`](https://github.com/paritytech/soketto/blob/38f5254ad230f2d16ea0d087be527abb21321500/src/handshake/server.rs#L219) which does not contain any header info at all. 

Supporting request filtering based on headers will require a change in `soketto`.

### DIFF
--- a/ws-server/src/server.rs
+++ b/ws-server/src/server.rs
@@ -27,6 +27,7 @@
 use futures_channel::mpsc;
 use futures_util::io::{BufReader, BufWriter};
 use futures_util::stream::StreamExt;
+use jsonrpsee_types::TEN_MB_SIZE_BYTES;
 use soketto::handshake::{server::Response, Server as SokettoServer};
 use std::{net::SocketAddr, sync::Arc};
 use tokio::net::{TcpListener, ToSocketAddrs};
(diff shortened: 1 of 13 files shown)


ws-server/src/server.rs
```diff
@@ -168,3 +172,29 @@ async fn background_task(
 		}
 	}
 }
+
+/// Builder to configure and create a JSON-RPC Websocket server
+#[derive(Debug)]
+pub struct Builder {
```
dvdplm, line 178: We should probably split this into a `Builder` and `Settings` to keep things sane.

maciejhirsz (approved):
LGTM, I think I'd prefer `WsServerBuilder::new("127.0.0.1:0").build()`, but I also realize that's slightly more work for arbitrary reason of keeping the design pattern pure-ish :upside_down_face:.

niklasad1:
> LGTM, I think I'd prefer WsServerBuilder::new("127.0.0.1:0").build(), but I also realize that's slightly more work for arbitrary reason of keeping the design pattern pure-ish upside_down_face.

Right, I think I did like this in the clients because `the lifetime on the str slice` but I don't really remember but `fn new(&str)` should be easy to add as addition to this.

niklasad1 (approved):

ws-server/src/server.rs
```diff
@@ -95,8 +92,9 @@ impl Server {
 
 async fn background_task(
 	socket: tokio::net::TcpStream,
-	methods: Arc<Methods>,
 	conn_id: ConnectionId,
+	methods: Arc<Methods>,
+	cfg: Arc<Builder>,
```
niklasad1, line 97: why is this an `Arc`? 

Builder should be able implement Copy because it's cheap but for future additions maybe?

PR #370, opened 2021-06-07
Concat -> simple push

maciejhirsz:
Was looking into batching out of curiosity to see if we need to change our sinks from using `String` to `Box<RawValue>` (we don't) and just found this small thing that could be better.

### DIFF
--- a/utils/src/server/helpers.rs
+++ b/utils/src/server/helpers.rs
@@ -44,8 +44,8 @@ pub async fn collect_batch_response(rx: mpsc::UnboundedReceiver<String>) -> Stri
 	let mut buf = String::with_capacity(2048);
 	buf.push('[');
 	let mut buf = rx
-		.fold(buf, |mut acc, response| async {
-			acc = [acc, response].concat();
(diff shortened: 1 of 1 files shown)


utils/src/server/helpers.rs
```diff
@@ -44,8 +44,8 @@ pub async fn collect_batch_response(rx: mpsc::UnboundedReceiver<String>) -> Stri
 	let mut buf = String::with_capacity(2048);
 	buf.push('[');
 	let mut buf = rx
-		.fold(buf, |mut acc, response| async {
-			acc = [acc, response].concat();
```
maciejhirsz, line 48: I assumed this would alloc since `Concat` is implemented on `Borrow<str>`, and just checking ASM produced for this, it does indeed alloc on every call here, so the capacity isn't even being used.

niklasad1 (approved):

dvdplm (approved):
Good catch, I wrongly assumed `concat`, being specially built for the purpose, would be faster.

Issue #346, opened 2021-05-27
[ws client] Failed to connect to sockaddr: 127.0.0.1:9944 with err: Rejected { status_code: 403 } 


emostov:
I am trying to use the ws client via substrate's `remote-externalities`. I can connect fine to `wss://rpc.polkadot.io` and if I add `.handshake_url("/public-ws".into())` to the `WsClientBuilder` I can also connect to `wss://polkadot.api.onfinality.io/public-ws`.

But if I try connecting to a local node, `ws://127.0.0.1:9944`, I get the following error:

```bash
zeke substrate % RUST_LOG=debug cargo run --features=try-runtime --  try-runtime offchain-worker \
 --header-at 0x99007ac1e6cbff7efbacff7da6bb1b58480ba091c2f3eccc1648e5d22b92bc42 \
live ws://127.0.0.1:9944
    Running `target/debug/substrate try-runtime offchain-worker --header-at 0x99007ac1e6cbff7efbacff7da6bb1b58480ba091c2f3eccc1648e5d22b92bc42 live 'ws://127.0.0.1:9944'`
2021-05-27 11:03:53.843   INFO main remote-ext: initializing remote client to "ws://127.0.0.1:9944"    
2021-05-27 11:03:53.901  DEBUG main jsonrpsee_ws_client::transport: Failed to connect to sockaddr: 127.0.0.1:9944 with err: Rejected { status_code: 403 }    
Error: Input("failed to build ws client")
```

Exact error comes from here: https://github.com/paritytech/substrate/blob/fe9ced0c6048b45857e8806f8780a9a9a686421e/utils/frame/remote-externalities/src/lib.rs#L439-L443

I think `ws://127.0.0.1:9944`should be valid because I can use it to connect to the polkadot-js apps UI

emostov:
FWIW nmap shows the port open, although not sure if it means much since I am running against localhost

```bash
zeke ~ % nmap -p T:9944 127.0.0.1
Starting Nmap 7.91 ( https://nmap.org ) at 2021-05-27 11:15 PDT
Nmap scan report for localhost (127.0.0.1)
Host is up (0.00017s latency).

PORT     STATE SERVICE
9944/tcp open  unknown

Nmap done: 1 IP address (1 host up) scanned in 0.07 seconds
```

TarikGul:
> although not sure if it means much since I am running against localhost

I good way to test that might be to have nmap within a cloud instance, and then targeting your ip from the outside.

[...]

PR #375, opened 2021-06-10
Cross-origin protection

maciejhirsz:
Depends on https://github.com/paritytech/soketto/pull/35

No filters for `Host`, Substrate doesn't seem to care, it's more of interest to the likes of nginx reverse proxy for setting up virtual hosts. IIRC it was always confusing in `parity-ethereum`?

### DIFF
--- a/ws-server/src/server.rs
+++ b/ws-server/src/server.rs
@@ -250,13 +250,20 @@ impl Builder {
 	}
 
 	/// Set a list of allowet `Origin` headers, connections comming in with a different
-	/// origin will be denied.
+	/// origin will be denied. Values should include protocol.
 	///
-	/// Values should include protocol: `"protocol://hostname"`.
+	/// ```rust
+	/// # let mut builder = jsonrpsee_ws_server::WsServerBuilder::default();
+	/// builder.set_allowed_origins(vec!["https://example.com"]);
+	/// ```
 	///
-	/// By default allows any Origin.
-	pub fn set_allowed_origins(mut self, list: impl AsRef<[String]>) -> Self {
-		self.settings.cors = Cors::AllowList(list.as_ref().into());
+	/// By default allows any `Origin`.
+	pub fn set_allowed_origins<Origin, List>(mut self, list: List) -> Self
+	where
+		List: IntoIterator<Item = Origin>,
+		Origin: Into<String>,
+	{
+		self.settings.cors = Cors::AllowList(list.into_iter().map(Into::into).collect());
 		self
 	}
 
--- a/ws-server/src/server.rs
+++ b/ws-server/src/server.rs
@@ -255,6 +255,8 @@ impl Builder {
 	/// ```
 	///
 	/// By default allows any `Origin`.
+	///
+	/// Will return an error if `list` is empty. Use [`allow_all_origins`](Builder::allow_all_origins) to restore the default.
 	pub fn set_allowed_origins<Origin, List>(mut self, list: List) -> Result<Self, Error>
 	where
 		List: IntoIterator<Item = Origin>,
@@ -269,6 +271,13 @@ impl Builder {
 		self.settings.cors = Cors::AllowList(list);
 
 		Ok(self)
+	}
+
+	/// Restores the default behavior of allowing connections with `Origin` header
+	/// containing any value. This will undo any list set by [`set_allowed_origins`](Builder::set_allowed_origins).
+	pub fn allow_all_origins(mut self) -> Self {
+		self.settings.cors = Cors::AllowAny;
+		self
 	}
 
 	/// Finalize the configuration of the server. Consumes the [`Builder`].
--- a/ws-server/src/server.rs
+++ b/ws-server/src/server.rs
@@ -75,12 +75,14 @@ impl Server {
 	pub async fn start(self) {
 		let mut incoming = TcpListenerStream::new(self.listener);
 		let methods = Arc::new(self.methods);
-		// let cfg = self.cfg;
 		let mut id = 0;
 
 		while let Some(socket) = incoming.next().await {
 			if let Ok(socket) = socket {
-				socket.set_nodelay(true).unwrap_or_else(|e| panic!("Could not set NODELAY on socket: {:?}", e));
+				if let Err(e) = socket.set_nodelay(true) {
+					log::error!("Could not set NODELAY on socket: {:?}", e);
+					continue;
+				}
 
 				if Arc::strong_count(&methods) > self.cfg.max_connections as usize {
 					log::warn!("Too many connections. Try again in a while");
@@ -108,7 +110,7 @@ async fn background_task(
 	let key = {
 		let req = server.receive_request().await?;
 
-		cfg.cors.verify_origin(req.headers().origin).map(|_| req.key())
+		cfg.allowed_origins.verify(req.headers().origin).map(|()| req.key())
 	};
 
 	match key {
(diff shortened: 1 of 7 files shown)


ws-server/src/server.rs
```diff
@@ -179,18 +189,44 @@ async fn background_task(
...
+
+impl Cors {
+	fn verify_origin(&self, origin: Option<&[u8]>) -> Result<(), Error> {
+		if let (Cors::AllowList(list), Some(origin)) = (self, origin) {
+			if !list.iter().any(|o| o.as_bytes() == origin) {
+				let error = format!("Origin denied: {}", String::from_utf8_lossy(origin));
+				log::warn!("{}", error);
+				return Err(Error::Request(error));
+			}
+		}
```
jsdw, line 206: Do we want/need to handle "*" patterns as well in the allowed origins?

ws-server/src/server.rs
```diff
@@ -213,6 +249,24 @@ impl Builder {
 		self
 	}
 
+	/// Set a list of allowet `Origin` headers, connections comming in with a different
```
jsdw, line 252:
```suggestion
	/// Set a list of allowed `Origin` headers, connections coming in with a different
```

ws-server/src/server.rs
```diff
@@ -213,6 +249,24 @@ impl Builder {
...
+	/// origin will be denied. Values should include protocol.
+	///
+	/// ```rust
+	/// # let mut builder = jsonrpsee_ws_server::WsServerBuilder::default();
+	/// builder.set_allowed_origins(vec!["https://example.com"]);
+	/// ```
+	///
+	/// By default allows any `Origin`.
+	pub fn set_allowed_origins<Origin, List>(mut self, list: List) -> Self
```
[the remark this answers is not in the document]
jsdw: I'm ok with the current function myself (I'd just add a note in the comment that setting an empty list will deny all origins). Perhaps if we added an `allow_all_origins` function to toggle to `AllowAny` it would become a little clearer by contrast? This would also allow one to programmatically undo the effect of calling `set_allowed_origins`, which doesn't feel super necessary, but for the sake of completeness it feels nice to be able to set the full range of possible values.

ws-server/src/server.rs
niklasad1: yeah, I think so

ws-server/src/server.rs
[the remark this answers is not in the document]
jsdw: (To elaborate a little; if we had an `add_allowed_origins` instead that pushed to the list, it would feel slightly odd to me that you could add an origin to go from allowing everything (nothing added yet) to only allowing what's been added. Whereas, "set_allowed_origins" feels clear to me and doesn't have that slightly odd transition in my mind)

ws-server/src/server.rs
dvdplm: Yes we do.

ws-server/src/server.rs
dvdplm: nitpick: to me "header" means the full `key: value` string, whereas here we are not setting the **header** but rather the set of hosts (with wildcards) that will be checked against the `Origin` headers sent in the client handshake.
So perhaps something like: "Set the list of allowed origins. During the handshake the `Origin` header of incoming connection requests will be checked against this list. Values should include the protocol. Wildcards are allowed, as per spec [here](link-to-spec)."

ws-server/src/server.rs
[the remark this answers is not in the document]
dvdplm: My two cents: having a single `set_` method is enough. We do not expect the list to change once the server is started, and even if it is technically possible to dynamically add/remove allowed origins, I think it's fine to be opinionated here.

Wrt to the semantics of setting the allowed origins to the empty set, I think we should pick one of these:
1. empty list => everyone can connect, i.e. switch to `Cors::AllowAny`
2. empty list => return error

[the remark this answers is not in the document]
jsdw: I'd be more in favour of 2: I wouldn't be super keen on an empty list (perhaps by mistake) loosening restrictions and allowing all origins, so getting an error back feels more helpful.

ws-server/src/server.rs
maciejhirsz: There are no wildcards in the `Origin` headers (you can set one in `Access-Control-Allow-Origin`, but that doesn't apply to `WebSocket`. The only special `Origin` value is `null`, which the end user might pass as a string here, I don't think we need to build extra support for that here, and from what I've read the general wisdom is that you should never allow connections with `null` `Origin`, unless you are fine accepting it from any `Origin`.

maciejhirsz:
Updated the comment on `set_allowed_origins`, and made it return an error if the list is empty. It might still be a good idea to include `allow_all_origins()` method on the builder that restores the default, I can imagine that during the setup on runtime someone might want to set their default allowed origins, and then override them to allow all with a CLI setting.

ws-server/src/server.rs
maciejhirsz: There are 3 possible states here:

1. There is no `Origin` header, which equates to the request being done either on the same domain (no cross-origin shenanigans), or the request isn't coming from a browser at all (in which case the client can spoof the header to whatever it wants). We always allow these.
2. The header has a `protocol://hostname` value, if so we check it against the list (unless we allow all origins).
3. The header has a `null` value, which can be explicitly allowed by the list, but is generally advised against.

There is no `*` origin.

jsdw: I think you explained this in a previous comment which makes sense to me now :)

My initial thinking was that CORS responses can contain "*" in allowed origins, so perhaps we need to handle something like this.

I did a little reading about WebSockets (since I didn't really know anything about the WebSocket upgrade protocol) and, indeed, there is no such thing as CORS really when talking about WS connections by the sounds of it, and so the origin checking here can essentially take whatever form we like.

Given that, I have no problem with not handling "*" in an allowed origin; it could potentially be a future enhancement (so that you can say configure this to allow eg any connections from "https://*.mydomain.com") but offhand I don't see why it needs to exist now.

ws-server/src/server.rs
```diff
@@ -75,7 +75,7 @@ impl Server {
 	pub async fn start(self) {
 		let mut incoming = TcpListenerStream::new(self.listener);
 		let methods = Arc::new(self.methods);
-		let cfg = self.cfg;
+		// let cfg = self.cfg;
```
jsdw, line 78: Nit: remove this commented out line?

jsdw (approved):
Looking good to me; thanks for explaining the Origins stuff!

A couple of nits, but happy to approve :)

niklasad1:
> No filters for Host, Substrate doesn't seem to care, it's more of interest to the likes of nginx reverse proxy for setting up virtual hosts. IIRC it was always confusing in parity-ethereum?

hmm, I don't really follow are referring to [this](https://github.com/paritytech/substrate/blob/master/client/rpc-servers/src/lib.rs#L151-#L161) in substrate?

when I did [this](https://github.com/paritytech/jsonrpsee/issues/346#issuecomment-849964203) the request got denied for bad host in the `host header` how did that happen without host filtering?

ws-server/src/server.rs
```diff
@@ -179,18 +189,40 @@ async fn background_task(
 	}
 }
 
+#[derive(Debug, Clone)]
+enum Cors {
```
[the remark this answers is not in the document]
maciejhirsz: Happy to change it. I think I named it `Cors` since that's what the analog enum is already called in substrate, but no reason for us to confuse things, plus abbreviations are confusing almost by definition.

maciejhirsz:
> > No filters for Host, Substrate doesn't seem to care, it's more of interest to the likes of nginx reverse proxy for setting up virtual hosts. IIRC it was always confusing in parity-ethereum?
> 
> hmm, I don't really follow are you referring to [this](https://github.com/paritytech/substrate/blob/master/client/rpc-servers/src/lib.rs#L151-#L161) in substrate?
> 
> when I did [this](https://github.com/paritytech/jsonrpsee/issues/346#issuecomment-849964203) the request got denied for bad host in the `host header` how did that happen without host filtering?

Ha, curious. I didn't see it in the CLI when I checked it, but you might be right. I'll double check to make sure.

niklasad1 (approved):

maciejhirsz:
Will do a follow-up PR with `Host` filtering.

PR #383, opened 2021-06-15
Method aliases + RpcModule: Clone

### LINKED ISSUE #328: [rpc module]: possibility to re-use modules on different servers.
The `RpcModule` doesn't implement `Clone` and ideally we should some way to construct it once and re-use it to register on several servers such as running the servers on different ports or something similar.

```rust
let module = RpcModule::new();
ws_server.register_module(module.clone());
http_server.register_module(module.clone());
```

Best way is probably to provide some builder that merges the modules when `server.start` is called.

### LINKED ISSUE #378: [rpc module]: support aliases for registered methods and subscriptions
Feature request to support alias for a registered method or subscription.

Currently, users has to duplicate these for it to work properly.

maciejhirsz:
+ Added `register_alias` (closes #378).
+ `RpcModule` and `Methods` both now implement `Clone` (closes #328).
+ `Methods` is now using an opaque internal `Arc`, `ws_server` no longer wraps `Methods` in `Arc`.

### DIFF
--- a/utils/src/server/rpc_module.rs
+++ b/utils/src/server/rpc_module.rs
@@ -83,9 +83,9 @@ impl Debug for MethodCallback {
 }
 
 /// Collection of synchronous and asynchronous methods.
-#[derive(Default, Debug)]
+#[derive(Default, Debug, Clone)]
 pub struct Methods {
-	callbacks: FxHashMap<&'static str, MethodCallback>,
+	callbacks: Arc<FxHashMap<&'static str, MethodCallback>>,
 }
 
 impl Methods {
@@ -102,15 +102,22 @@ impl Methods {
 		Ok(())
 	}
 
+	/// Helper for obtaining a mut ref to the callbacks HashMap.
+	fn mut_callbacks(&mut self) -> &mut FxHashMap<&'static str, MethodCallback> {
+		Arc::make_mut(&mut self.callbacks)
+	}
+
 	/// Merge two [`Methods`]'s by adding all [`MethodCallback`]s from `other` into `self`.
 	/// Fails if any of the methods in `other` is present already.
-	pub fn merge(&mut self, other: Methods) -> Result<(), Error> {
+	pub fn merge(&mut self, mut other: Methods) -> Result<(), Error> {
 		for name in other.callbacks.keys() {
 			self.verify_method_name(name)?;
 		}
 
-		for (name, callback) in other.callbacks {
-			self.callbacks.insert(name, callback);
+		let callbacks = self.mut_callbacks();
+
+		for (name, callback) in other.mut_callbacks().drain() {
+			callbacks.insert(name, callback);
 		}
 
 		Ok(())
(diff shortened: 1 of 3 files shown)


maciejhirsz:
If we change the design a bit so that the server uses `Methods` instead of `RpcModule`, we can remove `Clone` from `RpcModule` and get rid of the whole `Arc::make_mut` I reckon.

utils/src/server/rpc_module.rs
```diff
@@ -132,22 +140,43 @@ impl Methods {
 	pub fn method_names(&self) -> Vec<&'static str> {
 		self.callbacks.keys().copied().collect()
 	}
+
+	/// Get the reference count this instance of `Methods`.
```
jsdw, line 144:
```suggestion
	/// Get the reference count for this instance of `Methods`.
```

jsdw (approved):
Looks good to me! 

I hadn't used `Arc::make_mut` before; that's a handy function to know about!

utils/src/server/rpc_module.rs
```diff
@@ -101,15 +102,22 @@ impl Methods {
 		Ok(())
 	}
 
+	/// Helper for obtaining a mut ref to the callbacks HashMap.
+	fn mut_callbacks(&mut self) -> &mut FxHashMap<&'static str, MethodCallback> {
+		Arc::make_mut(&mut self.callbacks)
```
niklasad1, line 107: neat TIL.

utils/src/server/rpc_module.rs
```diff
@@ -308,15 +337,16 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 		Ok(())
 	}
 
-	/// Convert a module into methods. Consumes self.
-	pub fn into_methods(self) -> Methods {
-		self.methods
-	}
+	/// Register an `alias` name for an `existing_method`.
```
niklasad1, line 340: :+1:

niklasad1:
> If we change the design a bit so that the server uses Methods instead of RpcModule, we can remove Clone from RpcModule and get rid of the whole Arc::make_mut I reckon. What do you think @niklasad1?

I like it as long as we don't have to provide duplicated methods on it as before `i.e, register_method, register_subscriptions and so on`

ws-server/src/server.rs
```diff
@@ -74,15 +74,15 @@ impl Server {
...
+		let methods = self.methods;
 		let cfg = self.cfg;
 		let mut id = 0;
 
 		while let Some(socket) = incoming.next().await {
 			if let Ok(socket) = socket {
 				socket.set_nodelay(true).unwrap_or_else(|e| panic!("Could not set NODELAY on socket: {:?}", e));
 
-				if Arc::strong_count(&methods) > self.cfg.max_connections as usize {
+				if methods.ref_count() > self.cfg.max_connections as usize {
```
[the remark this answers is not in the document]
maciejhirsz: Yeah, good point, will probably need to track this separately (maybe spawn a dummy `Arc<()>`?). I'm also happy about removing `ref_count` from the API since that's a bleeding abstraction.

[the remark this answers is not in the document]
niklasad1: yeah, we could have a separate `AtomicCounter` in the WsServer to keep track of the number of concurrent connections

niklasad1 (changes requested):
LGTM modulo the `Arc::ref_count` comment

niklasad1 (approved):

PR #385, opened 2021-06-16
Use criterion's async bencher

popzxc:
Doesn't change much in terms of both diff and bench results, but IMHO makes the code a bit more clear (by removing the necessity of `block_on` in every single bench).

Note that we still need to pre-create runtime and manually pass it everywhere, since we need the server to be spawned outside of the bench and run in the same runtime.

Also, I rewrote `run_concurrent_round_trip` in a functional style to avoid allocating `Vec` inside of the bench body.

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -134,10 +134,8 @@ impl RequestBencher for AsyncBencher {
 
 fn run_round_trip(rt: &TokioRuntime, crit: &mut Criterion, client: Arc<impl Client>, name: &str, request: RequestType) {
 	crit.bench_function(&request.group_name(name), |b| {
-		b.iter(|| {
-			rt.block_on(async {
-				black_box(client.request::<String>(request.method_name(), JsonRpcParams::NoParams).await.unwrap());
-			})
+		b.to_async(rt).iter(|| async {
+			black_box(client.request::<String>(request.method_name(), JsonRpcParams::NoParams).await.unwrap());
 		})
 	});
 }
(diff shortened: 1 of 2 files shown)


benches/bench.rs
```diff
@@ -134,40 +135,37 @@ impl RequestBencher for AsyncBencher {
...
+		b.to_async(rt).iter_with_setup(
 			|| {
-				rt.block_on(async {
-					client
-						.subscribe::<String>(SUB_METHOD_NAME, JsonRpcParams::NoParams, UNSUB_METHOD_NAME)
-						.await
-						.unwrap()
+				// We have to use `block_in_place` here since `b.to_async(rt)` automatically enters the
+				// runtime context and simply calling `block_on` here will cause the code to panic.
+				tokio::task::block_in_place(|| {
```
niklasad1, line 158: hmm, I don't understand this.

Why doesn't this work:

```rust
async {
   client.subscribe::<String>(SUB_METHOD_NAME, JsonRpcParams::NoParams, UNSUB_METHOD_NAME)
      .await
	  .unwrap()
}
```

benches/bench.rs
```diff
@@ -224,21 +222,19 @@ fn run_concurrent_round_trip<C: 'static + Client + Send + Sync>(
...
+							);
+						})
 					});
-					tasks.push(task);
-				}
-				for task in tasks {
-					rt.block_on(task).unwrap();
-				}
-			})
+					join_all(tasks).await;
```
niklasad1, line 235: :+1:

benches/bench.rs
popzxc: Because for some reason in `iter_with_setup` the first closure [must be sync](https://github.com/bheisler/criterion.rs/blob/master/src/bencher.rs#L517), so the suggested way will just result in a future being returned (but not polled).

niklasad1 (approved):

benches/bench.rs
niklasad1: ok, that makes sense.

maciejhirsz (approved):

Issue #377, opened 2021-06-15
[servers]: API to stop the servers


niklasad1:
Feature request to provide functionality to shutdown the servers.

For `http` hyper has a feature to do this but for the websocket server it might require a little but more code.

PR #386, opened 2021-06-16
Add a way to stop servers

popzxc:
I didn't dig into stopping all the active subscriptions, since I'm not sure whether it must be a part of the public interface (it's not a trivial matter, since the interface provides flexibility to implement subscription maintenance in any way you want, and doing something without proper notification to user may result in a bad DevEx).

### DIFF
--- a/http-server/src/server.rs
+++ b/http-server/src/server.rs
@@ -115,7 +115,7 @@ impl Default for Builder {
 	}
 }
 
-/// Handle that is able to stop the running server.
+/// Handle used to stop the running server.
 #[derive(Debug, Clone)]
 pub struct StopHandle {
 	stop_sender: mpsc::Sender<()>,
--- a/http-server/src/server.rs
+++ b/http-server/src/server.rs
@@ -26,7 +26,7 @@
 
 use crate::{response, AccessControl, TEN_MB_SIZE_BYTES};
 use futures_channel::mpsc;
-use futures_util::stream::StreamExt;
+use futures_util::{lock::Mutex, stream::StreamExt, SinkExt};
 use hyper::{
 	server::{conn::AddrIncoming, Builder as HyperBuilder},
 	service::{make_service_fn, service_fn},
@@ -95,12 +95,16 @@ impl Builder {
 		let local_addr = listener.local_addr().ok();
 
 		let listener = hyper::Server::from_tcp(listener)?;
+
+		let stop_pair = mpsc::channel(1);
 		Ok(Server {
 			listener,
 			local_addr,
 			methods: Methods::default(),
 			access_control: self.access_control,
 			max_request_body_size: self.max_request_body_size,
+			stop_pair,
+			stop_handle: Arc::new(Mutex::new(())),
 		})
 	}
 }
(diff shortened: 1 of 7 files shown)


popzxc:
Related to https://github.com/paritytech/jsonrpsee/issues/377 (not sure whether resolves or not)

niklasad1:
> Related to #377 (not sure whether resolves or not)

Cool,
Indeed, this PR closes #377 and is what we want.

I will take a look at later today or tomorrow.

ws-server/src/server.rs
```diff
@@ -71,26 +82,48 @@ impl Server {
...
+							}
+							let methods = methods.clone();
 
-				tokio::spawn(background_task(socket, id, methods, cfg));
+							tokio::spawn(background_task(socket, id, methods, cfg));
 
-				id += 1;
+							id += 1;
+					} else {
+						break;
```
niklasad1, line 118: hmm, we shouldn't kill the entire server because the `incoming.next()` failed? 

Can you explain this?

http-server/src/server.rs
```diff
@@ -124,6 +147,10 @@ pub struct Server {
 	max_request_body_size: u32,
 	/// Access control
 	access_control: AccessControl,
+	/// Pair of channels to stop the server.
+	stop_pair: (mpsc::Sender<()>, mpsc::Receiver<()>),
+	/// Stop handle that indicates whether server has been stopped.
+	stop_handle: Arc<Mutex<()>>,
```
niklasad1, line 153: Is this mutex really necessary, can't we just call `Sender::is_connected` instead?

As I understand it, the server keeps the `receiver` and when it's closed it's dropped? Thus, then `Sender::is_connected` will fail? Or am I missing something?

niklasad1 (approved):
LGTM,

For `WebSockets` the connection will be reset and that should be sufficient I think, then for `subscriptions` we will send a message when each subscription is closed but I guess we could test it with our `WebSocket client` to be sure how it's handled when the server is closed.

For HTTP then `hyper` will take care of that for us and I suppose all pending requests/futures then are "waited" to completed then it's terminated (no new requests are accepted)?

http-server/src/server.rs
popzxc: `is_connected` is not async, unfortunately. Thus it'd be a burden on the user's shoulders to wait for it, and most likely will result in timeout-based polling, which is not the nicest thing in the world.
From my experience, using mutexes as semaphores is kinda normal practice, is there something wrong with it?

popzxc:
> I suppose all pending requests/futures are completed then it's terminated (no new requests are accepted)

Yep, that's my understanding as well.

http-server/src/server.rs
niklasad1: Alright, it makes sense to have an async API for this.

> From my experience, using mutexes as semaphores is kinda normal practice, is there something wrong with it?

It's fine it's just that `Sender` is kind of already a "semaphore" (it's a built on top of AtomicUsize)

http-server/src/tests.rs
```diff
@@ -308,3 +314,23 @@ async fn can_register_modules() {
...
+	// First `unwrap` is timeout, second is `JoinHandle`'s one, third is the server future result.
+	join_handle
+		.with_default_timeout()
+		.await
+		.expect("Timeout")
+		.expect("Join error")
+		.expect("Server stopped with an error");
+
+	// After server was stopped, attempt to stop it again should result in an error.
+	assert!(stop_handle.stop().with_default_timeout().await.unwrap().is_err());
```
dvdplm, line 335: Can we assert that we actually get the `AlreadyStopped` error here? I get that it is the only possible error, but if the code is ever changed in the future this may not be true anymore.

ws-server/src/server.rs
```diff
@@ -50,6 +56,11 @@ pub struct Server {
 	methods: Methods,
 	listener: TcpListener,
 	cfg: Settings,
+
```
dvdplm, line 59:
```suggestion
```

dvdplm (approved):
There's a bit of duplication happening now between ws and http servers. Not something we need to tackle here ofc, but it'd be good to start thinking about what a better way would be.

niklasad1 (changes requested):
sorry I didn't realize that  `StopHandle` was not exported when I reviewed the PR initially because `mod server` is private thus those must be exported to be usable.

Follow-up for us to investigate is if it makes sense to cancel the background task for that already started connection too.

niklasad1 (approved):

PR #387, opened 2021-06-17
New proc macro

popzxc:
One macro to rule them all 😎 

There is a vast more to be done; this is very far from being completed.
However, I'm creating this PR just in case I'm moving in a direction quite orthogonal from what you guys expect.

As for now, you can get a glimpse of how it'll look in the [main macro doc-comment](https://github.com/paritytech/jsonrpsee/blob/e319a7ad96d97cfcc19e28cd5af49e4db394065a/proc-macros/src/lib.rs#L115).
Feedback will be appreciated.

### DIFF
--- a/proc-macros/src/lib.rs
+++ b/proc-macros/src/lib.rs
@@ -174,7 +174,7 @@ pub fn rpc_client_api(input_token_stream: TokenStream) -> TokenStream {
 ///     fn sync_method(&self) -> String;
 ///
 ///     // Note that `subscription_sink` was added automatically.
-///     fn sub(&self, subscription_sink: SubscriptionSing);
+///     fn sub(&self, subscription_sink: SubscriptionSink);
 ///
 ///     fn into_rpc(self) -> Result<Self, jsonrpsee::types::error::Error> {
 ///         // Actual implementation stripped, but inside we will create
@@ -198,7 +198,7 @@ pub fn rpc_client_api(input_token_stream: TokenStream) -> TokenStream {
 ///     }
 ///     
 ///     // Subscription method returns `Subscription` object in case of success.
-///     async fn sub(&self, subscription_sink: SubscriptionSing) -> Result<Subscription<String>, Error> {
+///     async fn sub(&self, subscription_sink: SubscriptionSink) -> Result<Subscription<String>, Error> {
 ///         // ...
 ///     }
 /// }
--- a/proc-macros/src/lib.rs
+++ b/proc-macros/src/lib.rs
@@ -114,39 +114,38 @@ pub fn rpc_client_api(input_token_stream: TokenStream) -> TokenStream {
 
 /// Main RPC macro. TODO: Add docs
 #[proc_macro_attribute]
-pub fn rpc(_attr: TokenStream, item: TokenStream) -> TokenStream {
-	match rpc_impl(item) {
+pub fn rpc(attr: TokenStream, item: TokenStream) -> TokenStream {
+	let attr = proc_macro2::TokenStream::from(attr);
+
+	let rebuilt_rpc_attribute = syn::Attribute {
+		pound_token: syn::token::Pound::default(),
+		style: syn::AttrStyle::Outer,
+		bracket_token: syn::token::Bracket::default(),
+		path: syn::Ident::new("rpc", proc_macro2::Span::call_site()).into(),
+		tokens: quote! { (#attr) },
+	};
+
+	match rpc_impl(rebuilt_rpc_attribute, item) {
 		Ok(tokens) => tokens,
 		Err(err) => err.to_compile_error(),
 	}
 	.into()
 }
 
+#[derive(Debug)]
+struct RpcAttributes {
+	attrs: Vec<syn::Attribute>,
+}
+
+impl syn::parse::Parse for RpcAttributes {
+	fn parse(input: syn::parse::ParseStream) -> Result<Self, syn::Error> {
+		Ok(Self { attrs: input.call(syn::Attribute::parse_outer)? })
+	}
+}
+
 /// Convenience form of `rpc` that may use `?` for error handling to avoid boilerplate.
-fn rpc_impl(item: TokenStream) -> Result<proc_macro2::TokenStream, syn::Error> {
+fn rpc_impl(attr: syn::Attribute, item: TokenStream) -> Result<proc_macro2::TokenStream, syn::Error> {
 	let trait_data: syn::ItemTrait = syn::parse(item)?;
-	let rpc = RpcDescription::from_item(trait_data)?;
+	let rpc = RpcDescription::from_item(attr, trait_data)?;
 	rpc.render()
-}
-
-/// Marker for a method in the RPC trait definition.
-#[proc_macro_attribute]
-pub fn method(_attr: TokenStream, item: TokenStream) -> TokenStream {
-	// We don't modify the input stream, since this attribute only
-	// provides additional metadata for `rpc` attribute.
-	//
-	// This however should be a `proc_macro_attribute`, so rust compiler won't complain about
-	// unknown attribute.
-	item
-}
-
-/// Marker for a subscription in the RPC trait definition.
-#[proc_macro_attribute]
-pub fn subscription(_attr: TokenStream, item: TokenStream) -> TokenStream {
-	// We don't modify the input stream, since this attribute only
-	// provides additional metadata for `rpc` attribute.
-	//
-	// This however should be a `proc_macro_attribute`, so rust compiler won't complain about
-	// unknown attribute.
-	item
 }
--- a/proc-macros/src/lib.rs
+++ b/proc-macros/src/lib.rs
@@ -208,7 +208,158 @@ pub fn rpc_client_api(input_token_stream: TokenStream) -> TokenStream {
 ///
 /// ## Attributes
 ///
-/// To be done...
+/// ### `rpc` attribute
+///
+/// `rpc` attribute is applied to a trait in order to turn it into an RPC implementation.
+///
+/// **Arguments:**
+///
+/// - `server`: generate `<Trait>Server` trait for the server implementation.
+/// - `client`: generate `<Trait>Client` extension trait that makes RPC clients to invoke a concrete RPC
+///   implementation methods conveniently.
+/// - `namespace`: add a prefix to all the methods and subscriptions in this RPC. For example, with namespace
+///   `foo` and method `spam`, the resulting method name will be `foo_spam`.
+///
+/// **Trait requirements:**
+///
+/// Trait wrapped with an `rfc` attribute **must not**:
+///
+/// - have associated types or constants;
+/// - have Rust methods not marked with either `method` or `subscription` attribute;
+/// - be empty.
+///
+/// At least one of the `server` or `client` flags must be provided, otherwise the compilation will err.
+///
+/// ### `method` attribute
+///
+/// `method` attribute is used to define an RPC method.
+///
+/// **Arguments:**
+///
+/// - `name` (mandatory): name of the RPC method. Does not have to be the same as the Rust method name.
+///
+/// **Method requirements:**
+///
+/// Rust method marked with `method` attribute, **may**:
+///
+/// - be either `async` or not;
+/// - have input parameters or not;
+/// - have return value or not (in the latter case, it will be considered a notification method).
+///
+/// ### `subscription` attribute
+///
+/// **Arguments:**
+///
+/// - `name` (mandatory): name of the RPC method. Does not have to be the same as the Rust method name.
+/// - `unsub` (mandatory): name of the RPC method to unsubscribe from the subscription. Must not be the same as `name`.
+/// - `item` (mandatory): type of items yielded by the subscription. Note that it must be the type, not string.
+///
+/// **Method requirements:**
+///
+/// Rust method marked with `subscription` attribute **must**:
+///
+/// - be synchronous;
+/// - not have return value.
+///
+/// Rust method marked with `subscription` attribute **may**:
+///
+/// - have input parameters or not.
+///
+/// ## Full workflow example
+///
+/// ```rust
+/// //! Example of using proc macro to generate working client and server.
+///
+/// use std::net::SocketAddr;
+///
+/// use futures_channel::oneshot;
+/// use jsonrpsee::{ws_client::*, ws_server::WsServerBuilder};
+///
+/// // RPC is moved into a separate module to clearly show names of generated entities.
+/// mod rpc_impl {
+///     use jsonrpsee::{proc_macros::rpc, types::async_trait, ws_server::SubscriptionSink};
+///
+///     // Generate both server and client implementations, prepend all the methods with `foo_` prefix.
+///     #[rpc(client, server, namespace = "foo")]
+///     pub trait Rpc {
+///         #[method(name = "foo")]
+///         async fn async_method(&self, param_a: u8, param_b: String) -> u16;
+///
+///         #[method(name = "bar")]
+///         fn sync_method(&self) -> u16;
+///
+///         #[subscription(name = "sub", unsub = "unsub", item = String)]
+///         fn sub(&self);
+///     }
+///
+///     // Structure that will implement `RpcServer` trait.
+///     // In can have fields, if required, as long as it's still `Send + Sync + 'static`.
+///     pub struct RpcServerImpl;
+///
+///     #[async_trait]
+///     impl RpcServer for RpcServerImpl {
+///         async fn async_method(&self, _param_a: u8, _param_b: String) -> u16 {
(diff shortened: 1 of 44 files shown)


niklasad1 (commented):
awesome but I don't have time to look closer at it today probably some time next week

proc-macros/src/lib.rs
```diff
@@ -111,212 +110,127 @@ pub fn rpc_client_api(input_token_stream: TokenStream) -> TokenStream {
...
+///
+/// ```ignore
+/// #[async_trait]
+/// pub trait RpcServer {
+///     // RPC methods are usual methods and can be either sync or async.
+///     async fn async_method(&self, param_a: u8, param_b: String) -> u16;
+///     fn sync_method(&self) -> String;
+///
+///     // Note that `subscription_sink` was added automatically.
+///     fn sub(&self, subscription_sink: SubscriptionSing);
```
dvdplm, line 177:
```suggestion
///     fn sub(&self, subscription_sink: SubscriptionSink);
```

proc-macros/src/lib.rs
```diff
@@ -111,212 +110,127 @@ pub fn rpc_client_api(input_token_stream: TokenStream) -> TokenStream {
...
+///     async fn async_method(&self, param_a: u8, param_b: String) -> Result<u16, Error> {
+///         // Actual implementations are stripped, but inside a corresponding `Client` or
+///         // `SubscriptionClient` method is called.
+///     }
+///     async fn sync_method(&self) -> Result<String, Error> {
+///         // ...
+///     }
+///     
+///     // Subscription method returns `Subscription` object in case of success.
+///     async fn sub(&self, subscription_sink: SubscriptionSing) -> Result<Subscription<String>, Error> {
```
dvdplm, line 201:
```suggestion
///     async fn sub(&self, subscription_sink: SubscriptionSink) -> Result<Subscription<String>, Error> {
```

dvdplm (commented):
This looks very good. I'll be back with a proper review in the coming days, it'll take me a while to go through it all.

popzxc:
I've added an [integration test](https://github.com/paritytech/jsonrpsee/blob/1667d277bcfaf48375cb10f225ced73b850df246/tests/tests/new_proc_macros.rs) that (after fixing a few bugs) even seems to be working.
Also, a set of UI tests has been implemented.

I guess, at this point, the *skeleton* of the macro is pretty well-formed, and basic functionality works.

PR is already pretty big, and I understand that it may be hard to review it, so the question is:

Should I freeze this PR as-is, resolve issues, get this merged, and only then start adding new functionality; or would it be better to implement everything first to review the full picture?

niklasad1:
> Should I freeze this PR as-is, resolve issues, get this merged, and only then start adding new functionality; or would it be better to implement everything first to review the full picture?

Let's freeze it for now and we will need to look into it closer and provide feedback (hopefully soon) such that you don't keep working on this and we decide to take another route. In general it looks great so far.

I would like to have @maciejhirsz thoughts on this because he had some ideas for the macros :)

popzxc:
> Let's freeze it for now 

Yeah, OK. I've finished it in a form so it can get merged (e.g. finished the documentation), but will not add anything else unless directly requested.

Looking forward to seeing feedback.

proc-macros/src/lib.rs
```diff
@@ -111,212 +110,278 @@ pub fn rpc_client_api(input_token_stream: TokenStream) -> TokenStream {
...
+///
+/// - `server`: generate `<Trait>Server` trait for the server implementation.
+/// - `client`: generate `<Trait>Client` extension trait that makes RPC clients to invoke a concrete RPC
+///   implementation methods conveniently.
+/// - `namespace`: add a prefix to all the methods and subscriptions in this RPC. For example, with namespace
+///   `foo` and method `spam`, the resulting method name will be `foo_spam`.
+///
+/// **Trait requirements:**
+///
+/// Trait wrapped with an `rfc` attribute **must not**:
```
niklasad1, line 225:
```suggestion
/// Trait wrapped with an `rpc` attribute **must not**:
```

proc-macros/src/lib.rs
```diff
@@ -111,212 +110,278 @@ pub fn rpc_client_api(input_token_stream: TokenStream) -> TokenStream {
...
+///
+/// use futures_channel::oneshot;
+/// use jsonrpsee::{ws_client::*, ws_server::WsServerBuilder};
+///
+/// // RPC is moved into a separate module to clearly show names of generated entities.
+/// mod rpc_impl {
+///     use jsonrpsee::{proc_macros::rpc, types::async_trait, ws_server::SubscriptionSink};
+///
+///     // Generate both server and client implementations, prepend all the methods with `foo_` prefix.
+///     #[rpc(client, server, namespace = "foo")]
```
niklasad1, line 283: I like the namespace attribute, Is it possible to add any number of traits for the server to implement?

proc-macros/src/lib.rs
```diff
@@ -111,212 +110,278 @@ pub fn rpc_client_api(input_token_stream: TokenStream) -> TokenStream {
...
+///         #[subscription(name = "sub", unsub = "unsub", item = String)]
+///         fn sub(&self);
+///     }
+///
+///     // Structure that will implement `RpcServer` trait.
+///     // In can have fields, if required, as long as it's still `Send + Sync + 'static`.
+///     pub struct RpcServerImpl;
+///
+///     #[async_trait]
+///     impl RpcServer for RpcServerImpl {
```
niklasad1, line 300: the trait is named `Rpc`?!

```suggestion
///     impl Rpc for RpcServerImpl {
```

proc-macros/src/lib.rs
```diff
@@ -111,212 +110,278 @@ pub fn rpc_client_api(input_token_stream: TokenStream) -> TokenStream {
...
+///
+/// // RPC is moved into a separate module to clearly show names of generated entities.
+/// mod rpc_impl {
+///     use jsonrpsee::{proc_macros::rpc, types::async_trait, ws_server::SubscriptionSink};
+///
+///     // Generate both server and client implementations, prepend all the methods with `foo_` prefix.
+///     #[rpc(client, server, namespace = "foo")]
+///     pub trait Rpc {
+///         #[method(name = "foo")]
+///         async fn async_method(&self, param_a: u8, param_b: String) -> u16;
```
niklasad1, line 286: I haven't read the code but is it possible to name this method to whatever I want such as `async fn get_me_some_food(&self,  portions: u8, dish: String)` in the case instead of `async_method?

proc-macros/src/lib.rs
```diff
@@ -111,212 +110,278 @@ pub fn rpc_client_api(input_token_stream: TokenStream) -> TokenStream {
...
+/// #[tokio::main]
+/// async fn main() {
+///     let server_addr = websocket_server().await;
+///     let server_url = format!("ws://{}", server_addr);
+///     // Note that we create the client as usual, but thanks to the `use rpc_impl::RpcClient`,
+///     // the client object will have all the methods to interact with the server.
+///     let client = WsClientBuilder::default().build(&server_url).await.unwrap();
+///
+///     // Invoke RPC methods.
+///     assert_eq!(client.async_method(10, "a".into()).await.unwrap(), 42);
```
niklasad1, line 352: I really like this much less clunky than the old version.

proc-macros/src/new/render_client.rs
```diff
@@ -0,0 +1,134 @@
...
+
+		// Encoded parameters for the request.
+		let parameters = if !method.params.is_empty() {
+			let serde_json = self.jrps_client_item(quote! { __reexports::serde_json });
+			let params = method.params.iter().map(|(param, _param_type)| {
+				quote! { #serde_json::to_value(&#param)? }
+			});
+
+			quote! {
+				vec![ #(#params),* ].into()
```
niklasad1, line 73: so, this only supports parameters by position i.e, vec/array? 

I think that is alright because I think `params by name` is not used that often, maybe we could have an attribute for named params.

Currently, `named params` is the default in the proc macros but I think positional parameters is better as default.

proc-macros/src/new/render_server.rs
```diff
@@ -0,0 +1,234 @@
...
+
+		// Parameters encoded as a tuple (to be parsed from array).
+		let (params_fields_seq, params_types_seq): (Vec<_>, Vec<_>) = params.iter().cloned().unzip();
+		let params_types = quote! { (#(#params_types_seq),*) };
+		let params_fields = quote! { (#(#params_fields_seq),*) };
+
+		// Code to decode sequence of parameters from a JSON array.
+		let decode_array = {
+			let decode_fields = params.iter().enumerate().map(|(id, (name, ty))| {
+				if is_option(ty) {
```
niklasad1, line 134: am I reading this correct; this actually provides `None` if the param is missing in the request? If so, neat.

proc-macros/src/new/render_server.rs
```diff
@@ -0,0 +1,234 @@
...
+
+				#(#methods)*
+				#(#subscriptions)*
+
+				Ok(rpc)
+			}
+		})
+	}
+
+	fn render_params_decoding(&self, params: &[(syn::PatIdent, syn::Type)]) -> (TokenStream2, TokenStream2) {
```
niklasad1, line 118: I believe this allocates when using `serde_json::from_value` whereas [RpcParams doesn't](https://github.com/paritytech/jsonrpsee/blob/master/types/src/v2/params.rs#L63), 

I think our zero-alloc friend @maciejhirsz isn't super happy about that but we could take care of that in another pr :man_shrugging:

proc-macros/src/lib.rs
```diff
@@ -111,212 +110,278 @@ pub fn rpc_client_api(input_token_stream: TokenStream) -> TokenStream {
...
+///     async fn async_method(&self, param_a: u8, param_b: String) -> Result<u16, Error> {
+///         // Actual implementations are stripped, but inside a corresponding `Client` or
+///         // `SubscriptionClient` method is called.
+///     }
+///     async fn sync_method(&self) -> Result<String, Error> {
+///         // ...
+///     }
+///     
+///     // Subscription method returns `Subscription` object in case of success.
+///     async fn sub(&self, subscription_sink: SubscriptionSink) -> Result<Subscription<String>, Error> {
```
niklasad1, line 201: right, I don't understand `subscription_sink: SubscriptionSink` that should be for the server?

It should work to do `async fn sub_with_input(&self, a: usize, b: usize) -> Result<Subscription<String>, Error>` right?

Basically to tell the server to use params as input for the subscription.

proc-macros/tests/ui/correct/basic.rs
```diff
@@ -0,0 +1,72 @@
...
+
+#[rpc(client, server, namespace = "foo")]
+pub trait Rpc {
+	#[method(name = "foo")]
+	async fn async_method(&self, param_a: u8, param_b: String) -> u16;
+
+	#[method(name = "bar")]
+	fn sync_method(&self) -> u16;
+
+	#[subscription(name = "sub", unsub = "unsub", item = String)]
```
niklasad1, line 19: would super useful with a test/example for `subscriptions` with param(s),

I didn't managed to get it working:

```rust
  --> examples/proc_macro.rs:35:1
   |
35 | #[rpc(client, server, namespace = "foo")]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   | |
   | expected enum `jsonrpsee::types::Error`, found enum `CallError`
   | help: try using a variant of the expected enum: `Call(#[rpc(client, server, namespace = "foo")])`
   |
note: return type inferred to be `jsonrpsee::types::Error` here
  --> examples/proc_macro.rs:35:1
   |
35 | #[rpc(client, server, namespace = "foo")]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

I suspect that `CallError` is returned but it should be `Error`

niklasad1 (commented):
I like this very much and the overall design makes sense to me, thanks for the documentation/examples which made this much easier to understand.

proc-macros/src/lib.rs
popzxc: > Is it possible to add any number of traits for the server to implement?

Can you elaborate on what do you mean?

proc-macros/src/lib.rs
popzxc: No, the trait is named `RpcServer`. I think it's more clear, because one `rpc` can generate *two* traits:

- Server trait to be implemented for some item.
- Client extension trait.

Otherwise, it will be confusing why the server trait is named `Rpc` and the client trait is named `RpcClient`.

This test is actually mostly copied from the [integration test](https://github.com/paritytech/jsonrpsee/blob/ffbac8f9abd9e2b45662a01e177a79b6a055b79b/tests/tests/new_proc_macros.rs) you can play with.

proc-macros/src/new/render_client.rs
popzxc: It's the client implementation, so since parameters are encoded under the hood I thought it's not that important to support encoding by name as well. If that's required, we sure can.

proc-macros/src/new/render_server.rs
popzxc: Yep, I thought it will be strictly incorrect to explicitly require optional arguments. Missing optional arguments can either be provided as `null` or not be included at all.

proc-macros/src/new/render_server.rs
popzxc: *I believe* that for the first iteration it's not that important to be fully optimal here. Generic non-allocating implementation will be **much** less trivial to implement, and if the performance is the key, there always a way to optimize things manually.

Generally speaking, by writing the code yourself you are more likely to get better performance.

proc-macros/src/lib.rs
popzxc: Great catch! Yeah, the sink is for the server only.

proc-macros/tests/ui/correct/basic.rs
popzxc: Sure thing, will add!

popzxc: Must be fixed now.

proc-macros/src/lib.rs
popzxc: Ah, do you mean whether one `struct` can implement several server traits (like `#[rpc(server)] trait Foo` and `#[rpc(server) trait Bar`)? If so, it sure can.

proc-macros/src/lib.rs
niklasad1: right my bad

proc-macros/src/lib.rs
niklasad1: exactly.

so I suppose it's not possible to conflicting `functions names` in different traits then?

For example:

```rust
pub trait RpcNamespaceFoo {
	#[method(name = "foo")]
	async fn async_method(&self, param_a: u8, param_b: String) -> u16;
}

#[rpc(client, server, namespace = "bar")]
pub trait RpcNamespaceBar {
	#[method(name = "foo")]
	async fn async_method(&self, param_a: u8, param_b: String, x: usize) -> u16;
}
```

popzxc: As long as prefixes are different, I don't think there should be a problem. When you will call `RpcNamespaceFoo::into_rpc(self)`, it will register method `foo_foo`, and when you will call `RpcNamespaceBar::into_rpc(self)`, it will register method `bar_foo`, so there will be no name conflict.

popzxc: The same applies to the client. You will be able to call `RpcNamespaceFooClient::async_method(&client, 10, "a".into())` and `RpcNamespaceBarClient::async_method(&client, 10,  "a".into(), 20)`, and it will call `foo_foo` and `bar_foo` correspondingly.

niklasad1: all right, you are right I just tried and it works.

FWIW, the `server` in this context is `RpcServerImpl` or the type that implements the traits or one of them. I got that wrong myself

niklasad1 (approved):
I'm super happy about this PR, there are some minor things such as:

-  `alias attribute` for methods and subscription.
- avoid some allocations

However, I'm in favor of merging this as it is and improve down the line.

Epic work thanks alot :1st_place_medal:

maciejhirsz (approved):
The API looks good to me, as does the code. I do have nitpicks about param parsing as @niklasad1 expected, but I don't think we need to nail that one right now, we can solve that in a future PR :).

popzxc:
> we can solve that in a future PR

Sure! I think it will make sense to create a (tracking?) issue with the design discussion for the macro, so different problems/inconsistencies will be resolved as separate not-so-big PRs.

dvdplm:
@popzxc As a small thank you for your contributions we wanted to give you a "tip", but in order to do so you need to comment here with your polkadot address. :)

PR #388, opened 2021-06-18
Synchronization-less async connections in ws-server

maciejhirsz:
An experiment of sorts that will run all incoming ws connections on a single task/thread. This has multiple advantages in that we don't incur synchronization costs or atomic ref counting, so things like the `Methods` or `Settings` don't have to be wrapped in `Arc`s (methods can be shared between connections) or use `Arc`s internally (lists of allowed origins / hosts).

If we really need to use all cores for handling RPC parsing (as opposed to spawning individual method calls on their own tokio tasks), the `ws-server` could start a thread pool and poll for incoming connections on each thread (`TcpListener::poll_accept` only needs `&self`, so we can cheaply share `Arc<TcpListener>` between threads), although that would incur some costs for keeping track of the number of open connections.

### DIFF
--- a/ws-server/src/server.rs
+++ b/ws-server/src/server.rs
@@ -24,13 +24,17 @@
 // IN background_task WITH THE SOFTWARE OR THE USE OR OTHER
 // DEALINGS IN THE SOFTWARE.
 
+use std::future::Future;
+use std::pin::Pin;
+use std::task::{Context, Poll};
+use std::{net::SocketAddr, sync::Arc};
+
 use futures_channel::mpsc;
 use futures_util::future::{join_all, FutureExt};
 use futures_util::io::{BufReader, BufWriter};
 use futures_util::stream::StreamExt;
 use jsonrpsee_types::TEN_MB_SIZE_BYTES;
 use soketto::handshake::{server::Response, Server as SokettoServer};
-use std::{net::SocketAddr, sync::Arc};
 use tokio::net::{TcpListener, TcpStream, ToSocketAddrs};
 use tokio_util::compat::{Compat, TokioAsyncReadCompatExt};
 
@@ -86,8 +90,8 @@ impl Server {
 						continue;
 					}
 
-					if driver.count() >= self.cfg.max_connections as usize {
-						log::warn!("Too many connections. Try again in a while");
+					if driver.connection_count() >= self.cfg.max_connections as usize {
+						log::warn!("Too many connections. Try again in a while.");
 						continue;
 					}
 
@@ -106,10 +110,6 @@ impl Server {
 	}
 }
 
-use std::future::Future;
-use std::pin::Pin;
-use std::task::{Context, Poll};
-
 /// This is a glorified select `Future` that will attempt to drive all
 /// connection futures `F` to completion on each `poll`, while also
 /// handling incoming connections.
@@ -126,7 +126,7 @@ where
 		ConnDriver { listener, connections: Vec::new() }
 	}
 
-	fn count(&self) -> usize {
+	fn connection_count(&self) -> usize {
 		self.connections.len()
 	}
 
@@ -160,8 +160,6 @@ where
 		this.listener.poll_accept(cx)
 	}
 }
-
-impl<F: Unpin> Unpin for ConnDriver<F> {}
 
 async fn handshake(
 	socket: tokio::net::TcpStream,
--- a/ws-server/src/server.rs
+++ b/ws-server/src/server.rs
@@ -148,6 +148,9 @@ where
 
 		while i < this.connections.len() {
 			if this.connections[i].poll_unpin(cx).is_ready() {
+				// Using `swap_remove` since we don't care about ordering
+				// but we do care about removing being `O(1)`.
+				//
 				// We don't increment `i` in this branch, since we now
 				// have a shorter length, and potentially a new value at
 				// current index
--- a/ws-server/src/server.rs
+++ b/ws-server/src/server.rs
@@ -106,8 +106,8 @@ impl Server {
 						continue;
 					}
 
-					let methods = methods.clone();
-					let cfg = self.cfg.clone();
+					let methods = &methods;
+					let cfg = &self.cfg;
 
 					running.push(Box::pin(background_task(socket, id, methods, cfg)));
 				},
@@ -163,8 +163,8 @@ where
 async fn background_task(
 	socket: tokio::net::TcpStream,
 	conn_id: ConnectionId,
-	methods: Methods,
-	cfg: Settings,
+	methods: &Methods,
+	cfg: &Settings,
 ) -> Result<(), Error> {
 	// For each incoming background_task we perform a handshake.
 	let mut server = SokettoServer::new(BufReader::new(BufWriter::new(socket.compat())));
--- a/ws-server/src/server.rs
+++ b/ws-server/src/server.rs
@@ -120,7 +120,7 @@ impl Server {
 
 					driver.add(Box::pin(handshake(socket, id, methods, cfg, &shutdown, &self.stop_handle)));
 
-					id += 1;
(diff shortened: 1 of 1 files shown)


niklasad1 (approved):
neat, LGTM

ws-server/src/server.rs
```diff
@@ -73,43 +73,101 @@ impl Server {
...
+			} else {
+				i += 1;
 			}
 		}
+
+		this.listener.poll_accept(cx)
 	}
 }
 
+impl<F: Unpin> Unpin for ConnDriver<F> {}
```
jsdw, line 164: Nit: I think that this is implemented automatically anyway?

jsdw (approved):
Looks good to me!

ws-server/src/server.rs
```diff
@@ -73,43 +73,101 @@ impl Server {
...
+		loop {
+			match Pin::new(&mut driver).await {
+				Ok((socket, _addr)) => {
+					if let Err(e) = socket.set_nodelay(true) {
+						log::error!("Could not set NODELAY on socket: {:?}", e);
+						continue;
+					}
+
+					if driver.count() >= self.cfg.max_connections as usize {
+						log::warn!("Too many connections. Try again in a while");
```
dvdplm, line 90:
```suggestion
						log::warn!("Too many connections. Try again in a while.");
```

ws-server/src/server.rs
```diff
@@ -73,43 +73,101 @@ impl Server {
...
-				let counter = conn_counter.clone();
-				let cfg = self.cfg.clone();
+			}
+		}
+	}
+}
+
+use std::future::Future;
+use std::pin::Pin;
+use std::task::{Context, Poll};
```
dvdplm, line 111: on top?

ws-server/src/server.rs
```diff
@@ -73,43 +73,101 @@ impl Server {
...
+
+		loop {
+			match Pin::new(&mut driver).await {
+				Ok((socket, _addr)) => {
+					if let Err(e) = socket.set_nodelay(true) {
+						log::error!("Could not set NODELAY on socket: {:?}", e);
+						continue;
+					}
+
+					if driver.count() >= self.cfg.max_connections as usize {
```
dvdplm, line 89: This should maybe be named the full `connection_count()` to avoid confusion?

ws-server/src/server.rs
```diff
@@ -73,43 +73,101 @@ impl Server {
...
+{
+	type Output = std::io::Result<(TcpStream, SocketAddr)>;
+
+	fn poll(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
+		let this = Pin::into_inner(self);
+
+		let mut i = 0;
+
+		while i < this.connections.len() {
+			if this.connections[i].poll_unpin(cx).is_ready() {
```
dvdplm, line 150: So when we get `Poll::Ready` here it means the background task has finished/exited?

ws-server/src/server.rs
```diff
@@ -73,43 +73,101 @@ impl Server {
...
+				// We don't increment `i` in this branch, since we now
+				// have a shorter length, and potentially a new value at
+				// current index
+				this.connections.swap_remove(i);
```
dvdplm, line 154: To help slow readers maybe we can do `let _ = this.connection.swap_remove(i);` here, to emphasize that we don't care about the swapping aspect here (it's just a faster way to do `remove()` right?).

dvdplm (commented):
I think this is slightly harder to read than what we had (but it's likely just me struggling with reading async code).

Do you think you could elaborate a bit more on the tradeoff here, single-threaded RPC parsing sounds like it should be fine, but maybe we need to be careful with the amount of work we do in the parsing stage?

Finally: I imagine we need the same treatment for the http server?

ws-server/src/server.rs
niklasad1: Yes, swap remove is O(1) but it doesn't preserve the order.

`remove()` requires shifting all element after the removed element to the left.

ws-server/src/server.rs
maciejhirsz: Correct.

ws-server/src/server.rs
```diff
@@ -91,64 +95,114 @@ impl Server {
...
+			} else {
+				i += 1;
+			}
+		}
+
+		if let Poll::Ready(Some(())) = this.stop_receiver.next().poll_unpin(cx) {
+			return Poll::Ready(Err(DriverError::Shutdown));
+		}
```
maciejhirsz, line 192: This bit replaces the select that was added on `master`.

ws-server/src/server.rs
```diff
@@ -170,6 +224,27 @@ async fn background_task(
...
+	tokio::spawn(background_task(
+		server,
+		conn_id,
+		methods.clone(),
+		cfg.max_request_body_size,
+		shutdown.clone(),
+		stop_handle.clone(),
+	))
+	.await
+	.unwrap()
```
dvdplm, line 236: Is this here on purpose?

ws-server/src/server.rs
```diff
@@ -91,64 +95,114 @@ impl Server {
...
+						continue;
 					}
-				},
-				stop = stop_receiver.next() => {
-					if stop.is_some() {
-						break;
+
+					if driver.connection_count() >= self.cfg.max_connections as usize {
+						log::warn!("Too many connections. Try again in a while.");
+						continue;
```
dvdplm, line 115: This is where we'd send a `CLOSE` with a reason right?

ws-server/src/server.rs
```diff
@@ -91,64 +95,114 @@ impl Server {
...
 					}
-				},
-				complete => break,
+
+					let methods = &methods;
+					let cfg = &self.cfg;
+
+					driver.add(Box::pin(handshake(socket, id, methods, cfg, &shutdown, &self.stop_handle)));
+
+					id += 1;
```
niklasad1, line 123: I would prefer to use `wrapping_add` here but I don't how likely it's that a server reaches 2^32 connections though ^

ws-server/src/server.rs
dvdplm: It would be good to document how the shutdown process works and what connected clients should expect to happen to their pending requests.

ws-server/src/server.rs
maciejhirsz: I believe release build will make this into a wrapping add regardless, but might be a good idea to make it explicit.

ws-server/src/server.rs
niklasad1: No, we just deny the socket.

The close reason is only sent after the `websocket handshake` has been completed AFAIU

maciejhirsz: I think you're confusing this PR with telemetry :). At this stage the WS connection hasn't been established yet, so you can't send a frame. A nice thing to do would be to send a HTTP response with appropriate status code.

dvdplm: >  A nice thing to do would be to send a HTTP response with appropriate status code.

Indeed I'm confused. And strong yes for returning a proper http status code.

ws-server/src/server.rs
dvdplm: So I'll try again: is **this** where we should send a `CLOSE` frame?

maciejhirsz: Yeah, I reckon @niklasad1's refactor will make the code much easier to follow here too.

ws-server/src/server.rs
dvdplm: Proper status code is `429 Too Many Requests`? Should we include a `Retry-After` header too?

maciejhirsz: Let's do that in a separate PR, since that would need extra changes to `soketto` too.

PR #389, opened 2021-06-21
[jsonrpsee types]: unify a couple of types + more tests

### LINKED ISSUE #276: [types]: jsonrpc_v2 types write serialize/deserialize tests.

niklasad1:
This PR adds some additional tests to close #276 (remaining is the ignored tests for `RpcParams` that are untyped but as long we provide `RpcParams::parse` for it, it should be fine)

Then I was annoyed of writing so much tests so I unified `JsonRpcNotifResponse and JsonRpcNotification` and `JsonRpcNotificationParamsAlloc and `JsonRpcNotificationParams`.

Thus, I replaced `RawValue` with `T` it will require allocations when deserializing but the benefit is that we don't have convert values to `RawValue` when sending stuff via the `SubscriptionSink`

### DIFF
--- a/types/src/v2/params.rs
+++ b/types/src/v2/params.rs
@@ -4,22 +4,12 @@ use beef::Cow;
 use serde::de::{self, Deserializer, Unexpected, Visitor};
 use serde::ser::Serializer;
 use serde::{Deserialize, Serialize};
-use serde_json::{value::RawValue, Value as JsonValue};
+use serde_json::Value as JsonValue;
 use std::fmt;
 
 /// JSON-RPC parameter values for subscriptions.
 #[derive(Serialize, Deserialize, Debug)]
-pub struct JsonRpcNotificationParams<'a> {
-	/// Subscription ID
-	pub subscription: u64,
-	/// Result.
-	#[serde(borrow)]
-	pub result: &'a RawValue,
-}
-
-/// JSON-RPC parameter values for subscriptions with support for number and strings.
-#[derive(Deserialize, Debug)]
-pub struct JsonRpcNotificationParamsAlloc<T> {
+pub struct JsonRpcSubscriptionParams<T> {
 	/// Subscription ID
 	pub subscription: SubscriptionId,
 	/// Result.
@@ -245,7 +235,9 @@ impl<'a> From<Id<'a>> for OwnedId {
 
 #[cfg(test)]
 mod test {
-	use super::{Cow, Id, JsonRpcParams, JsonValue, RpcParams, SubscriptionId, TwoPointZero};
+	use super::{
+		Cow, Id, JsonRpcParams, JsonRpcSubscriptionParams, JsonValue, RpcParams, SubscriptionId, TwoPointZero,
+	};
 
 	#[test]
 	fn id_deserialization() {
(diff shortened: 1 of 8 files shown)


dvdplm (approved):
LGTM

types/src/v2/params.rs
```diff
@@ -335,4 +327,24 @@ mod test {
 			assert_eq!(&serialized, initial_ser);
 		}
 	}
+
+	#[test]
+	fn subscription_params_serialize_work() {
+		let ser =
+			serde_json::to_string(&JsonRpcSubscriptionParams { subscription: SubscriptionId::Num(12), result: "goal" })
+				.unwrap();
+		assert_eq!(ser, r#"{"subscription":12,"result":"goal"}"#);
```
TarikGul, line 336: Super tiny nit pick, more of a consistency suggestion. In the other tests you save the expected result to `let exp` and then plug it into `assert_eq!` which I think is definitely clean. 

I just noticed here you plug the expected result directly in the test macro. Up to you but it would make it more consistent to save it to a `exp` here as well.

types/src/v2/params.rs
```diff
@@ -335,4 +327,24 @@ mod test {
...
+	}
+
+	#[test]
+	fn subscription_params_deserialize_work() {
+		let ser = r#"{"subscription":"9","result":"offside"}"#;
+		assert!(
+			serde_json::from_str::<JsonRpcSubscriptionParams<()>>(ser).is_err(),
+			"invalid type should not be deser"
```
TarikGul, line 344: Will the average user know that `deser` is `deserialize`?

utils/src/server/rpc_module.rs
```diff
@@ -404,14 +407,8 @@ impl SubscriptionSink {
...
-			let result =
-				to_raw_value(&SubscriptionClosedError::from(close_reason)).expect("valid json infallible; qed");
-			let msg = serde_json::to_string(&JsonRpcSubscriptionResponse {
-				jsonrpc: TwoPointZero,
-				method: self.method,
-				params: JsonRpcNotificationParams { subscription: self.uniq_sub.sub_id, result: &*result },
-			})
-			.expect("valid json infallible; qed");
+			let msg =
+				self.build_message(&SubscriptionClosedError::from(close_reason)).expect("valid json infallible; qed");
```
TarikGul, line 411: Whats `qed`?

TarikGul (approved):
Few questions and one tiny nit pick, but overall lgtm.

utils/src/server/rpc_module.rs
niklasad1: https://en.wikipedia.org/wiki/Q.E.D.

basically, hand-wavy proof that a given `unwrap/expect` is infallible

types/src/v2/params.rs
niklasad1: fair enough, it saved me one line :)

types/src/v2/params.rs
niklasad1: changed it, probably better to verbose in this case thanks.

utils/src/server/rpc_module.rs
dvdplm: It's latin, old-europe snobbish way of saying "tada!"

types/src/v2/params.rs
dvdplm: Always good to be explicit, but yeah, most readers would be familiar with the abbreviation.

Issue #380, opened 2021-06-15
`RpcParams::parse` to handle optional params & validate JSON-RPC spec


maciejhirsz:
Currently `RpcParams` is just using `serde_json::from_str`, which creates two problems:

1. There is no way to handle deserializing a tuple with optional params, e.g.: `(A, Option<B>)` from json `["some type that creates A"]`.
2. It is technically possible to allow param types other than those allowed in the spec (`"params"` must be either `null`, an object with named params, or an array).

One way to do that would be to create tuple types for different numbers of params (`Params1<A>(A)`, `Params2<A, B>(A, B)`, etc., up to some fixed limit like 10), and then manually implementing `Deserialize` on them (with macros for all the variants). This would also remove the need to have `RpcParams::one` separate from `RpcParams::parse`.

maciejhirsz:
Solved by #401.

PR #391, opened 2021-06-22
[WIP]: support optional params in `RpcParams::parse`

niklasad1:
Rough draft in order to close https://github.com/paritytech/jsonrpsee/issues/380

I had some issues with the conflicting `serde impls` for implementing it directly on the "tuples" such that I had to introduce `new type pattern` for ParamsOne, .. ParamsFive instead of tuples.

Thus, a bit clunky for users to use `ParamsOne` then to access the inner tuple however it seems to work except an edge case when you entered a list ["x", "y"] then if you expect [String, Option<String>, String] it won't work because the second parameter will still be parsed as `Option<String>` but that should acceptable I think.

Also, for custom types it's awkward for users to implement the trait `MaybeOptionalParams` for their types.

//cc @maciejhirsz

### DIFF
--- a/tests/tests/integration_tests.rs
+++ b/tests/tests/integration_tests.rs
@@ -32,10 +32,15 @@ mod helpers;
 use helpers::{http_server, websocket_server, websocket_server_with_subscription};
 use jsonrpsee::{
 	http_client::{traits::Client, Error, HttpClientBuilder},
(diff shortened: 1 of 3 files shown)


niklasad1:
Closing in favor of #401

PR #392, opened 2021-06-24
[ci]: test each individual crate's manifest

niklasad1:
We have been bitten by that the CI just checks all crates compiles in the workspace (shares Cargo.lock) and features.

Thus, if a feature for an individual is not enabled for specific crate then this is not detected by CI so this PR changes that each crate using its manifest instead.

In addition:
- Fixes the `ws-server` to compile outside the workspace
- Added to run CI on macos and Windows.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -120,7 +120,7 @@ jobs:
         uses: actions-rs/cargo@v1.0.3
         with:
           command: check
-          args: --manifest-path utils/Cargo.toml
+          args: --manifest-path proc-macros/Cargo.toml
 
       - name: Cargo check test utils
         uses: actions-rs/cargo@v1.0.3
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,7 +62,7 @@ jobs:
       - name:                      Rust Cache
         uses:                      Swatinem/rust-cache@v1.3.0
 
-      - name: Cargo check all targets
+      - name: Cargo check all targets (use Cargo.toml in workspace)
         uses: actions-rs/cargo@v1.0.3
         with:
           command: check
@@ -73,6 +73,42 @@ jobs:
         with:
           command: check
           args: --manifest-path http-client/Cargo.toml --no-default-features --features tokio02
+
+      - name: Cargo check HTTP client
+        uses: actions-rs/cargo@v1.0.3
+        with:
+          command: check
+          args: --manifest-path http-client/Cargo.toml
+
+      - name: Cargo check WS client
+        uses: actions-rs/cargo@v1.0.3
+        with:
+          command: check
+          args: --manifest-path ws-client/Cargo.toml
+
+      - name: Cargo check WS client with tokio02
+        uses: actions-rs/cargo@v1.0.3
+        with:
+          command: check
+          args: --manifest-path ws-client/Cargo.toml --no-default-features --features tokio02
+
+      - name: Cargo check types
+        uses: actions-rs/cargo@v1.0.3
+        with:
+          command: check
+          args: --manifest-path types/Cargo.toml
+
+      - name: Cargo check utils
+        uses: actions-rs/cargo@v1.0.3
+        with:
+          command: check
+          args: --manifest-path utils/Cargo.toml
+
+      - name: Cargo check proc macros
+        uses: actions-rs/cargo@v1.0.3
+        with:
+          command: check
+          args: --manifest-path utils/Cargo.toml
 
   tests:
     name: Run tests
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -110,9 +110,77 @@ jobs:
           command: check
           args: --manifest-path utils/Cargo.toml
 
+      - name: Cargo check test utils
+        uses: actions-rs/cargo@v1.0.3
+        with:
+          command: check
+          args: --manifest-path test-utils/Cargo.toml
+
+      - name: Cargo check examples
+        uses: actions-rs/cargo@v1.0.3
+        with:
+          command: check
+          args: --manifest-path examples/Cargo.toml
+
   tests:
-    name: Run tests
+    name: Run tests Ubuntu
     runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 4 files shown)


ws-server/src/tests.rs
```diff
@@ -173,8 +173,6 @@ async fn can_set_max_connections() {
 	assert!(conn3.is_err());
 	let err = conn3.unwrap_err();
 	assert!(err.to_string().contains("WebSocketHandshake failed"));
-	assert!(err.to_string().contains("Connection reset by peer"));
-	// Err(Io(Os { code: 54, kind: ConnectionReset, message: \"Connection reset by peer\" }))");
```
niklasad1, line 177: windows returns https://support.microsoft.com/en-us/topic/an-application-may-receive-the-10054-error-when-the-application-receives-data-from-a-connection-on-a-computer-that-is-running-windows-7-or-windows-server-2008-r2-if-a-tdi-filter-driver-is-installed-73fd74a9-51c2-3663-e965-e1187b1cfc92

so let's remove the assertion it would be better if we used `io::Error` in the test client I suppose

.github/workflows/ci.yml
```diff
@@ -101,3 +161,59 @@ jobs:
...
+
+      - name: Install Rust stable toolchain
+        uses: actions-rs/toolchain@v1.0.7
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
+
+      - name:                      Rust Cache
+        uses:                      Swatinem/rust-cache@v1.3.0
```
jsdw, line 180: I'm curious; I've seen this spacing between the prop and value elsewhere for these cache lines and wasn't sure why; is there a reason for it?

niklasad1: haha, we got a PR from the CI and he did this way but I can fix it ^

.github/workflows/ci.yml
```diff
@@ -74,8 +74,68 @@ jobs:
           command: check
           args: --manifest-path http-client/Cargo.toml --no-default-features --features tokio02
```
jsdw, line 75: Instead of checking every crate individually, could we just run `cargo check` without arguments? My experience is that that checks every crate in the workspace by default (though you can also add `--workspace` to, I assume, be sure).

Or do we need to check individually because `http-client` needs the `tokio02` feature setting?

.github/workflows/ci.yml
```diff
@@ -101,3 +161,59 @@ jobs:
...
+          override: true
+
+      - name:                      Rust Cache
+        uses:                      Swatinem/rust-cache@v1.3.0
+
+      - name: Cargo build
+        uses: actions-rs/cargo@v1.0.3
+        with:
+          command: build
+          args: --workspace
```
jsdw, line 186: Do we need to have a build step before the `test` step? (I'll need to look into these rust github actions more!)

.github/workflows/ci.yml
jsdw: Aah; I think I removed the spaces when I copied some CI bits from jsonrpsee but I've been curious ever since :D Makes sense!

ws-server/Cargo.toml
```diff
@@ -12,7 +12,7 @@ documentation = "https://docs.rs/jsonrpsee-ws-server"
 [dependencies]
 thiserror = "1"
 futures-channel = "0.3.14"
-futures-util = { version = "0.3.14", default-features = false, features = ["io"] }
+futures-util = { version = "0.3.14", default-features = false, features = ["io", "async-await-macro"] }
```
jsdw, line 15: Nit: It doesn't look like we've added any code that requires this feature; do we want it?

jsdw (approved):
Looks good to me; I have a few questions but nothing major and mainly just for my own curiosity!

ws-server/Cargo.toml
niklasad1: it was introduced in another PR, thus doesn't compile standalone such as if you `cd ws-server && cargo check` outside the workspace it doesn't work. Because, features in Rust are additive (the union of all features) so if another crate adds that feature it's enabled for all in workspace if that makes sense.

TLDR; on master `cd ws-server && cargo check` doesn't work but `cargo build -p ws-server` works :P

.github/workflows/ci.yml
niklasad1: yeah, we did do `cargo check --workspace` before this PR but that's the problem and why I added `cargo check --manifest-path crate`, see my comment below.

.github/workflows/ci.yml
niklasad1: no you are right, we don't have too but the previous checks only run `cargo check` so we have want make sure nothing weird happens when building the binary but it could be another action for clarity I suppose.

ws-server/Cargo.toml
jsdw: Aah gotcha! Thanks for the explanation!

.github/workflows/ci.yml
jsdw: My first thought was that `cargo test` builds the binary anyway, but of course it builds it with `cfg(test)` enabled, so I can see why a separate build step is useful to make sure it works "normally"!

.github/workflows/ci.yml
jsdw: Thanks, that all makes sense now :)

ws-server/src/tests.rs
```diff
@@ -171,10 +171,12 @@ async fn can_set_max_connections() {
...
-	let err = conn3.unwrap_err();
-	assert!(err.to_string().contains("WebSocketHandshake failed"));
-	assert!(err.to_string().contains("Connection reset by peer"));
-	// Err(Io(Os { code: 54, kind: ConnectionReset, message: \"Connection reset by peer\" }))");
+
+	let err = match conn3 {
+		Err(soketto::handshake::Error::Io(err)) => err,
+		_ => panic!("Invalid error kind; expected std::io::Error"),
+	};
+	assert_eq!(err.kind(), std::io::ErrorKind::ConnectionReset);
```
dvdplm, line 179: Nice, ty for fixing this.

dvdplm (approved):

PR #393, opened 2021-06-25
WIP: [rpc errors]: user customized error codes

niklasad1:
The motivation behind this change is that we need to provide functionality for users to configure their own error messages; that means `error code`, `error message` and `data`, https://www.jsonrpc.org/specification#error-object

Another solution in a more generic way is to have a marker trait such as:

```rust

/// Marker trait
trait JsonRpcErrorT: Send + Sync + std::error::Error + JsonRpcErrorObjectT;

impl<T: Send + Sync + std::error::Error + JsonRpcErrorObject> for T {}

/// JSON-RPC error object trait
trait JsonRpcErrorObjectT {
   fn code(&self) -> i32;
   fn message(&self) -> &str;
   fn data(&self) -> &Option<RawValue>;
}

enum CallError {
   InvalidParams,
   Failed(Box<dyn JsonRpcErrorT>))
}
```

but I had some issues getting it to work for `anyhow::Error` however it might possible to avoid the introduced allocations in this PR.

### DIFF
--- a/types/src/error.rs
+++ b/types/src/error.rs
@@ -1,6 +1,8 @@
 use serde::{Deserialize, Serialize};
 use std::fmt;
 
+use crate::v2::error::JsonRpcErrorObjectOwned;
+
 /// Convenience type for displaying errors.
 #[derive(Clone, Debug, PartialEq)]
 pub struct Mismatch<T> {
@@ -24,7 +26,7 @@ pub enum CallError {
 	InvalidParams,
 	#[error("RPC Call failed: {0}")]
 	/// The call failed.
-	Failed(#[source] Box<dyn std::error::Error + Send + Sync>),
+	Failed(JsonRpcErrorObjectOwned),
 }
 
 /// Error type.
--- a/types/src/v2/error.rs
+++ b/types/src/v2/error.rs
@@ -25,6 +25,27 @@ impl<'a> fmt::Display for JsonRpcError<'a> {
 }
 
 /// JSON-RPC error object.
+#[derive(Debug, Deserialize, Serialize, Clone, PartialEq)]
+#[serde(deny_unknown_fields)]
+pub struct JsonRpcErrorObjectOwned {
+	/// Code
+	pub code: JsonRpcErrorCode,
+	/// Message
+	pub message: String,
+	/// Optional data
+	#[serde(skip_serializing_if = "Option::is_none")]
+	pub data: Option<serde_json::Value>,
+}
+
+impl fmt::Display for JsonRpcErrorObjectOwned {
+	fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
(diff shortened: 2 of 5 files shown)


types/src/error.rs
```diff
@@ -24,7 +26,7 @@ pub enum CallError {
 	InvalidParams,
 	#[error("RPC Call failed: {0}")]
 	/// The call failed.
-	Failed(#[source] Box<dyn std::error::Error + Send + Sync>),
+	Failed(JsonRpcErrorObjectOwned),
```
dvdplm, line 29: Hmm. Isn't this going to make things messy in substrate? As it is now, with a `Box<std::error::Error>`, it is fairly ergonomic to work with app specific errors (as long as they implement `dyn Error` it's all good). Won't we have a lot of pesky error conversion code with this?

It is true that in some places, `substrate` does this verbose verbose error handling, e.g. [here](https://github.com/paritytech/substrate/blob/e6cd6b1240ce741f06b05ad1c59f5399e52e717a/frame/transaction-payment/rpc/src/lib.rs#L103-L112).

So yeah, maybe we have to do this? :(

types/src/v2/error.rs
```diff
@@ -38,6 +59,17 @@ pub struct JsonRpcErrorObject<'a> {
 	pub data: Option<&'a RawValue>,
 }
 
+impl<'a> JsonRpcErrorObject<'a> {
+	/// Get owned version of the JsonRpcObject.
+	pub fn to_owned(&self) -> JsonRpcErrorObjectOwned {
+		JsonRpcErrorObjectOwned {
+			code: self.code,
+			message: self.message.to_owned(),
+			data: self.data.and_then(|data| Some(serde_json::Value::from(data.get()))),
```
dvdplm, line 68: clippy says to use `map`

types/src/v2/error.rs
```diff
@@ -24,6 +24,27 @@ impl<'a> fmt::Display for JsonRpcError<'a> {
 	}
 }
 
+/// JSON-RPC error object.
+#[derive(Debug, Deserialize, Serialize, Clone, PartialEq)]
+#[serde(deny_unknown_fields)]
+pub struct JsonRpcErrorObjectOwned {
```
dvdplm, line 30: I think this should be named just `RpcError`. It being owned is an implementation detail imo.

types/src/error.rs
niklasad1: indeed, that's why we need this but if you could convince that we don't need those application specific error codes and so on, I would happy to kill this PR :)

However, each rpc module in substrate has their own specific error codes, messages and data that we don't support currently

So you are right, it will be very boiler platey users of this API I really liked what had :(

I will investigate whether a trait design is better in this case, but we provide a macro for this or something.

niklasad1:
superseeded by #394

PR #394, opened 2021-06-25
feat: customizable JSON-RPC error codes via new enum variant on `CallErrror`

### LINKED ISSUE #299: [server]: error types when register methods

niklasad1:
The motivation behind this change is that we need to provide functionality for users to configure their own error messages; that means error code, error message and data, https://www.jsonrpc.org/specification#error-object

Superseeds #393 

Closes #299

### DIFF
--- a/http-server/src/tests.rs
+++ b/http-server/src/tests.rs
@@ -8,8 +8,35 @@ use jsonrpsee_test_utils::helpers::*;
 use jsonrpsee_test_utils::types::{Id, StatusCode, TestContext};
 use jsonrpsee_test_utils::TimeoutFutureExt;
 use jsonrpsee_types::error::{CallError, Error};
+use jsonrpsee_types::traits::JsonRpcErrorT;
 use serde_json::Value as JsonValue;
 use tokio::task::JoinHandle;
+
+#[derive(Debug, thiserror::Error)]
+enum TestError {
+	#[error("One")]
+	One,
+	#[error("Two")]
+	Two,
+}
+
+impl JsonRpcErrorT for TestError {
+	fn code(&self) -> i32 {
+		match self {
+			Self::One => 1,
+			Self::Two => 2,
+		}
+	}
+	fn message(&self) -> String {
+		match self {
+			Self::One => "error one".to_string(),
+			Self::Two => "error two".to_string(),
+		}
+	}
+	fn data(&self) -> Option<&serde_json::value::RawValue> {
+		None
+	}
+}
 
 async fn server() -> SocketAddr {
 	server_with_handles().await.0
@@ -39,21 +66,21 @@ async fn server_with_handles() -> (SocketAddr, JoinHandle<Result<(), Error>>, St
 	module.register_method("notif", |_, _| Ok("")).unwrap();
 	module
 		.register_method("should_err", |_, ctx| {
-			let _ = ctx.err().map_err(|e| CallError::Failed(e.into()))?;
+			let _ = ctx.err().map_err(|_| CallError::Failed(Box::new(TestError::One)))?;
 			Ok("err")
 		})
 		.unwrap();
 
 	module
 		.register_method("should_ok", |_, ctx| {
-			let _ = ctx.ok().map_err(|e| CallError::Failed(e.into()))?;
+			let _ = ctx.ok().map_err(|_| CallError::Failed(Box::new(TestError::One)))?;
 			Ok("ok")
 		})
 		.unwrap();
 	module
 		.register_async_method("should_ok_async", |_p, ctx| {
 			async move {
-				let _ = ctx.ok().map_err(|e| CallError::Failed(e.into()))?;
+				let _ = ctx.ok().map_err(|_| CallError::Failed(Box::new(TestError::Two)))?;
 				Ok("ok")
 			}
 			.boxed()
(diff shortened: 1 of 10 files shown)


types/src/error.rs
```diff
@@ -17,14 +18,24 @@ impl<T: fmt::Display> fmt::Display for Mismatch<T> {
 }
 
 /// Error that occurs when a call failed.
-#[derive(Debug, thiserror::Error)]
```
niklasad1, line 20: `thiserror` didn't work with the introduced marker trait

types/src/traits.rs
```diff
@@ -55,3 +55,20 @@ pub trait SubscriptionClient: Client {
...
+/// Trait for customizable [JSON-RPC error object](https://www.jsonrpc.org/specification#error-object)
+pub trait JsonRpcErrorT {
+	/// Error code (-32768 to -32000 are reserved and should not be used)
+	fn code(&self) -> i32;
+	/// Short description of the error.
+	// NOTE(niklasad1): this is a String because the underlying error may not be an constant
+	// and need to be included in the error.
+	fn message(&self) -> String;
+	/// A Primitive or Structured value that contains additional information about the error.
+	fn data(&self) -> Option<&serde_json::value::RawValue>;
```
niklasad1, line 73: might be annoying `serde_json::RawValue` here but at least we rexport it from types

dvdplm:
cc @maciejhirsz Thoughts on this? 
For context: it seems like substrate only adheres to the spec (which mandates a full ["Error Object"](https://www.jsonrpc.org/specification#error_object)) some of the time? Possibly because it's verbose and annoying.

niklasad1:
> cc @maciejhirsz Thoughts on this?
For context: it seems like substrate only adheres to the spec (which mandates a full "Error Object") some of the time? Possibly because it's verbose and annoying.

[example of this can be found here for context](https://github.com/paritytech/substrate/blob/master/client/rpc-api/src/author/error.rs#L106-L184) but I think these error codes is still quite useful in the context of substrate so we should really support it.

But I was thinking of two variants of `register_method`:

1. like old one, just require `StdError + Send + Sync` and just let jsonrpsee assign the error code
2. `register_method_with_custom_error`, require the error to impl `RpcErrorTrait`

http-server/src/tests.rs
```diff
@@ -8,9 +8,36 @@ use jsonrpsee_test_utils::helpers::*;
...
+}
+
+impl JsonRpcErrorT for TestError {
+	fn code(&self) -> i32 {
+		match self {
+			Self::One => 1,
+			Self::Two => 2,
+		}
+	}
+	fn message(&self) -> String {
```
maciejhirsz, line 30: Instead of message, can't we just require `JsonRpcErrorT: Display` bound?

maciejhirsz (commented):
This looks good to me, but I'm wondering if trait is the right approach. I think it might be better to have an enum like:

```rust
enum JsonRpcError {
    User(Box<dyn std::error::Error>),
    ParseError,
    InvalidError,
    // ...
}
```

With a `From<Box<dyn std::error::Error>>` impl, so we can auto cast all user errors into the `User` variant and give it some default error code, while also having the option to set codes for things like params parsing fail etc.

niklasad1:
> With a `From<Box<dyn std::error::Error>>` impl, so we can auto cast all user errors into the `User` variant and give it some default error code, while also having the option to set codes for things like params parsing fail etc.

Right, that would be nice but the custom error messages won't work but if we could do:

```rust
enum JsonRpcError {
    // provide default error code
    User(Box<dyn std::error::Error>),
    // user provides error code and similar
    CustomUser { code: i32, message: String, data: JsonValue },
    // some specific error code
    ParseError,
    // some specific error code
    InvalidError,
    // ...
}
```

that would be quite flexible and better, what do you think @maciejhirsz? :)

maciejhirsz:
Ye, that works I think.

types/src/error.rs
```diff
@@ -19,12 +20,22 @@ impl<T: fmt::Display> fmt::Display for Mismatch<T> {
 /// Error that occurs when a call failed.
 #[derive(Debug, thiserror::Error)]
 pub enum CallError {
```
niklasad1, line 22: we might want to put `non_exhaustive` on this but a little weird to just add some error code for non defined enum variant :P

types/src/error.rs
```diff
@@ -19,12 +20,22 @@ impl<T: fmt::Display> fmt::Display for Mismatch<T> {
...
+	Failed(Box<dyn std::error::Error + Send + Sync>),
+	/// Custom error with specific JSON-RPC error code, message and data.
+	#[error("RPC Call failed: code: {code}, message: {message}, data: {data:?}")]
+	Custom {
+		/// JSON-RPC error code
+		code: i32,
+		/// Short description of the error.
+		message: String,
+		/// A primitive or structured value that contains additional information about the error.
+		data: Option<Box<RawValue>>,
```
niklasad1, line 37: this might save allocations of the value is already a String

dvdplm (approved):
This should work.

maciejhirsz (approved):

Issue #395, opened 2021-06-28
[ws server]: `stop API` just denies new connections.


niklasad1:
Currently we expose a stop handle both the HTTP and WS servers but for the WS server it does only deny new connections. The already established connections will run forever or until the executors terminates (as `server.start().await` will block the current thread)

Because of
1. The subscriptions are not cancel-able currently
2. The background task that is spawned for each connection must be canceled and close the connection

niklasad1:
Closed by #396

PR #396, opened 2021-06-28
[ws server]: terminate already established connection(s) when the server is stopped

niklasad1:
Follow up on #386, closing #395 

This PR actually terminates the background tasks when `StopHandle::stop` is called, I re-used `inner StopHandle sender` that is used by the background tasks to determine when the server has been stopped.

Then everything taken care of including the subscriptions because we drop the `receiver` of the `SubscriptionSink`, so this should be sufficient.

It's a little bit uglier/complex but should be okay IMHO 

Might overlap with #388 

Thoughts?

### DIFF
--- a/examples/ws_subscription.rs
+++ b/examples/ws_subscription.rs
@@ -27,7 +27,7 @@
 use jsonrpsee::{
 	types::Error,
 	ws_client::{traits::SubscriptionClient, v2::params::JsonRpcParams, Subscription, WsClientBuilder},
-	ws_server::{RpcModule, WsServerBuilder},
+	ws_server::{RpcModule, WsServerBuilder, WsStopHandle},
 };
 use std::net::SocketAddr;
 
@@ -36,24 +36,31 @@ const NUM_SUBSCRIPTION_RESPONSES: usize = 10;
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
 	env_logger::init();
-	let addr = run_server().await?;
+	let (addr, mut handle) = run_server().await?;
 	let url = format!("ws://{}", addr);
 
 	let client = WsClientBuilder::default().build(&url).await?;
-	let mut subscribe_hello: Subscription<String> =
+	let mut sub: Subscription<String> =
 		client.subscribe("subscribe_hello", JsonRpcParams::NoParams, "unsubscribe_hello").await?;
 
-	let mut i = 0;
-	while i <= NUM_SUBSCRIPTION_RESPONSES {
-		let r = subscribe_hello.next().await;
-		log::debug!("received {:?}", r);
-		i += 1;
+	let r = sub.next().await;
+	log::debug!("received {:?}", r);
+
+	handle.stop().await.unwrap();
+	handle.wait_for_stop().await;
+
+	loop {
+		match sub.next().await {
+			Ok(None) => break,
+			Ok(Some(r)) => log::debug!("recv {:?}", r),
+			Err(e) => log::debug!("recv err: {:?}", e),
+		}
 	}
 
 	Ok(())
 }
 
-async fn run_server() -> anyhow::Result<SocketAddr> {
+async fn run_server() -> anyhow::Result<(SocketAddr, WsStopHandle)> {
 	let mut server = WsServerBuilder::default().build("127.0.0.1:0").await?;
 	let mut module = RpcModule::new(());
 	module.register_subscription("subscribe_hello", "unsubscribe_hello", |_, mut sink, _| {
@@ -65,8 +72,9 @@ async fn run_server() -> anyhow::Result<SocketAddr> {
 		});
 		Ok(())
 	})?;
+	let handle = server.stop_handle();
 	server.register_module(module).unwrap();
 	let addr = server.local_addr()?;
 	tokio::spawn(async move { server.start().await });
-	Ok(addr)
+	Ok((addr, handle))
 }
--- a/examples/ws_subscription.rs
+++ b/examples/ws_subscription.rs
@@ -27,7 +27,7 @@
 use jsonrpsee::{
 	types::Error,
 	ws_client::{traits::SubscriptionClient, v2::params::JsonRpcParams, Subscription, WsClientBuilder},
-	ws_server::{RpcModule, WsServerBuilder, WsStopHandle},
+	ws_server::{RpcModule, WsServerBuilder},
 };
 use std::net::SocketAddr;
 
@@ -36,31 +36,24 @@ const NUM_SUBSCRIPTION_RESPONSES: usize = 10;
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
 	env_logger::init();
-	let (addr, mut handle) = run_server().await?;
+	let addr = run_server().await?;
 	let url = format!("ws://{}", addr);
 
 	let client = WsClientBuilder::default().build(&url).await?;
-	let mut sub: Subscription<String> =
+	let mut subscribe_hello: Subscription<String> =
 		client.subscribe("subscribe_hello", JsonRpcParams::NoParams, "unsubscribe_hello").await?;
 
-	let r = sub.next().await;
-	log::debug!("received {:?}", r);
-
-	handle.stop().await.unwrap();
-	handle.wait_for_stop().await;
-
-	loop {
-		match sub.next().await {
-			Ok(None) => break,
-			Ok(Some(r)) => log::debug!("recv {:?}", r),
-			Err(e) => log::debug!("recv err: {:?}", e),
-		}
+	let mut i = 0;
+	while i <= NUM_SUBSCRIPTION_RESPONSES {
+		let r = subscribe_hello.next().await;
+		log::debug!("received {:?}", r);
+		i += 1;
 	}
 
 	Ok(())
 }
 
-async fn run_server() -> anyhow::Result<(SocketAddr, WsStopHandle)> {
+async fn run_server() -> anyhow::Result<SocketAddr> {
 	let mut server = WsServerBuilder::default().build("127.0.0.1:0").await?;
 	let mut module = RpcModule::new(());
 	module.register_subscription("subscribe_hello", "unsubscribe_hello", |_, mut sink, _| {
@@ -72,9 +65,8 @@ async fn run_server() -> anyhow::Result<(SocketAddr, WsStopHandle)> {
 		});
 		Ok(())
 	})?;
-	let handle = server.stop_handle();
 	server.register_module(module).unwrap();
 	let addr = server.local_addr()?;
 	tokio::spawn(async move { server.start().await });
-	Ok((addr, handle))
+	Ok(addr)
 }
--- a/tests/tests/integration_tests.rs
+++ b/tests/tests/integration_tests.rs
@@ -277,5 +277,6 @@ async fn ws_close_pending_subscription_when_server_terminated() {
 
 	// no new request should be accepted.
 	assert!(matches!(sub2, Err(_)));
+	// the already established subscription should also be closed.
 	assert!(matches!(sub.next().await, Ok(None)));
 }
--- a/tests/tests/integration_tests.rs
+++ b/tests/tests/integration_tests.rs
@@ -39,7 +39,7 @@ use std::time::Duration;
 
 #[tokio::test]
 async fn ws_subscription_works() {
-	let server_addr = websocket_server_with_subscription().await;
+	let (server_addr, _) = websocket_server_with_subscription().await;
 	let server_url = format!("ws://{}", server_addr);
 	let client = WsClientBuilder::default().build(&server_url).await.unwrap();
 	let mut hello_sub: Subscription<String> =
@@ -57,7 +57,7 @@ async fn ws_subscription_works() {
 
 #[tokio::test]
 async fn ws_subscription_with_input_works() {
-	let server_addr = websocket_server_with_subscription().await;
+	let (server_addr, _) = websocket_server_with_subscription().await;
 	let server_url = format!("ws://{}", server_addr);
 	let client = WsClientBuilder::default().build(&server_url).await.unwrap();
 	let mut add_one: Subscription<u64> =
@@ -89,7 +89,7 @@ async fn http_method_call_works() {
 
 #[tokio::test]
 async fn ws_subscription_several_clients() {
-	let server_addr = websocket_server_with_subscription().await;
+	let (server_addr, _) = websocket_server_with_subscription().await;
 	let server_url = format!("ws://{}", server_addr);
 
 	let mut clients = Vec::with_capacity(10);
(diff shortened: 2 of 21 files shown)


examples/ws_subscription.rs
```diff
@@ -36,24 +36,31 @@ const NUM_SUBSCRIPTION_RESPONSES: usize = 10;
...
+	let mut sub: Subscription<String> =
 		client.subscribe("subscribe_hello", JsonRpcParams::NoParams, "unsubscribe_hello").await?;
 
-	let mut i = 0;
-	while i <= NUM_SUBSCRIPTION_RESPONSES {
-		let r = subscribe_hello.next().await;
-		log::debug!("received {:?}", r);
-		i += 1;
+	let r = sub.next().await;
+	log::debug!("received {:?}", r);
```
niklasad1, line 47: this example should be moved to test, if we agree on this :)

ws-server/src/server.rs
```diff
@@ -107,27 +108,28 @@ impl Server {
...
 						let methods = methods.clone();
-						let counter = conn_counter.clone();
 						let cfg = self.cfg.clone();
 
 						tokio::spawn(async move {
-							let r = background_task(socket, id, methods, cfg).await;
-							drop(counter);
-							r
+							background_task(socket, id, methods, cfg, tx).await
+							// TODO: remove from connections.
```
niklasad1, line 123: this is a little bit tricky, the straightforward way would be just to put the in connections in `Arc<Mutex<>>` however that won't work when it's cloned into the background thread without any further syncronization. TLDR, Arc::strong_count > 1 then drop has no effect.

ws-server/src/server.rs
```diff
@@ -228,6 +239,7 @@ async fn background_task(
 			send_error(id, &tx, code.into());
 		}
 	}
+	Err(Error::Custom("Server is terminated".to_string()))
```
popzxc, line 242: Why is it an error? Logically, stopping the server is the normal way to exit the function.
And now, if I didn't miss anything, this function never returns an `Ok` value.

popzxc (commented):
I think [this comment](https://github.com/paritytech/jsonrpsee/blob/7b95f7e779c153461e3da1d38e68e43346377ee4/ws-server/src/server.rs#L370) should be updated now to explain how server stopping will affect currently open subscriptions.

ws-server/src/server.rs
niklasad1: good point, will change

tests/tests/integration_tests.rs
```diff
@@ -256,3 +256,26 @@ async fn server_should_be_able_to_close_subscriptions() {
...
+
+	let c1 = WsClientBuilder::default().build(&server_url).await.unwrap();
+
+	let mut sub: Subscription<String> =
+		c1.subscribe("subscribe_hello", JsonRpcParams::NoParams, "unsubscribe_hello").await.unwrap();
+
+	assert!(matches!(sub.next().await, Ok(Some(_))));
+
+	handle.stop().await.unwrap();
+	handle.wait_for_stop().await;
```
jsdw, line 273: Offhand, it looks like wait_for_stop won't wait for existing connections to close as well as it'll just wait for the loop accepting new connections to end. Is this OK?

tests/tests/integration_tests.rs
```diff
@@ -256,3 +256,26 @@ async fn server_should_be_able_to_close_subscriptions() {
...
+
+	let sub2: Result<Subscription<String>, _> =
+		c1.subscribe("subscribe_hello", JsonRpcParams::NoParams, "unsubscribe_hello").await;
+
+	// no new request should be accepted.
+	assert!(matches!(sub2, Err(_)));
+	assert!(matches!(sub.next().await, Ok(None)));
```
jsdw, line 280: I think this line is also testing that the first subscription is closed, too?

niklasad1: exactly, I can add a comment for this,

It is to ensure that already established subscriptions are canceled too because they are handled a bit different but because the sink is sharing the `receiver` from the background loop and once it's dropped it causes the `sender/sink` [to fail too here](https://github.com/paritytech/jsonrpsee/blob/master/utils/src/server/rpc_module.rs#L391) I realized now that error message is wrong, I can fix that 

`Ok(None)` means that the stream was terminated in the case.

jsdw (approved):
A couple of questions, but this look good to me!

One alternate approach that comes to mind offhand (building on your nice use of closing channels) is that instead of having a locked HashMap, we could declare a single `mpsc::channel`, and clone the send side before handing it to each background_task.

If the receiver is dropped, as now, all of the cloned senders will get `is_closed() == true` and the background tasks will finish up. You'd no longer need to remove individual connections from the hashmap when they finish, and the explicit locking stuff would go away.

But, in exchange, you would need to count connections as was done previously.

tests/tests/integration_tests.rs
jsdw: That makes sense! A comment would be handy for someone like me, just so it's obvious that the test is also aiming to check that.

tests/tests/integration_tests.rs
niklasad1: right good catch, I think the `stop mutex` must be passed the background thread too for this to work properly and only be unlocked until all connections are closed.

dvdplm (commented):
It's a bit odd to me that there is nothing writing or reading from the [channel](https://github.com/paritytech/jsonrpsee/pull/396/files#diff-af87809821893f3c14406d6ab445437c729078b8283da138e9c85a80c4230418R116). If I understand this correctly the server [receives a stop signal](https://github.com/paritytech/jsonrpsee/pull/396/files#diff-af87809821893f3c14406d6ab445437c729078b8283da138e9c85a80c4230418R134) which drops the connections `HashMap`, which drops the `rx` end of the channel, which means `is_closed()` returns true so we exit the background task?

Why is this different from having an `Arc<RwLock<bool>>` flag to know if we're stopped or running?

tests/tests/integration_tests.rs
jsdw: Instead of a mutex, I think you'd need a RwLock which acquired `read`'s on the various background tasks and writes on `wait_for_stop`.

Alternately you could do what RwLock does internally using Semaphores (create a new Semaphore with say u32::MAX permits, and then acquire a permit for each background task, and on `wait_for_stop`, acquire_many u32::MAX permits (which will block until all read permits are handed back when the reads finish)

jsdw:
> It's a bit odd to me that there is nothing writing or reading from the [channel](https://github.com/paritytech/jsonrpsee/pull/396/files#diff-af87809821893f3c14406d6ab445437c729078b8283da138e9c85a80c4230418R116). If I understand this correctly the server [receives a stop signal](https://github.com/paritytech/jsonrpsee/pull/396/files#diff-af87809821893f3c14406d6ab445437c729078b8283da138e9c85a80c4230418R134) which drops the connections `HashMap`, which drops the `rx` end of the channel, which means `is_closed()` returns true so we exit the background task?
> 
> Why is this different from having an `Arc<RwLock<bool>>` flag to know if we're stopped or running?

Your understanding sounds right to me!

Good question! (or even better, an `Arc<AtomicBool>` to avoid the locking).

niklasad1:
> It's a bit odd to me that there is nothing writing or reading from the channel. If I understand this correctly the server receives a stop signal which drops the connections HashMap, which drops the rx end of the channel, which means is_closed() returns true so we exit the background task?

Yes, or the background tasks

> Why is this different from having an Arc<RwLock<bool>> flag to know if we're stopped or running?

Yeah, it's a bit stupid as you and @jsdw pointed out, but we will actually have one background task per connection so `Arc<RwLock<bool>>` would do it but if move back the `Arc<()>` conn counter that would do it.

dvdplm:
DQ: would explicitly shutting down the `tokio::runtime::Runtime` with e.g. `shutdown_timeout()` work here?

niklasad1:
> DQ: would explicitly shutting down the tokio::runtime::Runtime with e.g. shutdown_timeout() work here?

I don't think it would work because we actually need handle to the runtime to shut it down right? The user will initilize their runtime and we have no way of cancel it AFAIU. Also, I'm a little afraid of `SokettoSender` which doesn't send the `close` on drop.

jsdw:
> but I still need a Vec<AtomicBool> or Vec<Receiver<()> to shutdown all background tasks.

Wouldn't the `Arc<AtomicBool>` approach work? You could pass a clone of the bool to each background task, and check that it's false on each loop. Once it's true, the background tasks would all see that and shut down.

If you want `stop_and_wait` to wait for everything to close, you'd still need something like a `Semaphore` or `RwLock` to be handed to each background task so that `stop_and_wait` can block until everything has completely finished up. 

I was wondering whether a channel approach could satisfy both use cases (drop the receiver to signal that background tasks should close, or block on a receiver in `stop_and_wait` to wait for things to finish) but the tokio `broadcast` channel doesn't seem to have what you'd need to pull it off, so it's probably not worth finding a version that does!

ws-server/src/server.rs
```diff
@@ -95,7 +95,8 @@ impl Server {
 
 		let mut incoming = TcpListenerStream::new(self.listener).fuse();
 		let methods = self.methods;
-		let connections = Arc::new(RwLock::new(HashMap::new()));
+		let conn_counter = Arc::new(());
+		let is_conn = Arc::new(AtomicBool::new(true));
```
dvdplm, line 99: How about calling this `shutdown`? Without the `is_` or mixing in `conn` that could mean "connections", "connected", "connecting" etc?

ws-server/src/server.rs
```diff
@@ -108,20 +109,19 @@ impl Server {
...
-						if connections.read().await.len() >= self.cfg.max_connections as usize {
+						if Arc::strong_count(&conn_counter) > self.cfg.max_connections as usize {
 							log::warn!("Too many connections. Try again in a while");
 							continue;
 						}
 
-						let (tx, rx) = mpsc::channel::<()>(1);
-						connections.write().await.insert(id, rx);
+						let counter = conn_counter.clone();
+						let conn = is_conn.clone();
```
dvdplm, line 118: I don't have a hard rule to suggest here, but I think it's better to avoid renaming the clone to something that is not obviously a variant of the original. If I can't come up with a good name, I actually prefer numbering the clones, so `is_conn2` here.

dvdplm (approved):
I have a few naming grumbles but otherwise lgtm.

ws-server/src/server.rs
```diff
@@ -169,19 +173,27 @@ async fn background_task(
...
 		}
+		// terminate connection.
+		let _ = sender.close().await;
 	});
 
 	// Buffer for incoming data.
 	let mut data = Vec::with_capacity(100);
 
-	loop {
+	while is_conn.load(std::sync::atomic::Ordering::SeqCst) {
```
popzxc, line 196: Why `SeqCst`? Are ordering rules important here?
IMHO `Relaxed` will do just fine.

ws-server/src/server.rs
niklasad1: fair enough, will change.

ws-server/src/server.rs
niklasad1: right, I think it could be same sense to call them `conn_counter_rc` or something but let's going with numbering for now it's better then old stuff.

ws-server/src/server.rs
niklasad1: I have screwed up atomic ordering so bad so for now I use strongest order by default.

We can fine-tune it's too slow

popzxc: Makes sense, why not :)

ws-server/src/server.rs
```diff
@@ -89,9 +90,9 @@ impl Server {
 
 	/// Start responding to connections requests. This will block current thread until the server is stopped.
 	pub async fn start(self) {
-		// Lock the stop mutex so existing stop handles can wait for server to stop.
-		// It will be unlocked once this function returns.
-		let _stop_handle = self.stop_handle.lock().await;
+		// Acquire read access the lock such that additional reader(s) may share this lock.
```
niklasad1, line 92:
```suggestion
		// Acquire read access to the lock such that additional reader(s) may share this lock.
```

ws-server/src/server.rs
```diff
@@ -88,15 +88,16 @@ impl Server {
 
 	/// Start responding to connections requests. This will block current thread until the server is stopped.
 	pub async fn start(self) {
-		// Lock the stop mutex so existing stop handles can wait for server to stop.
-		// It will be unlocked once this function returns.
-		let _stop_handle = self.stop_handle.lock().await;
+		// Acquire read access to the lock such that additional reader(s) may share this lock.
+		// Write access to this lock will only occur until the server and all background tasks has been stopped.
```
niklasad1, line 92:
```suggestion
		// Write access to this lock will only be possible after that the server and all background tasks has been stopped.
```

ws-server/src/server.rs
```diff
@@ -89,7 +89,7 @@ impl Server {
 	/// Start responding to connections requests. This will block current thread until the server is stopped.
 	pub async fn start(self) {
 		// Acquire read access to the lock such that additional reader(s) may share this lock.
-		// Write access to this lock will only occur until the server and all background tasks has been stopped.
+		// Write access to this lock will only be possible after that the server and all background tasks has been stopped.
```
dvdplm, line 92:
```suggestion
		// Write access to this lock will only be possible after the server and all background tasks have stopped.
```

utils/src/server/rpc_module.rs
```diff
@@ -416,13 +416,12 @@ impl SubscriptionSink {
...
 	fn drop(&mut self) {
-		self.close(format!("Subscription: {} closed by the server", self.uniq_sub.sub_id));
+		self.close(format!("Subscription: {} is closed and dropped", self.uniq_sub.sub_id));
 	}
 }
 
-fn subscription_closed_by_client() -> Error {
-	const CLOSE_REASON: &str = "Subscription closed by the client";
-	Error::SubscriptionClosed(CLOSE_REASON.to_owned().into())
+fn subscription_closed_err(sub_id: u64) -> Error {
```
niklasad1, line 423: this error message was wrong, it's just the receiver was dropped in the server loop but before it was not possible to cancel, thus this fixes that.

jsdw (approved):
Looks good to me! 

We could still go for the `Arc<AtomicBool>` approach instead of the `stop_pair` channel I think (instead of the `select!` we'd check for the bool being true at the beginning of each loop) but it's not necessary and maybe a future tweak if we decide it's worth it!

niklasad1:
> We could still go for the Arc<AtomicBool> approach instead of the stop_pair channel I think (instead of the select! we'd check for the bool being true at the beginning of each loop) but it's not necessary and maybe a future tweak if we decide it's worth it!

Yeah, either way it is a little bit noisy/unclean for my taste but we could investigate if it's some performance gain there.

However, I was thinking about the refactoring `async fn start(self)` and do tokio::spawn `fn start(self) -> Handle { tokio::task::JoinHandle, mpsc::Channel/AtomicBool}` but we still need to have `channel/AtomicBool` to cancel to loop and background tasks but we could at least the rid of the `Mutex/RwLock` in this way

dvdplm (approved):
It's not pretty but it'll do for now. :)

PR #397, opened 2021-06-30
ws client redirections

### LINKED ISSUE #339: [ws client]: support HTTP redirections.

niklasad1:
Closes #339


ws-client/src/transport.rs
```diff
@@ -24,11 +24,13 @@
 // IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 // DEALINGS IN THE SOFTWARE.
 
+use crate::stream::EitherStream;
 use crate::tokio::{TcpStream, TlsStream};
 use futures::io::{BufReader, BufWriter};
 use futures::prelude::*;
 use soketto::connection;
 use soketto::handshake::client::{Client as WsRawClient, ServerResponse};
```
dvdplm, line 32: Can we use `SokettoClient` rather than `WsRawClient`?

ws-client/src/transport.rs
```diff
@@ -174,64 +176,80 @@ impl<'a> WsTransportClientBuilder<'a> {
...
+								Mode::Plain => None,
+							};
 						}
-					}
+						// redirection is relative, either `/baz` or `bar`.
+						Err(_) => {
+							// replace the entire path if `location` is `/`.
+							if location.starts_with('/') {
+								path = location.into();
+							} else {
```
dvdplm, line 241: This, in combination with `Err(_)` seems like it could be a bit dangerous. What if the `location` is the empty string such that we end up with the same path as we had before the redirect? Or if the [error](https://docs.rs/url/2.2.2/url/enum.ParseError.html) is `Overflow` and we append 4Gb of data here? Maybe I'm being paranoid but I would feel better if we did explicit matching on the `ParseError` variants we know are safe to use here?

ws-client/src/transport.rs
```diff
@@ -174,64 +176,80 @@ impl<'a> WsTransportClientBuilder<'a> {
...
-							TlsOrPlain::Tls(tls_stream)
+		let mut sockaddrs = self.target.sockaddrs;
+		let mut path = PathBuf::from(self.target.path);
+		let mut host = self.target.host;
+		let mut host_header = self.target.host_header;
+
+		let mut err = Err(None);
+
+		let client = loop {
+			let sockaddr = match sockaddrs.pop() {
```
dvdplm, line 194: dq: shouldn't we try all the addresses in `sockaddrs` in case the `connect` call fails?

ws-client/src/transport.rs
```diff
@@ -174,64 +176,80 @@ impl<'a> WsTransportClientBuilder<'a> {
...
-							let tls_stream = connector.connect(dns_name, socket).await?;
-							TlsOrPlain::Tls(tls_stream)
+		let mut sockaddrs = self.target.sockaddrs;
+		let mut path = PathBuf::from(self.target.path);
+		let mut host = self.target.host;
+		let mut host_header = self.target.host_header;
+
+		let mut err = Err(None);
+
+		let client = loop {
```
dvdplm, line 193: I think we should have a `max_redirects` threshold set to a rather low value by default (5?) and refuse to follow further redirects. I'm ok having a constant too. I'd rather err on the side of too few redirects to risk exposing the client to redirect loops.

dvdplm:
Left a few thoughts!

ws-client/src/transport.rs
niklasad1: > What if the location is the empty string such that we end up with the same path as we had before the redirect? 

That must be some violation of HTTP spec to send back the same path as in the original request (but I haven't checked it...)

> Or if the error is Overflow and we append 4Gb of data here? Maybe I'm being paranoid but I would feel better if we did explicit matching on the ParseError variants we know are safe to use here?

fair enough, I agree with matching on the error and just make this assumption if the error is `RelativeUrl_x` then.

ws-client/src/transport.rs
```diff
@@ -241,6 +259,34 @@ impl<'a> WsTransportClientBuilder<'a> {
...
+async fn connect(
+	sockaddr: SocketAddr,
+	timeout_dur: Duration,
+	host: &str,
+	tls_connector: &Option<crate::tokio::TlsConnector>,
+) -> Result<EitherStream<TcpStream, TlsStream<TcpStream>>, WsHandshakeError> {
+	let socket = TcpStream::connect(sockaddr);
+	let timeout = crate::tokio::sleep(timeout_dur);
+	futures::pin_mut!(socket, timeout);
+	Ok(match future::select(socket, timeout).await {
```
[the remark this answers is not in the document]
niklasad1: sure, it should work

ws-client/src/transport.rs
niklasad1: right thanks

ws-client/src/transport.rs
dvdplm: > That must be some violation of HTTP spec to send back the same path as in the original request (but I haven't checked it...)

I am sure it is indeed a violation. And also that there are many many misconfigured redirects out there, servers ping-ponging requests in eternity...

ws-client/src/transport.rs
niklasad1: I was wrong here, basically if pop fails it means that we have tried all sockaddrs.

dvdplm: Hmm, when `pop()` succeeds here we proceed to `connect()` (L199), but if that fails we exit the `try_connect()` function (because of `.await?`). Shouldn't we instead try with the next item in `sockaddrs`?

ws-client/src/transport.rs
niklasad1: renamed it WsHandshakeClient I think it's more descriptive...

ws-client/src/transport.rs
```diff
@@ -179,74 +183,118 @@ impl<'a> WsTransportClientBuilder<'a> {
...
-							let tls_stream = connector.connect(dns_name, socket).await?;
-							TlsOrPlain::Tls(tls_stream)
-						}
-					}
-				}
-				future::Either::Right((_, _)) => return Err(WsHandshakeError::Timeout(self.timeout)),
+		let mut sockaddrs = self.target.sockaddrs;
+		let mut path_and_query = PathBuf::from(self.target.path_and_query);
+		let mut host = self.target.host;
+		let mut host_header = self.target.host_header;
```
maciejhirsz, line 196: Why not just:

```rust
let mut target = self.target;
```

ws-client/src/transport.rs
```diff
@@ -179,74 +183,118 @@ impl<'a> WsTransportClientBuilder<'a> {
...
+				ServerResponse::Redirect { status_code, location } => {
+					log::trace!("recv redirection: status_code: {}, location: {}", status_code, location);
+					match url::Url::parse(&location) {
+						// redirection with absolute path => need to lookup.
+						Ok(url) => {
+							let target = Target::parse(url)?;
+							sockaddrs = target.sockaddrs;
+							path_and_query = PathBuf::from(target.path_and_query);
+							host = target.host;
+							host_header = target.host_header;
```
maciejhirsz, line 236: With the change above this could be just:

```rust
target = Target::parse(url)?;
```

Would also switch `path_and_query` to be `PathBuf` on the `Target` struct so there is no need to cast it here.

maciejhirsz (commented):
Some grumbles

ws-client/src/transport.rs
niklasad1: yupp, that makes sense fixed.

ws-client/src/transport.rs
niklasad1: ty

test-utils/Cargo.toml
```diff
@@ -15,6 +15,6 @@ hyper = { version = "0.14.10", features = ["full"] }
 log = "0.4"
 serde = { version = "1", default-features = false, features = ["derive"] }
 serde_json = "1"
-soketto = "0.6"
+soketto = { version = "0.7", features = ["http"] }
```
dvdplm, line 18: Do you think we should upgrade `soketto` in other places as well?

test-utils/src/types.rs
```diff
@@ -314,3 +314,57 @@ async fn connection_task(socket: tokio::net::TcpStream, mode: ServerMode, mut ex
 		}
 	}
 }
+
```
dvdplm, line 317: This file is called `types.rs` which isn't really what it is about anymore. Perhaps merge it with `helpers.rs`?

test-utils/src/types.rs
```diff
@@ -314,3 +314,57 @@ async fn connection_task(socket: tokio::net::TcpStream, mode: ServerMode, mut ex
...
+// Run a `WebSocket server` that performs some re-directions for testing.
```
dvdplm, line 318:
```suggestion
// Run a WebSocket server running on localhost that redirects requests for testing.
// Requests to any url except for `/myblock/two` will redirect one or two times (HTTP 301) and eventually end up in `/myblock/two`.
```

ws-client/src/tests.rs
```diff
@@ -263,3 +263,18 @@ fn assert_error_response(err: Error, exp: ErrorObject) {
...
+		.with_default_timeout()
+		.await
+		.unwrap();
+
+	let server_url = format!("ws://{}", server.local_addr());
+	let redirect_url = jsonrpsee_test_utils::types::ws_server_with_redirect(server_url);
+
+	// The client will first connect to a server that only performs re-directions and finally
+	// redirect to another server to complete the handshake.
+	assert!(matches!(WsClientBuilder::default().build(&redirect_url).with_default_timeout().await, Ok(Ok(_))));
```
dvdplm, line 279: We should probably assert on the expected outcome as well, whatever is returned by `with_hardcoded_response`.

ws-client/src/transport.rs
```diff
@@ -37,7 +38,10 @@ use tokio_rustls::{
 	TlsConnector,
 };
 
-type TlsOrPlain = crate::stream::EitherStream<TcpStream, TlsStream<TcpStream>>;
+type TlsOrPlain = EitherStream<TcpStream, TlsStream<TcpStream>>;
+
+// TODO(niklasad1): make this configurable.
```
dvdplm, line 43: Add an issue and toss it on the `0.5` tracking ticket? :)

ws-client/src/transport.rs
```diff
@@ -184,74 +188,137 @@ impl<'a> WsTransportClientBuilder<'a> {
...
+					continue;
 				}
+			};
+			let mut client = WsHandshakeClient::new(
+				BufReader::new(BufWriter::new(tcp_stream)),
+				&target.host_header,
+				&target.path_and_query,
+			);
+			if let Some(origin) = self.origin_header.as_ref() {
+				client.set_origin(origin);
```
dvdplm, line 228: This breaks in `soketto` 0.7, not sure if upgrading in this PR is appropriate or not.

dvdplm:
@niklasad1 Struggling to get the redirect test working with windows, it fails with `"thread 'tests::redirections' panicked at 'WsClient builder failed with: Transport(No IP address found for this hostname: 0)', ws-client\src\tests.rs:287:23"`. 
I'll ignore it on windows for now.

dvdplm (approved):
I'm happy with this. The redirect test fails on windows and this PR pulls in `soketto` v0.7 in addition to v0.6, but those grumbles aside it lgtm.

test-utils/Cargo.toml
niklasad1: yeah, we probably should. It just needs some refactoring because the soketto API now requires `httparse::Header` instead of `&str`

ws-client/src/tests.rs
```diff
@@ -263,3 +263,34 @@ fn assert_error_response(err: Error, exp: ErrorObject) {
...
+	// redirect to another server to complete the handshake.
+	let client = WsClientBuilder::default().build(&redirect_url).with_default_timeout().await;
+	// It's an ok client
+	let client = match client {
+		Ok(Ok(client)) => client,
+		Ok(Err(e)) => panic!("WsClient builder failed with: {:?}", e),
+		Err(e) => panic!("WsClient builder timed out with: {:?}", e),
+	};
+	// It's connected
+	assert!(client.is_connected());
```
niklasad1, line 292: :+1:

niklasad1:
> @niklasad1 Struggling to get the redirect test working with windows, it fails with "thread 'tests::redirections' panicked at 'WsClient builder failed with: Transport(No IP address found for this hostname: 0)', ws-client\src\tests.rs:287:23".
I'll ignore it on windows for now.

Weird, especially that the hostname is `0`. I would expect it to be `127.0.0.1` ^

niklasad1:
> Weird, especially that the hostname is 0. I would expect it to be 127.0.0.1 ^

Ah, I figured it out basically `0`just prints 0 this was a `thiserror nit` :see_no_evil:

ws-client/src/transport.rs
```diff
@@ -184,74 +187,135 @@ impl<'a> WsTransportClientBuilder<'a> {
...
+							Err(
+								url::ParseError::RelativeUrlWithoutBase
+								| url::ParseError::RelativeUrlWithCannotBeABaseBase,
+							) => {
+								// replace the entire path if `location` is `/`.
+								if location.starts_with('/') {
+									target.path_and_query = location;
+								} else {
+									// join paths such that the leaf is replaced with `location`.
+									let strip_last_child = Path::new(&target.path_and_query)
```
niklasad1, line 268: we need to have a slightly modified version of this because on Windows the paths are appended with `\\` instead of `/` which is not according to the HTTP relative path.

ws-client/src/transport.rs
```diff
@@ -184,74 +195,132 @@ impl<'a> WsTransportClientBuilder<'a> {
...
+											Some(Arc::new(client_config).into())
+										}
+										Mode::Plain => None,
+									};
+									break;
+								}
+								// Relative URI.
+								else {
+									// Replace the entire path_and_query if `location` starts with `/` or `//`.
+									if location.starts_with('/') {
```
niklasad1, line 267: TODO: what if the `URL` had a query part?

Should that be added by us here too?

ws-client/Cargo.toml
```diff
@@ -21,12 +21,13 @@ jsonrpsee-types = { path = "../types", version = "0.3.0" }
 log = "0.4"
 serde = "1"
 serde_json = "1"
-soketto = "0.6"
+soketto = "0.7"
 pin-project = "1"
 thiserror = "1"
-url = "2"
+http = "0.2"
```
niklasad1, line 27: Switched to the `http crate` instead for URI parsing because it supports both `absolute URI` and `relative resource URI`

Useful when getting the location header back...

ws-client/src/tests.rs
```diff
@@ -263,3 +263,35 @@ fn assert_error_response(err: Error, exp: ErrorObject) {
 		e => panic!("Expected error: \"{}\", got: {:?}", err, e),
 	};
 }
+
+//#[cfg_attr(target_os = "windows", ignore)]
```
niklasad1, line 267:
```suggestion
```

ws-client/src/transport.rs
```diff
@@ -184,74 +195,132 @@ impl<'a> WsTransportClientBuilder<'a> {
...
+										target.path_and_query = location;
+									} else {
+										match target.path_and_query.rfind('/') {
```
niklasad1, line 270: //cc @maciejhirsz I had to do this instead of `Path::join` because on `Windows` the path is joined by `\` which is a faulty `URI relative resource`.

This will work for both `//` and `/` FWIW.

test-utils/src/mocks.rs
```diff
@@ -314,3 +312,60 @@ async fn connection_task(socket: tokio::net::TcpStream, mode: ServerMode, mut ex
...
+
+/// Handle incoming HTTP Requests.
+async fn handler(
+	req: hyper::Request<Body>,
+	other_server: String,
+) -> Result<hyper::Response<Body>, soketto::BoxedError> {
+	if is_upgrade_request(&req) {
+		log::debug!("{:?}", req);
+
+		match req.uri().path() {
```
TarikGul, line 345: Ahh okay the above description makes sense after reading this! 👍

TarikGul (approved):
lgtm

maciejhirsz (approved):
Looking pretty clean!

PR #401, opened 2021-07-01
Streaming RpcParams parsing

maciejhirsz:
This is an alternative to #391.

`RpcParams` now has two new methods, `next` and `optional_next`, which yields the next value from the JSON array `RpcParams` was created from. `optional_next` will return `None` for `null` or for missing elements (it will still error when parsing a wrong type or invalid JSON).

### DIFF
--- a/types/src/v2/params.rs
+++ b/types/src/v2/params.rs
@@ -133,10 +133,10 @@ impl<'a> RpcParams<'a> {
 	/// let mut params = RpcParams::new(Some(r#"[1, 2, null]"#));
 	///
 	/// let params: [Option<u32>; 4] = [
-	/// 	params.optional_next().unwrap(),
-	/// 	params.optional_next().unwrap(),
-	/// 	params.optional_next().unwrap(),
-	/// 	params.optional_next().unwrap(),
+	///     params.optional_next().unwrap(),
+	///     params.optional_next().unwrap(),
+	///     params.optional_next().unwrap(),
+	///     params.optional_next().unwrap(),
 	/// ];;
 	///
 	/// assert_eq!(params, [Some(1), Some(2), None, None]);
--- a/types/src/v2/params.rs
+++ b/types/src/v2/params.rs
@@ -165,8 +165,7 @@ impl<'a> RpcParams<'a> {
 	where
 		T: Deserialize<'a>,
 	{
-		let params = self.0.unwrap_or("null");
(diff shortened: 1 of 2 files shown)


types/src/v2/params.rs
```diff
@@ -68,7 +68,90 @@ impl<'a> RpcParams<'a> {
...
+	/// let mut params = RpcParams::new(Some(r#"[1, 2, null]"#));
+	///
+	/// let params: [Option<u32>; 4] = [
+	///     params.optional_next().unwrap(),
+	///     params.optional_next().unwrap(),
+	///     params.optional_next().unwrap(),
+	///     params.optional_next().unwrap(),
+	/// ];;
+	///
+	/// assert_eq!(params, [Some(1), Some(2), None, None]);
```
dvdplm, line 142: Does this work when calling `parse()` directly? Like

```rust
let in_params = RpcParams::new(Some(r#"[1, 2, null]"#));
let params: [Option<u32>; 4] = in_params.parse().unwrap();
assert_eq!(params, [Some(1), Some(2), None, None]);
```
?

dvdplm (approved):

types/src/v2/params.rs
maciejhirsz: We'd need trait specialization to be able to do that unfortunately.

types/src/v2/params.rs
```diff
@@ -68,7 +68,90 @@ impl<'a> RpcParams<'a> {
...
+	{
+		let mut json = self.0?.trim_start();
+
+		match json.as_bytes().get(0)? {
+			b']' => {
+				self.0 = None;
+
+				return None;
+			}
+			b'[' | b',' => json = &json[1..],
```
niklasad1, line 83: this could panic if the `str slice` is something like `[`, `,`, `[∀`

technically I think, a nasty client could send it erroneous json objects as params and we would accept them?

niklasad1: looks like I was wrong

niklasad1 (approved):

niklasad1:
Clean as always but we need add some additional code for this in macros to work smoothly ^

types/src/v2/params.rs
maciejhirsz: Slicing `"["` at offset `1` gives you an empty slice, and there is no issue slicing in front of a non-ascii character since it's at boundary there.

niklasad1: yeah right, I guess I'm getting tired of code reading today.

PR #367, opened 2021-06-04
[client]: add timeouts to all request kinds + default timeout

niklasad1:
Add default timeout on the all the clients requests to 60 seconds.

The PR itself is !DRY of  in a few places.

PR #406, opened 2021-07-06
Timeouts for all requests

dvdplm:
Variant of #367

This PR takes a more opinionated stance than #367, where timeouts are optional. In this PR I suggest we make a all requests use a timeout and only let users choose the length.

### DIFF
--- a/ws-client/Cargo.toml
+++ b/ws-client/Cargo.toml
@@ -11,18 +11,18 @@ documentation = "https://docs.rs/jsonrpsee-ws-client"
 
 [dependencies]
 # Tokio v1 deps
-tokioV1 = { package="tokio", version = "1", features = ["net", "time", "rt-multi-thread"], optional = true }
+tokioV1 = { package="tokio", version = "1", features = ["net", "time", "rt-multi-thread", "macros"], optional = true }
 tokioV1-rustls = { package="tokio-rustls", version = "0.22", optional = true }
 tokioV1-util = { package="tokio-util", version = "0.6", features = ["compat"], optional = true }
 
 # Tokio v0.2 deps
-tokioV02 = { package="tokio", version = "0.2", features = ["net", "time", "rt-threaded", "sync"], optional = true }
+tokioV02 = { package="tokio", version = "0.2", features = ["net", "time", "rt-threaded", "sync", "macros"], optional = true }
 tokioV02-rustls = { package="tokio-rustls", version = "0.15", optional = true }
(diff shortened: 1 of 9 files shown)


ws-client/src/client.rs
```diff
@@ -313,7 +315,16 @@ impl Client for WsClient {
...
+
+		let mut sender = self.to_back.clone();
+		let fut = sender.send(FrontToBack::Notification(raw));
+
+		let timeout = crate::tokio::sleep(self.request_timeout);
+		futures::pin_mut!(fut, timeout);
+		let res = match futures::future::select(fut, timeout).await {
+			futures::future::Either::Left((res, _)) => res,
+			futures::future::Either::Right((_, _)) => return Err(Error::RequestTimeout),
+		};
```
maciejhirsz, line 327: I reckon using `futures::select!` would improve readability here.

ws-client/src/helpers.rs
```diff
@@ -188,3 +189,16 @@ pub fn process_error_response(manager: &mut RequestManager, err: JsonRpcError) -
...
 	}
 }
+
+/// Wait for a stream to complete within the given timeout.
+pub async fn call_with_timeout<T>(
+	timeout: Duration,
+	rx: oneshot::Receiver<Result<T, Error>>,
+) -> Result<Result<T, Error>, oneshot::Canceled> {
+	let timeout = crate::tokio::sleep(timeout);
+	futures::pin_mut!(rx, timeout);
```
maciejhirsz, line 199: ditto about `select!` macro

maciejhirsz (approved):
Looks good, minor grumble.

ws-client/Cargo.toml
```diff
@@ -22,7 +22,7 @@ tokioV02-util = { package="tokio-util", version = "0.3", features = ["compat"],
 
 async-trait = "0.1"
 fnv = "1"
-futures = { version = "0.3.14", default-features = false, features = ["std"] }
+futures = { version = "0.3.14", default-features = false, features = ["std", "async-await"] }
```
dvdplm, line 25: @maciejhirsz I wonder if this was the reason for not using `select!`, to keep the required features down?

maciejhirsz: Interesting. The `select!` in futures also seems to have some weird requirements. Tokio has a `select!` macro that works identical, but doesn't require any `.fuse()` or `pin_mut!()` calls.

Issue #407, opened 2021-07-07
Access CallError without macros


That3Percent:
If you need to access `CallError` (which is used in the signature of `register_async_method`) the only path to do so is through `types::error::CallError`. But, the `types` crate is only available if the `macros` feature is included.

This is undesirable. Instead, `CallError` should be re-exported from the `http_server` crate, or the `types` crate should be available when the `server` feature is specified.

dvdplm:
I agree and have had the exact same thought.

PR #409, opened 2021-07-09
Rexport types for servers

dvdplm:
Include the "macros" feature in the the "server" feature so users don't have to import "jsonrpsee-types" explicitly.

Also re-export the `SubscriptionSink`.

This is possibly slightly contentious. I think it's fair to assume that people writing servers will want to use the "macros" feature (which implies "types"). The downside of this is that "macros" includes the client side macros and "types" also includes types used in clients more.

Alternative solutions would be
- include the "types" feature in "server" instead (downside: you get both [client](https://github.com/paritytech/jsonrpsee/blob/dd4a17ba5e92091458805ad8017c2926cd28d1fa/types/src/client.rs) and server types)
- granularly re-export common types like `CallError` or `SubscriptionSink` (downside: more cfg-fu)
- `<insert-yours />`

Fixes https://github.com/paritytech/jsonrpsee/issues/407

Substrate "companion": https://github.com/paritytech/substrate/compare/dp-jsonrpsee-integration-2...dp-tweak-reexports?expand=1


jsonrpsee/Cargo.toml
```diff
@@ -20,6 +20,6 @@ types = { path = "../types", version = "0.2.0", package = "jsonrpsee-types", opt
 
 [features]
 client = ["http-client", "ws-client"]
-server = ["http-server", "ws-server", "utils"]
+server = ["http-server", "ws-server", "utils", "macros"]
```
niklasad1, line 23: I think types should be sufficient here, no?

jsonrpsee/Cargo.toml
```diff
@@ -20,6 +20,6 @@ types = { path = "../types", version = "0.2.0", package = "jsonrpsee-types", opt
 
 [features]
 client = ["http-client", "ws-client"]
```
niklasad1, line 22: perhaps we should add types here too? however types are re-exported from the clients

niklasad1 (commented):
I think we should re-export the types from the servers too inorder for the crates to be used as independent crates but the most common usage should be from the facade crate anyway.

jsonrpsee/Cargo.toml
dvdplm: "types" is sufficient, but I'm a bit concerned that [this](https://github.com/paritytech/jsonrpsee/blob/dd4a17ba5e92091458805ad8017c2926cd28d1fa/jsonrpsee/src/lib.rs#L28-L35) will become messy. I'd like to be able to look at `lib.rs` and see at a glance where everything is coming from.

http-server/src/lib.rs
```diff
@@ -40,7 +40,7 @@ pub use access_control::{
 	hosts::{AllowHosts, Host},
 	AccessControl, AccessControlBuilder,
 };
-pub use jsonrpsee_types::{Error, TEN_MB_SIZE_BYTES};
+pub use jsonrpsee_types::*;
```
maciejhirsz, line 43: Wouldn't it make more sense to re-export types in namespace like: `pub use jsonrpsee_types as types`?

dvdplm: Yes I agree. I went ahead and did that so it's a lot more noise here now.

maciejhirsz (approved):

jsonrpsee/src/lib.rs
```diff
@@ -18,14 +18,14 @@ pub use ws_server;
...
-pub use utils::server::rpc_module::RpcModule;
+pub use utils::server::rpc_module::{RpcModule, SubscriptionSink};
 
 /// Procedural macros for JSON RPC implementations.
 #[cfg(feature = "macros")]
 pub use proc_macros;
 
 /// Common types used to implement JSON RPC server and client.
-#[cfg(feature = "macros")]
+#[cfg(any(feature = "types", feature = "macros"))]
```
niklasad1, line 28: :+1:

ws-client/src/lib.rs
```diff
@@ -27,4 +27,4 @@ mod tokio;
 mod tests;
 
 pub use client::{WsClient, WsClientBuilder};
-pub use jsonrpsee_types::*;
+pub use jsonrpsee_types as types;
```
niklasad1, line 30: :+1

niklasad1 (approved):

PR #410, opened 2021-07-09
Pass OwnedRpcParams to async methods

### LINKED ISSUE #408: Cannot parse `RpcParams` on another thread
JSON parsing is a surprisingly expensive operation. It is desirable to be able to parse JSON in the thread which handles the request, rather than the "main" thread producing requests. Otherwise, throughput is limited.

I would like to be able to write:
```rust
module.register_async_method("challenge", move |rpc_params, context| {
    return (async move {
        let outcome = tokio::spawn(async move {
            let challenge = rpc_params.parse()?;
```

But instead need to write:
```rust
module.register_async_method("challenge", move |rpc_params, context| {
    let challenge = rpc_params.parse();
    return (async move {
        let outcome = tokio::spawn(async move {
            let challenge = challenge?;
```
because of the lifetime in `RpcParams`.

One could `.to_owned()` to get an `OwnedRpcParams` but this is an unnecessary performance drain in it's own right.

The `register_async_method` implementation already has access to an `OwnedRpcParams` which it does not make use of. It would be preferable to transfer ownership to the callback to avoid the unnecessary performance overhead.

dvdplm:
Pass `OwnedRpcParams` to async methods to allow deserializing to happen asynchronously as well.

Closes #408﻿

### DIFF
--- a/types/src/v2/params.rs
+++ b/types/src/v2/params.rs
@@ -58,7 +58,10 @@ impl Serialize for TwoPointZero {
 	}
 }
 
-/// Parameters sent with the RPC request
+/// Parameters sent with the RPC request.
+///
+/// The data containing the params is a `Cow<&str>` and can either be a borrowed `&str` of JSON from an incoming [`JsonRpcRequest`] (which in
+/// turn borrows it from the input buffer that is shared between requests); or, it can be an owned `String`.
 #[derive(Clone, Debug)]
 pub struct RpcParams<'a>(Option<Cow<'a, str>>);
 
@@ -68,12 +71,15 @@ impl<'a> RpcParams<'a> {
 		Self(raw.map(Into::into))
 	}
 
-	/// Obtain a sequence parser
+	/// Obtain a sequence parser, [`RpcParamsSequence`].
+	///
+	/// This allows sequential parsing of the incoming params, using an `Iterator`-style API and is useful when the RPC
+	/// request has optional parameters at the tail that may or may not be present.
 	pub fn sequence(&self) -> RpcParamsSequence {
 		RpcParamsSequence(self.0.as_ref().map(AsRef::as_ref).unwrap_or(""))
 	}
 
-	/// Attempt to parse all parameters as array or map into type `T`
+	/// Attempt to parse all parameters as an array or map into type `T`.
 	pub fn parse<T>(&'a self) -> Result<T, CallError>
 	where
 		T: Deserialize<'a>,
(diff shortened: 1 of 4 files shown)


utils/src/server/rpc_module.rs
```diff
@@ -215,7 +215,7 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 	pub fn register_async_method<R, F>(&mut self, method_name: &'static str, callback: F) -> Result<(), Error>
 	where
 		R: Serialize + Send + Sync + 'static,
-		F: Fn(RpcParams, Arc<Context>) -> BoxFuture<'static, Result<R, CallError>> + Copy + Send + Sync + 'static,
+		F: Fn(OwnedRpcParams, Arc<Context>) -> BoxFuture<'static, Result<R, CallError>> + Copy + Send + Sync + 'static,
```
dvdplm, line 218: This is the important change.

types/src/v2/params.rs
```diff
@@ -167,23 +167,28 @@ impl<'a> RpcParams<'a> {
...
 /// Owned version of [`RpcParams`].
 #[derive(Clone, Debug)]
 pub struct OwnedRpcParams(Option<String>);
 
 impl OwnedRpcParams {
-	/// Converts `OwnedRpcParams` into borrowed [`RpcParams`].
-	pub fn borrowed(&self) -> RpcParams<'_> {
-		RpcParams(self.0.as_ref().map(|s| s.as_ref()))
+	/// Attempt to parse all parameters as array or map into type `T`
+	pub fn parse<T>(self) -> Result<T, CallError>
```
dvdplm, line 178: The reason this is a draft is that I'm not sure if there's a better way to do this than duplicating code from `RpcParams`.

maciejhirsz:
Might have to mess a bit more with lifetimes, or do an extra quasi iterator for parsing using `next|optional_next`, because as things are now this is not viable:

```rust
let a: &str = params.next();
let b: &str = params.next(); // a is using a lifetime that's bound to a mutable borrow, hence borrowck forbids this thinking it's aliasing
```

types/src/v2/params.rs
```diff
@@ -152,49 +164,30 @@ impl<'a> RpcParams<'a> {
 	}
 
 	/// Attempt to parse all parameters as array or map into type `T`
-	pub fn parse<T>(self) -> Result<T, CallError>
+	pub fn parse<T>(&'a self) -> Result<T, CallError>
 	where
 		T: Deserialize<'a>,
 	{
-		let params = self.0.unwrap_or("null");
+		let params = self.json.as_ref().map(AsRef::as_ref).unwrap_or("null");
```
dvdplm, line 166: Will the `unwrap_or` allocate here if we're a `Cow::Owned`?

maciejhirsz: No, it's casting it to `Option<&str>` and then unwrapping.

types/src/v2/params.rs
```diff
@@ -399,6 +398,16 @@ mod test {
...
 	}
 
+	#[test]
+	fn params_sequence_borrows() {
+		let params = RpcParams::new(Some(r#"["foo", "bar"]"#));
+		let mut seq = params.sequence();
+
+		assert_eq!(seq.next::<&str>().unwrap(), "foo");
+		assert_eq!(seq.next::<&str>().unwrap(), "bar");
+		assert!(seq.next::<&str>().is_err());
```
maciejhirsz, line 408: This is the important test and the reason for adding a separate struct for handling sequence parsing: `seq` borrows the json from `params`, which allows it to produce slices with that lifetime, while `&mut self` borrows are transient.

PR #412, opened 2021-07-11
Module API refactor

maciejhirsz:
Since we want to support both HTTP and WS servers sharing the same `Methods` internally, it makes more sense to force the user to pre-merge all their modules into one before cloning it and passing it to their respective servers. This PR changes some of the module API to make this the expected behavior.

+ Removed `register_module` on both WS and HTTP server, instead when starting a server you can pass a single `Into<Methods>` type.
+ `RpcModule<T>` implements `Deref<Target = Methods>` and `DerefMut` to keep module merging logic DRY.
+ Creating modules via proc macro should be infallible (can be checked statically) so `into_rpc` need not return a `Result`.
+ `method_names` returns an `Iterator` instead of a `Vec`.


dvdplm (commented):
Still a draft but here are my thought so far:
- I think that in the back of our heads we had in mind a setup flow where a ref to the `Server` would be passed around to different subsystems and each would `register_module()` to add "their" RPC calls. Instead what substrate does is the opposite, the subsystems are all asked for their sets of RPC calls and when they're all collected they are added to the server in a single place. (This architecture is what is making initialization tricky to do in more than one place, such as adding specific RPC modules on the basis of config files that are processed in different places. This is hackfix-solved by passing around closures.)
- This work here makes it more obvious that setting up the server and the methods to call are quite separate things. I think it makes sense (and the DRYing is welcome obviously). 
- A UI-test for a bad macro setup would be nice, one proving that the static method check works as expected.
- I might be wrong but I'm not sure we have good tests for setting up and starting both a ws and http server; would be good to have one.

proc-macros/src/new/render_server.rs
```diff
@@ -47,70 +48,99 @@ impl RpcDescription {
...
+			})
+			.collect::<Vec<_>>();
+
+		let subscriptions = self
+			.subscriptions
+			.iter()
+			.map(|sub| {
+				// Rust method to invoke (e.g. `self.<foo>(...)`).
+				let rust_method_name = &sub.signature.sig.ident;
+				// Name of the RPC method to subscribe (e.g. `foo_sub`).
```
dvdplm, line 105:
```suggestion
				// Name of the RPC method to subscribe to (e.g. `foo_sub`).
```

proc-macros/src/new/render_server.rs
```diff
@@ -47,70 +48,99 @@ impl RpcDescription {
...
+					#(#methods)*
+					#(#subscriptions)*
 
-				#(#methods)*
-				#(#subscriptions)*
+					Ok(rpc)
+				};
 
-				Ok(rpc)
+				inner().expect("Proc macro method names should never conflict")
```
dvdplm, line 143:
```suggestion
				inner().expect("RPC macro method names should never conflict")
```

proc-macros/tests/ui/incorrect/rpc/rpc_name_conflict.rs
```diff
@@ -0,0 +1,13 @@
+use jsonrpsee::proc_macros::rpc;
+
+// Associated items are forbidden.
+#[rpc(client, server)]
+pub trait MethodNameConflict {
+	#[method(name = "foo")]
+	async fn foo(&self) -> u8;
+
+	#[method(name = "foo")]
```
dvdplm, line 9: Ok, but this is the easy case. Let's do two separate modules?

dvdplm (approved):
I like it. Minor grumbles.

proc-macros/tests/ui/incorrect/rpc/rpc_name_conflict.rs
maciejhirsz: Separate modules need to be merged on runtime and will return errors, there is no way to check them on compile time.

dvdplm: Good pt!

PR #414, opened 2021-07-12
Only return the crate name when the crate is `jsonrpsee` as well

dvdplm:
I missed one place where "types" was added to the crate name instead of when the types are actually used. Only pops up when the crate name searched for is "jsonrpsee".﻿


jsdw (commented):
Offhand (though github is being slow for me), it looks like CI caught the problem with the last PR? If that's the case then this looks good given that CI has passed, but I haven't locally tested the various path changes (they all look minor and like things that should be caught though).

maciejhirsz (approved):

dvdplm:
> it looks like CI caught the problem with the last PR?

Yup, I did all the wrong things there, all of them!

PR #419, opened 2021-07-15
Sniff the first byte to glean if the incoming request is a single or batch request

### LINKED ISSUE #296: `#[serde(untagged)]` doesn't work with `RawValue`
For reasons outlined [here](https://github.com/serde-rs/json/issues/497), `RawValue` can't be
used with untagged enums at the moment. 
This means we can't use an `SingleOrBatch` untagged enum to handle incoming requests that can be one of:
- single JSONRPC request, i.e. `{"jsonrpc":"2.0","method":"bar","id":1}`
- a batch of requests, i.e. `[{"jsonrpc":"2.0","method":"baz","id":1}, {"jsonrpc":"2.0","method":"baz","id":2}]`
- a single notification, i.e. `{"jsonrpc":"2.0","method":"bezz"}`
- a batch of notifications, i.e. `[{"jsonrpc":"2.0","method":"bezz"}, {"jsonrpc":"2.0","method":"bozz"}, {"jsonrpc":"2.0","method":"bizz"}]`

(We do not support notifications, but if we ever were to do that to be fully compliant with the spec, this problem would make things worse)

Instead we have to try each case individually: first the single request case, then the batch case and then one time more to check for the error. For the worst case – unparseable input – we make three calls to [`serde_json::from_slice`] which is pretty annoying.

There is a PoC fix for this in https://github.com/serde-rs/serde/pull/1354

Suggestion: revisit this when/if https://github.com/serde-rs/serde/pull/1354 lands.

dvdplm:
This works around the serde limitations around `untagged` enums and `RawValue` by instead looking at the first byte of the request: if it's a `[` then it's a batch request; if it's a `{` then it's a single request.

The http case is less nice because there we have notifications (i.e. requests without an id, requiring no response).

Also see #420 as an alternative to this.

Closes #296

### DIFF
--- a/http-server/src/server.rs
+++ b/http-server/src/server.rs
@@ -217,9 +217,9 @@ impl Server {
 						let mut single = true;
 						// TODO: Cut down on some noise. Is this a terrible idea?
 						type Notif<'a> = JsonRpcNotification<'a, Option<&'a RawValue>>;
-						match body[0] {
+						match body.get(0) {
 							// Single request or notification
-							b'{' => {
+							Some(b'{') => {
 								if let Ok(req) = serde_json::from_slice::<JsonRpcRequest>(&body) {
 									// NOTE: we don't need to track connection id on HTTP, so using hardcoded 0 here.
 									methods.execute(&tx, req, 0).await;
@@ -231,7 +231,7 @@ impl Server {
 								}
 							}
 							// Bacth of requests or notifications
-							b'[' => {
+							Some(b'[') => {
 								if let Ok(batch) = serde_json::from_slice::<Vec<JsonRpcRequest>>(&body) {
 									if !batch.is_empty() {
 										single = false;
--- a/http-server/src/server.rs
+++ b/http-server/src/server.rs
@@ -37,13 +37,13 @@ use jsonrpsee_types::{
 	v2::{
 		error::JsonRpcErrorCode,
 		params::Id,
-		request::{JsonRpcInvalidRequest, JsonRpcNotification, JsonRpcRequest},
+		request::{JsonRpcNotification, JsonRpcRequest},
 	},
 	TEN_MB_SIZE_BYTES,
 };
 use jsonrpsee_utils::hyper_helpers::read_response_to_body;
 use jsonrpsee_utils::server::{
-	helpers::{collect_batch_response, send_error},
+	helpers::{collect_batch_response, prepare_error, send_error},
 	rpc_module::Methods,
 };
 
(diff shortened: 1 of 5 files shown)


http-server/src/server.rs
```diff
@@ -215,43 +215,43 @@ impl Server {
...
-							methods.execute(&tx, req, 0).await;
-						} else if let Ok(_req) = serde_json::from_slice::<JsonRpcNotification<Option<&RawValue>>>(&body)
-						{
-							return Ok::<_, HyperError>(response::ok_response("".into()));
-						} else if let Ok(batch) = serde_json::from_slice::<Vec<JsonRpcRequest>>(&body) {
-							if !batch.is_empty() {
-								single = false;
-								for req in batch {
+						// TODO: Cut down on some noise. Is this a terrible idea?
+						type Notif<'a> = JsonRpcNotification<'a, Option<&'a RawValue>>;
```
dvdplm, line 219: Thoughts on this? I'm doing this to avoid having this: `serde_json::from_slice::<Vec<JsonRpcNotification<Option<&RawValue>>>>(&body)` which is a mouthful...

http-server/src/server.rs
```diff
@@ -215,43 +215,43 @@ impl Server {
...
+						match body[0] {
```
niklasad1, line 220: I think the body could be empty so it's safer to do `match body.get(0) { ... }` or something

http-server/src/server.rs
niklasad1: I like it, almost that we should re-export it from `types`, we would want it for the ws-server too then?

utils/src/server/helpers.rs
```diff
@@ -38,6 +39,15 @@ pub fn send_error(id: Id, tx: &MethodSink, error: JsonRpcErrorObject) {
 	}
 }
 
+/// Figure out if this is a sufficiently complete request that we can extract an [`Id`] out of, or just plain
+/// unparseable garbage.
+pub fn prepare_error<'a>(data: &'a Vec<u8>) -> (Id<'a>, JsonRpcErrorCode) {
```
niklasad1, line 44: I think `rustc` can deduce this lifetime for you and use `&[u8]` instead of `&Vec<u8>`

http-server/src/server.rs
niklasad1: Nah, right we don't care about notifications in ws-server

utils/src/server/helpers.rs
dvdplm: cheers!

ws-server/src/server.rs
```diff
@@ -275,40 +271,42 @@ async fn background_task(
...
-
-				join_all(batch.into_iter().map(|req| methods.execute(&tx_batch, req, conn_id))).await;
-
-				// Closes the receiving half of a channel without dropping it. This prevents any further messages from
-				// being sent on the channel.
-				rx_batch.close();
-				let results = collect_batch_response(rx_batch).await;
-				if let Err(err) = tx.unbounded_send(results) {
-					log::error!("Error sending batch response to the client: {:?}", err)
+		match data[0] {
```
maciejhirsz, line 274: Is there a guarantee that data is not empty here?

maciejhirsz (approved):
Looks good, just a question on one bit that might cause panic on runtime?

ws-server/src/server.rs
dvdplm: good catch, fixing.

http-server/src/server.rs
```diff
@@ -215,43 +215,42 @@ impl Server {
...
-							log::error!(
-								"[service_fn], Cannot parse request body={:?}",
-								String::from_utf8_lossy(&body[..cmp::min(body.len(), 1024)])
-							);
-							let (id, code) = match serde_json::from_slice::<JsonRpcInvalidRequest>(&body) {
-								Ok(req) => (req.id, JsonRpcErrorCode::InvalidRequest),
-								Err(_) => (Id::Null, JsonRpcErrorCode::ParseError),
-							};
-							send_error(id, &tx, code.into());
+							// Bacth of requests or notifications
```
niklasad1, line 232:
```suggestion
							// Batch of requests or notifications
```

utils/src/server/helpers.rs
```diff
@@ -38,6 +39,15 @@ pub fn send_error(id: Id, tx: &MethodSink, error: JsonRpcErrorObject) {
 	}
 }
 
+/// Figure out if this is a sufficiently complete request that we can extract an [`Id`] out of, or just plain
+/// unparseable garbage.
+pub fn prepare_error(data: &[u8]) -> (Id<'_>, JsonRpcErrorCode) {
+	match serde_json::from_slice::<JsonRpcInvalidRequest>(&data) {
```
niklasad1, line 45:
```suggestion
	match serde_json::from_slice::<JsonRpcInvalidRequest>(data) {
```

niklasad1 (approved):
LGTM,

Did you notice any performance gains compared to master?

dvdplm:
> Did you notice any performance gains compared to master?

I did not run the benchmarks tbh. I should have.

PR #420, opened 2021-07-15
Don't allocate until we know it's worth it

### LINKED ISSUE #296: `#[serde(untagged)]` doesn't work with `RawValue`

dvdplm:
For http server, check first byte before allocating space for the body.

Also re-work error handling  and prefer sending JSON-RPC-style errors (JSON doc, spec'd error codes) combined with http status codes.

Alternative to #419 

Closes #296


utils/src/hyper_helpers.rs
```diff
@@ -29,24 +29,39 @@
...
 
-/// Read a hyper response with configured `HTTP` settings.
+/// Read a data from a [`hyper::Body`] and return the data if it is valid and within the allowed size range.
 ///
-/// Returns `Ok(bytes)` if the body was in valid size range.
+/// Returns `Ok((bytes, single))` if the body was in valid size range; and a bool indicating whether the JSON-RPC
+/// request is a single or a batch.
 /// Returns `Err` if the body was too large or the body couldn't be read.
-pub async fn read_response_to_body(
+pub async fn read_body(
```
dvdplm, line 37: The relevant changes are in this function.

http-server/src/server.rs
```diff
@@ -201,61 +201,61 @@ impl Server {
 						}
 
 						let (parts, body) = request.into_parts();
-						let body = match read_response_to_body(&parts.headers, body, max_request_body_size).await {
-							Ok(body) => body,
-							Err(GenericTransportError::TooLarge) => {
-								return Ok::<_, HyperError>(response::too_large("The request was too large"))
-							}
+
+						let (body, mut is_single) = match read_body(&parts.headers, body, max_request_body_size).await {
```
dvdplm, line 205: This is the relevant change of the PR.

utils/src/hyper_helpers.rs
```diff
@@ -29,24 +29,39 @@
...
 	}
 
+	let first_chunk =
+		body.next().await.ok_or(GenericTransportError::Malformed)?.map_err(GenericTransportError::Inner)?;
+
+	if first_chunk.len() > max_request_body_size as usize {
+		return Err(GenericTransportError::TooLarge);
+	}
+
+	let single = match first_chunk[0] {
```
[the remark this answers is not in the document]
dvdplm: I _think_ `body.next().await` will return `None` for empty bodies (and become a `GenericTransportError::Malformed` error to us), but better to use `.get` anyway.

http-server/src/tests.rs
```diff
@@ -199,7 +199,7 @@ async fn async_method_call_with_ok_context() {
 }
 
 #[tokio::test]
-async fn valid_batched_method_calls() {
+async fn valid_batched_method_callsx() {
```
niklasad1, line 202: hmm, is this change intended?

ws-server/src/tests.rs
```diff
@@ -218,6 +218,7 @@ async fn can_set_max_connections() {
 
 #[tokio::test]
 async fn single_method_calls_works() {
+	println!("[test] START single");
```
niklasad1, line 221: make these `log!` or remove?!

niklasad1 (approved):

Issue #403, opened 2021-07-01
Proc macro tweaks


maciejhirsz:
+ [x] Remove old macro code, update examples to new macro code.
+ [x] Remove allocations in param parsing by using `RpcParams::next` and `RpcParams::optional_next` (#401) in the generated code.

CC @popzxc

popzxc:
>  Remove allocations in param parsing by using `RpcParams::next` and `RpcParams::optional_next` ([Streaming RpcParams parsing #401](https://github.com/paritytech/jsonrpsee/pull/401)) in the generated code.

If I understood the linked PR correctly, it only works for the array params. What about object params? Should we fall back to the allocation-based deserialization in that case?

popzxc:
I'd add the following to the list:

```markdown
- [ ] Add more UI tests.
- [ ] Add tests for missing parameter deserialization (if the parameter is an `Option`, it should still work).
- [ ] Improve auto-generated doc-comments:
  - Subscriptions descriptions should include the type of yielded items.
  - Provide a way for users to extend auto-generated doc-comments (there must be a way to add descriptions to both server- and client-side parts).
- [ ] Rewrite `no_run` examples in doc-comments to the actual doc-tests (otherwise they'll become outdated at some point).
- [ ] Add a way to set up an associated type for the RPC-specific `Error` type.
```

dvdplm:
>     Add a way to set up an associated type for the RPC-specific Error type.

@popzxc Can you elaborate a bit on this, not sure I follow?

popzxc:
> Can you elaborate a bit on this, not sure I follow?

Currently, in the RPC server implementation, you may return the `Result<OkVal, CustomErr>` from your method, but it still will be deserialized as `<CallError>::Ok(<CustomErr>::Err(YourErrorType { ... }))`.

While it will work and (kinda) serve its purpose, it's still a clunky way to deal with errors, and it's not how RPC is supposed to work.

I want the macro to be able to handle something like:

```rust
#[rpc_error] // <- Implies `serde::{Serialize, Deserialize}`, `Into<CallError>` and `derive(thiserror::Error)`.
enum CustomError {
    #[error("Bad number: {0}")] // <- error message is derived from the `thiserror::Error` implementation
    BadNumber(u32) = 42, // <- error code.
                        // ^- thanks to `serde`, this object will go into the `data` field.
}

#[rpc(client, server)
trait Rpc {
    type Error = CustomError; // <- Now we can return our own errors.
[...]

PR #421, opened 2021-07-21
Proc macro params optimizations and tests.

### LINKED ISSUE #404: Support direct RPC calls
In substrate, the [`RpcSession`](https://github.com/paritytech/substrate/blob/c078c00881b806ae9720afd713f6c696fb08c9a0/client/service/src/lib.rs#L425) and [`RpcHandlers`](https://github.com/paritytech/substrate/blob/c078c00881b806ae9720afd713f6c696fb08c9a0/client/service/src/lib.rs#L99) types are used to run RPC queries against the running node without passing over the normal http/ws machinery. Instead, JSON requests are sent via the [`rpc_query`](https://github.com/paritytech/substrate/blob/c078c00881b806ae9720afd713f6c696fb08c9a0/client/service/src/lib.rs#L111) method directly to the running server. We currently do not have any way to do this but I suspect we could make good use of it for testing.

This mechanism is used by polkadot (for [testing](https://github.com/paritytech/polkadot/blob/ac39b076ca4e31f8511799770cdb80794ec16a42/node/test/service/src/lib.rs#L297)) and [cumulus](https://github.com/paritytech/cumulus/blob/2384feb0714d32b80896c9b12a0229291d634c43/test/service/src/lib.rs#L660). The substrate-in-a-browser project also uses this ([here](https://github.com/paritytech/substrate/blob/c078c00881b806ae9720afd713f6c696fb08c9a0/utils/browser/src/lib.rs#L165-L218)). NOTE: only S-in-B seems to use this for non-test code.

The current substrate integration branch removes these types (happened [here](https://github.com/paritytech/substrate/pull/9266/files#diff-a1a9155116bf8f393b5a385a515f29f6108294a1e7b442fbd108a624eb95be5bR290)) and it needs to be re-implemented.

maciejhirsz:
Fixes allocation in proc macro param parsing (#403). Closes #404.


proc-macros/src/new/lifetimes.rs
```diff
@@ -0,0 +1,118 @@
...
+	match ty {
+		Type::Path(p) => {
+			for segment in p.path.segments.iter_mut() {
+				if let PathArguments::AngleBracketed(ref mut ab) = segment.arguments {
+					let mut lifetimes = 0;
+
+					for arg in ab.args.iter_mut() {
+						match arg {
+							GenericArgument::Lifetime(lt) => {
+								*lt = Lifetime::new(LIFETIME, lt.span());
```
jsdw, line 25: Given that we repalce every lifetime we run into with "'a", I guess this will cause an error if there is more than one lifetime present.

Perhaps we could throw a nice compile error if we run into a second lifetime to make it clear what the issue in in that case?

proc-macros/src/new/lifetimes.rs
```diff
@@ -0,0 +1,118 @@
...
+								*replaced = true;
+								lifetimes += 1;
+							}
+							// Stop iterating on first non-lifetime generic argument
+							// TODO: Replace by `.iter_mut().map_while(...).count()` when it's stabilized
+							_ => break,
+						}
+					}
+
+					// Check if the type is a `Cow<str>` or similar with lifetime elision,
```
jsdw, line 35: Nitpick: "or similar"? we only check for exactly Cow, so I feel like those words can go :D

proc-macros/src/new/lifetimes.rs
```diff
@@ -0,0 +1,118 @@
...
+					// if so insert the lifetime as first argument.
+					if lifetimes == 0 && segment.ident == "Cow" {
```
jsdw, line 37: Could `lifetimes` be a bool? If I'm understanding this right, we can only swap at most one lifetime param out, and after that we'd ideally throw some sort of error saying that multiple lifetimes aren't supported or something.

proc-macros/src/new/lifetimes.rs
```diff
@@ -0,0 +1,118 @@
...
+			};
+
+			r.lifetime = Some(Lifetime::new(LIFETIME, span));
+			*replaced = true;
+		}
+		_ => (),
+	}
+}
+
+pub fn traverse_type(ty: &mut Type, replaced: &mut bool, f: fn(&mut Type, &mut bool)) {
```
jsdw, line 60: I like this separation; def makes the code cleaner!

tests/tests/new_proc_macros.rs
```diff
@@ -47,6 +48,16 @@ mod rpc_impl {
 
 		#[subscription(name = "echo", unsub = "no_more_echo", item = u32)]
 		fn sub_with_params(&self, val: u32);
+
+		#[method(name = "params")]
+		fn params(&self, a: u8, b: &str) -> String {
```
jsdw, line 53: I reckon it'd be nice to have some more tests to exercise this lisetime stuff (if that makes sense); like, trying a couple of custom types with generics (elided or not) and a type with more than one lifetime to check that we hit a compile error (and preferably a nice one).

jsdw:
Looks awesome! If I understand right, a big change in this PR is actually adding support for lifetimes when using the RPC macro? A couple of questions came up (assuming that's true) around adding more tests to exercise this lifetime stuff, and throwing a friendly error when more than one lifetime is encountered.

proc-macros/src/new/lifetimes.rs
maciejhirsz: I was thinking about it, but I don't think that's going to come up in practice, you'd have to have two lifetimes defined on the method signature, and you could still produce a conflict by having them on separate types:

```rust
fn foo<'a, 'b>(&self, a: &'a str, b: &'b str) -> some_type
```

Those would be then put into the struct with lifetime `'a` and passed through, and I'm not even convinced there would be any error there necessarily.

proc-macros/src/new/lifetimes.rs
maciejhirsz: I think we can drop it altogether, I wanted to be able to handle `Cow<str>`, but compiler throws an error if you do that regardless and requires you put in at least `Cow<'_, str>` :).

jsdw: Ah interesting!

proc-macros/src/new/lifetimes.rs
maciejhirsz: Correction: since we are using `async-trait` all lifetimes have to be either `'_`, or just one specific lifetime, so this isn't much of an issue.

proc-macros/src/new/lifetimes.rs
maciejhirsz: Keeping it in for now, I'll see if we can get away from using a trait so that we can actually accept `Cow<str>` in params :).

tests/tests/new_proc_macros.rs
maciejhirsz: Added a bunch of tests for that purpose :)

jsdw (approved):
Looks good to me!

TarikGul (approved):
Logic, and syntax lgtm. Not super familiar with the bigger picture here, but overall no grumbles from me.

PR #423, opened 2021-07-22
[ws server] async method calls

### LINKED ISSUE #422: [ws server] processing calls sequentially
Hi folks,

I'm working on a [personal project](https://github.com/FelipeRosa/ipchess) that uses `jsonrpsee` to build a Websocket server for an API.

### Problem
The problem I'm facing is that when I make multiple calls to async methods each call is getting blocked by the ones coming before it. I made an example to show what I mean: https://github.com/FelipeRosa/jsonrpsee-bug-triage.

Did I get anything wrong about async methods?

FelipeRosa:
Closes #422 

Hey folks,

I thought of this fix for the problem I reported in #422. 

I don't know whether you have preferences of where that `tokio::spawn` should occur. Since `JsonRpcRequest` is borrowed, I decided making the `tokio::spawn` call before parsing the incoming request so `data` gets moved inside the `async` block.

Feel free to make any edits you think are appropriate.

### DIFF
--- a/ws-server/src/server.rs
+++ b/ws-server/src/server.rs
@@ -261,11 +261,9 @@ async fn background_task(
 		let _ = sender.close().await;
 	});
 
-	// Buffer for incoming data.
-	let mut data = Vec::with_capacity(100);
-
 	while !shutdown.is_closed() {
-		data.clear();
+		// Buffer for incoming data.
+		let mut data = Vec::with_capacity(100);
 
 		receiver.receive_data(&mut data).await?;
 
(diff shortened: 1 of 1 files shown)


niklasad1 (commented):
thanks looks good, but I want to hear if @maciejhirsz has any objections on the overhead of `tokio::spawn` in this context or if we should `tokio::spawn` inside execute/rpc module instead?!

maciejhirsz:
I think between this and #408 it would make most sense to just always spawn async calls on tasks. We already need have `RpcParams` for that so there is no need to allocate a separate buffer for each message (plus using a shared buffer saves us from reallocation issues).

maciejhirsz:
@FelipeRosa thanks for the PR and bringing our attention to this! I think now that #424 is merged it should have solved your use case so I'll close this PR. Feel free to re-open it or file an issue if there is anything we might have missed.

Issue #400, opened 2021-06-30
[benches]: add/fix benchmarks for concurrent connections


niklasad1:
The concurrent benchmarks that we have currently are only testing concurrent requests on **a single client** and we need to have similar benches for concurrent connections. This is particularly a problem for the ws benches.

niklasad1:
Fixed by #430

PR #424, opened 2021-07-28
Concurrent polling on async methods

### LINKED ISSUE #422: [ws server] processing calls sequentially

maciejhirsz:
Fixes #422, alternative to #423.

Tested a couple different options here and settled on this solution.

Everything on a single connection is still single-threaded, but it uses a generic `FutureDriver` (logic extracted out from the `ConnDriver`) so that multiple futures can be polled at the same time.

Performance is on par with current master, there are some minor speedups here and there, and some oddball regressions (mostly in http batch requests, but I reckon there is something else happening there). Additionally it gives us a clear way to count concurrent calls per connection for future throttling.

In case we still want to spawn async methods on tasks, we can use the `JoinHandle` instead of boxed futures to keep the driver lean, though from what I've benched so far spawning tasks is a performance hit to running on single thread without synchronization costs.

IMG_0653


http-server/src/server.rs
```diff
@@ -224,16 +225,16 @@ impl Server {
...
 						} else if let Ok(_req) = serde_json::from_slice::<JsonRpcNotification<Option<&RawValue>>>(&body)
 						{
 							return Ok::<_, HyperError>(response::ok_response("".into()));
 						} else if let Ok(batch) = serde_json::from_slice::<Vec<JsonRpcRequest>>(&body) {
 							if !batch.is_empty() {
 								single = false;
-								for req in batch {
-									methods.execute(&tx, req, 0).await;
-								}
+								join_all(batch.into_iter().filter_map(|req| methods.execute(&tx, req, 0))).await;
```
TarikGul, line 237: Interesting, so here you don't need to attach the `.await` to each executed method but instead can attach it to the `join_all`, and it will treat each request as an async method? Very cool.

ws-server/src/server.rs
```diff
@@ -263,11 +233,12 @@ async fn background_task(
 
 	// Buffer for incoming data.
 	let mut data = Vec::with_capacity(100);
+	let mut method_executors = FutureDriver::default();
 
 	while !shutdown.is_closed() {
 		data.clear();
 
-		receiver.receive_data(&mut data).await?;
+		method_executors.select_with(receiver.receive_data(&mut data)).await?;
```
jsdw, line 241: If a client has a WS connection open and so `receive_data` here is waiting for incoming data from them, we won't loop around and notice that we've been `shutdown.is_closed()` is true and we should finish.

So, will shutdown not complete until all open connections from clients have been voluntarily closed?

Should we `select` on shutdown + receive_data here so that we can see as soon as we've shutdown and can end this loop (which I think will also end the task spawned above gracefully)? (receive_data isn't cancel safe, but I don't think it matters if we're shutting down anyway.. at least I made the same assumption in telemetry!)

jsdw (commented):
A couple of comments/questions, at least one of them unrelaetd to your actual review changes :D

utils/src/server/rpc_module.rs
```diff
@@ -51,29 +48,33 @@ pub enum MethodCallback {
...
+		conn_id: ConnectionId,
+	) -> Option<BoxFuture<'static, ()>> {
 		let id = req.id.clone();
 		let params = RpcParams::new(req.params.map(|params| params.get()));
 
-		let result = match self {
-			MethodCallback::Sync(callback) => (callback)(req.id.clone(), params, tx, conn_id),
+		match self {
+			MethodCallback::Sync(callback) => {
+				(callback)(id, params, tx, conn_id);
```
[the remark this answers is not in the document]
jsdw: This is just how you have to call a variable that happens to be a function IIRC :D

http-server/src/server.rs
jsdw: It's basically the equivalent of `Promise.all` :)

TarikGul: Ahhh, awesome thanks. That makes a ton of sense :)

utils/src/server/rpc_module.rs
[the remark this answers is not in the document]
TarikGul: Ok cool, this totally makes sense, now, its just calling the callback and passing in the params, I was distracted by the syntax for a second.. Thanks for clarification, definitely overlooked that.

ws-server/src/future.rs
```diff
@@ -0,0 +1,132 @@
+// Copyright 2019 Parity Technologies (UK) Ltd.
```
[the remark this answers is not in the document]
niklasad1: We haven't updated those headers yet, I think it's just a copy-paste thingy.

I guess we should run a script and update all headers in another PR :)

niklasad1 (commented):
Looks generally good but I'm a little scared of that `soketto` is not cancel-safe when running select on the futures. We had some issues with it before such as #154 

I think we could either put the `soketto::Receiver` in a stream that lives as long the connection or do `tokio::spawn`

ws-server/src/future.rs
```diff
@@ -0,0 +1,132 @@
...
+		self.futures.push(future);
+	}
+}
+
+impl<F> FutureDriver<F>
+where
+	F: Future + Unpin,
+{
+	pub(crate) async fn select_with<S: Future>(&mut self, selector: S) -> S::Output {
+		tokio::pin!(selector);
```
niklasad1, line 67: tokio is brought in a dependency just for `pin`?

is there any difference between `futures::pin` and `futures::pin`?

niklasad1 (approved):
Some questions to resolve, but overall I think this is good even if the logic is a bit more complicated to follow.

I benched with #400 and it was slightly faster than master

ws-server/src/future.rs
jsdw: I had a gander, and both `tokio::pin` and `futures::pin_mut` have identical impls for this usage; the tokio version just supports an alternate usage as well taht isn't useful here.

dvdplm (approved):
(Apologies for dropping the late response here)

I think the code looks fine modulo the concerns/nits already expressed by others. I don't know that we have enough real-world usage data to actually settle the conversation we had on our call about how the library will be used; the trade-offs of this solution might be exactly what we need, or maybe not. We just don't know.
For the time being I think it addresses a real concern of an actual user and while it complicates the code a bit it's not too bad. So it's a 👍 from me.

jsdw:
It sounds like this benchmarks OK, and while I think it would be really good to have a couple of benchmarks to test this with a large number of concurrent connections/requests (which might be around in another branch; I haven't looked carefully), I'd have no issues with this merging as it is!

https://github.com/paritytech/jsonrpsee/pull/424#discussion_r679241602 still could potentially be an issue (I'd need to look more carefully to see if those connections are handled elsewhere), but if so, it wasn't caused by the work here anyway.

jsdw (approved):

PR #425, opened 2021-07-29
Tidy `StopHandle`

maciejhirsz:
Draft since it builds up on #424 and there are some things I want to check to make sure this does what I want it to do.

This replaces the `StopHandle` mechanism that used a `Arc<RwLock<()>>` with a combination with an async channel to handle graceful shutdown, with a single `Arc` containing an `AtomicBoolean` and `AtomicWaker`.


ws-server/src/future.rs
```diff
@@ -130,3 +136,72 @@ where
...
+/// Handle that is able to stop the running server.
+#[derive(Debug, Clone)]
+pub struct StopHandle(Weak<MonitorInner>);
+
+impl StopHandle {
+	/// Requests server to stop. Returns an error if server was already stopped.
+	///
+	/// Returns a future that can be awaited for when the server shuts down.
+	pub fn stop(self) -> Result<ShutdownWaiter, Error> {
+		if let Some(arc) = Weak::upgrade(&self.0) {
```
jsdw, line 181: So, when `stop` is called, we add 1 to the `strong_count` of the arc (if it's still around).

So there exists a brief period of time I think in which, if the last `StopMonitor` is dropped, its `Drop` impl won't wake the waker because, at the same time, a `StopHandle` is holding another strong ref to it.

If that manages to be the case (a very slim chance of course) then you'll have no `StopMonitors` left, and the `ShutdownWaiter` future will wait forever for a waker that will never be woken, I think.

ws-server/src/future.rs
```diff
@@ -130,3 +136,72 @@ where
...
+pub struct ShutdownWaiter(Weak<MonitorInner>);
+
+impl Future for ShutdownWaiter {
+	type Output = ();
+
+	fn poll(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
+		match Weak::upgrade(&self.0) {
+			None => Poll::Ready(()),
+			Some(arc) => {
+				arc.waker.register(cx.waker());
```
jsdw, line 202: Similar to the above, is there a chance that all of the `StopMonitor`s are dropped after we upgrade our ref here, and before we let go of it again? This would mean that none of the `StopMonitor`s would wake the waker and we'd be stuck forever I think.

jsdw: One approach might be to not rely on just the Arc, and to keep track of the number of `stop_waiters` separately with an AtomicUsize or something. That way, you don't have to worry about these (probably very unlikely) issues I think, since nothing would be (even for a little bit) bumping that count up again when it shouldn't?

jsdw: (I had a think about whether you could do all of this with a `tokio::Semaphore`, but you'd run into an issue telling things to shutdown (by closing the Semaphore) because that would immediately notify any `ShutdownMonitors` that the thing has closed and wouldn't wait for tidy up)

ws-server/src/future.rs
maciejhirsz: I think it should be fine because the temporary arc is going to be dropped before the waiter is returned and gets a chance to be polled.

jsdw: Ah that's true!

ws-server/src/future.rs
jsdw: (Maybe a simpler solution: after registering the waker, drop the upgraded `arc` and then check the `strong_count` again; if it's 0 then return Ready immediately?)

tests/tests/integration_tests.rs
```diff
@@ -274,14 +274,13 @@ async fn ws_close_pending_subscription_when_server_terminated() {
...
+	handle.stop().unwrap().await;
 
 	let sub2: Result<Subscription<String>, _> =
 		c1.subscribe("subscribe_hello", JsonRpcParams::NoParams, "unsubscribe_hello").await;
 
 	// no new request should be accepted.
 	assert!(matches!(sub2, Err(_)));
 	// the already established subscription should also be closed.
-	assert!(matches!(sub.next().await, Ok(None)));
+	// assert!(matches!(sub.next().await, Ok(None)));
```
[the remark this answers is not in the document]
niklasad1: I see, it worked as the background task created a channel to send & receive data, then when the receiver was dropped [here](https://github.com/paritytech/jsonrpsee/blob/master/ws-server/src/server.rs#L245-#L262) caused the `SubscriptionSink::send` to fail and send a `SubscriptionClosed message` to the server.

However, it seems like the `Server` terminates the connection before that happens on `master` (not what I remember).

In your case it could simply be timing thingy or some regression with the changed `StopHandle`. I can test it on your branch later.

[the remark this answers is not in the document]
niklasad1: I just debugged this and on this branch it actually sends on two responses on the subscription before the connection is terminated. Such that the `assertion on line 285 will be Some(_)` but after polling it one more time it will be return `Ok(None)`. Thus, timing issue/badly written test ^^

```bash
[2021-08-12T15:12:18Z DEBUG jsonrpsee_ws_client::transport] send: {"jsonrpc":"2.0","method":"subscribe_hello","id":0,"params":null}
[2021-08-12T15:12:18Z DEBUG jsonrpsee_ws_server::server] recv: JsonRpcRequest { jsonrpc: TwoPointZero, id: Number(0), method: "subscribe_hello", params: None }
[2021-08-12T15:12:18Z DEBUG jsonrpsee_ws_server::server] send: {"jsonrpc":"2.0","result":3287776652631530,"id":0}
[2021-08-12T15:12:18Z DEBUG jsonrpsee_ws_server::server] send: {"jsonrpc":"2.0","method":"subscribe_hello","params":{"subscription":3287776652631530,"result":"hello from subscription"}}
[2021-08-12T15:12:18Z DEBUG jsonrpsee_ws_client::client] [backend]: recv method_call JsonRpcResponse { jsonrpc: TwoPointZero, result: Number(3287776652631530), id: Number(0) }
[2021-08-12T15:12:18Z DEBUG jsonrpsee_ws_client::client] [backend]: recv subscription JsonRpcNotification { jsonrpc: TwoPointZero, method: "subscribe_hello", params: JsonRpcSubscriptionParams { subscription: Num(3287776652631530), result: String("hello from subscription") } }
[2021-08-12T15:12:18Z DEBUG jsonrpsee_ws_server::server] send: {"jsonrpc":"2.0","method":"subscribe_hello","params":{"subscription":3287776652631530,"result":"hello from subscription"}}
[2021-08-12T15:12:18Z DEBUG jsonrpsee_ws_client::client] [backend]: recv subscription JsonRpcNotification { jsonrpc: TwoPointZero, method: "subscribe_hello", params: JsonRpcSubscriptionParams { subscription: Num(3287776652631530), result: String("hello from subscription") } }
[2021-08-12T15:12:18Z ERROR jsonrpsee_ws_client::client] Error: Connection(Closed) terminating client
```

on master we get this:

```bash
[2021-08-12T15:17:24Z DEBUG jsonrpsee_ws_client::transport] Connecting to target: Target { sockaddrs: [127.0.0.1:39995], host: "127.0.0.1", host_header: "127.0.0.1:39995", mode: Plain, path_and_query: "/" }
[2021-08-12T15:17:24Z DEBUG jsonrpsee_ws_client::transport] send: {"jsonrpc":"2.0","method":"subscribe_hello","id":0,"params":null}
[2021-08-12T15:17:24Z DEBUG jsonrpsee_ws_server::server] recv: JsonRpcRequest { jsonrpc: TwoPointZero, id: Number(0), method: "subscribe_hello", params: None }
[2021-08-12T15:17:24Z DEBUG jsonrpsee_ws_server::server] send: {"jsonrpc":"2.0","result":2896007276381328,"id":0}
[2021-08-12T15:17:24Z DEBUG jsonrpsee_ws_server::server] send: {"jsonrpc":"2.0","method":"subscribe_hello","params":{"subscription":2896007276381328,"result":"hello from subscription"}}
[2021-08-12T15:17:24Z DEBUG jsonrpsee_ws_client::client] [backend]: recv method_call JsonRpcResponse { jsonrpc: TwoPointZero, result: Number(2896007276381328), id: Number(0) }
[2021-08-12T15:17:24Z DEBUG jsonrpsee_ws_client::client] [backend]: recv subscription JsonRpcNotification { jsonrpc: TwoPointZero, method: "subscribe_hello", params: JsonRpcSubscriptionParams { subscription: Num(2896007276381328), result: String("hello from subscription") } }
[2021-08-12T15:17:24Z DEBUG jsonrpsee_ws_client::transport] send: {"jsonrpc":"2.0","method":"subscribe_hello","id":2,"params":null}
[2021-08-12T15:17:24Z ERROR jsonrpsee_ws_client::client] Error: Connection(Closed) terminating client
test ws_close_pending_subscription_when_server_terminated ... ok
```

Not really sure why but maybe you have an idea?! Maybe the closing stuff is slower?

ws-server/src/future.rs
maciejhirsz: @jsdw did the last suggestion now :)

tests/tests/integration_tests.rs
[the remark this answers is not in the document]
maciejhirsz: Ok, I've updated the test to allow for one message before the server shuts down. Kind of makes sense now that everything is 100% gracefully shutting down.

ws-server/src/future.rs
```diff
@@ -130,3 +136,77 @@ where
...
+	fn poll(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
+		match Weak::upgrade(&self.0) {
+			None => return Poll::Ready(()),
+			Some(arc) => {
+				arc.waker.register(cx.waker());
+				drop(arc);
+			}
+		}
+
+		match Weak::strong_count(&self.0) {
```
jsdw, line 207: Looks good! IS it worth adding a small comment explaining the reasoning for this, so that we remember whe nwe look at the code again in the future?

jsdw (approved):

ws-server/src/future.rs
```diff
@@ -130,3 +136,80 @@ where
...
+pub struct StopHandle(Weak<MonitorInner>);
+
+impl StopHandle {
+	/// Requests server to stop. Returns an error if server was already stopped.
+	///
+	/// Returns a future that can be awaited for when the server shuts down.
+	pub fn stop(self) -> Result<ShutdownWaiter, Error> {
+		if let Some(arc) = Weak::upgrade(&self.0) {
+			// We proceed only if the previous value of the flag was `false`
+			if arc.shutdown_requested.swap(true, Ordering::Relaxed) == false {
```
niklasad1, line 183: I'm not sure about the memory ordering here,  can't the `if let Some(arc...` be ordered any given order by the compiler or are we relying on that the `Weak::upgrade` itself is doing `atomic stuff`?

ws-server/src/server.rs
```diff
@@ -25,26 +25,20 @@
...
 
-use crate::future::FutureDriver;
+use crate::future::{FutureDriver, StopHandle, StopMonitor};
 use crate::types::{
 	error::Error, v2::error::JsonRpcErrorCode, v2::params::Id, v2::request::JsonRpcRequest, TEN_MB_SIZE_BYTES,
 };
 use futures_channel::mpsc;
-use futures_util::future::FutureExt;
+use futures_util::io::{BufReader, BufWriter};
+// use futures_util::future::FutureExt;
```
niklasad1, line 38:
```suggestion
```

niklasad1 (commented):
Some questions regarding `Atomic::Ordering` otherwise LGTM.

How hard do you think it's to modify the `start() -> StopHandle/StopMonitor` instead of having to call handle on the server? Could be a follow-up together with #402

ws-server/src/future.rs
```diff
@@ -130,3 +136,80 @@ where
...
+	}
+}
+
+impl StopMonitor {
+	pub(crate) fn new() -> Self {
+		StopMonitor(Arc::new(MonitorInner { shutdown_requested: AtomicBool::new(false), waker: AtomicWaker::new() }))
+	}
+
+	pub(crate) fn shutdown_requested(&self) -> bool {
+		self.0.shutdown_requested.load(Ordering::Relaxed)
```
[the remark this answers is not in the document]
dvdplm: The way I read this was it's ok to read the "wrong" value here; it'd simply be polled again on the next iteration.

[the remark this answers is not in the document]
maciejhirsz: @niklasad1: @dvdplm has the right of it, it's being polled in loops (both for the socket listener and background tasks) so ordering doesn't really matter, and it's better if it's faster since those are relatively hot loops :)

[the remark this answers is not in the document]
niklasad1: ok got it, makes sense :)

ws-server/src/future.rs
niklasad1, line 183: make clippy happy?!

```suggestion
			if !arc.shutdown_requested.swap(true, Ordering::Relaxed) {
```

niklasad1 (approved):
LGTM, would be good if you could document the Atomic ordering stuff so people like me understand it :brain:

PR #429, opened 2021-08-04
fix(ws client): use query part of URL.

### LINKED ISSUE #428: Client seems to ignore the query parameters of a WSS address
I'm trying to connect to an OnFinality RPC service and failing with `401 - Unauthorized`. The endpoint is in the format `wss://kusama.api.onfinality.io/ws?apikey={API_KEY}`, and I'm able to connect to the endpoint using a simple websocket client, which makes me think that the `API_KEY` query parameter is being ignored when the client is trying to connect to the service. I can connect to the simple RPC endpoints using the same code (e.g. `wss://kusama-rpc.polkadot.io`).

niklasad1:
Fixes #428, we didn't take that query part of the URL into account and it wasn't sent in GET request because of that.

### DIFF
--- a/ws-client/src/transport.rs
+++ b/ws-client/src/transport.rs
@@ -218,6 +218,7 @@ impl<'a> WsTransportClientBuilder<'a> {
 			}
 		};
 
+		log::debug!("Connecting to target: {:?}", self.target);
 		let mut client =
 			WsRawClient::new(BufReader::new(BufWriter::new(tcp_stream)), &self.target.host_header, &self.target.path);
 		if let Some(origin) = self.origin_header.as_ref() {
--- a/ws-client/src/transport.rs
+++ b/ws-client/src/transport.rs
@@ -219,8 +219,11 @@ impl<'a> WsTransportClientBuilder<'a> {
 		};
 
 		log::debug!("Connecting to target: {:?}", self.target);
-		let mut client =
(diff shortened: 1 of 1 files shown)


ws-client/src/transport.rs
```diff
@@ -294,9 +295,14 @@ impl Target {
 			url.host_str().map(ToOwned::to_owned).ok_or_else(|| WsHandshakeError::Url("No host in URL".into()))?;
 		let port = url.port_or_known_default().ok_or_else(|| WsHandshakeError::Url("No port number in URL".into()))?;
 		let host_header = format!("{}:{}", host, port);
+		let mut path = url.path().to_owned();
+		if let Some(query) = url.query() {
+			path.push('?');
+			path.push_str(query);
+		}
```
jsdw, line 302: This looks sane to me (without a wider context of how the path `String` is used elsewhere)!

I was wondering whether percent-encoded URLs would end up being double percent-encoded when returned (because Url::query percent-encodes) but I did this:

```
use url::Url;

fn main() {
    let foo = Url::parse("ws://localhost/foo?bar=hello%20world").unwrap();
    println!("{:?}", foo.query());
}
```

in the rust playground and the output is "bar=hello%20world" as I'd hope for :)

jsdw (approved):

PR #430, opened 2021-08-05
benches: add benchmark for concurrent connections

niklasad1:
The code is a little bit !DRY but that's because the benchmark itself needs to construct the concrete client.

Interestingly, `HTTP server` throughput seems to be worse than the `WS server` for example if one compares `one round trip` with `two connections` it's around 15x slower on my machine. 

When it comes to `WebSockets` the actual initial "connection" is excluded but it should at least show how efficient/slow polling each connections is...  It's around 10x slower on my machine .

Concurrent connections with one roundtrip
```
sync/http_concurrent_connections/2
                        time:   [1.3427 ms 1.3654 ms 1.3898 ms]
                        change: [-2.9408% -1.1348% +0.6740%] (p = 0.24 > 0.05)
                        No change in performance detected.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild
sync/http_concurrent_connections/4
                        time:   [7.3308 ms 8.0426 ms 8.7470 ms]
                        change: [+159.46% +186.47% +212.97%] (p = 0.00 < 0.05)
                        Performance has regressed.
sync/http_concurrent_connections/8
                        time:   [24.833 ms 25.393 ms 26.038 ms]
                        change: [+388.94% +401.38% +414.71%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 11 outliers among 100 measurements (11.00%)
  4 (4.00%) high mild
  7 (7.00%) high severe
sync/http_concurrent_connections/16
                        time:   [9.3473 ms 9.4424 ms 9.5436 ms]
                        change: [-9.7526% -7.9180% -6.2010%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) high severe
sync/http_concurrent_connections/32
                        time:   [18.828 ms 19.172 ms 19.545 ms]
                        change: [-8.1828% -5.7520% -3.2709%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 7 outliers among 100 measurements (7.00%)
  6 (6.00%) high mild
  1 (1.00%) high severe
sync/http_concurrent_connections/64
                        time:   [38.144 ms 38.908 ms 39.692 ms]
                        change: [-12.634% -9.7696% -6.9251%] (p = 0.00 < 0.05)
                        Performance has improved.

```

Roundtrip with one client
```
sync/http_round_trip    time:   [85.041 us 95.603 us 107.21 us]
                        change: [-48.697% -41.492% -32.649%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 6 outliers among 100 measurements (6.00%)
  4 (4.00%) high mild
  2 (2.00%) high severe
```


benches/bench.rs
```diff
@@ -239,3 +241,66 @@ fn run_concurrent_round_trip<C: 'static + Client + Send + Sync>(
 	}
 	group.finish();
 }
+
+fn run_ws_concurrent_connections(rt: &TokioRuntime, crit: &mut Criterion, url: &str, name: &str, request: RequestType) {
+	let mut group = crit.benchmark_group(request.group_name(name));
+	for conns in [2, 4, 8, 16, 32, 64].iter() {
```
maciejhirsz, line 247: If we are not supporting old rust versions you can do `for conns in [2, 4, 8, 16, 32, 64] {` :)

maciejhirsz (approved):
LGTM.

Out of curiosity, does the HTTP client preserve the connection to the server, or does it make a new connection on each request?

niklasad1:
> Out of curiosity, does the HTTP client preserve the connection to the server, or does it make a new connection on each request?

Good question, the answer is that [it depends](https://github.com/hyperium/hyper/blob/master/src/client/client.rs#L337-#L350) but we create the new clients in the `bench setup (not included in bench)` so it might or might not reuse the existing connection because the clients are used several times.  If the client was only used for one request then I think we could be sure that the client didn't reuse the an already existing connection.

Issue #417, opened 2021-07-14
Refine server proc macros


dvdplm:
- [x] remove old macros
- [x] test the new ones in substrate
- [x] support generic type params in in trait definition #426
- [x] Add more UI tests.
- [x] Add tests for missing parameter deserialization (if the parameter is an `Option`, it should still work).
- [x] Improve auto-generated doc-comments:
  - Subscriptions descriptions should include the type of yielded items.
  - Provide a way for users to extend auto-generated doc-comments (there must be a way to add descriptions to both server- and client-side parts).
- [x] Rewrite `no_run` examples in doc-comments to the actual doc-tests (otherwise they'll become outdated at some point).
- [x] Add a way to set up an associated type for the RPC-specific `Error` type or just return the actual error instead of putting it inside `Ok(Err(err))` which causes the error type to have the `Serialize bound`, https://github.com/paritytech/jsonrpsee/issues/403#issuecomment-891947052
- [x] Require subscriptions to return an error, just like for normal methods

(do add more details here as needed, draft ticket)

PR #431, opened 2021-08-05
[proc macros]: remove old code and tests.

niklasad1:
Closing #403

However, there still some pieces missing such as generic support, alias but it's tracked by #417


TarikGul (approved):
LGTM, i read through the tests thoroughly and tried to think of some edge cases, but couldnt think of any! Looks great.

niklasad1:
> LGTM, i read through the tests thoroughly and tried to think of some edge cases, but couldnt think of any! Looks great.

Yeah, they `new macro code` has some edge-cases that are not properly tested yet such as generics but this PR just removes some old code so should be quite straight forward :)

insipx (approved):
lgtm

PR #432, opened 2021-08-09
[clients]: kill tokio0.2 support

niklasad1:
I propose that from release `v0.4` we remove support for tokio 0.2 and users that wants tokio0.2 could still use `v0.3`?!

Thoughts?


maciejhirsz (approved):

dvdplm (approved):
Let's do it.

PR #433, opened 2021-08-09
Fix errors with generics when using the proc macro

maciejhirsz:
The `inner` closure was causing issues when parent types were using generics, since those types aren't usable inside closures (unless said closure would be made to be generic).

Since all we used the closure for is collect all errors to `.expect` on them at the end, this PR strips the closure altogether, and instead handles errors in place: ignoring them and panicking in debug builds if the macro somehow were to produce conflicting names.

### DIFF
--- a/proc-macros/src/new/render_server.rs
+++ b/proc-macros/src/new/render_server.rs
@@ -48,7 +48,6 @@ impl RpcDescription {
 	}
 
 	fn render_into_rpc(&self) -> Result<TokenStream2, syn::Error> {
-		let jrps_error = self.jrps_server_item(quote! { types::Error });
 		let rpc_module = self.jrps_server_item(quote! { RpcModule });
 
 		let mut registered = HashSet::new();
@@ -61,6 +60,18 @@ impl RpcDescription {
 				registered.insert(name);
 			}
 		};
+
+		/// Helper that will ignore results of `register_*` method calls, and panic
+		/// if there have been any errors in debug builds.
+		///
+		/// The debug assert is a safeguard should the contract that guarantees the method
+		/// names to never conflict in the macro be broken in the future.
+		fn handle_register_result(tokens: TokenStream2) -> TokenStream2 {
+			quote! {{
+				let res = #tokens;
+				debug_assert!(res.is_ok(), "RPC macro method names should never conflict, this is a bug, please report it.");
+			}}
+		}
 
 		let methods = self
 			.methods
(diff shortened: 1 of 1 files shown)


niklasad1:
hmm suppose I want to have this:

```rust
#[rpc(server, namespace = "state")]
pub trait Rpc<Hash> {
	#[method(name = "getPairs")]
	async fn storage_pairs(&self, hash: Hash) -> Vec<usize>;
}

pub struct RpcServerImpl;

#[async_trait]
impl RpcServer<()> for RpcServerImpl {
	async fn storage_pairs(&self, _hash: ()) -> Vec<usize> {
		vec![1, 2, 3, 4]
	}
}
```

That won't work right? 
First we don't have any [generics in the impl](https://github.com/paritytech/jsonrpsee/blob/master/proc-macros/src/new/render_server.rs#L22) then there are the serde impls that I think needs the concrete type?!

Otherwise, please show an example how to use this :pray:

niklasad1 (approved):
code looks good but it's not quite want I had in mind but I guess a step in the right direction?

see my previous comment

proc-macros/src/new/render_server.rs
```diff
@@ -62,6 +61,18 @@ impl RpcDescription {
...
 
+		/// Helper that will ignore results of `register_*` method calls, and panic
+		/// if there have been any errors in debug builds.
+		///
+		/// The debug assert is a safeguard should the contract that guarantees the method
+		/// names to never conflict in the macro be broken in the future.
+		fn handle_register_result(tokens: TokenStream2) -> TokenStream2 {
+			quote! {{
+				let res = #tokens;
+				debug_assert!(res.is_ok(), "RPC macro method names should never conflict, this is a bug, please report it.");
```
dvdplm, line 72: Why make this a debug-only assert? Isn't it worth the cost to catch these violations in all builds?

Also, I'd clarify that the bug is in the user code, in not `jsonrpsee`:
```suggestion
				debug_assert!(res.is_ok(), "RPC names must never conflict, please ensure all method names are unique.");
```

niklasad1: This is just related to `jsonrpsee proc macros` that actually ensures that each method is unique so not the user should not have to do it, it would just be a compile time error then.

But this is just assertion if we introduce any bugs in that logic and I think Maciej doesn't like overhead of `expect` in release builds :P

Alternatively we could just add a test for this

jsdw (approved):
Looks sane to me, and fairly simple/easy to adjust if needed going forwards!

proc-macros/src/new/render_server.rs
dvdplm: > Alternatively we could just add a test for this

How would that work? Wouldn't it have to be in the users' code?

niklasad1: Yeah, it has to be in users code but we could just do a UI test:

```rust
#[rpc(client, server, namespace = "state")]
pub trait Rpc {
	#[method(name = "getPairs")]
	async fn storage_pairs(&self, prefix: usize, hash: Option<u128>) -> Vec<usize>;

	#[method(name = "getPairs")]
	async fn dup(&self, prefix: usize, hash: Option<u128>) -> Vec<usize>;
}

pub struct RpcServerImpl;

#[async_trait]
impl RpcServer for RpcServerImpl {
	async fn storage_pairs(&self, _prefix: usize, _hash: Option<u128>) -> Vec<usize> {
		vec![1, 2, 3, 4]
	}

	async fn dup(&self, _prefix: usize, _hash: Option<u128>) -> Vec<usize> {
		vec![1, 2, 3, 4]
	}
}
```

expect error: `error: "state_getPairs" is already defined`

niklasad1: We already got this: https://github.com/paritytech/jsonrpsee/blob/master/proc-macros/tests/ui/incorrect/rpc/rpc_name_conflict.rs so I merged this PR

PR #435, opened 2021-08-13
[proc macros] force proc macro api to return `Result`

niklasad1:
A little controversial solution to https://github.com/paritytech/jsonrpsee/issues/403#issuecomment-891947052, to require each RPC method in the `proc macro API` to require the method to return `JsonRpcResult<T, jsonrpsee_types::Error>`


types/src/error.rs
```diff
@@ -38,6 +38,18 @@ pub enum CallError {
 	},
 }
 
+// TODO(niklasad1): doesn't work probably conflicting with `thiserror::Error`
```
niklasad1, line 41: I was wrong, essentially you can't have this if the type implements `std::error::Error` :cold_face:

niklasad1: hrm, go with `From<anyhow::Error>` here?

types/src/error.rs
```diff
@@ -38,6 +38,14 @@ pub enum CallError {
 	},
 }
 
+// NOTE(niklasad1): this `From` impl is a bit opinionated to regard all generic errors as `CallError`.
+// The most common use case is when register method calls on the servers.
+impl From<anyhow::Error> for Error {
```
niklasad1, line 43: the annoying thing with this is that if a user provides their own error type:

then one have to this:

```rust
let err = MyError;
// This might not work because rustc can't deduce the first `into` as anyhow::Error
let call_err: CallError = err.into().into();

// Then users of this API must use `anyhow::anyhow!` themselves such as (that we forces them to do):
let call_err: CallError = anyhow!(err).into();
```

niklasad1: at least this was the case when I tested it in substrate.

niklasad1: alternatively, we could have helper function:

```rust
impl CallError {
   fn from_std_error<std::error::Error + Send + Sync + 'static>(err: E) -> CallError {
     CallError::Failed(e)
   }
}
```

maciejhirsz: Does it need to be `anyhow::Error`? Can't we just accept anything that's `std::fmt::Debug`?

niklasad1: I guess so but we would need to change it to a `String` [here](https://github.com/paritytech/jsonrpsee/blob/master/types/src/error.rs#L28) then or what's your idea?

utils/src/server/rpc_module.rs
```diff
@@ -204,11 +206,12 @@ impl<Context> From<RpcModule<Context>> for Methods {
...
 impl<Context: Send + Sync + 'static> RpcModule<Context> {
 	/// Register a new synchronous RPC method, which computes the response with the given callback.
-	pub fn register_method<R, F>(&mut self, method_name: &'static str, callback: F) -> Result<(), Error>
+	pub fn register_method<R, F, E>(&mut self, method_name: &'static str, callback: F) -> Result<(), Error>
 	where
 		Context: Send + Sync + 'static,
 		R: Serialize,
-		F: Fn(RpcParams, &Context) -> Result<R, CallError> + Send + Sync + 'static,
+		F: Fn(RpcParams, &Context) -> Result<R, E> + Send + Sync + 'static,
+		E: Into<Error>,
```
dvdplm, line 214: Now that we have three generics is perhaps a good time to use more explicit naming. I also think the more logical order would be `F, R, E` perhaps?
How about:
`F` –> `Method`,
`R` –> `RetVal`, 
`E` –> `MethodError`

dvdplm (approved):

utils/src/server/rpc_module.rs
niklasad1: if we put `F, R, E`, you will probably be annoyed

Because `F` the closure that you provide is always known by the compiler so in practice you would have to do this:

`module.register_method::<_, (), Error>("hello_world", |_: RpcParams, _| Ok(())).unwrap();`

isn't `<R,E, F>` better then?

niklasad1: I decided the revert the `Into<Error>` as this doesn't really work out-of-the-box when having the `client` in the API.

Basically, `Client::request -> Error` then we would require the custom error type to implement both `From<CustomError>` and `From<jsonrpsee::Error`, I prefer to tackle it another PR to actually make it work properly.

maciejhirsz (approved):

PR #436, opened 2021-08-20
[WIP]: proc macros support generic type params

niklasad1:
So this the first step for support generic params on the `trait` in the `proc macros`

For now I have ignored the `JSON Map` support as it currently not is used that widely at least by us (substrate etc) and looks quite annoying to fix the `serde impls`

However, I have added an example for this in `examples/proc_macros.rs` where you can see that users need to add some trait bounds for it work but I lean towards that we should do it for the users instead? Thoughts?

### DIFF
--- a/examples/proc_macro.rs
+++ b/examples/proc_macro.rs
@@ -28,31 +28,25 @@ use jsonrpsee::{
 	proc_macros::rpc,
 	types::{async_trait, error::Error},
 	ws_client::WsClientBuilder,
-	ws_server::{RpcModule, SubscriptionSink, WsServerBuilder},
+	ws_server::{SubscriptionSink, WsServerBuilder},
 };
+use serde::de::DeserializeOwned;
+use serde::ser::Serialize;
 use std::net::SocketAddr;
 
 #[rpc(server, client, namespace = "state")]
-pub trait Rpc {
+pub trait Rpc<Hash: DeserializeOwned + Serialize + Send + Sync + 'static> {
 	/// Async method call example.
 	#[method(name = "getPairs")]
-	async fn storage_pairs(&self, prefix: usize, hash: Option<u128>) -> Result<Vec<usize>, Error>;
-
-	/// Subscription that take `Option<Vec<u8>>` as input and produces output `Vec<usize>`.
-	#[subscription(name = "subscribeStorage", unsub = "unsubscribeStorage", item = Vec<usize>)]
-	fn subscribe_storage(&self, keys: Option<Vec<u8>>);
+	async fn storage_pairs(&self, prefix: usize, hash: Hash) -> Result<Vec<usize>, Error>;
 }
 
 pub struct RpcServerImpl;
 
 #[async_trait]
-impl RpcServer for RpcServerImpl {
-	async fn storage_pairs(&self, _prefix: usize, _hash: Option<u128>) -> Result<Vec<usize>, Error> {
+impl RpcServer<Vec<u8>> for RpcServerImpl {
+	async fn storage_pairs(&self, _prefix: usize, _hash: Vec<u8>) -> Result<Vec<usize>, Error> {
 		Ok(vec![1, 2, 3, 4])
-	}
-
-	fn subscribe_storage(&self, mut sink: SubscriptionSink, keys: Option<Vec<u8>>) {
-		sink.send(&keys.unwrap_or_default()).unwrap();
 	}
 }
 
@@ -64,18 +58,13 @@ async fn main() -> anyhow::Result<()> {
 	let url = format!("ws://{}", server_addr);
 
 	let client = WsClientBuilder::default().build(&url).await?;
-	assert_eq!(client.storage_pairs(10, None).await.unwrap(), vec![1, 2, 3, 4]);
-
-	let mut sub = client.subscribe_storage(None).await.unwrap();
-	assert_eq!(Some(vec![]), sub.next().await.unwrap());
+	assert_eq!(client.storage_pairs(10, vec![1, 2, 3, 4]).await.unwrap(), vec![1, 2, 3, 4]);
 
 	Ok(())
 }
 
 async fn run_server() -> anyhow::Result<SocketAddr> {
 	let server = WsServerBuilder::default().build("127.0.0.1:0").await?;
-	let mut module = RpcModule::new(());
-	module.register_method("state_getPairs", |_, _| Ok(vec![1, 2, 3]))?;
 
 	let addr = server.local_addr()?;
 	tokio::spawn(async move { server.start(RpcServerImpl.into_rpc()).await });
--- a/examples/proc_macro.rs
+++ b/examples/proc_macro.rs
@@ -35,16 +35,20 @@ use serde::ser::Serialize;
 use std::net::SocketAddr;
 
 #[rpc(server, client, namespace = "state")]
-pub trait Rpc<Hash: DeserializeOwned + Serialize + Send + Sync + 'static> {
+pub trait Rpc<
+	Hash: DeserializeOwned + Serialize + Send + Sync + 'static,
+	Prefix: DeserializeOwned + Serialize + Send + Sync + 'static,
+>
+{
 	/// Async method call example.
 	#[method(name = "getPairs")]
-	async fn storage_pairs(&self, prefix: usize, hash: Hash) -> Result<Vec<usize>, Error>;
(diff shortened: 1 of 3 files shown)


examples/proc_macro.rs
```diff
@@ -28,32 +28,30 @@ use jsonrpsee::{
...
+	ws_server::{SubscriptionSink, WsServerBuilder},
 };
+use serde::de::DeserializeOwned;
+use serde::ser::Serialize;
 use std::net::SocketAddr;
 
 #[rpc(server, client, namespace = "state")]
-pub trait Rpc {
+pub trait Rpc<
+	Hash: DeserializeOwned + Serialize + Send + Sync + 'static,
```
niklasad1, line 39: These trait bounds are required when both need to the client and server but we could add them behind the scenes instead.

proc-macros/src/new/render_server.rs
```diff
@@ -220,11 +220,12 @@ impl RpcDescription {
 
 		// Parsing of `serde_json::Value`.
 		let parsing = quote! {
-			let (#params_fields) = if params.is_object() {
+			/*let (#params_fields) = if params.is_object() {
 				#decode_map
```
niklasad1, line 224: no JSON Map support ^

examples/proc_macro.rs
```diff
@@ -28,31 +28,30 @@ use jsonrpsee::{
...
 
 #[rpc(server, client, namespace = "state")]
-pub trait Rpc {
+pub trait Rpc<Hash: std::fmt::Debug, Prefix> {
 	/// Async method call example.
 	#[method(name = "getPairs")]
-	async fn storage_pairs(&self, prefix: usize, hash: Option<u128>) -> Result<Vec<usize>, Error>;
-
+	async fn storage_pairs(&self, prefix: Option<Prefix>, hash: Hash) -> Result<Vec<usize>, Error>;
 	/// Subscription that take `Option<Vec<u8>>` as input and produces output `Vec<usize>`.
```
jsdw, line 40: This needs a tweak :)

dvdplm:
```suggestion
	/// Subscription that takes a [`Prefix`] as input and produces output `Vec<usize>`.
```

examples/proc_macro.rs
```diff
@@ -26,33 +26,40 @@
...
+		storage_key: ExampleStorageKey,
+		_hash: Option<ExampleHash>,
+	) -> Result<Vec<ExampleStorageKey>, Error> {
+		Ok(vec![storage_key])
 	}
 
-	fn subscribe_storage(&self, mut sink: SubscriptionSink, keys: Option<Vec<u8>>) {
-		sink.send(&keys.unwrap_or_default()).unwrap();
+	fn subscribe_storage(&self, mut sink: SubscriptionSink, _keys: Option<Vec<ExampleStorageKey>>) {
+		sink.send(&vec![[0; 32]]).unwrap();
```
dvdplm, line 62: Maybe produce a few more items here?

examples/proc_macro.rs
```diff
@@ -64,18 +71,17 @@ async fn main() -> anyhow::Result<()> {
...
 
 	let client = WsClientBuilder::default().build(&url).await?;
-	assert_eq!(client.storage_pairs(10, None).await.unwrap(), vec![1, 2, 3, 4]);
+	assert_eq!(client.storage_keys(vec![1, 2, 3, 4], None::<ExampleHash>).await.unwrap(), vec![vec![1, 2, 3, 4]]);
 
-	let mut sub = client.subscribe_storage(None).await.unwrap();
-	assert_eq!(Some(vec![]), sub.next().await.unwrap());
+	/*let mut sub: Subscription<Vec<ExampleHash>> =
+		RpcClient::<ExampleStorageKey, _>::subscribe_storage(&client, None).await.unwrap();
+	assert_eq!(Some(vec![[0; 32]]), sub.next().await.unwrap());*/
```
dvdplm, line 78: work-in-progress I assume?

proc-macros/src/helpers.rs
```diff
@@ -31,3 +38,113 @@ fn find_jsonrpsee_crate(http_name: &str, ws_name: &str) -> Result<proc_macro2::T
...
+impl<'ast> Visit<'ast> for FindTyParams {
+	fn visit_attribute(&mut self, attr: &'ast syn::Attribute) {
+		use syn::NestedMeta;
+
+		if attr.path.is_ident("subscription") {
+			if let Ok(args) = attr.parse_args_with(Punctuated::<NestedMeta, Token![,]>::parse_terminated) {}
+		}
+	}
+
+	fn visit_type_param(&mut self, ty_param: &'ast syn::TypeParam) {
```
dvdplm, line 124:
```suggestion
    /// Walk the tree and collect all generics.
	fn visit_type_param(&mut self, ty_param: &'ast syn::TypeParam) {
```

proc-macros/src/helpers.rs
```diff
@@ -31,3 +38,113 @@ fn find_jsonrpsee_crate(http_name: &str, ws_name: &str) -> Result<proc_macro2::T
...
+#[derive(Default)]
+struct FindTyParams {
+	trait_generics: HashSet<syn::Ident>,
+	input_params: HashSet<syn::Ident>,
+	ret_params: HashSet<syn::Ident>,
+	visiting_return_type: bool,
+	visiting_fn_arg: bool,
+}
+impl<'ast> Visit<'ast> for FindTyParams {
+	fn visit_attribute(&mut self, attr: &'ast syn::Attribute) {
```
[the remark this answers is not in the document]
niklasad1: I unintentionally pushed it, I was trying to have an visitor to fetch the item from the `#[subscription(item = type)]` but noticed that we already have it in `RpcSubscription`

examples/proc_macro.rs
niklasad1: yes, the entire example doesn't work.

I have some issues with adding trait bounds to `subscription output` i.e, `item = SomeType` which can be contain generics.

proc-macros/src/new/render_server.rs
```diff
@@ -137,9 +140,19 @@ impl RpcDescription {
...
+		let mut sub_items: Vec<_> = self.subscriptions.clone().into_iter().map(|s| s.item).collect();
+
+		for it in sub_items {
+			let tokens = it.to_token_stream();
+			//panic!("error {:?}", ty);
+		}
+
+		let where_clause = server_generate_where_clause(&self.trait_def);
+
+		// NOTE(niklasad1): empty where clause is valid rust syntax.
```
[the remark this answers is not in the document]
jsdw: TIL too! (I do like how Rust syntax is generally very amenable to code gen)

proc-macros/src/helpers.rs
```diff
@@ -31,3 +38,113 @@ fn find_jsonrpsee_crate(http_name: &str, ws_name: &str) -> Result<proc_macro2::T
...
+
+	fn visit_type_param(&mut self, ty_param: &'ast syn::TypeParam) {
+		self.trait_generics.insert(ty_param.ident.clone());
+	}
+
+	fn visit_return_type(&mut self, return_type: &'ast syn::ReturnType) {
+		self.visiting_return_type = true;
+		visit::visit_return_type(self, return_type);
+		self.visiting_return_type = false
+	}
```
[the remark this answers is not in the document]
niklasad1: I have "borrowed" this code from `jsonrpc` and I think it's to make it possible to know whether it's a return type or param when visiting the `Ident` in `fn visit_ident`

examples/proc_macro.rs
```diff
@@ -26,33 +26,40 @@
...
+		storage_key: ExampleStorageKey,
+		_hash: Option<ExampleHash>,
+	) -> Result<Vec<ExampleStorageKey>, Error> {
+		Ok(vec![storage_key])
 	}
 
-	fn subscribe_storage(&self, mut sink: SubscriptionSink, keys: Option<Vec<u8>>) {
-		sink.send(&keys.unwrap_or_default()).unwrap();
+	fn subscribe_storage(&self, mut sink: SubscriptionSink, _keys: Option<Vec<ExampleStorageKey>>) {
+		sink.send(&vec![[0; 32]]).unwrap();
```
niklasad1, line 62: technically, we don't have any compile time check that actually only `Vec<Hash>` can sent over the sink because send has T: Serialize bound.... not sure if we want ensure that at compile-time

Issue #440, opened 2021-08-25
[proc macros]: add alias attribute


niklasad1:
We want the `proc macros` to support aliases for a certain method or subscription such as:

```
pub trait Rpc {
	/// The server should respond to both `getKeys` and `getKeysWithAlias` with some method registered on the server.
	#[method(name = "getKeys", alias = "getKeysWithAlias")]
	async fn get_keys(&self) -> Vec<usize>
}
```

niklasad1:
Fixed by #442

PR #442, opened 2021-08-25
feat: alias attribute for proc macros

niklasad1:
Closing #440 

The new format was suggested by @dvdplm:

```rust
#[rpc(server, namespace = "state")]
pub trait StateApi<Hash> {
    …
    #[subscription(
      name = "subscribeRuntimeVersion", 
      item = RuntimeVersion, 
      /// may one or several separated by comma
      aliases = "chain_subscribeRuntimeVersion", 
      /// may one or several separated by comma
      unsubscribe_aliases = "state_unsubscribeRuntimeVersion, chain_unsubscribeRuntimeVersion"
    )]
    fn subscribe_runtime_version(&self, keys: Option<Vec<u8>>);
}
```

The unsubscribe method naming is handled by jsonrpsee and is `unsubscribeRuntimeVersion`.
If the subscribe method starts with `subscribe` we just push `un` to the string otherwise `unsubscribe`
Then users can provide their own aliases if they don't like default unsubscribe name.

After some discussion we agreed that aliases should be kept outside the namespace (if one exist) because `aliases` are often mistake and might be a PITA to refactor the entire trait just to manually add the namespace to each method to work properly.


proc-macros/src/attributes.rs
```diff
@@ -63,15 +63,19 @@ impl Rpc {
 pub(crate) struct Method {
 	/// Method name
 	pub name: syn::LitStr,
+	/// Alias for the method.
```
dvdplm, line 66:
```suggestion
	/// Aliases for the method.
```

(and same below)

proc-macros/src/render_server.rs
```diff
@@ -79,12 +79,12 @@ impl RpcDescription {
 
 		let mut registered = HashSet::new();
```
dvdplm, line 80: This tracks method names right? Maybe name it `registered_methods` or `seen_method_names`? And that brings the question, do we need to check aliases too? I guess it's not a big problem if users define duplicate aliases but might be good to keep things tight.
EDIT: nvm, you do check aliases too.

dvdplm (approved):
lgtm, let's see how this works in substrate.

proc-macros/src/render_server.rs
niklasad1: Exactly, the code is little !DRY but that's mainly because of that inner closure `check_names`, should ok here

niklasad1: I think it makes sense to have this check otherwise it would panic with duplicate names or something, also good to have if you unintentionally register the same alias for two different methods.

proc-macros/src/attributes.rs
```diff
@@ -63,7 +63,7 @@ impl Rpc {
 pub(crate) struct Method {
 	/// Method name
 	pub name: syn::LitStr,
-	/// Alias for the method.
+	/// Aliases for the method.
```
dvdplm, line 66: Same thing below, lines 77 and 79. And The `alias` member should be in the plural as well.

proc-macros/src/rpc_macro.rs
```diff
@@ -241,3 +247,14 @@ fn has_attr(attrs: &[Attribute], ident: &str) -> bool {
 	}
 	false
 }
+
+fn build_unsubscribe_method(existing_method: &str) -> String {
+	let method = existing_method.trim();
+	let mut new_method = String::from("unsubscribe");
+	if method.starts_with("subscribe") {
+		new_method.extend(method.chars().skip(9));
```
TarikGul, line 255: Just for clarification, you are just appending anything after "subscribe" to the "unsubscribe" new method, and this method name originates from the subscriptions attributes, which can technically be any string?

niklasad1: in a nutshell if the existing string is named `subscribeTarik` the unsubscribe method is named `unsubscribeTarik` hence the `skip(9)`

otherwise only `unsubcribe` is pushed at the start of the `subscribe method` for example `someSub` -> `unsubscribesomeSub`

It can be any string but if a duplicate string is used then it will end-up in compile-time error

TarikGul (approved):
LGTM

proc-macros/src/rpc_macro.rs
TarikGul: Cool, makes sense!

PR #443, opened 2021-08-26
Add a test illustrating how to use the `call` convenience method

dvdplm:
This is just me learning the API, not much to see here.﻿


utils/src/server/rpc_module.rs
```diff
@@ -574,4 +574,12 @@ mod tests {
 		assert!(module.method("hello_world").is_some());
 		assert!(module.method("hello_foobar").is_some());
 	}
+
+	#[tokio::test]
+	async fn calling_method_without_server() {
+		let mut module = RpcModule::new(());
+		module.register_method("boo", |_: RpcParams, _| Ok(String::from("boo!"))).unwrap();
+		let result = &module.call("boo", None).await;
+		assert_eq!(result, &Some(String::from("{\"jsonrpc\":\"2.0\",\"result\":\"boo!\",\"id\":0}")));
```
maciejhirsz, line 583: Wouldn't this work:

```rust
assert_eq!(result.as_ref(), Some("{\"jsonrpc\":\"2.0\",\"result\":\"boo!\",\"id\":0}"));
```

Or, since it's a test:

```rust
assert_eq!(result.unwrap(), "{\"jsonrpc\":\"2.0\",\"result\":\"boo!\",\"id\":0}");
```

maciejhirsz (approved):
Looks good, the assert could be less boilerplate-y I think :)

niklasad1 (commented):
would be good to a test with the `proc macro` and param parsing with the issues that you ran into recently

dvdplm:
Yeah I'm keeping this open as a "scratch pad" for things I run into. :)

Issue #441, opened 2021-08-25
param parsing: empty array should be deserialized as no parameters.


niklasad1:
After testing out the `proc macros` I found that the following outcome when having a parameter as `Option<SomeType`>.

```
2021-08-25 10:23:42.280 DEBUG substrate jsonrpc ws server jsonrpsee_ws_server::server: recv: JsonRpcRequest { jsonrpc: TwoPointZero, id: Number(11), method: "chain_getBlock", params: Some(RawValue([])) }
2021-08-25 10:23:42.281 DEBUG substrate jsonrpc ws server jsonrpsee_ws_server::server: send: {"jsonrpc":"2.0","error":{"code":-32602,"message":"Invalid params"},"id":11}
```
That should be deserialized to `None` instead of `Some<OtherType>`

niklasad1:
Fixed by #446

PR #446, opened 2021-08-27
Make it possible to treat empty JSON response as no params

niklasad1:
Closing #441

### DIFF
--- a/proc-macros/src/helpers.rs
+++ b/proc-macros/src/helpers.rs
@@ -149,3 +149,15 @@ fn visit_trait(item_trait: &syn::ItemTrait, sub_tys: &[syn::Type]) -> FindAllPar
 	visitor.visit_item_trait(item_trait);
 	visitor
 }
+
+/// Checks whether provided type is an `Option<...>`.
+pub(crate) fn is_option(ty: &syn::Type) -> bool {
+	if let syn::Type::Path(path) = ty {
+		// TODO: Probably not the best way to check whether type is an `Option`.
+		if path.path.segments.iter().any(|seg| seg.ident == "Option") {
+			return true;
+		}
+	}
+
+	false
+}
--- a/proc-macros/src/helpers.rs
+++ b/proc-macros/src/helpers.rs
@@ -154,8 +154,8 @@ fn visit_trait(item_trait: &syn::ItemTrait, sub_tys: &[syn::Type]) -> FindAllPar
 pub(crate) fn is_option(ty: &syn::Type) -> bool {
 	if let syn::Type::Path(path) = ty {
 		let mut it = path.path.segments.iter().peekable();
-
 		while let Some(seg) = it.next() {
+			// The leaf segment should be `Option` with or without angled brackets.
 			if seg.ident == "Option" && it.peek().is_none() {
 				return true;
 			}
--- a/proc-macros/src/helpers.rs
+++ b/proc-macros/src/helpers.rs
@@ -153,7 +153,8 @@ fn visit_trait(item_trait: &syn::ItemTrait, sub_tys: &[syn::Type]) -> FindAllPar
 /// Checks whether provided type is an `Option<...>`.
 pub(crate) fn is_option(ty: &syn::Type) -> bool {
 	if let syn::Type::Path(path) = ty {
-		// TODO: Probably not the best way to check whether type is an `Option`.
+		// TODO: https://github.com/paritytech/jsonrpsee/issues/447
+		// Probably not the best way to check whether type is an `Option`.
 		if path.path.segments.iter().any(|seg| seg.ident == "Option") {
 			return true;
 		}
--- a/proc-macros/src/helpers.rs
+++ b/proc-macros/src/helpers.rs
@@ -153,12 +153,29 @@ fn visit_trait(item_trait: &syn::ItemTrait, sub_tys: &[syn::Type]) -> FindAllPar
 /// Checks whether provided type is an `Option<...>`.
 pub(crate) fn is_option(ty: &syn::Type) -> bool {
 	if let syn::Type::Path(path) = ty {
-		// TODO: https://github.com/paritytech/jsonrpsee/issues/447
-		// Probably not the best way to check whether type is an `Option`.
-		if path.path.segments.iter().any(|seg| seg.ident == "Option") {
-			return true;
+		let mut it = path.path.segments.iter().peekable();
+
+		while let Some(seg) = it.next() {
+			if seg.ident == "Option" && it.peek().is_none() {
+				return true;
+			}
 		}
 	}
 
 	false
+}
+
+#[cfg(test)]
+mod tests {
+	use super::is_option;
+	use syn::parse_quote;
+
+	#[test]
+	fn is_option_works() {
+		assert!(is_option(&parse_quote!(Option<T>)));
+		// could be a type alias.
+		assert!(is_option(&parse_quote!(Option)));
+		assert!(is_option(&parse_quote!(std::option::Option<R>)));
+		assert!(!is_option(&parse_quote!(foo::bar::Option::Booyah)));
+	}
 }
--- a/proc-macros/src/render_server.rs
+++ b/proc-macros/src/render_server.rs
@@ -197,7 +197,8 @@ impl RpcDescription {
 			let decode_fields = params.iter().map(|(name, ty)| {
 				if is_option(ty) {
 					quote! {
-						let #name: #ty = seq.optional_next()?;
+						// treat empty array or map as no params.
+						let #name: #ty = seq.optional_next_ignore_empty_json()?;
 					}
 				} else {
 					quote! {
--- a/proc-macros/src/render_server.rs
+++ b/proc-macros/src/render_server.rs
@@ -208,7 +208,7 @@ impl RpcDescription {
 			});
 
 			quote! {
-				let mut seq = params.sequence_ignore_empty();
+				let mut seq = params.sequence();
 				#(#decode_fields);*
 			}
 		};
--- a/proc-macros/src/render_server.rs
+++ b/proc-macros/src/render_server.rs
@@ -198,7 +198,7 @@ impl RpcDescription {
 				if is_option(ty) {
 					quote! {
 						// treat empty array or map as no params.
-						let #name: #ty = seq.optional_next_ignore_empty_json()?;
+						let #name: #ty = seq.optional_next()?;
 					}
 				} else {
 					quote! {
@@ -208,7 +208,7 @@ impl RpcDescription {
 			});
 
 			quote! {
-				let mut seq = params.sequence();
+				let mut seq = params.sequence_ignore_empty();
 				#(#decode_fields);*
 			}
 		};
--- a/proc-macros/src/render_server.rs
+++ b/proc-macros/src/render_server.rs
@@ -189,22 +189,6 @@ impl RpcDescription {
 			return (TokenStream2::default(), TokenStream2::default());
 		}
 
-		// We do this check instead for input of arity one:
-		// 	i) could be an single type passed in array that should be decoded as `T`
-		//  ii) could be an entire array that should be decoded as `Vec<T>` or something similar.
-		// TODO(niklasad1): or should we require array parameters to be passed as `[[12, 12, 00]]` instead?.
-		if params.len() == 1 {
-			let (name, ty) = params[0].clone();
-			let parse = quote! {
-				let #name: #ty = match params.one() {
-					Ok(one) => one,
-					Err(_) => params.parse()?,
-				};
-			};
-			let field = quote!(#name);
-			return (parse, field);
-		}
-
 		let params_fields_seq = params.iter().map(|(name, _)| name);
 		let params_fields = quote! { #(#params_fields_seq),* };
 
--- a/proc-macros/src/render_server.rs
+++ b/proc-macros/src/render_server.rs
@@ -197,7 +197,6 @@ impl RpcDescription {
 			let decode_fields = params.iter().map(|(name, ty)| {
 				if is_option(ty) {
 					quote! {
-						// treat empty array or map as no params.
 						let #name: #ty = seq.optional_next()?;
 					}
 				} else {
--- a/proc-macros/src/render_server.rs
+++ b/proc-macros/src/render_server.rs
@@ -189,9 +189,18 @@ impl RpcDescription {
 			return (TokenStream2::default(), TokenStream2::default());
 		}
 
+		// We do this check instead for input of arity one:
+		// 	i) could be an single type passed in array that should be decoded as `T`
+		//  ii) could be an entire array that should be decoded as `Vec<T>` or something similar.
+		// TODO(niklasad1): or should we require array parameters to be passed as `[[12, 12, 00]]` instead?.
 		if params.len() == 1 {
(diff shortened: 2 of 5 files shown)


types/src/v2/params.rs
```diff
@@ -482,4 +506,26 @@ mod test {
...
+		let mut seq = params.sequence();
+
+		assert_eq!(seq.optional_next_ignore_empty_json::<&str>().unwrap(), Some("foo"));
+		assert_eq!(seq.optional_next_ignore_empty_json::<&str>().unwrap(), Some("bar"));
+
+		let params = RpcParams::new(Some(r#"[]"#));
+		let mut seq = params.sequence();
+		assert_eq!(seq.optional_next_ignore_empty_json::<&str>().unwrap(), None);
+
+		let params = RpcParams::new(Some(r#"[12, "[]"]"#));
```
dvdplm, line 522: Curious about why you added `"` around the empty element here. Does this work the same without it? I.e. `r#"[12, []]"#`

types/src/v2/params.rs
```diff
@@ -150,12 +150,17 @@ impl<'a> RpcParams<'a> {
 pub struct RpcParamsSequence<'a>(&'a str);
 
 impl<'a> RpcParamsSequence<'a> {
-	fn next_inner<T>(&mut self) -> Option<Result<T, CallError>>
+	fn next_inner<T>(&mut self, treat_empty_json_as_no_params: bool) -> Option<Result<T, CallError>>
 	where
 		T: Deserialize<'a>,
 	{
 		let mut json = self.0.trim_start();
```
dvdplm, line 157: What do you think of doing the trimming in `RpcParams::new` instead?

types/src/v2/params.rs
```diff
@@ -221,15 +226,34 @@ impl<'a> RpcParamsSequence<'a> {
...
 		T: Deserialize<'a>,
 	{
-		match self.next_inner::<Option<T>>() {
+		match self.next_inner::<Option<T>>(false) {
+			Some(result) => result,
+			None => Ok(None),
+		}
+	}
+
+	/// Similar to `RpcParamsSequence::optional_next` but regards empty JSON Array and JSON Object
```
dvdplm, line 243: An alternative to this would be to pass the `bool` to `optional_next()` directly, i.e. have the signature `pub fn optional_next<T>(&mut self, empty_means_none: bool) -> Result<Option<T>, CallError>`. Would that work?

types/src/v2/params.rs
```diff
@@ -482,4 +506,26 @@ mod test {
...
+		assert_eq!(seq.optional_next_ignore_empty_json::<&str>().unwrap(), Some("bar"));
+
+		let params = RpcParams::new(Some(r#"[]"#));
+		let mut seq = params.sequence();
+		assert_eq!(seq.optional_next_ignore_empty_json::<&str>().unwrap(), None);
+
+		let params = RpcParams::new(Some(r#"[12, "[]"]"#));
+		let mut seq = params.sequence();
+		assert_eq!(seq.optional_next_ignore_empty_json::<u64>().unwrap(), Some(12));
+		assert_eq!(seq.optional_next_ignore_empty_json::<&str>().unwrap(), Some("[]"));
```
dvdplm, line 525: We could extend the test to illustrate the above a bit better, like so:

```rust
		let params = RpcParams::new(Some(r#"[12, "[]", []]"#));
		let mut seq = params.sequence();
		assert_eq!(seq.optional_next_ignore_empty_json::<u64>().unwrap(), Some(12));
		assert_eq!(seq.optional_next_ignore_empty_json::<&str>().unwrap(), Some("[]"));
		assert_eq!(seq.optional_next_ignore_empty_json::<Vec<u8>>().unwrap(), Some(vec![]));
```

types/src/v2/params.rs
```diff
@@ -482,4 +506,26 @@ mod test {
...
+
+		let params = RpcParams::new(Some(r#"{}"#));
+		let mut seq = params.sequence();
+		assert_eq!(seq.optional_next_ignore_empty_json::<&str>().unwrap(), None);
+	}
```
dvdplm, line 530: I added a test here while playing with this, not sure if it makes sense but here it is, your call. :
```suggestion
	}

	#[test]
	fn parse_optional_nested_params() {
		let nested = RpcParams::new(Some(r#"[1, [2], [3, 4], [[5], [6, 7], []]"#));
		let mut seq = nested.sequence();
		assert_eq!(seq.optional_next_ignore_empty_json::<i8>().unwrap(), Some(1));
		assert_eq!(seq.optional_next_ignore_empty_json::<[i8; 1]>().unwrap(), Some([2]));
		assert_eq!(seq.optional_next_ignore_empty_json::<Vec<u16>>().unwrap(), Some(vec![3, 4]));
		assert_eq!(seq.optional_next_ignore_empty_json::<Vec<Vec<u32>>>().unwrap(), Some(vec![vec![5], vec![6, 7], vec![]]));
	}	
	
```

types/src/v2/params.rs
```diff
@@ -150,12 +150,17 @@ impl<'a> RpcParams<'a> {
 pub struct RpcParamsSequence<'a>(&'a str);
 
 impl<'a> RpcParamsSequence<'a> {
-	fn next_inner<T>(&mut self) -> Option<Result<T, CallError>>
+	fn next_inner<T>(&mut self, treat_empty_json_as_no_params: bool) -> Option<Result<T, CallError>>
```
dvdplm, line 153:
```suggestion
	fn next_inner<T>(&mut self, empty_means_none: bool) -> Option<Result<T, CallError>>
```

dvdplm (commented):
Need some docs here I think, and maybe less long names?

types/src/v2/params.rs
niklasad1: good point, it was mostly that I was concerned about a edge case, empty array vs empty element.

cool thanks, we should test both.

types/src/v2/params.rs
niklasad1: I don't see a common use case for `optional_next` where the result is an empty json array or json map/object but I kept it to be on the safe-side.

In general I think we should avoid having a public API that has bool flags to do different things, it's so hard to read and easy to get wrong IMHO.

types/src/v2/params.rs
niklasad1: Yeah that we would much cleaner or in `RpcParams::sequence` but that won't be possible to opt-out from unless we did `RpcParams::sequence_ignore_empty` or something like that.

dvdplm: Why would you want to opt out of it? I think we can be opinionated and just enforce no surrounding whitespace.

niklasad1: I don't have a good reason for opt-out but kept it for being on the safe-side, remove it?

dvdplm: Then let's move it to `new` and enforce it for everyone. :)

proc-macros/src/render_server.rs
```diff
@@ -197,6 +197,7 @@ impl RpcDescription {
 			let decode_fields = params.iter().map(|(name, ty)| {
 				if is_option(ty) {
 					quote! {
+						// treat empty array or map as no params.
```
niklasad1, line 200:
```suggestion
```

types/src/v2/params.rs
```diff
@@ -89,40 +89,39 @@ impl Serialize for TwoPointZero {
...
+/// Regards empty `JSON array` and `JSON object` as no parameters provided.
 #[derive(Clone, Debug)]
-pub struct RpcParams<'a>(Option<Cow<'a, str>>);
+pub struct RpcParams<'a>(Cow<'a, str>);
 
 impl<'a> RpcParams<'a> {
 	/// Create params
 	pub fn new(raw: Option<&'a str>) -> Self {
-		Self(raw.map(Into::into))
+		Self(match raw.unwrap_or("").trim() {
```
niklasad1, line 100: Moved this check to the `constructor` instead of doing it in every method, also we used `unwrap_or("null")` in `fn parse` but that is not needed anymore.

types/src/v2/params.rs
```diff
@@ -89,40 +89,39 @@ impl Serialize for TwoPointZero {
...
 
 	/// Returns true if the contained JSON is an object
 	pub fn is_object(&self) -> bool {
-		let json: &str = match self.0 {
-			Some(ref cow) => cow,
-			None => return false,
-		};
-
-		json.trim_start().starts_with('{')
+		self.0.starts_with('{')
```
niklasad1, line 108: trimmed in the `new`

maciejhirsz: With the string always being pre-trimmed, you can also move the trim in `next_inner` from line 156 to 172 (`trim_start` next subslice), which will skip one trim for the first param.

maciejhirsz (approved):
Looks good, just one suggestion for the trims :)

types/src/v2/params.rs
```diff
@@ -153,7 +153,7 @@ impl<'a> RpcParamsSequence<'a> {
 	where
 		T: Deserialize<'a>,
 	{
-		let mut json = self.0.trim_start();
+		let mut json = self.0;
```
maciejhirsz, line 156: You need to add it back on line 172:

```rust
self.0 = &json[iter.byte_offset()..].trim_start();
```

niklasad1: ty

types/src/v2/params.rs
```diff
@@ -89,40 +89,39 @@ impl Serialize for TwoPointZero {
...
 #[derive(Clone, Debug)]
-pub struct RpcParams<'a>(Option<Cow<'a, str>>);
+pub struct RpcParams<'a>(Cow<'a, str>);
 
 impl<'a> RpcParams<'a> {
 	/// Create params
 	pub fn new(raw: Option<&'a str>) -> Self {
-		Self(raw.map(Into::into))
+		Self(match raw.unwrap_or("").trim() {
+			"{}" | "[]" => "".into(),
```
maciejhirsz, line 101: This just occurred to me: `{ }` or `[ ]` are valid JSON but will still run into an error here.

types/src/v2/params.rs
niklasad1: //cc @maciejhirsz :point_up: regarding that that we ignore valid JSON `{} and []`

types/src/v2/params.rs
niklasad1: https://github.com/paritytech/jsonrpsee/pull/446#discussion_r699070793

niklasad1: lemme add a test for this,  but your point is that:

```rust
RpcParams::new("[]").parse::<JsonValue>() -> Err(InvalidParams)
RpcParams::new("{}").parse::<JsonValue>() -> Err(InvalidParams)
```

It should be treated as valid param right?
So are we back at having to different `RpcParams::sequence` and `RpcParams::sequence_ignore_empty`?

niklasad1: Right, there is also another edge-case here when you have a single array/map as a parameter and call `sequence.next()` it will be rejected as an invalid param.

The simplest would just to provide `None` for `optional_next` in the proc macro code but it's not a good idea because parameters will ignored without telling the caller.

proc-macros/tests/ui/correct/basic.rs
```diff
@@ -71,6 +94,27 @@ async fn main() {
...
+
+	assert_eq!(client.array_params(vec![1]).await.unwrap(), 1);
+	assert_eq!(
+		client
+			.request::<u64>("foo_array_params", vec![to_json_value(Vec::<u64>::new()).unwrap()].into())
+			.await
+			.unwrap(),
+		0
+	);
+	// TODO(niklasad1): do we want this to accepted as empty array?! doesn't makes sense to me
```
niklasad1, line 108:
```suggestion
```

proc-macros/tests/ui/correct/basic.rs
```diff
@@ -71,6 +94,27 @@ async fn main() {
...
+	assert!(client.request::<u64>("foo_array_params", JsonRpcParams::NoParams).await.is_err());
```
niklasad1, line 109:
```suggestion
```

niklasad1:
@dvdplm @maciejhirsz can you have another look at this now?

I confused myself with a faulty test but I moved the `[]` -> `""` to RpcParamsSequence such that RpcParams still works for `[]` and `{}` should be ready to go

proc-macros/src/helpers.rs
```diff
@@ -149,3 +149,15 @@ fn visit_trait(item_trait: &syn::ItemTrait, sub_tys: &[syn::Type]) -> FindAllPar
 	visitor.visit_item_trait(item_trait);
 	visitor
 }
+
+/// Checks whether provided type is an `Option<...>`.
+pub(crate) fn is_option(ty: &syn::Type) -> bool {
+	if let syn::Type::Path(path) = ty {
+		// TODO: Probably not the best way to check whether type is an `Option`.
```
jsdw, line 156: I guess a slightly better version would be to check only the last segment? (I'd have to read the `syn` manual to remind myself what these things are!)

proc-macros/tests/ui/correct/basic.rs
```diff
@@ -13,6 +13,15 @@ pub trait Rpc {
 	#[method(name = "foo")]
 	async fn async_method(&self, param_a: u8, param_b: String) -> JsonRpcResult<u16>;
 
+	#[method(name = "optional_params")]
+	async fn optional_params(&self, a: Option<u8>, b: String) -> JsonRpcResult<bool>;
```
jsdw, line 17: Not sure why anybody would, but it may also be useful to test that `std::option::Option` works, and maybe test that something stupid like `foo::Option::Bar` doesn't work?

dvdplm (approved):
lgtm

types/src/v2/params.rs
```diff
@@ -104,23 +104,35 @@ impl<'a> RpcParams<'a> {
...
+	/// NOTE: there is an edge-case in JSON-RPC context where you would want to
+	/// decode `[1]` as either `Vec<u64>` or `u64` depending on specific method.
+	/// Thus, we decode `[1]` as `Vec<u64>` and not as `u64`.
+	///
+	/// If you need specific behaviour use `RpcParams::parse` or `RpcParams::one`.
 	pub fn sequence(&self) -> RpcParamsSequence {
-		RpcParamsSequence(self.0.as_ref().map(AsRef::as_ref).unwrap_or(""))
+		let json = match self.0.as_ref() {
+			// It's assumed that params is `[a,b,c]`, if empty regard as no params.
+			Some(json) if json == "[]" => "",
```
jsdw, line 123: Aah, so this is basically done because by the looks of it, `RpcParamsSequence` doesn't properly handle "[]" as an empty sequence if I understand right (it tries to parse "]" and errors, or something like that).

I guess this is essentially the constructor for `RpcParamsSequence`; something inmy brain would prefer this logic to be on that type, but yeah, it's probably a bit nitpicky :)

jsdw (approved):
Looks good to me; a couple of comments but nothing important :)

types/src/v2/params.rs
niklasad1: indeed :)

proc-macros/src/helpers.rs
niklasad1: sure but nothing is introduced in this PR, I can add an issue for this ^^

Issue #449, opened 2021-09-03
[proc macros]: doc comments on the RPC trait doesn't work


niklasad1:
The `rpc trait definition` should read the docs and put it on the generated traits.

dvdplm:
Came here to ask for this! :)

PR #453, opened 2021-09-07
fix(proc macros): generate documentation for trait methods.

niklasad1:
Closing #449

Note, we are not using the docs on the entire trait instead we are manually inserting documentation there but for that we need to parse the `TraitItem` for that so out-of-scope for this PR.

To elaborate if you do:

```rust
/// My import API but these comments are not used anywhere
#[rpc(client, server)]
pub trait MyApi {
	/// API.
	#[method(name = "api"]
	async fn api(&self) -> jsonrpsee::types::JsonRpcResult<()>;
}
```

Expands to
```rust
///Server trait implementation for the `Rpc` RPC API.
pub trait MyApiServer: Sized + Send + Sync + 'static
{
    /// API
    async fn api(&self) { ....}
}

///Client implementation for the `Rpc` RPC API.
pub trait RpcClient: jsonrpsee::types::traits::Client
{
    /// API
    async fn api(&self) { ....}
}
```

### DIFF
--- a/proc-macros/src/helpers.rs
+++ b/proc-macros/src/helpers.rs
@@ -172,15 +172,12 @@ pub(crate) fn is_option(ty: &syn::Type) -> bool {
 pub(crate) fn extract_doc_comments(attrs: &[syn::Attribute]) -> TokenStream2 {
 	let docs = attrs.iter().filter_map(|attr| {
 		if let Ok(syn::Meta::NameValue(meta)) = attr.parse_meta() {
-			if meta.path.get_ident().map_or(false, |ident| ident == "doc") {
-				if let syn::Lit::Str(_lit) = &meta.lit {
-					Some(attr)
-				} else {
-					None
-				}
-			} else {
-				None
-			}
+			match &meta.lit {
+				syn::Lit::Str(_) => (),
+				_ => return None,
+			};
+
+			meta.path.get_ident().and_then(|ident| if ident == "doc" { Some(attr) } else { None })
 		} else {
 			None
 		}
--- a/proc-macros/src/helpers.rs
+++ b/proc-macros/src/helpers.rs
@@ -170,15 +170,9 @@ pub(crate) fn is_option(ty: &syn::Type) -> bool {
 /// Note that `doc comments` are expanded into `#[doc = "some comment"]`
 /// Thus, if the attribute starts with `doc` => it's regarded as a doc comment.
 pub(crate) fn extract_doc_comments(attrs: &[syn::Attribute]) -> TokenStream2 {
-	let docs = attrs.iter().filter_map(|attr| {
-		match attr.path.segments.first() {
-			Some(syn::PathSegment { ident, .. }) if ident == "doc" => (),
-			_ => return None,
-		};
-		match attr.parse_meta() {
-			Ok(syn::Meta::NameValue(meta)) if matches!(&meta.lit, syn::Lit::Str(_)) => Some(attr),
-			_ => None,
-		}
+	let docs = attrs.iter().filter(|attr| attr.path.is_ident("doc")).filter(|attr| match attr.parse_meta() {
+		Ok(syn::Meta::NameValue(meta)) if matches!(&meta.lit, syn::Lit::Str(_)) => true,
+		_ => false,
 	});
 	quote! ( #(#docs)* )
 }
--- a/proc-macros/src/helpers.rs
+++ b/proc-macros/src/helpers.rs
@@ -165,6 +165,20 @@ pub(crate) fn is_option(ty: &syn::Type) -> bool {
 	false
 }
 
+/// Iterates over all Attribute's and parses only the attributes that are doc comments.
+///
+/// Note that `doc comments` are expanded into `#[doc = "some comment"]`
+/// Thus, if the attribute starts with `doc` => it's regarded as a doc comment.
+pub(crate) fn extract_doc_comments(attrs: &[syn::Attribute]) -> Vec<syn::Attribute> {
+	attrs
+		.iter()
+		.filter_map(|attr| match attr.path.segments.first() {
+			Some(syn::PathSegment { ident, .. }) if ident == "doc" => Some(attr.to_owned()),
+			_ => None,
+		})
+		.collect()
+}
+
 #[cfg(test)]
 mod tests {
 	use super::is_option;
--- a/proc-macros/src/helpers.rs
+++ b/proc-macros/src/helpers.rs
@@ -25,7 +25,7 @@
 // DEALINGS IN THE SOFTWARE.
 
 use crate::visitor::{FindAllParams, FindSubscriptionParams};
-use proc_macro2::Span;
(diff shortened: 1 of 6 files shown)


proc-macros/src/helpers.rs
```diff
@@ -165,6 +165,20 @@ pub(crate) fn is_option(ty: &syn::Type) -> bool {
...
 }
 
+/// Iterates over all Attribute's and parses only the attributes that are doc comments.
+///
+/// Note that `doc comments` are expanded into `#[doc = "some comment"]`
+/// Thus, if the attribute starts with `doc` => it's regarded as a doc comment.
+pub(crate) fn extract_doc_comments(attrs: &[syn::Attribute]) -> Vec<syn::Attribute> {
+	attrs
+		.iter()
+		.filter_map(|attr| match attr.path.segments.first() {
```
dvdplm, line 175: In `scale-info` this is done in a slightly different way: https://github.com/paritytech/scale-info/blob/45bee5967b0fb1b282f326ffb1aa02ceddac7834/derive/src/lib.rs#L264 – might be a little more robust? Not sure if we need to strip leading whitespace here though.

proc-macros/src/rpc_macro.rs
```diff
@@ -35,6 +35,7 @@ use syn::Attribute;
 #[derive(Debug, Clone)]
 pub struct RpcMethod {
 	pub name: String,
+	pub doc: Vec<syn::Attribute>,
```
dvdplm, line 38:
```suggestion
	pub docs: Vec<syn::Attribute>,
```

proc-macros/src/rpc_macro.rs
```diff
@@ -46,6 +47,7 @@ impl RpcMethod {
 		let attributes = attributes::Method::from_attributes(&method.attrs).respan(&method.attrs.first())?;
 		let sig = method.sig.clone();
 		let name = attributes.name.value();
+		let doc = extract_doc_comments(&method.attrs);
```
dvdplm, line 50:
```suggestion
		let docs = extract_doc_comments(&method.attrs);
```

proc-macros/src/rpc_macro.rs
```diff
@@ -67,13 +69,14 @@ impl RpcMethod {
...
 
-		Ok(Self { aliases, name, params, returns, signature: method })
+		Ok(Self { aliases, name, params, returns, signature: method, doc })
 	}
 }
 
 #[derive(Debug, Clone)]
 pub struct RpcSubscription {
 	pub name: String,
+	pub doc: Vec<syn::Attribute>,
```
dvdplm, line 79:
```suggestion
	pub docs: Vec<syn::Attribute>,
```

dvdplm (approved):
lgtm!

proc-macros/src/helpers.rs
niklasad1: replaced it with the scale-info code without stripping

dvdplm (approved):

proc-macros/src/helpers.rs
```diff
@@ -165,6 +165,20 @@ pub(crate) fn is_option(ty: &syn::Type) -> bool {
 	false
 }
 
+/// Iterates over all Attribute's and parses only the attributes that are doc comments.
+///
+/// Note that `doc comments` are expanded into `#[doc = "some comment"]`
+/// Thus, if the attribute starts with `doc` => it's regarded as a doc comment.
+pub(crate) fn extract_doc_comments(attrs: &[syn::Attribute]) -> TokenStream2 {
+	let docs = attrs.iter().filter_map(|attr| match attr.parse_meta() {
```
maciejhirsz, line 173: I'd flip the order here: first check if `attr.path` is a `"doc"` ident, then do the more expensive parse meta to see if it's a name value pair with literal string value.

maciejhirsz (approved):
Looks good, just one nit.

proc-macros/src/helpers.rs
niklasad1: ty, will fix

proc-macros/src/helpers.rs
```diff
@@ -170,11 +170,15 @@ pub(crate) fn is_option(ty: &syn::Type) -> bool {
 /// Note that `doc comments` are expanded into `#[doc = "some comment"]`
 /// Thus, if the attribute starts with `doc` => it's regarded as a doc comment.
 pub(crate) fn extract_doc_comments(attrs: &[syn::Attribute]) -> TokenStream2 {
-	let docs = attrs.iter().filter_map(|attr| match attr.parse_meta() {
-		Ok(syn::Meta::NameValue(meta)) if matches!(&meta.lit, syn::Lit::Str(_)) => {
-			meta.path.get_ident().and_then(|ident| if ident == "doc" { Some(attr) } else { None })
+	let docs = attrs.iter().filter_map(|attr| {
+		match attr.path.segments.first() {
```
maciejhirsz, line 174: I think you can do `attr.path.is_ident("doc")` here

proc-macros/src/helpers.rs
```diff
@@ -170,11 +170,15 @@ pub(crate) fn is_option(ty: &syn::Type) -> bool {
 /// Note that `doc comments` are expanded into `#[doc = "some comment"]`
 /// Thus, if the attribute starts with `doc` => it's regarded as a doc comment.
 pub(crate) fn extract_doc_comments(attrs: &[syn::Attribute]) -> TokenStream2 {
-	let docs = attrs.iter().filter_map(|attr| match attr.parse_meta() {
-		Ok(syn::Meta::NameValue(meta)) if matches!(&meta.lit, syn::Lit::Str(_)) => {
-			meta.path.get_ident().and_then(|ident| if ident == "doc" { Some(attr) } else { None })
+	let docs = attrs.iter().filter_map(|attr| {
```
maciejhirsz, line 173: This can be just a `filter` now :)

proc-macros/src/helpers.rs
```diff
@@ -165,6 +165,18 @@ pub(crate) fn is_option(ty: &syn::Type) -> bool {
 	false
 }
 
+/// Iterates over all Attribute's and parses only the attributes that are doc comments.
+///
+/// Note that `doc comments` are expanded into `#[doc = "some comment"]`
+/// Thus, if the attribute starts with `doc` => it's regarded as a doc comment.
+pub(crate) fn extract_doc_comments(attrs: &[syn::Attribute]) -> TokenStream2 {
+	let docs = attrs.iter().filter(|attr| attr.path.is_ident("doc")).filter(|attr| match attr.parse_meta() {
```
niklasad1, line 173: I'm not super happy with the formatting of this one but it's better then previous version at least.....

maciejhirsz: Could have been `.filter(|attr| attr.path.is_ident("doc") && match ...` :see_no_evil:

niklasad1: :cold_face: I can fix in an follow up PR

PR #455, opened 2021-09-08
fix(proc macros): subscriptions must return result

niklasad1:
Similar to #435 that adds the same restrictions to subscriptions too.
To avoid having faulty trait bounds on when the subscription actually returns Result.

However, the most common use-case is probably to run the subscription in background task or something where returning the error might not make sense.


dvdplm (approved):
lgtm, this is good for consistency.

maciejhirsz (approved):

Issue #398, opened 2021-06-30
Send HTTP 429 response if connection limit is reached


maciejhirsz:
Currently we just drop the connection without any response, it would be nice to send a rejection with status code `429 Too Many Requests` within the handshake.

Ideally `soketto` would also allow us to attach extra headers to rejections, so we can add the `Retry-After` header.

PR #456, opened 2021-09-09
Reject overflowing connection with status code 429

maciejhirsz:
Addresses #398.

Needs an extra PR in soketto to add the `Retry-After` header, although I think this might be a case of perfect is the enemy of good here.

### DIFF
--- a/ws-server/src/server.rs
+++ b/ws-server/src/server.rs
@@ -35,7 +35,6 @@ use crate::types::{
 };
 use futures_channel::mpsc;
 use futures_util::io::{BufReader, BufWriter};
-// use futures_util::future::FutureExt;
 use futures_util::stream::StreamExt;
 use soketto::handshake::{server::Response, Server as SokettoServer};
 use tokio::net::{TcpListener, TcpStream, ToSocketAddrs};
@@ -157,6 +156,11 @@ async fn handshake(socket: tokio::net::TcpStream, mode: HandshakeMode<'_>) -> Re
 			// Forced rejection, don't need to read anything from the socket
 			let reject = Response::Reject { status_code };
 			server.send_response(&reject).await?;
+
+			let (mut sender, _) = server.into_builder().finish();
+
+			// Gracefully shut down the connection
+			sender.close().await?;
 
 			Ok(())
 		}
--- a/ws-server/src/server.rs
+++ b/ws-server/src/server.rs
@@ -86,13 +86,17 @@ impl Server {
 
 					if connections.count() >= self.cfg.max_connections as usize {
 						log::warn!("Too many connections. Try again in a while.");
+						connections.add(Box::pin(handshake(socket, Handshake::Reject { status_code: 429 })));
 						continue;
 					}
 
 					let methods = &methods;
 					let cfg = &self.cfg;
 
-					connections.add(Box::pin(handshake(socket, id, methods, cfg, &stop_monitor)));
+					connections.add(Box::pin(handshake(
+						socket,
+						Handshake::Accept { conn_id: id, methods, cfg, stop_monitor: &stop_monitor },
+					)));
 
 					id = id.wrapping_add(1);
 				}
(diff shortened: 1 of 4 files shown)


ws-server/src/server.rs
```diff
@@ -139,49 +143,59 @@ impl<'a> Future for Incoming<'a> {
...
-	methods: &Methods,
-	cfg: &Settings,
-	stop_monitor: &StopMonitor,
-) -> Result<(), Error> {
+enum Handshake<'a> {
+	Reject { status_code: u16 },
+	Accept { conn_id: ConnectionId, methods: &'a Methods, cfg: &'a Settings, stop_monitor: &'a StopMonitor },
+}
+
+async fn handshake(socket: tokio::net::TcpStream, details: Handshake<'_>) -> Result<(), Error> {
```
niklasad1, line 151: I'm struggling a little with the readability of `details`, can we rename to `handshake_input`, `handshake` or something like that?

ws-server/src/server.rs
```diff
@@ -139,49 +143,59 @@ impl<'a> Future for Incoming<'a> {
...
-		Err(error) => {
-			let reject = Response::Reject { status_code: 403 };
+	match details {
+		Handshake::Reject { status_code } => {
+			// Forced rejection, don't need to read anything from the socket
+			let reject = Response::Reject { status_code };
 			server.send_response(&reject).await?;
 
-			return Err(error);
+			Ok(())
```
niklasad1, line 161: what does `Ok(())` vs `Err(error)` imply here? 

I suppose we don't have any error for it in `Incoming` and return `Ok(())` is probably fine because it's already logged once the connection limit is exceeded when `Handshake::Reject` is detected.

niklasad1 (approved):
LGTM, a couple questions :)

ws-server/src/server.rs
maciejhirsz: Yup, returning 429 rejection isn't an error here, it's expected behavior (but we can still get an io error when sending the response for whatever reason).

ws-server/src/server.rs
maciejhirsz: Yeah, I'm open for suggestions for naming scheme here, I'm thinking maybe naming the enum `HandshakeKind`, so it would be `kind: HandshakeKind<'_>`. Alternatively `-Mode`?

niklasad1: mode or kind is works :+1:

ws-server/src/server.rs
```diff
@@ -139,49 +142,64 @@ impl<'a> Future for Incoming<'a> {
...
 }
 
-async fn handshake(
-	socket: tokio::net::TcpStream,
-	conn_id: ConnectionId,
-	methods: &Methods,
-	cfg: &Settings,
-	stop_monitor: &StopMonitor,
-) -> Result<(), Error> {
+enum HandshakeMode<'a> {
```
dvdplm, line 145: Maybe `HandshakeResponse` is more appropriate?

ws-server/src/server.rs
```diff
@@ -86,13 +85,17 @@ impl Server {
 
 					if connections.count() >= self.cfg.max_connections as usize {
 						log::warn!("Too many connections. Try again in a while.");
+						connections.add(Box::pin(handshake(socket, HandshakeMode::Reject { status_code: 429 })));
```
dvdplm, line 88: Why do we have to `add()` the connection here? Wouldn't calling `handshake(socket, HandshakeMode::Reject { status_code: 429 })` be enough here?

maciejhirsz: You'd have to `.await` it, which means you'd block incoming connections while you're sending the response here. It would be fine to `.await`, but you could potentially lock the loop if the receiving end of the socket is unusually slow?

dvdplm: Got it, ty.

dvdplm (approved):
lgtm, modulo possibly the name, up to you @maciejhirsz.

PR #458, opened 2021-09-09
[rpc module] test helper for calling and converting types to JSON-RPC params  

niklasad1:
Follow-up on #443

A little bit hacky to do like this but I think it will work in order to avoid users to manually serializing  their parameters to `serde::RawValue` and manually add it to be serialized to a sequence `"[ .... ]"` to work properly if it's not already the case.

See examples in the introduced tests


utils/src/server/rpc_module.rs
```diff
@@ -180,6 +180,18 @@ impl Methods {
 		}
 	}
 
+	/// Helper alternative to `execute`, useful for writing unit tests without having to spin
+	/// a server up.
+	///
+	/// Converts the params to an array for you if it's not already serialized to a sequence.
+	pub async fn call_with<T: Serialize>(&self, method: &str, params: &T) -> Option<String> {
+		let params = serde_json::to_string(params).ok().map(|json| {
+			let json = if json.starts_with("[") && json.ends_with("]") { json } else { format!("[{}]", json) };
```
niklasad1, line 189:
```suggestion
			let json = if json.starts_with('[') && json.ends_with(']') { json } else { format!("[{}]", json) };
```

utils/src/server/rpc_module.rs
```diff
@@ -180,6 +180,18 @@ impl Methods {
 		}
 	}
 
+	/// Helper alternative to `execute`, useful for writing unit tests without having to spin
+	/// a server up.
+	///
+	/// Converts the params to an array for you if it's not already serialized to a sequence.
+	pub async fn call_with<T: Serialize>(&self, method: &str, params: &T) -> Option<String> {
```
dvdplm, line 187: I think this (and `call`) should be `#[cfg(test)]`, wdyt?

utils/src/server/rpc_module.rs
```diff
@@ -574,4 +588,61 @@ mod tests {
 		assert!(module.method("hello_world").is_some());
 		assert!(module.method("hello_foobar").is_some());
 	}
+
+	#[tokio::test]
+	async fn calling_method_without_server() {
+		// Call sync method with no params
+		let mut module = RpcModule::new(());
+		module.register_method("boo", |_: RpcParams, _| Ok(String::from("boo!"))).unwrap();
+		let result = &module.call_with("boo", &None::<()>).await.unwrap();
```
dvdplm, line 597: dq: why take a reference here? Wouldn't `module.call_with("boo", &None::<()>).await.unwrap();` work?

dvdplm (approved):

utils/src/server/rpc_module.rs
niklasad1: #[cfg(test)] doesn't work to shared with other crates IIRC, you need to hide behind some feature such `test-helper` or something like that, but was an anti-pattern with the `feature leakage`

dvdplm: Ofc, I'm dumb. That's annoying.

utils/src/server/rpc_module.rs
niklasad1: do you mean `module.call_with("boo", None::<()>).await.unwrap();`?

I think I messed up the trait bounds for that to work.

utils/src/server/rpc_module.rs
niklasad1: I can rename the function to `test_call_rpc_method` if you think it's more clear but I prefer not to feature-gate it.

dvdplm: For the subscription-side of this I called it [`test_subscription()`](https://github.com/paritytech/jsonrpsee/pull/459/files#diff-7cf339924bf9f55b7b2a32c481d69d8e70114e91a0c074b7228bf7f80f964f8aR233). Not a very inspired name, but it gets the point across.

(I need to remove the `#[cfg(test)]` there though.)

utils/src/server/rpc_module.rs
```diff
@@ -184,8 +184,8 @@ impl Methods {
 	/// a server up.
 	///
 	/// Converts the params to an array for you if it's not already serialized to a sequence.
-	pub async fn call_with<T: Serialize>(&self, method: &str, params: &T) -> Option<String> {
-		let params = serde_json::to_string(params).ok().map(|json| {
+	pub async fn call_with<T: Serialize>(&self, method: &str, params: T) -> Option<String> {
```
dvdplm, line 187: 👍

utils/src/server/rpc_module.rs
```diff
@@ -594,7 +594,7 @@ mod tests {
 		// Call sync method with no params
 		let mut module = RpcModule::new(());
 		module.register_method("boo", |_: RpcParams, _| Ok(String::from("boo!"))).unwrap();
-		let result = &module.call_with("boo", &None::<()>).await.unwrap();
+		let result = &module.call_with("boo", None::<()>).await.unwrap();
```
dvdplm, line 597:
```suggestion
		let result = module.call_with("boo", None).await.unwrap();
```

utils/src/server/rpc_module.rs
niklasad1: yepp, I think it's better here with test_*

utils/src/server/rpc_module.rs
```diff
@@ -180,6 +180,17 @@ impl Methods {
 		}
 	}
 
+	/// Test helper to call a method on the `RPC module` without having to spin a server up.
+	///
+	/// Converts the params to a stringified array for you if it's not already serialized to a sequence.
+	pub async fn test_call<T: Serialize>(&self, method: &str, params: T) -> Option<String> {
+		let params = serde_json::to_string(&params).ok().map(|json| {
+			let json = if json.starts_with('[') && json.ends_with(']') { json } else { format!("[{}]", json) };
```
jsdw, line 188: What if there's only one param provided to the call, and it's something that already serializes to an array like `Vec<u8>`?

utils/src/server/rpc_module.rs
```diff
@@ -180,6 +180,17 @@ impl Methods {
 		}
 	}
 
+	/// Test helper to call a method on the `RPC module` without having to spin a server up.
+	///
+	/// Converts the params to a stringified array for you if it's not already serialized to a sequence.
+	pub async fn test_call<T: Serialize>(&self, method: &str, params: T) -> Option<String> {
```
jsdw, line 186: Nitpick: `test_call` makes it sound like this should only be used for tests (perhaps true, but there may be other uses, and it's available all of the time), but I wonder whether something like `local_call` would be a better description of what it does (allow you to call a method locally rather than over the network interface)?

niklasad1: It should "only" used for tests to unit tests your `RPC module`, see https://github.com/paritytech/jsonrpsee/pull/458#discussion_r706087979

So it's a test helper so might be a little weird but I think it's useful and I didn't want to introduce a `test-helper` feature it.

Does it make sense?

utils/src/server/rpc_module.rs
niklasad1: It would already be serialized to an array then we won' insert `[` json `]`, see if expression above :)

However, it could be weird if you put in a `Vec<_>` and expect it to be serialized as `[[my vec]]`, lemme double

jsdw: That's the case I was picturing; if there's an RPC call that accepts 1 param, a `Vec<u8>`, then I guess I couldn't use `test_call`, because serializing that would lead to params like `[1,2,3]` instead of a single param like `[[1,2,3]]`

utils/src/server/rpc_module.rs
jsdw: Ah interesting, ok! I guess I could see uses for this beyond just testing. For example, if you wanted to put a RESTish API or GraphQL or something on top of your RPC interface, you could use `call_with`/`test_call` to make those calls without needing to go via the network interface.

But anyway, I'm happy either way really if you prefer `test_call` :)

utils/src/server/rpc_module.rs
niklasad1: yepp, you are right.

I did it this way to work with both `RpcParams::parse` and `RpcParams::Sequence::next()`, but it won't work with doing weird things or that. I guess I was fooled because I used `JsonValue` in the test....

Don't remember if I had an issue with the parser to get this work.

niklasad1: I think this is an chicken and egg problem it will always be an edge-case for this; so no good solution AFAIU.

Perhaps, I should revisit to try to add `into_rpc_test_module` or something from because this essentially becomes a footgun unless we do `if params == edge_case then panic!("not supported")`

jsdw: A call like `test_call` could accept something like `T: IntoRpcParams`, which is implemented for a bunch of tuple sizes (and maybe even slices and arrays) like:

```
impl <A: Serialize> IntoRpcParams for (A,) { .. }
impl <A: Serialize, B: Serialize> IntoRpcParams for (A,B) { .. }
impl <const N: usize, A: Serialize> IntoRpcParams for [A;N] { .. }
```

That way, You could call eg `.test_call("foo", (a, b, c))` or `.test_call("foo", [a])` or `.test_call("foo", (a,))` with low overhead?

(perhaps something like that exists in JsonRpsee already and I'm not being very helpful; I haven't checked! :))

niklasad1: that's good idea, I think that's better

utils/src/server/rpc_module.rs
niklasad1: ok, didn't realize that use-case renamed it now thanks

types/src/traits.rs
```diff
@@ -81,3 +83,57 @@ pub trait SubscriptionClient: Client {
...
+	fn to_rpc_params(&self) -> Result<Box<RawValue>, serde_json::Error> {
+		serde_json::to_string(&self).map(|json| RawValue::from_string(json).expect("JSON String; qed"))
+	}
+}
+
+impl<P: Serialize> ToRpcParams for &[P] {}
+
+impl<P: Serialize> ToRpcParams for Vec<P> {}
+
+macro_rules! array_impls {
```
niklasad1, line 102: const generics doesn't work because of `serde` I think.

jsdw (approved):
LGTM! :)

niklasad1:
I'll merge it so @dvdplm can continue with his work :)

utils/src/server/rpc_module.rs
dvdplm: I played around with tuples on a branch on saturday and it worked fine up to a point and then didn't. I ended up having to call some permutations with a string and some with a tuple, in the end I abandoned it because it felt like the user would have a deeper understanding of what's going on than I felt was warranted.

That said, it might work out better for you. :)

types/src/traits.rs
dvdplm: Ah, came here to ask about this.

PR #459, opened 2021-09-10
Add testing facilities for subscriptions

dvdplm:
Testing subscriptions

Log more information to help test authors to debug params parsing issues.
Add a way to interact with subscriptions without spinning up a server.


utils/src/server/rpc_module.rs
```diff
@@ -574,4 +637,82 @@ mod tests {
...
+			let my_sub = my_sub_stream.next().await.unwrap();
+			let my_sub = serde_json::from_str::<JsonRpcNotification<JsonRpcSubscriptionParams<char>>>(&my_sub).unwrap();
+			assert_eq!(my_sub.params.result, std::char::from_digit(i, 10).unwrap());
+			assert_eq!(my_sub.params.subscription, v2::params::SubscriptionId::Num(sub_id));
+		}
+
+		// The subscription is now closed
+		let my_sub = my_sub_stream.next().await.unwrap();
+		let my_sub =
+			serde_json::from_str::<JsonRpcNotification<JsonRpcSubscriptionParams<SubscriptionClosedError>>>(&my_sub)
```
niklasad1, line 714: :laughing: :

types/src/v2/params.rs
```diff
@@ -182,7 +186,13 @@ impl<'a> RpcParamsSequence<'a> {
 
 				Some(Ok(value))
 			}
-			Err(_) => {
+			Err(e) => {
+				log::error!(
```
niklasad1, line 190: this is super useful but I think we could include it in `InvalidParams(String)` too

utils/src/server/rpc_module.rs
```diff
@@ -489,6 +549,7 @@ impl SubscriptionSink {
 		serde_json::to_string(&JsonRpcNotification {
 			jsonrpc: TwoPointZero,
 			method: self.method,
+			// TODO: `params` is a bit weird here. It's a result payload not really a set of params?
```
niklasad1, line 552: This is basically the same as `jsonrpc crate` and you could not send it in another field without diverging for  spec but at little bit weird yes.

I think you could create an issue for this but should most likely be kept out-of this PR

niklasad1: https://www.jsonrpc.org/specification#notification

utils/src/server/rpc_module.rs
dvdplm: lol

utils/src/server/rpc_module.rs
dvdplm: I guess the weirdness comes from us using `JsonRpcNotification` as a response (subscription ––> subscriber). In the spec, notifications are sent from the client ––> server right?

dvdplm: > I think you could create an issue for this but should most likely be kept out-of this PR

Definitely.

types/src/v2/params.rs
dvdplm: Fair, but that's a different PR I think.

niklasad1: indeed

dvdplm: https://github.com/paritytech/jsonrpsee/issues/460

utils/src/server/rpc_module.rs
niklasad1: The server is sending subscription responses as "notifications", I don't really know what else to do without introduces stuff outside the jsonrpc spec.

utils/src/server/rpc_module.rs
```diff
@@ -199,6 +228,33 @@ impl Methods {
...
+		};
+
+		let (tx, mut rx) = mpsc::unbounded();
+
+		if let Some(fut) = self.execute(&tx, req, 0) {
+			fut.await;
+		}
+		let response = rx.next().await.expect("Could not establish subscription.");
+		let subscription_response = serde_json::from_str::<JsonRpcResponse<SubscriptionId>>(&response)
+			.expect(&format!("Could not deserialize subscription response {:?}", response));
```
niklasad1, line 253:
```suggestion
			unwrap_or_else(|| panic!("Could not deserialize subscription response {:?}", response));
```

utils/src/server/rpc_module.rs
```diff
@@ -574,4 +632,82 @@ mod tests {
...
+			.register_async_method("roo", |params, ctx| {
+				let ns: Vec<u8> = params.parse().expect("valid params please");
+				async move { Ok(ctx.roo(ns)) }.boxed()
+			})
+			.unwrap();
+		let result = module.call_with("roo", vec![12, 13]).await.unwrap();
+		assert_eq!(result, r#"{"jsonrpc":"2.0","result":25,"id":0}"#);
+	}
+
+	// TODO: expand with test using the proc macros
```
niklasad1, line 673: fix in this PR?

utils/src/server/rpc_module.rs
```diff
@@ -574,4 +632,82 @@ mod tests {
...
+	// TODO: expand with test provoking Invalid Params (e.g. requiring `Bytes`)
```
niklasad1, line 674: same here; fix in this PR?

examples/proc_macro.rs
```diff
@@ -30,6 +30,7 @@ use jsonrpsee::{
 	ws_client::WsClientBuilder,
 	ws_server::{SubscriptionSink, WsServerBuilder},
 };
+use log;
```
niklasad1, line 33: is this required?

niklasad1 (approved):

utils/Cargo.toml
```diff
@@ -40,3 +40,5 @@ server = [
 
 [dev-dependencies]
 serde_json = "1.0"
+env_logger = "0.9"
+tokio = { version = "1", features = ["rt", "macros"] }
```
niklasad1, line 44: you could remove `tokio` as a `dependency` above or wait for #458

utils/src/server/rpc_module.rs
dvdplm: Why is `unwrap_or_else` + `panic` better here?

examples/proc_macro.rs
dvdplm: Totally isn't, thanks.

utils/Cargo.toml
```diff
@@ -38,4 +38,10 @@ server = [
...
 [dev-dependencies]
 serde_json = "1.0"
-tokio = { version = "1", features = ["macros", "rt"] }
\ No newline at end of file
+env_logger = "0.9"
+tokio = { version = "1", features = ["macros", "rt"] }
+# TODO: (dp) seems like the macro doesn't work with the component crates, but users must import `jsonrpsee`?
+# jsonrpsee-proc-macros = { path = "../proc-macros" }
+# jsonrpsee-types = { path = "../types" }
+# jsonrpsee-ws-server = { path = "../ws-server" }
```
dvdplm, line 46: @niklasad1 Is this a known gotcha with the proc macro? I couldn't make the proc macro work without importing the full `jsonrpsee` crate.

niklasad1: ah right, I see that's expected.

Can't you move this test to integration-tests/proc macros instead?

PR #461, opened 2021-09-11
Let rustfmt wrap comments at the 120 width boundary

dvdplm:
One interesting rule is `format_code_in_doc_comments` but it's not stable yet so leaving that out for now.﻿

### DIFF
--- a/ws-server/src/tests.rs
+++ b/ws-server/src/tests.rs
@@ -57,8 +57,8 @@ async fn server() -> SocketAddr {
 /// It has the following methods:
 ///     sync methods: `say_hello` and `add`
 ///     async: `say_hello_async` and `add_sync`
-///     other: `invalid_params` (always returns `CallError::InvalidParams`), `call_fail` (always returns `CallError::Failed`), `sleep_for`
-/// Returns the address together with handles for server future and server stop.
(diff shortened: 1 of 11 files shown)


niklasad1 (approved):

ws-server/src/tests.rs
```diff
@@ -57,8 +57,8 @@ async fn server() -> SocketAddr {
 /// It has the following methods:
 ///     sync methods: `say_hello` and `add`
 ///     async: `say_hello_async` and `add_sync`
-///     other: `invalid_params` (always returns `CallError::InvalidParams`), `call_fail` (always returns `CallError::Failed`), `sleep_for`
-/// Returns the address together with handles for server future and server stop.
+///     other: `invalid_params` (always returns `CallError::InvalidParams`), `call_fail` (always returns
+/// `CallError::Failed`), `sleep_for` Returns the address together with handles for server future and server stop.
```
jsdw, line 61: I guess if this were a proper markdown list it would indent properly!

jsdw (approved):

ws-server/src/tests.rs
dvdplm: Yeah this is a bit annoying.

dvdplm:
Sigh, I messed up: this is a nightly-only feature of rustfmt.

PR #462, opened 2021-09-11
Rename and reorg types

### LINKED ISSUE #457: Rename and reorganise types a bit
Currently (v0.3) we have many types prefixed with `JsonRpc`, e.g. `JsonRpcRequest`, `JsonRpcInvalidRequest`, `JsonRpcResponse`, `JsonRpcSubscriptionParams` etc.

Rename such types without the `JsonRpc` prefix, `Request`, `InvalidRequest`, `Response` etc.

The rationale for adding the `JsonRpc` prefix in the first place was to signal that the type is part of [the spec](). It'd be good to continue having a clear separation between types stemming from the spec and types that are implementation specific. Today, most types in `jsonrpsee-types::v2` are from the spec.

While rummaging with the types it might be a good time to also figure out what to do about: 
- `v2::params::JsonRpcSubscriptionParams`: maybe not actually parameters? I think it's the subscription item response type?
- `v2::request::JsonRpcNotification`: isn't this a response types, and if yes, let's move it to `v2::response` (EDIT: see [the conversation here](https://github.com/paritytech/jsonrpsee/pull/459#discussion_r706199657))
- https://github.com/paritytech/jsonrpsee/blob/e62a0decd7cad25b7aea72e4e783d8322e99f7f6/types/src/v2/request.rs#L149

dvdplm:
⚠️  Breaking change ⚠️ 

Closes #457﻿.

The changes here mostly removes the `JsonRpc` prefix from the types that come from the spec. I have kept a prefix for some types with a very common stdlib version such as `Result` and `Error`. I also changed the serializable versions to consistently use the `*Ser` postfix (not loving this convention but have no better suggestion).

I'm tempted to make the `jsonrpsee-types` paths more shallow, i.e. re-export all the modules under `v2`, so that instead of doing `jsonrpsee-types::v2::error::RpcError` we'd have `jsonrpsee-types::v2::RpcError`; instead of `jsonrpsee-types::v2::response::Response` we'd have `jsonrpsee-types::v2::Response` etc. Thoughts?

### DIFF
--- a/utils/src/server/rpc_module.rs
+++ b/utils/src/server/rpc_module.rs
@@ -31,7 +31,7 @@ use futures_util::{future::BoxFuture, FutureExt, StreamExt};
 use jsonrpsee_types::error::{CallError, Error, SubscriptionClosedError};
 use jsonrpsee_types::v2::error::{ErrorCode, ErrorObject, CALL_EXECUTION_FAILED_CODE, UNKNOWN_ERROR_CODE};
 use jsonrpsee_types::v2::params::{
-	Id, RpcParams, SubscriptionId as JsonRpcSubscriptionId, SubscriptionParams, TwoPointZero,
+	Id, Params, SubscriptionId as JsonRpcSubscriptionId, SubscriptionParams, TwoPointZero,
 };
 use jsonrpsee_types::v2::request::{Notification, Request};
 
@@ -47,10 +47,9 @@ use std::sync::Arc;
 /// implemented as a function pointer to a `Fn` function taking four arguments:
 /// the `id`, `params`, a channel the function uses to communicate the result (or error)
 /// back to `jsonrpsee`, and the connection ID (useful for the websocket transport).
-pub type SyncMethod = Arc<dyn Send + Sync + Fn(Id, RpcParams, &MethodSink, ConnectionId)>;
+pub type SyncMethod = Arc<dyn Send + Sync + Fn(Id, Params, &MethodSink, ConnectionId)>;
 /// Similar to [`SyncMethod`], but represents an asynchronous handler.
-pub type AsyncMethod<'a> =
-	Arc<dyn Send + Sync + Fn(Id<'a>, RpcParams<'a>, MethodSink, ConnectionId) -> BoxFuture<'a, ()>>;
+pub type AsyncMethod<'a> = Arc<dyn Send + Sync + Fn(Id<'a>, Params<'a>, MethodSink, ConnectionId) -> BoxFuture<'a, ()>>;
 /// Connection ID, used for stateful protocol such as WebSockets.
 /// For stateless protocols such as http it's unused, so feel free to set it some hardcoded value.
 pub type ConnectionId = usize;
@@ -81,7 +80,7 @@ impl MethodCallback {
 	/// Execute the callback, sending the resulting JSON (success or error) to the specified sink.
 	pub fn execute(&self, tx: &MethodSink, req: Request<'_>, conn_id: ConnectionId) -> Option<BoxFuture<'static, ()>> {
 		let id = req.id.clone();
-		let params = RpcParams::new(req.params.map(|params| params.get()));
+		let params = Params::new(req.params.map(|params| params.get()));
 
 		match self {
 			MethodCallback::Sync(callback) => {
@@ -235,7 +234,7 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 	where
 		Context: Send + Sync + 'static,
 		R: Serialize,
-		F: Fn(RpcParams, &Context) -> Result<R, Error> + Send + Sync + 'static,
+		F: Fn(Params, &Context) -> Result<R, Error> + Send + Sync + 'static,
 	{
 		self.methods.verify_method_name(method_name)?;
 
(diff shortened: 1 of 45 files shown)


maciejhirsz:
How about `CallParams` instead of `ParamsSer`? Or, if we want to be slightly confusing but pedantically correct: `Arguments`.

dvdplm:
> How about `CallParams` instead of `ParamsSer`? Or, if we want to be slightly confusing but pedantically correct: `Arguments`.

GH threw away my reply to this, grr.

I like `Arguments` quite a lot, but I don't think it's wise to move away from "*params*" because the spec uses "parameter" throughout and I don't think it's good to ask users/readers to map from one to the other.

The downside of going with `CallParams` is that it doesn't help with the other `*Ser` types: `RequestSer` and `NotificationSer`. Using a post-fix has the advantage of clearly communicating what is peculiar about all three of them.

:/

niklasad1:
> I'm tempted to make the jsonrpsee-types paths more shallow, i.e. re-export all the modules under v2, so that instead of doing jsonrpsee-types::v2::error::RpcError we'd have jsonrpsee-types::v2::RpcError; instead of jsonrpsee-types::v2::response::Response we'd have jsonrpsee-types::v2::Response etc. Thoughts?

Sounds good to me

utils/src/server/rpc_module.rs
```diff
@@ -780,16 +759,15 @@ mod tests {
 		let (sub_id, mut my_sub_stream) = module.test_subscription("my_sub", None).await;
 		for i in (0..=2).rev() {
 			let my_sub = my_sub_stream.next().await.unwrap();
-			let my_sub = serde_json::from_str::<JsonRpcNotification<JsonRpcSubscriptionParams<char>>>(&my_sub).unwrap();
+			let my_sub = serde_json::from_str::<Notification<SubscriptionPayload<char>>>(&my_sub).unwrap();
```
niklasad1, line 762: can't you use the `SubscriptionResponse` type alias here instead?

utils/src/server/rpc_module.rs
```diff
@@ -780,16 +759,15 @@ mod tests {
...
 			assert_eq!(my_sub.params.result, std::char::from_digit(i, 10).unwrap());
 			assert_eq!(my_sub.params.subscription, v2::params::SubscriptionId::Num(sub_id));
 		}
 
 		// The subscription is now closed
 		let my_sub = my_sub_stream.next().await.unwrap();
 		let my_sub =
-			serde_json::from_str::<JsonRpcNotification<JsonRpcSubscriptionParams<SubscriptionClosedError>>>(&my_sub)
-				.unwrap();
+			serde_json::from_str::<Notification<SubscriptionPayload<SubscriptionClosedError>>>(&my_sub).unwrap();
```
niklasad1, line 770: can't you use the SubscriptionResponse type alias here instead?

ws-client/src/helpers.rs
```diff
@@ -73,19 +73,19 @@ pub fn process_batch_response(manager: &mut RequestManager, rps: Vec<JsonRpcResp
 ///
 /// Returns `Ok()` if the response was successfully sent to the frontend.
 /// Return `Err(None)` if the subscription was not found.
-/// Returns `Err(Some(msg))` if the subscription was full.
+/// Returns `Err(Some(msg))` if the channel to the frontend was full.
```
niklasad1, line 76: I don't mind this much but I would call the `subscription stream` as a "frontend" as it's part of the `Subscription` returned to user after a subscription has been established and "not a frontend" in the client....

utils/src/server/rpc_module.rs
```diff
@@ -506,7 +487,7 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 		Ok(())
 	}
 
-	/// Register an `alias` name for an `existing_method`.
+	/// Register an alias for an existing_method. Alias uniqueness is enforced at compile time.
```
niklasad1, line 490: This is only true for the proc macros and not in this context but for most use-cases your comment is true :P

```suggestion
	/// Register an alias for an existing_method.
```

niklasad1:
> How about CallParams instead of ParamsSer? Or, if we want to be slightly confusing but pedantically correct: Arguments.

We have used `Params` because it is was spec states. I think `CallParams` is slightly confusing because when the server receives the params it's still a call?! At least `ParamsSer` is super obvious too me even if it's a bit "verbose"

`Arguments` could work though....

niklasad1 (approved):
LGTM, some minor comments to address

In an followup PR, we could flatten the `v2 mod` and remove the child modules `request, response, params`.

utils/src/server/rpc_module.rs
dvdplm: Would `Alias uniqueness is enforced.` be correct?

dvdplm:
> In an followup PR, we could flatten the `v2 mod` and remove the child modules `request, response, params`.

https://github.com/paritytech/jsonrpsee/issues/468

utils/src/server/rpc_module.rs
dvdplm: Ofc I can, that's the very reason I added `SubscriptionResponse`! I messed up a merge I think. Ty!

utils/src/server/rpc_module.rs
niklasad1: Yes, a duplicated alias will return error

maciejhirsz (approved):
Aside form my initial bikeshed it looks good to me :)

PR #463, opened 2021-09-11
Propagate cause of `InvalidParams`

dvdplm:
Change `CallError::InvalidParams` to carry the actual error (as an `anyhow::Error`), enabling more descriptive errors for users.

One doubt I'd like reviewers feedback on is adding the `#[source]` attribute to `CallError::InvalidParams`. Not sure it's useful/correct.

Based on #458, hence the noise.


types/src/error.rs
```diff
@@ -48,7 +48,7 @@ impl<T: fmt::Display> fmt::Display for Mismatch<T> {
 pub enum CallError {
 	/// Invalid params in the call.
 	#[error("Invalid params in the call")]
```
niklasad1, line 50: you want to change this to `#[error(transparent)]` or `#[error("Invalid params in the call: {0}")]`

types/src/error.rs
```diff
@@ -48,7 +48,7 @@ impl<T: fmt::Display> fmt::Display for Mismatch<T> {
...
-	InvalidParams,
+	InvalidParams(#[source] anyhow::Error),
```
niklasad1, line 51: This looks good

types/src/error.rs
dvdplm: Good catch, ty!

utils/Cargo.toml
```diff
@@ -38,4 +38,10 @@ server = [
 
 [dev-dependencies]
 serde_json = "1.0"
-tokio = { version = "1", features = ["macros", "rt"] }
\ No newline at end of file
+env_logger = "0.9"
+tokio = { version = "1", features = ["macros", "rt"] }
+# TODO: (dp) seems like the macro doesn't work with the component crates, but users must import `jsonrpsee`?
```
niklasad1, line 43: I don't get this? Can you elaborate?

dvdplm: I tried writing a test using the proc macro and only the server, `#[rpc(server)]` and wanted to pull in only the minimum dependencies; I thought all I needed was `jsonrpsee-proc-macros` but that doesn't compile and so I tried a few different combinations but in the end I needed the full` jsonrpsee` façade crate. So the question was: is this a regression?

niklasad1: No, we assume that the facade crate is used there I think or maybe my recent refactoring broke that.

Should we still support both?! I still think that this test should be moved to `integration test crate` (not really a unit test IMO)

types/src/traits.rs
```diff
@@ -88,10 +88,10 @@ pub trait SubscriptionClient: Client {
 
 /// Marker trait for types that can be serialized as JSON array/sequence.
 ///
-/// If your type isn't a sequence such as `String`, `usize` or similar.
-/// You could insert it in a tuple, slice, array or Vec for it to work.
+/// If your type isn't a sequence, for example `String`, `usize` or similar
+/// you must insert it in a tuple, slice, array or Vec for it to work.
 pub trait ToRpcParams: Serialize {
-	/// Serialized the type as a JSON array.
+	/// Serialize the type as a JSON array.
```
niklasad1, line 94: :+1:

niklasad1 (commented):
LGTM expect that I think the `proc macro API` test should be moved into integration tests but feel free to convince me :)

proc-macros/src/render_server.rs
```diff
@@ -272,11 +273,23 @@ impl RpcDescription {
 			let decode_fields = params.iter().map(|(name, ty)| {
 				if is_option(ty) {
 					quote! {
-						let #name: #ty = seq.optional_next()?;
+						let #name: #ty = match seq.optional_next() {
+							Ok(v) => v,
+							Err(e) => {
+								log::error!("Error parsing optional {:?} as {:?}: {:?}", stringify!(#name), stringify!(#ty), e);
```
maciejhirsz, line 279: You can concatenate string literals at compile time with macros:

```suggestion
								log::error!(concat!("Error parsing optional \"", stringify!(#name), "\" as \"", stringify!(#ty), "\": {:?}"), e);
```

proc-macros/src/render_server.rs
```diff
@@ -272,11 +273,23 @@ impl RpcDescription {
...
+							}
+						};
 					}
 				} else {
 					quote! {
-						let #name: #ty = seq.next()?;
+						let #name: #ty = match seq.next() {
+							Ok(v) => v,
+							Err(e) => {
+								log::error!("Error parsing {:?} as {:?}: {:?}", stringify!(#name), stringify!(#ty), e);
```
maciejhirsz, line 289: Ditto about `concat!`

maciejhirsz (approved):
Looks good, you can turn some runtime concats into compile time concats :)

utils/Cargo.toml
dvdplm: > (not really a unit test IMO)

Well. I think I disagree. The test is unit-testing `call_with`. You are correct that it touches on a bunch of other stuff but that's because the setup is kind of long.

dvdplm:
> I think the `proc macro API` test should be moved into integration tests but feel free to convince me :)

I think it is a unit test of the `call_with` and `call` methods. You're right that the setup is big-ish but it *is* just setup, no asserting on the correctness or not of the macro generated code.

types/src/traits.rs
```diff
@@ -110,7 +110,7 @@ macro_rules! array_impls {
 }
 
 array_impls! {
-	1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16
+	0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16
```
maciejhirsz, line 113: This is kind of a tangent, but we could just use const generics now that they are available instead of the macro:

```rust
impl<P: Serialize, const N: usize> ToRpcParams for [P; N] {}
```

niklasad1: The problem I think is that serde impls for `[T; 0] .... [T; 32]` and `impl<P: Serialize, const N: usize> ToRpcParams for [P; N]` for 0...usize::MAX?

niklasad1: ok, maybe we should open a PR to serde then?!

maciejhirsz:
```rust
impl<P, const N: usize> ToRpcParams for [P; N] where [P; N]: Serialize {}
```
Works for me

niklasad1 (approved):

PR #464, opened 2021-09-12
Remove unstable rustfmt comment wrapping/limiting

dvdplm:
In #461 I missed that these options are not stable so it breaks our CI. This PR reverts that change while keeping the whitespace changes from #461.

Apologies for the noise.


niklasad1 (approved):
I don't mind requiring nightly for formatting FWIW, but sure I get the point

PR #465, opened 2021-09-14
Fix build warnings

dvdplm:
Looks like we had a few fields left over from recent refactorings.

The matter of `AccessControl::cors_max_age` is less clear-cut. I don't remember where we're at with preflight and CORS, but I think we support it and that we should set the `Access-Control-Max-Age` header [here](https://github.com/paritytech/jsonrpsee/blob/master/http-server/src/server.rs#L266), but only for pre-flight requests?

### DIFF
--- a/http-server/src/access_control/mod.rs
+++ b/http-server/src/access_control/mod.rs
@@ -41,6 +41,8 @@ use jsonrpsee_utils::http_helpers;
 pub struct AccessControl {
 	allow_hosts: AllowHosts,
 	cors_allow_origin: Option<Vec<AccessControlAllowOrigin>>,
+	// TODO: (dp) We never set this header in the response headers.
+	#[allow(dead_code)]
 	cors_max_age: Option<u32>,
 	cors_allow_headers: AccessControlAllowHeaders,
 	continue_on_invalid_cors: bool,
--- a/http-server/src/access_control/mod.rs
+++ b/http-server/src/access_control/mod.rs
@@ -41,8 +41,6 @@ use jsonrpsee_utils::http_helpers;
 pub struct AccessControl {
 	allow_hosts: AllowHosts,
 	cors_allow_origin: Option<Vec<AccessControlAllowOrigin>>,
-	// TODO: (dp) We never set this header in the response headers.
-	#[allow(dead_code)]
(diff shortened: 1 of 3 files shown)


http-server/src/access_control/mod.rs
```diff
@@ -41,6 +41,8 @@ use jsonrpsee_utils::http_helpers;
 pub struct AccessControl {
 	allow_hosts: AllowHosts,
 	cors_allow_origin: Option<Vec<AccessControlAllowOrigin>>,
+	// TODO: (dp) We never set this header in the response headers.
+	#[allow(dead_code)]
 	cors_max_age: Option<u32>,
```
dvdplm, line 46: ☝️

niklasad1:
ok, I removed it in #466 but fine we need to add `fn deny_max_cors_age` then

niklasad1 (approved):

http-server/src/access_control/mod.rs
maciejhirsz: I think we can drop it? The defaults in browsers seem sensible, and I don't suppose there really is a need to set them higher (it also appears you can't set them higher than whatever arbitrary cap a browser decides) or disable them. For most cases I assume ctrl+shift+R clears it anyway so it should be irrelevant for devs.

niklasad1: from what I understand `a preflight request must be sent as a HTTP Option request`?

jsonrpc crate seems to set it for all requests https://github.com/paritytech/jsonrpc/blob/master/http/src/handler.rs#L558-L563?!

maciejhirsz: Yes, if cached the browser will not send the preflight and either go straight to the HTTP request (if origin is allowed in cache) or block the request (if origin is not allowed in cache).

maciejhirsz (approved):

http-server/src/access_control/mod.rs
dvdplm: So to do this properly we'd need to keep some state so we know if a connecting host has to do the preflight or not?

dvdplm: I'm going to revert this change in this PR and remove it in #466 instead.

PR #466, opened 2021-09-14
[http server]: export acl types + remove cors_max_age

niklasad1:


dvdplm:
I think this is better than what I had in #465 but the question about where we send back this header should be. I don't think this fixes that, or…?

niklasad1:
> I think this is better than what I had in #465 but the question about where we send back this header should be. I don't think this fixes that, or…?

No, it just provides the "functionality" to fetch the cors_max_age. I can fix it in this PR or a follow-up doesn't really matter to me

niklasad1:
> I'm tempted to go with @maciejhirsz's suggestion and remove the field entirely (until we know we need it).

Yes, I like that idea :+1

dvdplm (approved):

PR #470, opened 2021-09-15
replace `array_impl macro` with const generics

niklasad1:
Fixes https://github.com/paritytech/jsonrpsee/pull/463#discussion_r708130440

Note, [this doesn't work for an empty array because of serde AFAIU](https://github.com/serde-rs/serde/blob/master/serde/src/ser/impls.rs#L163-#L168)

### DIFF
--- a/types/src/traits.rs
+++ b/types/src/traits.rs
@@ -98,20 +98,8 @@ pub trait ToRpcParams: Serialize {
 }
 
 impl<P: Serialize> ToRpcParams for &[P] {}
-
 impl<P: Serialize> ToRpcParams for Vec<P> {}
-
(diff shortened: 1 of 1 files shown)


maciejhirsz (approved):

jsdw:
LGTM! Since there doesn't appear to be a zero-length array of zero-tuple impl, It's a little harder to pass in 0 params. Is it worth impling the trait on an empty tuple as well?

(also, the `$len` part of the tuple impl isn't used so I wonder if it's worth keeping around?)

jsdw (approved):

dvdplm (approved):

PR #479, opened 2021-09-20
examples: remove weather

niklasad1:
The rationale behind this is `restjson` brings in many annoying dev-dependencies when using `cargo test`, so this speeds up the build `241` vs `294` on master.

Additional, `restjson` relies on tokio0.2 and panics with the new async API in 1.0 because we use tokio1 however we could use the blocking API...

//cc @dvdplm


dvdplm (approved):
I had the same thought. Ty.

maciejhirsz (approved):

Issue #481, opened 2021-09-21
`SubscriptionSink.send` does not return error for disconnected client.


kukabi:
Steps to reproduce:
1. Add a subscription method to the `RpcModule` with `register_subscription` . The loop in the inner spawned thread sends to the `SubscriptionSink` at regular intervals.
2. Subscribe from the client, server starts publishing values.
3. Disconnect the client without unsubscribing.
4. The spawned thread still can `send` to the `SubscriptionSink` and unexpectedly receives `Ok(())`.

If the client had unsubscribed then `send` expectedly returns `Err`.

_Note:_ Tested with `WsServerBuilder`.

niklasad1:
Thanks, yepp I could reproduce this and I agree.

Weird I thought I wrote a test for this but it was for the other way around.

niklasad1:
I checked this and it's a `ws client` bug when the client is dropped we should send unsubscribe requests to all active subscriptions then the server will remove the subscription and not continue to send out the messages that the client is not interested in anymore.

For now, you have to `drop` the subscriptions manually in order for this to work but I will fix this tomorrow, should be quite easy to fix.

kukabi:
Thanks a lot @niklasad1, I follow the edit in your comment, as the service provider I don't ever have access to how the clients handle their subscription/connection, and even if they handled the subscription properly they could also abruptly lose connection.

Not sure if it's ready but I just gave `na-fix-481` a try, there's a new error message `rx ws data failed: Closed; closing connection` when I cut the connection at the server side before any unsubscription, yet I'm still able to `send` to the `SubscriptionSink`. It returns `Ok`.

niklasad1:
Yeah, I can probably remove that error message is just that the client closed the connection we didn't have that log before.

The issue on the server side is basically when we create subscription the `mpsc::Sender` is cloned and put in the Hashmap in the `rpc module (outside the server struct called)` and that subscription will not be possible to close (because the strong count on mpsc::Sender will be bigger > 1) without sending an unsubscribe request to it. For instance when a connection is closed by the client those subscriptions will "be alive" because the all the senders of that channel is not dropped and `SubscriptionSink::send` will not return an error.

Thus, the correct fix would be better if the ws server would remove all subscriptions associated with the connection ID that got closed (the subscription IDs are not known by the server currently) then `SubscriptionSink::send` will return error.

But that requires some more refactoring and might that a little longer to fix.

niklasad1:
Fixed by #483

PR #483, opened 2021-09-22
fix: ws server terminate subscriptions when connection is closed by the client.

niklasad1:
Closing #481

The underlying issue is that `receiver` of the connection channel is not dropped when the connection is terminated and the correspond senders are stored in the subscription HashMap. 

With this change the `receiver` is dropped when the connection is terminated and when  `SubscriptionSink::send` is tried it will fail and cause the server the remove the subscription.


ws-server/src/server.rs
```diff
@@ -241,7 +241,13 @@ async fn background_task(
 	while !stop_monitor.shutdown_requested() {
 		data.clear();
 
-		method_executors.select_with(receiver.receive_data(&mut data)).await?;
+		if let Err(e) = method_executors.select_with(receiver.receive_data(&mut data)).await {
```
niklasad1, line 244: NOTE; this will kill a connection on any soketto::connection::Error,

In a followup PR I think we should try to do some graceful error handling

ws-server/src/server.rs
```diff
@@ -241,7 +241,13 @@ async fn background_task(
...
+			log::error!("rx ws data failed: {:?}; closing connection", e);
+			if let Ok(stop_fut) = stop_monitor.handle().stop() {
```
niklasad1, line 246: this is wrong.... it will terminate the entire server

types/src/client.rs
```diff
@@ -148,6 +148,10 @@ pub enum FrontToBack {
 	// Such operations will be blocked until a response is received or the background
 	// thread has been terminated.
 	SubscriptionClosed(SubscriptionId),
+	/// Terminate background thread.
+	///
+	/// This implies that all pending operations are ignored and the connection is terminated.
+	Terminate,
```
niklasad1, line 154: we need this message because once we have either pending request or subscriptions the background channel Sender is cloned and the drop impl won't work.

another option is keep it as it is and require users to explicitly drop subscriptions for this to work.

ws-client/src/helpers.rs
```diff
@@ -181,6 +181,20 @@ pub async fn stop_subscription(sender: &mut WsSender, manager: &mut RequestManag
 	}
 }
 
+/// Garbage collect all active subscriptions.
+pub async fn close_all_subscriptions(sender: &mut WsSender, mut manager: RequestManager) {
+	let subs = manager.subscriptions();
```
niklasad1, line 186: we could maybe batch these instead but the code will become event uglier w.r.t borrowing issues.

ws-server/src/server.rs
```diff
@@ -209,16 +211,19 @@ async fn background_task(
...
-	stop_monitor: StopMonitor,
+	stop_server: StopMonitor,
 ) -> Result<(), Error> {
 	// And we can finally transition to a websocket background_task.
 	let (mut sender, mut receiver) = server.into_builder().finish();
 	let (tx, mut rx) = mpsc::unbounded::<String>();
+	let stop_conn = Arc::new(AtomicBool::new(false));
+
+	let stop_server2 = stop_server.clone();
+	let stop_conn2 = stop_conn.clone();
```
niklasad1, line 222: we need this flag to stop the inner task that holds the `receiver` of the channel when the connection is dropped by the client.

ws-server/src/server.rs
```diff
@@ -209,16 +211,19 @@ async fn background_task(
...
 
-	let stop_monitor2 = stop_monitor.clone();
 	// Send results back to the client.
 	tokio::spawn(async move {
-		while !stop_monitor2.shutdown_requested() {
+		while !stop_server2.shutdown_requested() && !stop_conn2.load(Ordering::SeqCst) {
```
niklasad1, line 226: it could probably be ok to use `Relaxed` here....

maciejhirsz (approved):

ws-client/src/client.rs
```diff
@@ -656,19 +662,22 @@ async fn background_task(
...
 			Either::Right((None, _)) => {
 				log::error!("[backend]: WebSocket receiver dropped; terminate client");
 				let _ = front_error.send(Error::Custom("WebSocket receiver dropped".into()));
-				return;
+				break;
 			}
 		}
 	}
+
+	// Send WebSocket close reason to the server (this might fail if the server already terminated the connection).
```
dvdplm, line 681: The comment seems to indicate we actually send a close reason, which it's not, really (it's always hard coded to `1000` I think?).

tests/tests/integration_tests.rs
```diff
@@ -295,3 +294,45 @@ async fn ws_close_pending_subscription_when_server_terminated() {
...
+
+	let mut sub: Subscription<usize> =
+		client.subscribe("subscribe_hello", ParamsSer::NoParams, "unsubscribe_hello").await.unwrap();
+
+	let res = sub.next().await.unwrap();
+
+	assert_eq!(res.as_ref(), Some(&1));
+	drop(client);
+	// assert that the server received `SubscriptionClosed` after the client was dropped.
+	assert!(rx.next().await.is_some());
```
dvdplm, line 337: This is a bit opaque unless one knows exactly what is going on.

Maybe we can do something like this?

```
	let sub_closed_err = rx.next().await.expect("Server received `SubscriptionClosed` after the client was dropped.");
	assert!(format!("{:?}", sub_closed_err).contains("is closed but not yet dropped"));
```

It's a bit annoying that `SubscriptionClosedErr` has a private member (and that there's no way for the client to know what the subscription ID is) or we could have used `assert_matches` here for better readability...

dvdplm (approved):

ws-client/src/client.rs
niklasad1: yepp, bad comment I will re-phrase.

tests/tests/integration_tests.rs
niklasad1: Yepp, I think we could replace SubscriptionClosed(String) with a better type.

Perhaps with a `kind enum`, to distinguish why the subscription was closed the error message is quite useful atm

tests/tests/integration_tests.rs
```diff
@@ -334,5 +334,5 @@ async fn ws_server_should_stop_subscription_after_client_drop() {
 	assert_eq!(res.as_ref(), Some(&1));
 	drop(client);
 	// assert that the server received `SubscriptionClosed` after the client was dropped.
-	assert!(rx.next().await.is_some());
+	assert!(matches!(rx.next().await.unwrap(), SubscriptionClosedError(_)));
```
dvdplm, line 337: missing import here :)

PR #489, opened 2021-09-28
wrapper struct for test subscription

niklasad1:


dvdplm (approved):
LGTM, much cleaner than before.

PR #490, opened 2021-09-29
Share the request id code between the http and websocket clients

### LINKED ISSUE #290: [clients]: new type abstraction for `atomic request IDs`
We are using `AtomicU64` in the clients to get new `request ID(s)` and those are exposed as raw `AtomicU64`. 
There is nothing really that prevents `read/store` the values with inproper ordering  thus we have made them explicit `SeqCst` to be on the safe side.

It would be better to wrap those `AtomicU64` in a `new type` such as `AtomicCounter` to prevent misuse of them and I think that's the only proper use-case for `Ordering::Relaxed`

Also, see the discussion in #289 for further information

dvdplm:
Move the `RequestIdGuard` type to `jsonrpsee-types` and use it in both clients. This means the http client now also supports the `max_concurrent_requests`.

Closes #290

### DIFF
--- a/http-client/src/client.rs
+++ b/http-client/src/client.rs
@@ -28,12 +28,11 @@ use crate::transport::HttpTransportClient;
 use crate::types::{
 	traits::Client,
 	v2::{Id, NotificationSer, ParamsSer, RequestSer, Response, RpcError},
-	Error, TEN_MB_SIZE_BYTES,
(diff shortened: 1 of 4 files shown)


niklasad1 (approved):
:+1:

maciejhirsz (approved):

PR #495, opened 2021-10-01
Unbox async futures

maciejhirsz:
There is no need to box futures for async methods in user space, since the closure is captured and boxed internally already.

### DIFF
--- a/benches/helpers.rs
+++ b/benches/helpers.rs
@@ -1,5 +1,4 @@
 use futures_channel::oneshot;
-use futures_util::future::FutureExt;
 use jsonrpsee::{
 	http_server::HttpServerBuilder,
 	ws_server::{RpcModule, WsServerBuilder},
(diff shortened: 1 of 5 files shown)


dvdplm (approved):

benches/helpers.rs
```diff
@@ -32,7 +31,7 @@ pub async fn ws_server() -> String {
 		let server = WsServerBuilder::default().build("127.0.0.1:0").await.unwrap();
 		let mut module = RpcModule::new(());
 		module.register_method(SYNC_METHOD_NAME, |_, _| Ok("lo")).unwrap();
-		module.register_async_method(ASYNC_METHOD_NAME, |_, _| (async { Ok("lo") }).boxed()).unwrap();
+		module.register_async_method(ASYNC_METHOD_NAME, |_, _| (async { Ok("lo") })).unwrap();
```
jsdw, line 34: Do we still need the parens around the `async` block?

jsdw (approved):
Cool!

benches/helpers.rs
maciejhirsz: Right, don't think they were needed there to begin with!

jsdw: I thought that too when I saw that they weren't used below in a couple of places :D

PR #496, opened 2021-10-01
Add macro to build params

### LINKED ISSUE #275: [jsonrpsee types]: provide macro for constructing JsonRpcParams
The `JsonRpcParams` are a little boiler-platy to provide as we require the arguments to be `serde_json::Value`, 

Example:
```rust
// MyType doesn't implement `Into<serde_json::Value>`
let x: = MyType { };
let inner: &[JsonValue] = &[0_u32.into(), "hoo".to_string().into(), serde_json::to_value(x).unwrap()];
let params =  JsonRpcParams::ArrayRef(inner);
```

So it would be nice is to provide a macro such as `params_positional!` and `params_object` to avoid that.
Perhaps just provide a wrapper on-top of [serde_json::json!](https://docs.serde.rs/serde_json/macro.json.html)?

Note: Copy @kianenigma's stuff.

dvdplm:
Add a convenience macro to build `ParamsSer` for use in clients. Ty @kianenigma.

Closes #275

### DIFF
--- a/examples/ws_subscription.rs
+++ b/examples/ws_subscription.rs
@@ -25,28 +25,29 @@
 // DEALINGS IN THE SOFTWARE.
 
 use jsonrpsee::{
-	types::{traits::SubscriptionClient, v2::ParamsSer, Error, Subscription},
+	rpc_params,
+	types::{traits::SubscriptionClient, Error, Subscription},
 	ws_client::WsClientBuilder,
 	ws_server::{RpcModule, WsServerBuilder},
 };
 use std::net::SocketAddr;
 
-const NUM_SUBSCRIPTION_RESPONSES: usize = 10;
 
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
+	const NUM_SUBSCRIPTION_RESPONSES: usize = 5;
 	env_logger::init();
 	let addr = run_server().await?;
 	let url = format!("ws://{}", addr);
 
 	let client = WsClientBuilder::default().build(&url).await?;
 	let mut subscribe_hello: Subscription<String> =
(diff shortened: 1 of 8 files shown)


maciejhirsz (approved):
Looks clean!

examples/ws_subscription.rs
```diff
@@ -25,28 +25,28 @@
...
 	ws_client::WsClientBuilder,
 	ws_server::{RpcModule, WsServerBuilder},
 };
 use std::net::SocketAddr;
 
-const NUM_SUBSCRIPTION_RESPONSES: usize = 10;
-
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
+	const NUM_SUBSCRIPTION_RESPONSES: usize = 5;
```
insipx, line 37: why move const into the scope of a function, instead of just using a let/5 in the loop

insipx (approved):
lgtm other than minor nitpick

examples/ws_subscription.rs
dvdplm: You're right, this is dumb. Will fix.

utils/src/client.rs
```diff
@@ -0,0 +1,49 @@
...
+
+//! Shared utilities for `jsonrpsee` clients.
+
+#[doc(hidden)]
+pub mod __reexports {
+	pub use jsonrpsee_types::{to_json_value, v2::ParamsSer};
+}
+
+#[macro_export]
+/// Convert the given values to a [`ParamsSer`] as expected by a jsonrpsee Client (http or websocket).
```
niklasad1, line 35: I guess we could document that this panics if the serialization fails which is very unlikely but possible I guess....

PR #500, opened 2021-10-01
Resource Limiting

### LINKED ISSUE #203: Resource limiting
A common problem JSON-RPC servers run into is making sure the public facing API can be made resilient to DOS attacks (or just plain, unintentional overload). The following are solutions we'd like to see in jsonrpsee, following a chat with @gww-parity and @niklasad1.

# MVP (2.0 stable release)

The simplest solution is to provide a single cap of in-flight requests, either globally for the HTTP server, or per connection for the WS server, with an additional cap for the number of connections.

This is pretty crude since we can't distinguish between requests that are blocking on CPU, IO, or some other reason, but it should be a sufficient approximation given a limit being some factor of available CPU cores. The implementation is pretty low effort here since we are already using bounded channels internally, so providing an informative error objects when the limit is hit would be sufficient here. For the WS server adding a single atomic int to track resources globally should also work fine and not deteriorate performance too much given we are in macrosecond space already due to IO.

There is no defined error code for the server being busy, so we'll need to use either `-32603` for internal error, or define a custom error code in the `-32000` to `-32099` range.

# The dream solution (undefined future release)

Down the line we should be able to provide a more granular control over resources for advanced users. To make this performant on runtime we need to make sure that we are using a builder pattern for method definitions before starting a server (already the case for WS).

The user should be able to register a resource on the server, similarly to the way they register a method. For each resource the user defines it's name as a string, and an arbitrary integer value that reflects the maximum capacity of that resource, e.g.:

```rust
builder.register_resource("cpu", 100);
builder.register_resource("io", 500);
```

Each method registered on the builder would default to having all of its resource costs set to `1`. Alternatively, users should be able to set individual resources, either using a map, or with some method chaining in the api, e.g.:

```rust
builder.register_method("cpu_heavy", ...).set_resource("cpu", 20);
```

The limits are then checked (naively globally) for each defined resource before triggering the request, and if exhausted an error object is returned, same as the MVP implementation.

To keep things fast at runtime we map each defined resource name to an index in an array (there should be a limit to how many resources a user can register), so in the examples above `"cpu"` would be `0`, `"io"` would be `1`. On runtime each request would then simply have a fixed-size array of integers (constructed when the builder is finalized) to check against a global array of limits, which should be fairly quick and easy to optimize with vector instructions if need be (assuming compiler doesn't auto-vectorize it).

maciejhirsz:
Closes #203 (details in the issue).

Remaining TODOs:

+ [ ] proc macros.
+ [ ] HTTP server.


utils/src/server/resource_limiting.rs
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	/// Attempt to claim units for each resource, incrementing current totals.
+	/// If successful returns a `ResourceGuard` which decrements the totals by the same
+	/// amounts once dropped.
+	pub fn claim(&self, units: ResourceTable<u16>) -> Result<ResourceGuard, Error> {
+		let mut totals = self.totals.lock();
+		let mut sum = *totals;
+
+		for (idx, sum) in sum.iter_mut().enumerate() {
```
dvdplm, line 52: So this is where we'd use simd right?

utils/src/server/rpc_module.rs
```diff
@@ -43,18 +44,20 @@ use parking_lot::Mutex;
...
 use serde::Serialize;
 use serde_json::value::RawValue;
+use std::collections::hash_map::Entry;
 use std::fmt::Debug;
 use std::future::Future;
 use std::ops::{Deref, DerefMut};
 use std::sync::Arc;
 
-/// A `Method` is an RPC endpoint, callable with a standard JSON-RPC request,
+/// A `MethodCallback` is an RPC endpoint, callable with a standard JSON-RPC request,
```
dvdplm, line 53: Should be `SyncMethod` I think?

utils/src/server/rpc_module.rs
```diff
@@ -43,18 +44,20 @@ use parking_lot::Mutex;
...
 /// implemented as a function pointer to a `Fn` function taking four arguments:
 /// the `id`, `params`, a channel the function uses to communicate the result (or error)
 /// back to `jsonrpsee`, and the connection ID (useful for the websocket transport).
 pub type SyncMethod = Arc<dyn Send + Sync + Fn(Id, Params, &MethodSink, ConnectionId)>;
 /// Similar to [`SyncMethod`], but represents an asynchronous handler.
```
dvdplm, line 58:
```suggestion
/// Similar to [`SyncMethod`], but represents an asynchronous handler and takes an additional argument containing a [`ResourceGuard`] if configured.
```

utils/src/server/rpc_module.rs
```diff
@@ -74,21 +77,82 @@ struct SubscriptionKey {
...
 #[derive(Clone)]
-pub enum MethodCallback {
+enum MethodKind {
 	/// Synchronous method handler.
 	Sync(SyncMethod),
 	/// Asynchronous method handler.
 	Async(AsyncMethod<'static>),
 }
 
+/// Information about resources the method uses during its execution.
```
dvdplm, line 87:
```suggestion
/// Information about resources the method uses during its execution. Lazily initialized when the the server starts.
```

utils/src/server/rpc_module.rs
```diff
@@ -137,20 +201,57 @@ impl Methods {
...
+	fn verify_method_name(&mut self, name: &'static str) -> Result<(), Error> {
 		if self.callbacks.contains_key(name) {
 			return Err(Error::MethodAlreadyRegistered(name.into()));
 		}
 
 		Ok(())
 	}
 
+	/// Inserts the method callback for a given name, or returns an error if the name was already taken.
+	/// On success returns a mut reference to just inserted callback.
```
dvdplm, line 213:
```suggestion
	/// On success it returns a mut reference to the [`MethodCallback`] just inserted.
```

utils/src/server/rpc_module.rs
```diff
@@ -174,10 +275,35 @@ impl Methods {
...
-			Some(callback) => callback.execute(tx, req, conn_id),
+			Some(callback) => callback.execute(tx, req, conn_id, None),
+			None => {
+				send_error(req.id, tx, ErrorCode::MethodNotFound.into());
+				None
+			}
+		}
+	}
+
+	/// Attempt to execute a callback, sending the resulting JSON (success or error) to the specified sink.
```
dvdplm, line 289:
```suggestion
	/// Attempt to execute a callback while checking that the call does not exhaust the available resources, sending the resulting JSON (success or error) to the specified sink.
```

ws-server/src/server.rs
```diff
@@ -365,6 +396,14 @@ impl Builder {
 		self
 	}
 
+	/// Register a new resource type. Errors if `label` was already registered, or if number of
```
dvdplm, line 399:
```suggestion
	/// Register a new resource kind. Errors if `label` is already registered, or if number of
```

ws-server/src/tests.rs
```diff
@@ -540,12 +542,12 @@ async fn can_register_modules() {
...
 async fn stop_works() {
 	let _ = env_logger::try_init();
-	let (_addr, join_handle, stop_handle) = server_with_handles().with_default_timeout().await.unwrap();
+	let (_addr, stop_handle) = server_with_handles().with_default_timeout().await.unwrap();
 	stop_handle.clone().stop().unwrap().with_default_timeout().await.unwrap();
 
 	// After that we should be able to wait for task handle to finish.
 	// First `unwrap` is timeout, second is `JoinHandle`'s one.
-	join_handle.with_default_timeout().await.expect("Timeout").expect("Join error");
+	// join_handle.with_default_timeout().await.expect("Timeout").expect("Join error");
```
dvdplm, line 546: cleanup?

ws-server/src/server.rs
```diff
@@ -273,7 +300,10 @@ async fn background_task(
 						// complete batch response back to the client over `tx`.
 						let (tx_batch, mut rx_batch) = mpsc::unbounded::<String>();
 
-						for fut in batch.into_iter().filter_map(|req| methods.execute(&tx_batch, req, conn_id)) {
+						for fut in batch
```
dvdplm, line 303: Random thought: it might be possible to calculate the overall sum of required resources for the whole batch call and accept/refuse it accordingly. Like it is, users could get themselves in a bind if half their batch is overquota and the calls are not idempotent. One could reply "you should have thought of that before using batch calls!", but that would be a bit unfriendly...

Might be good to have a test that illustrates the behaviour for batches in the presence of resource limiting.

utils/src/server/rpc_module.rs
```diff
@@ -370,12 +498,14 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 							send_error(id, &tx, err)
 						}
 					};
+
+					drop(claimed);
```
dvdplm, line 502:
```suggestion
					// Release claimed resources
					drop(claimed);
```

…and add an explicit `drop` for the sync case as well, for readability?

dvdplm (commented):
I like it.

types/src/error.rs
```diff
@@ -142,6 +142,21 @@ pub enum Error {
...
+	/// Failed to execute a method because a resource was already at capacity
+	#[error("Resource at capacity: {0}")]
+	ResourceAtCapacity(&'static str),
+	/// Failed to register a resource due to a name conflict
+	#[error("Resource name already taken: {0}")]
+	ResourceNameAlreadyTaken(&'static str),
+	/// Failed to initialize resources for a method
+	#[error("Resource name `{0}` not found for method `{1}`")]
+	ResourceNameNotFoundForMethod(&'static str, &'static str),
+	/// Trying to claim resources for a method, but method has not been initialized
```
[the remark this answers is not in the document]
maciejhirsz: Yeah, more explicit is probably better.

utils/src/server/resource_limiting.rs
```diff
@@ -0,0 +1,82 @@
+use std::sync::Arc;
+
+use arrayvec::ArrayVec;
+use jsonrpsee_types::error::Error;
+use parking_lot::Mutex;
+
+const RESOURCE_COUNT: usize = 8;
+
+/// Fixed size table mapping a resource to some value
```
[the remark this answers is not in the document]
maciejhirsz: Having checked now, I think all uses of `ResourceTable` are using `u16`, so can remove the generic and make the comment clearer.

utils/src/server/resource_limiting.rs
```diff
@@ -0,0 +1,82 @@
...
+pub type ResourceTable<T> = [T; RESOURCE_COUNT];
+/// Variable size table mapping a resource to some value
+pub type ResourceVec<T> = ArrayVec<T, RESOURCE_COUNT>;
+
+/// User defined resources used by the JSON-RPC server.
```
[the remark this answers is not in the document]
maciejhirsz: There is no real point in defining a resource if it's infinite, and defining `0` is kind of pointless as it would only allow executing methods that cost `0`, and then allow infinite number of those. Not sure if those edge cases need describing or whether or not they are kind of intuitive :man_shrugging:?

utils/src/server/resource_limiting.rs
maciejhirsz: Yup!

utils/src/server/rpc_module.rs
maciejhirsz: Sans `Lazily`, it's doing opposite of lazy initialization atm.

ws-server/src/tests.rs
```diff
@@ -540,12 +538,12 @@ async fn can_register_modules() {
...
```
maciejhirsz, line 546:
```suggestion
```

tests/tests/resource_limiting.rs
```diff
@@ -0,0 +1,104 @@
...
+	ws_client::WsClientBuilder,
+	ws_server::{WsServerBuilder, WsStopHandle},
+	RpcModule,
+};
+use tokio::time::sleep;
+
+use std::time::Duration;
+use std::net::SocketAddr;
+
+pub async fn websocket_server() -> Result<(SocketAddr, WsStopHandle), Error> {
```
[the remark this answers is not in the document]
maciejhirsz: This one doesn't have to be `pub` at all :upside_down_face:.

utils/src/server/resource_limiting.rs
```diff
@@ -0,0 +1,83 @@
...
+
+// The number of kinds of resources that can be used for limiting.
+const RESOURCE_COUNT: usize = 8;
+
+/// Fixed size table, mapping a resource to a (unitless) value indicating the amount of the resource that is available to RPC calls.
+pub type ResourceTable<T> = [T; RESOURCE_COUNT];
+/// Variable size table, mapping a resource to a (unitless) value indicating the amount of the resource that is available to RPC calls.
+pub type ResourceVec<T> = ArrayVec<T, RESOURCE_COUNT>;
+
+/// User defined resources available to be used by calls on the JSON-RPC server. Each of the 8 possible resource kinds, for instance "cpu", "io", "nanobots", store a maximum `capacity` and a default. A value of `0` means no limits for the given resource.
```
niklasad1, line 15: nitpick; a little too long line :)

utils/src/server/rpc_module.rs
```diff
@@ -74,21 +77,82 @@ struct SubscriptionKey {
...
 	/// Asynchronous method handler.
 	Async(AsyncMethod<'static>),
 }
 
+/// Information about resources the method uses during its execution.
+#[derive(Clone, Debug)]
+enum MethodResources {
+	/// Unintialized resource table, mapping string label to units.
+	Uninitialized(Box<[(&'static str, u16)]>),
+	/// Intialized resource table containing units for each `ResourceId`.
```
niklasad1, line 92:
```suggestion
	/// Initialized resource table containing units for each `ResourceId`.
```

utils/src/server/rpc_module.rs
```diff
@@ -74,21 +77,82 @@ struct SubscriptionKey {
...
 	/// Synchronous method handler.
 	Sync(SyncMethod),
 	/// Asynchronous method handler.
 	Async(AsyncMethod<'static>),
 }
 
+/// Information about resources the method uses during its execution.
+#[derive(Clone, Debug)]
+enum MethodResources {
+	/// Unintialized resource table, mapping string label to units.
```
niklasad1, line 90:
```suggestion
	/// Uninitialized resource table, mapping string label to units.
```

ws-server/src/server.rs
```diff
@@ -68,9 +70,18 @@ impl Server {
 	}
 
 	/// Start responding to connections requests. This will block current thread until the server is stopped.
-	pub async fn start(self, methods: impl Into<Methods>) {
+	pub fn start(self, methods: impl Into<Methods>) -> Result<StopHandle, Error> {
```
niklasad1, line 73: ah, lovely

niklasad1 (commented):
looks really promising, I will have a look tomorrow when my head is fresh...

tests/tests/resource_limiting.rs
```diff
@@ -0,0 +1,221 @@
...
+	);
+
+	assert!(pass_cpu1.is_ok());
+	assert!(pass_cpu2.is_ok());
+	assert_server_busy(fail_cpu);
+	assert!(pass_mem.is_ok());
+	assert_server_busy(fail_mem);
+
+	// Client being active prevents the server from shutting down?!
+	drop(client);
```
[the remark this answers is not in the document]
maciejhirsz: It _should_ do that, but if you remove the drop the test runs forever, so it never actually does shut down :upside_down_face:

utils/src/server/resource_limiting.rs
```diff
@@ -0,0 +1,85 @@
...
+	/// Default value for all resource kinds; unless a method has a resource limit defined, this is the cost of a call (0 means no default limit)
+	pub defaults: ResourceTable<u16>,
+	/// Labels for every registered resource
+	pub labels: ResourceVec<&'static str>,
+}
+
+impl Resources {
+	/// Register a new resource kind. Errors if `label` is already registered, or if the total number of
+	/// registered resources would exceed 8.
+	pub fn register(&mut self, label: &'static str, capacity: u16, default: u16) -> Result<(), Error> {
```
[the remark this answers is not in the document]
maciejhirsz: We talked about it with @dvdplm at the retreat. The problem with % is that if you add hardware you can easily reconfigure max capacity, where as going in and adjusting every method to lower cpu or memory % would be quite a hassle.

ws-server/src/tests.rs
```diff
@@ -59,7 +58,7 @@ async fn server() -> SocketAddr {
 ///     other: `invalid_params` (always returns `CallError::InvalidParams`), `call_fail` (always returns
 /// 			`CallError::Failed`), `sleep_for` Returns the address together with handles for server future
 ///				and server stop.
-async fn server_with_handles() -> (SocketAddr, JoinHandle<()>, StopHandle) {
+async fn server_with_handles() -> (SocketAddr, StopHandle) {
```
niklasad1, line 61: :+1:

niklasad1 (approved):
nice work, looks clean as always

tests/tests/resource_limiting.rs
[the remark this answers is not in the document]
niklasad1: ah I see, do you mind creating an issue for this?

tests/tests/resource_limiting.rs
```diff
@@ -0,0 +1,221 @@
...
+		}
+
+		#[method(name = "memory_hog", resources("CPU" = 0, "MEM" = 8))]
+		async fn memory(&self) -> Result<&'static str, Error> {
+			sleep(Duration::from_millis(50)).await;
+			Ok("hello memory hog")
+		}
+	}
+
+	impl RpcServer for () {}
```
TarikGul, line 89: DQ: Whats happening here?

TarikGul (approved):
👍

proc-macros/src/attributes.rs
```diff
@@ -24,58 +24,157 @@
...
+
+pub(crate) struct Argument {
+	pub label: syn::Ident,
+	pub tokens: TokenStream2,
+}
+
+#[derive(Debug, Clone)]
+pub struct Resource {
+	pub name: syn::LitStr,
+	pub assign: Token![=],
```
jsdw, line 46: Is this actually needed in the struct?

types/src/error.rs
```diff
@@ -142,6 +142,21 @@ pub enum Error {
...
+	ResourceAtCapacity(&'static str),
+	/// Failed to register a resource due to a name conflict
+	#[error("Resource name already taken: {0}")]
+	ResourceNameAlreadyTaken(&'static str),
+	/// Failed to initialize resources for a method at startup
+	#[error("Resource name `{0}` not found for method `{1}`")]
+	ResourceNameNotFoundForMethod(&'static str, &'static str),
+	/// Trying to claim resources for a method execution, but the method resources have not been initialized
+	#[error("Method `{0}` has uninitialized resources")]
+	UninitializedMethod(Box<str>),
```
jsdw, line 156: Bike shedding: I wonder whether something like `ResourceUnitializedForMethod` might be a touch more consistent with the above resource related (and prefixed) errors?

proc-macros/src/rpc_macro.rs
```diff
@@ -121,8 +132,16 @@ pub struct RpcDescription {
 	pub(crate) jsonrpsee_client_path: Option<TokenStream2>,
 	/// Path to the `jsonrpsee` server types part.
 	pub(crate) jsonrpsee_server_path: Option<TokenStream2>,
-	/// Data about RPC declaration
-	pub(crate) attrs: attributes::Rpc,
+	/// Switch denoting that server trait must be generated.
+	/// Assuming that trait to which attribute is applied is named `Foo`, the generated
+	/// server trait will have `FooServer` name.
```
dvdplm, line 137:
```suggestion
	/// Switch denoting that the server trait must be generated.
	/// Assuming that the trait to which the attribute is applied is named `Foo`, the generated
	/// server trait will be named `FooServer`.
```

proc-macros/src/rpc_macro.rs
```diff
@@ -121,8 +132,16 @@ pub struct RpcDescription {
...
+	pub(crate) needs_server: bool,
+	/// Switch denoting that client extension trait must be generated.
+	/// Assuming that trait to which attribute is applied is named `Foo`, the generated
+	/// client trait will have `FooClient` name.
```
dvdplm, line 141:
```suggestion
	/// Switch denoting that the client extension trait must be generated.
	/// Assuming that the trait to which the attribute is applied is named `Foo`, the generated
	/// client trait will be named `FooClient`.
```

proc-macros/src/rpc_macro.rs
```diff
@@ -121,8 +132,16 @@ pub struct RpcDescription {
...
+	pub(crate) needs_client: bool,
+	/// Optional prefix for RPC namespace.
```
dvdplm, line 143:
```suggestion
	/// Optional prefix for the RPC namespace.
```

tests/tests/resource_limiting.rs
```diff
@@ -0,0 +1,221 @@
...
+			Ok("hello expensive call")
+		})?
+		.resource("CPU", 3)?;
+
+	module
+		.register_async_method("memory_hog", |_, _| async move {
+			sleep(Duration::from_millis(50)).await;
+			Ok("hello memory hog")
+		})?
+		.resource("CPU", 0)?
```
dvdplm, line 61: Add comment explaining that this is the same as omitting the resource, i.e. that `0` is the default?

ws-server/src/server.rs
```diff
@@ -365,6 +396,14 @@ impl Builder {
 		self
 	}
 
+	/// Register a new resource kind. Errors if `label` is already registered, or if number of
+	/// registered resources would exceed 8.
```
dvdplm, line 400: Please document the meaning of `capacity` and `default`.

I wonder if having a `ResourceBuilder` with `.with_cap` and `.with_default` methods would make the api more self-explanatory? Not a blocker but one to ponder.

jsdw (approved):
Looks great!

Definitely some of the more readable proc macro code I've come across, too :D

ws-server/src/server.rs
dvdplm: @maciejhirsz did we get a test&docs for this somewhere?

We had a conversation about this and came to the conclusion that while it's possible to calculate the batch "cost" up front, the semantics would become rather awkward and the overall behaviour of the server a bit hard to understand and reason about; on a busy server a batch could end up in a situation where it never gets enough resources to run or, conversely, a sequentially executed batch might block other requests from running.

Issue #485, opened 2021-09-24
[types]: SubscriptionClosedError


niklasad1:
Improve `SubscriptionClosedError` to include a more descriptive such as `Client disconnected`, `Server closed` and similar.

Would be better to separate the actual error message and subscription ID.

PR #504, opened 2021-10-06
improve SubscriptionClosed error

niklasad1:
Closing #485


dvdplm (approved):
LGTM.

On a higher level I wonder if closing a subscription should really be treated as an error. I think it'd be better to have a `ConnectionState` enum with `Active`, `ClientClosed`, `ServerClosed`, `Errored` variants. But that's a bigger refactoring.

PR #507, opened 2021-10-07
Reworking attribute macros for resource limiting

maciejhirsz:
To get the macros working nicely I'm trying to do a bit more generic manual parsing of `#[attributes]`, so that we can permit arbitrary syntax we need.

This PR replaces the `bae` "derived" `Method` and `Subscription` structs with a single `Attr` struct that contains arbitrary "arguments" (using the nomenclature from `bae`, naming is hard). Not sure why `non_exhaustive` on `ArgumentBody` enum doesn't seem to work, want to add different body types there:

+ `Value` (the only available right now): `foo = <tokens...>,`
+ `Group` with nested `foo(<values...>)`, so we can do stuff like `resource("cpu", 10)`
+ `Flag` unit variant: `foo,`

I also reckon it might be a good idea to bikeshed a bit on the syntax here, do we want one of these?

1. `#[method(name = "foo", resource("cpu", 2), resource("mem", 1))]`
1. `#[method(name = "foo", resources = ["cpu" 2, "mem" 1])]`
1. `#[method(name = "foo", resources("cpu" = 2, "mem" = 1))]`
1. Something else?

### DIFF
--- a/proc-macros/src/attributes.rs
+++ b/proc-macros/src/attributes.rs
@@ -24,57 +24,73 @@
 // IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 // DEALINGS IN THE SOFTWARE.
 
-use crate::helpers::is_punct;
-use proc_macro2::{Delimiter, TokenStream as TokenStream2, TokenTree};
+use proc_macro2::{TokenStream as TokenStream2, TokenTree};
 use std::fmt;
-use syn::{spanned::Spanned, Attribute, Error};
+use syn::{spanned::Spanned, Attribute, Error, Token};
+use syn::punctuated::Punctuated;
+use syn::parse::{Parse, Parser, ParseStream};
 
-#[derive(Debug)]
 pub(crate) struct AttributeMeta {
 	pub path: syn::Path,
-	pub arguments: Vec<Argument>,
+	pub arguments: Punctuated<Argument, Token![,]>,
 }
 
-#[derive(Debug)]
 pub(crate) struct Argument {
 	pub label: syn::Ident,
 	pub tokens: TokenStream2,
+}
+
+#[derive(Debug, Clone)]
+pub struct Resource {
+	pub name: syn::LitStr,
+	pub assign: Token![=],
+	pub value: syn::LitInt,
+}
+
+impl Parse for Argument {
+	fn parse(input: ParseStream) -> syn::Result<Self> {
+		let label = input.parse()?;
+
+		let mut tokens = Vec::new();
+
+		while !input.peek(Token![,]) {
+			match input.parse::<TokenTree>() {
+				Ok(token) => tokens.push(token),
+				Err(_) => break,
+			}
+		}
+
+		Ok(Argument {
+			label,
+			tokens: tokens.into_iter().collect(),
+		})
+	}
+}
+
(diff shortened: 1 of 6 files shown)


proc-macros/src/rpc_macro.rs
```diff
@@ -43,12 +46,15 @@ pub struct RpcMethod {
 }
 
 impl RpcMethod {
-	pub fn from_item(mut method: syn::TraitItemMethod) -> Result<Self, syn::Error> {
-		let attributes = attributes::Method::from_attributes(&method.attrs).respan(&method.attrs.first())?;
+	pub fn from_item(attr: Attribute, mut method: syn::TraitItemMethod) -> syn::Result<Self> {
+		let [aliases, name] = AttributeMeta::parse(attr)?.retain(["aliases", "name"])?;
+
+		let aliases = parse_aliases(aliases)?;
+		let name = name?.string()?;
```
maciejhirsz, line 53: I think I'm finally happy with the proc-macro-inside-proc-macro-less `bae` replacement here.

proc-macros/src/attributes.rs
```diff
@@ -24,58 +24,179 @@
...
 	}
 
-	/// Returns `true` if client implementation was requested.
-	pub(crate) fn needs_client(&self) -> bool {
-		self.client.is_some()
+	/// Attempt to get a list of `Argument`s from a list of names in order.
+	///
+	/// Errors if there is an argument with a name that's not on the list, or if there is a duplicate definition.
+	pub fn retain<const N: usize>(self, allowed: [&str; N]) -> syn::Result<[syn::Result<Argument>; N]> {
+		// TODO: is there a static assert for const generics?
```
niklasad1, line 110: maybe the static assertion crate will work for this?

proc-macros/src/attributes.rs
```diff
@@ -24,58 +24,179 @@
...
+			// 		},
+			// 		_ => None,
+			// 	}.unwrap_or_else(|| self.label.span());
+
+			// 	Err(Error::new(span, format!("Expected a value assignment for `{}`, but got a group instead", self.label)))
+			// }
+		}
+	}
+
+	/// Asserts tthat the argument is `key = "string"` and gets the value of the string
```
niklasad1, line 198:
```suggestion
	/// Asserts that the argument is `key = "string"` and gets the value of the string
```

proc-macros/src/attributes.rs
```diff
@@ -24,58 +24,179 @@
...
+	pub label: syn::Ident,
+	pub kind: ArgumentKind,
+}
+
+#[non_exhaustive]
+#[derive(Debug)]
+pub(crate) enum ArgumentKind {
+	Flag,
+	Value(TokenStream2),
+	// Group(Vec<TokenStream2>),
```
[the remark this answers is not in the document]
maciejhirsz: Yeah, just unused code warnings breaking UI tests made me comment it for now :)

proc-macros/src/lib.rs
```diff
@@ -35,7 +35,7 @@ mod helpers;
 mod lifetimes;
 mod render_client;
 mod render_server;
-mod respan;
+// mod respan;
```
niklasad1, line 38:
```suggestion
```

PR #509, opened 2021-10-11
Proc macro Argument parsing should permit commas inside angle brackets

maciejhirsz:
This is not necessarily super robust. I've got an idea how to get around the whole "prase till comma" issue with some generics that automatically parse to expected types, instead of doing this two step parsing, will follow up with a separate PR for that.

### DIFF
--- a/proc-macros/src/attributes.rs
+++ b/proc-macros/src/attributes.rs
@@ -54,6 +54,9 @@ impl Parse for Argument {
 		let mut tokens = TokenStream2::new();
 		let mut scope = 0usize;
 
+		// Need to read to till either the end of the stream,
+		// or the nearest comma token that's not contained
+		// inside angle brackets.
 		loop {
 			if scope == 0 && input.peek(Token![,]) {
 				break;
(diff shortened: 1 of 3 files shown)


tests/tests/resource_limiting.rs
```diff
@@ -91,6 +91,23 @@ fn module_macro() -> RpcModule<()> {
 	().into_rpc()
 }
 
+#[tokio::test]
+async fn parse_angle_braces() {
+	// This needs to compile
```
dvdplm, line 96: maybe this is better off as a `trybuild` test?

proc-macros/src/attributes.rs
```diff
@@ -51,16 +51,27 @@ impl Parse for Argument {
 	fn parse(input: ParseStream) -> syn::Result<Self> {
 		let label = input.parse()?;
 
-		let mut tokens = Vec::new();
+		let mut tokens = TokenStream2::new();
+		let mut scope = 0usize;
+
+		loop {
```
dvdplm, line 57: A comment here would be good.

PR #510, opened 2021-10-12
Document resource limiting

maciejhirsz:


ws-server/src/server.rs
```diff
@@ -404,9 +404,50 @@ impl Builder {
 
 	/// Register a new resource kind. Errors if `label` is already registered, or if number of
 	/// registered resources would exceed 8.
+	///
```
jsdw, line 407: The thing that worries me about these duplicated docs is keeping them both in sync in the event that things change. I don't _think_ that there is a nice way to share documentation though is there?

I'd be tempted to have one version of the docs link to the other version eg

```
/// Register a new reousrce kind. See [`crate::heep_server::Builder::register_resource()`] for example usage.
```

http-server/src/server.rs
```diff
@@ -86,6 +86,48 @@ impl Builder {
 
 	/// Register a new resource kind. Errors if `label` is already registered, or if number of
 	/// registered resources would exceed 8.
+	///
+	/// Every method will temporarily lock either the `default` or otherwise specified number of units during its
+	/// execution. Any method execution that would cause the total sum of locked resource units to exceed
+	/// the `capacity` of that resource will be denied execution, immediately returning JSON-RPC error object with code `-32604`.
```
dvdplm, line 92:
```suggestion
	/// Each method will claim the specified number of units (or the default) for the duration of its execution
	/// Any method execution that would cause the total sum of claimed resource units to exceed
	/// the `capacity` of that resource will be denied execution, immediately returning JSON-RPC error object with code `-32604`.
```

http-server/src/server.rs
```diff
@@ -86,6 +86,48 @@ impl Builder {
...
+	///
+	/// To specify a different than default number of units a method should use, use the `resources` argument in the
+	/// `#[method]` attribute:
+	///
+	/// ```
+	/// # use jsonrpsee::{types::RpcResult, proc_macros::rpc};
+	/// #
+	/// #[rpc(server)]
+	/// pub trait Rpc {
+	/// 	#[method(name = "my_expensive_method", resources("cpu" = 5, "mem" = 2))]
```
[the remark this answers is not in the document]
dvdplm: > Either way; perhaps it'd be useful to make this clear?

I agree. It is not immediately obvious. I think one important piece to understand this is to first grok the way resources are set up: **not** here, around the trait marked with `#[rpc]` or at the method definition site; resources – along with their capacity and default cost – are set up on the **server** when it is constructed.

The reason for this separation is that the server app might want to make the resource set up configurable (through CLI or config file) so that different instances of the server can use different caps/defaults (e.g. fast servers == higher caps). It is also technically tricky (impossible?) to have a resource defined on the rpc modules and then somehow coalesced into a server-wide singleton.

So there's a UX problem here, but one we can't really fix. I think the docs should explain this important piece, as it makes understanding the rest much easier.

ws-server/src/server.rs
dvdplm: > I don't _think_ that there is a nice way to share documentation though is there?

I don't know of a way either and it's quite common to see "look over here for details" in rustdocs.

http-server/src/server.rs
```diff
@@ -86,6 +86,48 @@ impl Builder {
...
+	/// 	async fn my_expensive_method(&self) -> RpcResult<&'static str> {
+	/// 		// Do work
```
dvdplm, line 106: "using tabs in doc comments is not recommended" says the bot, but I'm not sure what the danger is. Either way, let's silence it somehow?

ws-server/src/server.rs
```diff
@@ -404,9 +404,50 @@ impl Builder {
 
 	/// Register a new resource kind. Errors if `label` is already registered, or if number of
 	/// registered resources would exceed 8.
```
dvdplm, line 406:
```suggestion
	/// Register a new resource kind. Errors if `label` is already registered, or if the number of
	/// registered resources on this server instance would exceed 8.
```

utils/src/server/mod.rs
```diff
@@ -28,7 +28,6 @@
 
 /// Helpers.
 pub mod helpers;
-/// Resource limiting helpers
 pub mod resource_limiting;
```
dvdplm, line 31:
```suggestion
/// Resource limiting. Create generic "resources" and configure their limits to ensure servers are not overloaded.
pub mod resource_limiting;
```

utils/src/server/resource_limiting.rs
```diff
@@ -1,3 +1,85 @@
...
+// IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+// DEALINGS IN THE SOFTWARE.
+
+//! # Resource Limiting
+//!
+//! This module handles limiting the capacity of the server to respond to requests.
+//!
+//! `jsonrpsee` is agnostic about the types of resources available on the server, and the units used are arbitrary.
+//! The units are used to model the availability of a resource, be it anything mundane like CPU or Memory,
+//! or anything more exotic like API accesses to a 3rd party service, or a use of some external hardware
```
dvdplm, line 33:
```suggestion
//! The units are used to model the availability of a resource, be it something mundane like CPU or Memory,
//! or more exotic things like remote API access to a 3rd party service, or use of some external hardware
```

utils/src/server/resource_limiting.rs
```diff
@@ -1,3 +1,85 @@
...
+//! that's under the control of the server.
+//!
+//! To get the most out of this feature, we suggest benchmarking individual methods to see how many resources they
+//! consume, in particular anything critical that is expected to result in a lot of stress on the server,
+//! and then defining your units such that the limits can be adjusted for different hardware configurations.
```
dvdplm, line 38:
```suggestion
//! and then defining your units such that the limits (`capacity`) can be adjusted for different hardware configurations.
```

utils/src/server/resource_limiting.rs
```diff
@@ -1,3 +1,85 @@
...
+//!
+//! Resources can be defined using the [`WsServerBuilder::register_resource`](../../../jsonrpsee_ws_server/struct.WsServerBuilder.html#method.register_resource)
+//! or [`HttpServerBuilder::register_resource`](../../../jsonrpsee_ws_server/struct.WsServerBuilder.html#method.register_resource) method
+//! for the WebSocket and HTTP server respectively.
+//!
+//! Each method will claim the specified number of units (or the default) for the duration of its execution
```
dvdplm, line 44:
```suggestion
//! Each method will claim the specified number of units (or the default) for the duration of its execution. 
```

utils/src/server/resource_limiting.rs
```diff
@@ -1,3 +1,85 @@
...
+//! Any method execution that would cause the total sum of claimed resource units to exceed
+//! the `capacity` of that resource will be denied execution, immediately returning JSON-RPC error object with code `-32604`.
+//!
```
dvdplm, line 47: Missing:
- 8 resources max for a single server instance
- a value of 0 means "no limits"
- resource names must be unique across all RPC modules
- misconfigured resources is a runtime error

dvdplm (changes requested):

utils/src/server/resource_limiting.rs
maciejhirsz: >     * a value of 0 means "no limits"

That's not a thing atm, I think we talked about it, though if resources are to be configurable outside it might make sense to make it a thing.

dvdplm: How do you mean it's not a thing? I was referring to the "default default" being zero and setting a resource cost to `0` for a method means that it does not cost any of that resource to run.

maciejhirsz: Ah, right. I think I got confused by the word "limit". That said if we are expecting configuration from outside, it might make sense to allow resources to be "disabled" by setting cap to `0` too.

ws-server/Cargo.toml
```diff
@@ -24,3 +24,4 @@ tokio-util = { version = "0.6", features = ["compat"] }
 anyhow = "1"
 env_logger = "0.9"
 jsonrpsee-test-utils = { path = "../test-utils" }
+jsonrpsee = { path = "../jsonrpsee", features=["full"] }
```
niklasad1, line 27:
```suggestion
jsonrpsee = { path = "../jsonrpsee", features = ["full"] }
```

utils/src/server/resource_limiting.rs
dvdplm: My bad, poorly worded. You're right about the cap though, saw the commit. Either way, it proves the point: we should explain properly what `0` means. :)

ws-server/src/server.rs
maciejhirsz: I still have the duplicated comments on these, but the bulk of it has been moved to the `resource_limiting` module, link included, so I reckon this is resolved now?

PR #514, opened 2021-10-12
docs(release): improve release checklist

TarikGul:
Updates the release checklist.

### DIFF
--- a/RELEASE-CHECKLIST.md
+++ b/RELEASE-CHECKLIST.md
@@ -5,7 +5,6 @@
 1. In the `CHANGELOG.md` file, move everything under "Unreleased" to a new section named `## [vx.y.z] – YYYY-MM-DD`
 1. Make a dryrun like so:
 	1. Ensure you're in the project root dir
-	1. Run: `./scripts/publish.sh --dry-run`
 	Note: the script will publish the crates in the correct order and pause after each crate to ensure it's available at the crates registry before proceeding. This means the dry run isn't as useful and will end up in an infinite loop. If you're really unsure about the changes and want to do a dry run you should do a `cargo publish --dry-run` for each individual crate.
 1. Publish: `./scripts/publish.sh`
 1. Once published, make sure to "create a release" for the pushed tag on github. 
--- a/RELEASE-CHECKLIST.md
+++ b/RELEASE-CHECKLIST.md
@@ -7,4 +7,5 @@
 	1. Ensure you're in the project root dir
 	1. Run: `./scripts/publish.sh --dry-run`
 	Note: the script will publish the crates in the correct order and pause after each crate to ensure it's available at the crates registry before proceeding. This means the dry run isn't as useful and will end up in an infinite loop. If you're really unsure about the changes and want to do a dry run you should do a `cargo publish --dry-run` for each individual crate.
+1. Publish: `./scripts/publish.sh`
 1. Once published, make sure to "create a release" for the pushed tag on github. 
--- a/RELEASE-CHECKLIST.md
+++ b/RELEASE-CHECKLIST.md
@@ -5,5 +5,6 @@
 1. In the `CHANGELOG.md` file, move everything under "Unreleased" to a new section named `## [vx.y.z] – YYYY-MM-DD`
 1. Make a dryrun like so:
 	1. Ensure you're in the project root dir
-	1. Run: `./scripts/publish.sh --dry-run --allow-dirty` (If you are running this against the master branch, you may omit the `--allow-dirty` flag)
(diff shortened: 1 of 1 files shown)


RELEASE-CHECKLIST.md
```diff
@@ -5,4 +5,5 @@
 1. In the `CHANGELOG.md` file, move everything under "Unreleased" to a new section named `## [vx.y.z] – YYYY-MM-DD`
 1. Make a dryrun like so:
 	1. Ensure you're in the project root dir
-	1. Run: `./scripts/publish.sh --dry-run --allow-dirty`
+	1. Run: `./scripts/publish.sh --dry-run --allow-dirty` (If you are running this against the master branch, you may omit the `--allow-dirty` flag)
```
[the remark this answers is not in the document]
dvdplm: @maciejhirsz is right, and that's why I added it, to remove a potential gotcha (i.e. if the release person is working on a release-prep branch and wants to do a dry run to check things are dandy). I suspect it's more confusing than helpful though, so I think we should remove the `--allow-dirty` flag.

dvdplm, line 8:
```suggestion
	1. Run: `./scripts/publish.sh --dry-run`
	Note: the script will publish the crates in the correct order and pause after each crate to ensure it's available at the crates registry before proceeding. This means the dry run isn't as useful and will end up in an infinite loop. If you're really unsure about the changes and want to do a dry run you should do a `cargo publish --dry-run` for each individual crate.
```

RELEASE-CHECKLIST.md
```diff
@@ -5,4 +5,5 @@
...
+1. Once published, make sure to "create a release" for the pushed tag on github.
```
dvdplm, line 9:
```suggestion
1. Publish: `./scripts/publish.sh`
1. Once published, make sure to "create a release" for the pushed tag on github. 
```

dvdplm (changes requested):

dvdplm:
@TarikGul is this still on your radar? :)

TarikGul:
> @TarikGul is this still on your radar? :)

Completely forgot I had this PR up 😅  Ill get this wrapped up today :)

RELEASE-CHECKLIST.md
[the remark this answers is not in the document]
TarikGul: This should be all resolved now

RELEASE-CHECKLIST.md
```diff
@@ -5,4 +5,7 @@
 1. In the `CHANGELOG.md` file, move everything under "Unreleased" to a new section named `## [vx.y.z] – YYYY-MM-DD`
 1. Make a dryrun like so:
 	1. Ensure you're in the project root dir
-	1. Run: `./scripts/publish.sh --dry-run --allow-dirty`
+	1. Run: `./scripts/publish.sh --dry-run`
```
niklasad1, line 8: This not point is not useful anymore after your findings but the `Note` below is.

Out-of-scope for this PR: I think we should provide another script to perform dry-run for all the crates or update the existing one to work in dry-run mode too.

dvdplm: yeah, hmm, I just don't see how we can make it work, not without some crazy shenanigans. Do you?

TarikGul:
@niklasad1 Good to merge this?

RELEASE-CHECKLIST.md
niklasad1: we can just provide another script that performs alternatively have `CLI option` --dry-run that doesn't perform the complicated stuff such as reading remote version and so on.

```bash
for crate in ${ORDER}; do {
   cd $crate
   cargo publish --dry-run
   cd -
}
```

niklasad1, line 8:
```suggestion
```

niklasad1: //cc @TarikGul :point_up: see my previous comment.

I think `1. Run: /scripts/publish.sh --dry-run` does more harm than good because it doesn't work but the added NOTE is excellent

niklasad1 (dismissed):
I think 1. Run: /scripts/publish.sh --dry-run does more harm than good because it doesn't work but the added NOTE is excellent.

RELEASE-CHECKLIST.md
```diff
@@ -5,4 +5,6 @@
 1. In the `CHANGELOG.md` file, move everything under "Unreleased" to a new section named `## [vx.y.z] – YYYY-MM-DD`
 1. Make a dryrun like so:
 	1. Ensure you're in the project root dir
```
dvdplm, line 7:
```suggestion
```

dvdplm (approved):

Issue #487, opened 2021-09-24
Support for customized logging


crystalin:
Ideally, the library should allow the user to control how the logs are produced.
Here is the requirements we believe are needed for our parachain node:
- Support to log the request, including its payload*
- Support to log the response, including its payload* and the timing
- *Payload could be filtered (to include some fields, or based on the method used). The possibility to reduce the the content of the fields in the payout automatically (like an array of 1000 items displaying only the first and last item).

librelois:
One possibility proposed by @dvdplm would be to integrate `sc_tracing` and allow users to inject their own `tracing-subscriber`.

niklasad1:
This is partly fixed by #722

It adds a tracing span per `method name` with a limit just to truncate logs bigger than `n` characters.
What's is nice is that you can subscribe to just a method name that you are interested in  instead of doing it globally with middleware (unless you want to recompile and have some filter).


But we'll add more details in the middleware too soon so we can pick either because `tracing` is a bit less flexible for users as the logging output is simply truncated.

niklasad1:
Closed by #793 

jsonrpsee has both "tracing spans" and middleware for you customize your logging.

PR #525, opened 2021-10-15
deps: replace log with tracing

niklasad1:
Initial work on #487 

A follow-up would be to add `spans` on the individual methods to get `metrics` such as payload size, time it took to execute and so on.

I'm not super familiar with the different tracing subscriber libraries but https://crates.io/crates/tracing-timing looks interesting but that's up the user of the server to configure but we should test a couple of them to ensure that sufficient data is emitted.


maciejhirsz (approved):
LGTM

dvdplm (approved):

PR #526, opened 2021-10-15
Re-introduce object param parsing

### LINKED ISSUE #445: [proc macros]: support for named parameters (JSON Map/Object)
Currently in proc macros one implementation for named parameter(s) exist [here](https://github.com/paritytech/jsonrpsee/blob/master/proc-macros/src/render_server.rs#L301-#L336)

The reason why it's not used is because it won't work to have a serde implementation inside a trait impl with generic type parameters, thus one would a get compile error such as `can't use generic parameters from outer function`.

The reason for this is how the `RpcServer trait` is generated.

```rust
pub trait RpcServer<T> {
   // defined methods without impls
   async fn foo() -> Result<(), Error>;


  ///Collects all the methods and subscriptions defined in the trait and adds them into a single `RpcModule`.
  fn into_rpc(self) -> jsonrpsee::RpcModule<Self> {
       // registering methods and decoding params goes here (which is problematic for serde impls)
  }
}
```

maciejhirsz:
Fixes #445.

The trick is that instead of putting fn argument types (which may be generic) on the struct like it used to, e.g.:

```rust
#[derive(jsonrpsee :: types :: __reexports :: serde :: Deserialize)]
#[serde(crate = "jsonrpsee :: types :: __reexports :: serde")]
struct ParamsObject {
    gun: Gun,
    map: HashMap<u8, u8>,
}
let parsed: ParamsObject = params.parse()?;
```

We make a struct that's entirely generic, so that all types are always used in scope where they belong:

```rust
#[derive(jsonrpsee :: types :: __reexports :: serde :: Deserialize)]
#[serde(crate = "jsonrpsee :: types :: __reexports :: serde")]
struct ParamsObject<G0, G1> {
    gun: G0,
    map: G1,
}
let parsed: ParamsObject<Gun, HashMap<u8, u8>> = params.parse()?;
```

As a bonus this removes the need to inject lifetimes into references (no need to transform `&str` into `&'a str` etc.), so that whole thing can be removed.


dvdplm:
> As a bonus this removes the need to inject lifetimes into references (no need to transform `&str` into `&'a str` etc.), so that whole thing can be removed.

That is … magic!

dvdplm (approved):

Issue #531, opened 2021-10-16
Support for other headers besides Origin


except:
I would like to use this crate however a certain API requires a user to send an `Authorization` header before upgrading, will there be support for extra headers?

PR #535, opened 2021-10-19
feat(ws client): support custom headers.

niklasad1:
Closing #531


ws-client/src/client.rs
```diff
@@ -155,9 +175,9 @@ impl<'a> WsClientBuilder<'a> {
 		self
 	}
 
-	/// Set origin header to pass during the handshake.
-	pub fn origin_header(mut self, origin: Cow<'a, str>) -> Self {
-		self.origin_header = Some(origin);
+	/// Set custom headers to pass during the handshake.
```
jsdw, line 178: In Soketto there aren't any checks to see whether the headers that are set conflict with the oens it sets internally; I wonder whether it's worth noting that it's the users responsibility here to make sure that headers don't conflict

ws-client/src/client.rs
```diff
@@ -102,14 +102,34 @@ pub struct WsClient {
...
+///
+/// no_run```
+///
+/// use jsonrpsee_ws_client::{WsClientBuilder, Header};
+///
+/// #[tokio::main]
+/// async fn main() {
+///     // build client
+///     let client = WsClientBuilder::default()
+///          .custom_headers(&[Header { name: "Origin", value: "localhost".as_bytes() }])
```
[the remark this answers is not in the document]
niklasad1: I think that's fine in this case, you mean like having a `Vec<(&str, &str)>` internally? 

but you like `custom_header` over `custom_headers` because it's more readable?

ws-client/src/client.rs
```diff
@@ -102,14 +102,34 @@ pub struct WsClient {
 	id_guard: RequestIdGuard,
 }
 
-/// Configuration.
+/// Builder for [`WsClient`].
+///
+/// # Examples
+///
+/// no_run```
```
dvdplm, line 109: Shouldn't this be `'''no_run` (but with actual backticks)? It would be better if it was actually compiled though.

ws-client/src/client.rs
```diff
@@ -102,14 +102,34 @@ pub struct WsClient {
...
+///
+/// use jsonrpsee_ws_client::WsClientBuilder;
+///
+/// #[tokio::main]
+/// async fn main() {
+///     // build client
+///     let client = WsClientBuilder::default()
+///          .custom_header("Origin", "localhost")
```
dvdplm, line 117:
```suggestion
///          .custom_header("Any-Header-You-Like", "42")
```

ws-client/src/client.rs
```diff
@@ -155,9 +175,12 @@ impl<'a> WsClientBuilder<'a> {
 		self
 	}
 
-	/// Set origin header to pass during the handshake.
-	pub fn origin_header(mut self, origin: Cow<'a, str>) -> Self {
-		self.origin_header = Some(origin);
+	/// Set a custom header to pass during the handshake.
```
dvdplm, line 178:
```suggestion
	/// Set a custom header passed to the server during the handshake.
```

ws-client/src/client.rs
```diff
@@ -155,9 +175,12 @@ impl<'a> WsClientBuilder<'a> {
...
+	///
+	/// You have to ensure that headers doesn't conflict if this is called more than once.
+	/// Because it is not checked that the headers doesn't conflict.
```
dvdplm, line 181:
```suggestion
	/// The caller is responsible for checking that the headers do not conflict or are duplicated.
```

ws-client/src/client.rs
```diff
@@ -155,9 +175,12 @@ impl<'a> WsClientBuilder<'a> {
...
+	pub fn custom_header(mut self, header: &'a str, value: &'a str) -> Self {
```
dvdplm, line 182: I think `add_header` would be a better name. The HTTP spec doesn't define a set of headers afaik, so in a way all headers are "custom".

What do you think about changing the signature to take a `Header` directly? It'd be annoying for users that already have a `Header` to extract the name and value.

ws-client/src/transport.rs
```diff
@@ -71,9 +70,9 @@ pub struct WsTransportClientBuilder<'a> {
 	pub target: Target,
 	/// Timeout for the connection.
 	pub timeout: Duration,
-	/// `Origin` header to pass during the HTTP handshake. If `None`, no
-	/// `Origin` header is passed.
-	pub origin_header: Option<Cow<'a, str>>,
+	/// Custom headers to pass during the HTTP handshake. If `None`, no
+	/// custom header is passed.
+	pub custom_headers: Vec<Header<'a>>,
```
dvdplm, line 75:
```suggestion
	/// Headers passed to the server during the HTTP handshake.
	pub headers: Vec<Header<'a>>,
```

Could it be a slice, `&'a [Header<'a>]`, instead of a `Vec` you think?

ws-client/src/client.rs
dvdplm: Oh I missed this comment. Can we document the headers set internally here so users know what to not conflict with?

ws-client/src/transport.rs
niklasad1: it could be but then we need to have `add_headers(&[&str, &str])` instead that it slightly more error prone I think but would work...

ws-client/src/client.rs
niklasad1: > What do you think about changing the signature to take a Header directly? It'd be annoying for users that already have a Header to extract the name and value.

Haha, I thought the other way around that it's a pain of using `Header` directly and we would need to export it.... then I don't think it's that likely that users of the client has the `httparse crate` in their dependency tree.... but I could be wrong :)

jsdw: I was sugegsting the name + value function; yeah, I guess I also expect that whoever using this won't have a Header struct to hand; they will just know what header name+value pairs they want and this would make that easy!

ws-client/src/transport.rs
niklasad1: when we can avoid allocations without sacrificing "ergonomics" then it's a go but we should prioritize "ergonomics" especially in "builders" that only performed once.

but we have no dedicated rules rather than that @maciejhirsz likes to remove needless allocations :)

PR #541, opened 2021-10-25
WIP: tracing futures for the servers

niklasad1:
Continuation of #487

Example output:

```
Oct 26 15:00:28.093 DEBUG method_call{req.method=say_hello}: jsonrpsee_ws_server::server: rx_method_call_len=45
Oct 26 15:00:28.093 TRACE method_call{req.method=say_hello}: jsonrpsee_ws_server::server: rx_method_call=Request { jsonrpc: TwoPointZero, id: Number(0), method: "say_hello", params: None }
Oct 26 15:00:29.094 DEBUG method_call{req.method=say_hello}: jsonrpsee_utils::server::helpers: tx_method_call_len=38
Oct 26 15:00:29.094 DEBUG method_call{req.method=say_hello}: jsonrpsee_utils::server::helpers: tx_method_call="{\"jsonrpc\":\"2.0\",\"result\":\"lo\",\"id\":0}"
Oct 26 15:00:29.094 DEBUG method_call{req.method=say_hello}: jsonrpsee_utils::server::rpc_module: execution_time=1.000502295s
```

Should solve:
> Support to log the request, including its payload*
> Support to log the response, including its payload* and the timing

I have used `debug` to only print payload sizes and `trace` to print the entire payload.
Perhaps we should make it possible to truncate the payloads when it exceeds some limit....

### DIFF
--- a/http-server/src/server.rs
+++ b/http-server/src/server.rs
@@ -47,10 +47,7 @@ use jsonrpsee_utils::server::{
 
 use serde_json::value::RawValue;
 use socket2::{Domain, Socket, Type};
-use std::{
-	cmp,
-	net::{SocketAddr, TcpListener},
-};
+use std::net::{SocketAddr, TcpListener};
 use tracing_futures::Instrument;
 
 /// Builder to create JSON-RPC HTTP server.
@@ -238,8 +235,8 @@ impl Server {
 							if let Ok(req) = serde_json::from_slice::<Request>(&body) {
 								let span = tracing::span!(tracing::Level::DEBUG, "method_call", %req.method);
 								let _enter = span.enter();
-								tracing::debug!("recv {} bytes", body.len());
-								tracing::trace!("recv: {:?}", req);
+								tracing::debug!(rx_method_call_len = body.len());
+								tracing::trace!(rx_method_call = ?req);
 								// NOTE: we don't need to track connection id on HTTP, so using hardcoded 0 here.
 								if let Some(fut) = methods.execute_with_resources(&tx, req, 0, &resources) {
 									fut.await.in_current_span();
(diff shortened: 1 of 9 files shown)


http-server/src/server.rs
```diff
@@ -281,7 +287,8 @@ impl Server {
 						} else {
 							collect_batch_response(rx).await
 						};
-						tracing::debug!("[service_fn] sending back: {:?}", &response[..cmp::min(response.len(), 1024)]);
+						tracing::debug!(tx_batch_call_len = response.len());
```
dvdplm, line 290: We can still have a log message in here right? `tracing::debug!("[service_fn] sending back response.", tx_batch_call_len = response.len());` or something?

http-server/src/server.rs
```diff
@@ -281,7 +287,8 @@ impl Server {
...
+						tracing::trace!(tx_batch = ?response);
```
dvdplm, line 291: Same question here.

utils/src/server/rpc_module.rs
```diff
@@ -169,12 +163,6 @@ impl MethodCallback {
 				let tx = tx.clone();
 				let params = params.into_owned();
 				let id = id.into_owned();
-				tracing::trace!(
-					"[MethodCallback::execute] Executing async callback, params={:?}, req.id={:?}, conn_id={:?}",
-					params,
-					id,
-					conn_id
-				);
```
dvdplm, line 177: How come you want these gone?

niklasad1: those were a bit noisy because we both print `execute_with_resources` and `execute` but I can revert those

dvdplm: Your call, I don't mind either way.

niklasad1:
Closing this as we have introduced middleware for users to this kind of metrics.

PR #542, opened 2021-10-27
fix(ws server): `batch` wait until all methods has been executed.

niklasad1:
This fixes a bug in the ws-server when any `async methods` are included in batch, the futures were not awaited on that "may" cause method calls to be ignored. In worst case just send back `]`

### DIFF
--- a/tests/tests/integration_tests.rs
+++ b/tests/tests/integration_tests.rs
@@ -381,5 +381,5 @@ async fn ws_batch_works() {
 	batch.push(("say_hello", rpc_params![]));
 	batch.push(("slow_hello", rpc_params![]));
 
-	let response: Vec<String> = client.batch_request(batch).await.unwrap();
+	assert!(client.batch_request::<String>(batch).await.is_ok())
 }
--- a/tests/tests/integration_tests.rs
+++ b/tests/tests/integration_tests.rs
@@ -360,3 +360,26 @@ async fn ws_server_should_stop_subscription_after_client_drop() {
 	// assert that the server received `SubscriptionClosed` after the client was dropped.
 	assert!(matches!(rx.next().await.unwrap(), SubscriptionClosedError { .. }));
 }
+
+#[tokio::test]
+async fn ws_batch_works() {
+	let subscriber = tracing_subscriber::FmtSubscriber::builder()
+		// all spans/events with a level higher than TRACE (e.g, debug, info, warn, etc.)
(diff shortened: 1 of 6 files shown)


tests/tests/integration_tests.rs
```diff
@@ -360,3 +360,26 @@ async fn ws_server_should_stop_subscription_after_client_drop() {
...
+	let server_addr = websocket_server().await;
+	let server_url = format!("ws://{}", server_addr);
+	let client = Arc::new(WsClientBuilder::default().build(&server_url).await.unwrap());
+
+	let mut batch = Vec::new();
+
+	batch.push(("say_hello", rpc_params![]));
+	batch.push(("slow_hello", rpc_params![]));
+
+	assert!(client.batch_request::<String>(batch).await.is_ok())
```
jsdw, line 384: Just to be really safe, is it worth checking that the results are as expetced too (two "hello" values)?

ws-server/src/server.rs
```diff
@@ -296,34 +296,46 @@ async fn background_task(
...
+				let fut = async move {
+					// Batch responses must be sent back as a single message so we read the results from each
+					// request in the batch and read the results off of a new channel, `rx_batch`, and then send the
+					// complete batch response back to the client over `tx`.
+					let (tx_batch, mut rx_batch) = mpsc::unbounded();
+					if let Ok(batch) = serde_json::from_slice::<Vec<Request>>(&d) {
+						if !batch.is_empty() {
+							let futs = batch
+								.into_iter()
+								.filter_map(|req| m.execute_with_resources(&tx_batch, req, conn_id, r));
```
jsdw: Does this mean that responses could come back in a different order to the request order in the batch?

eg if we have a batch request to first `slow_hello` and then `say_hello`, and then collect those up via a channel, we'll get the `say_hello` response back first.

Do we care that the response order might not match the request order (if I'm even right about this!)?

niklasad1: good question

>The Response objects being returned from a batch call MAY be returned in any order within the Array. The Client SHOULD match contexts between the set of Request objects and the resulting set of Response objects based on the id member within each Object.

I should have put a comment in the code about this but it's kind of an optimization :)

jsdw: Ah awesome, gotcha!

jsdw (approved):
LGTM! Should the test confirm that 2 responses come back?

PR #543, opened 2021-10-27
clients: request ID as RAII guard

niklasad1:
The rationale behind this change is to reduce the technical debt, not to require us to remember to reclaim a request ID when a request was completed successful or not.

### DIFF
--- a/types/src/client.rs
+++ b/types/src/client.rs
@@ -30,7 +30,8 @@ use futures_channel::{mpsc, oneshot};
 use futures_util::{future::FutureExt, sink::SinkExt, stream::StreamExt};
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
 use serde_json::Value as JsonValue;
-use std::sync::atomic::{AtomicU64, AtomicUsize, Ordering};
+use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::Arc;
 
 /// Subscription kind
 #[derive(Debug)]
@@ -190,7 +191,7 @@ impl<Notif> Drop for Subscription<Notif> {
 /// Keep track of request IDs.
 pub struct RequestIdGuard {
 	// Current pending requests.
-	current_pending: AtomicUsize,
+	current_pending: Arc<()>,
 	/// Max concurrent pending requests allowed.
 	max_concurrent_requests: usize,
 	/// Get the next request ID.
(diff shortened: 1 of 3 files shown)


types/src/client.rs
```diff
@@ -200,52 +201,70 @@ pub struct RequestIdGuard {
...
-				if val >= self.max_concurrent_requests {
-					None
-				} else {
-					Some(val + 1)
-				}
-			})
-			.map(|_| ())
-			.map_err(|_| Error::MaxSlotsExceeded)
+	fn get_slot(&self) -> Result<Arc<()>, Error> {
+		// Strong count is 1 at start, so that's why we use `>` and not `>=`.
```
jsdw, line 208: Useful comment, thank you!

types/src/client.rs
```diff
@@ -200,52 +201,70 @@ pub struct RequestIdGuard {
...
+			ids.push(self.current_id.fetch_add(1, Ordering::SeqCst));
 		}
-		Ok(batch)
+		Ok(RequestId { _rc: rc, id: ids })
+	}
+}
+
+/// Reference counted request ID.
+#[derive(Debug)]
+pub struct RequestId<T> {
```
jsdw, line 240: Naming nit: perhaps `RequestIdGuard` would be clearer, in line with things like `MutexGuard` and so on?

types/src/client.rs
```diff
@@ -200,52 +201,70 @@ pub struct RequestIdGuard {
...
+	id: T,
+	/// Reference count decreased when dropped.
+	_rc: Arc<()>,
+}
+
+impl<T> RequestId<T> {
```
jsdw, line 246: I guess this could be a `Deref` impl too, but given how it's used I wouldn't mind either way!

types/src/client.rs
jsdw: Ooohh, the parent thing is called `RequestidGuard`! I see now :)

jsdw (approved):
LGTM!

types/src/client.rs
niklasad1: yeah, or we rename `RequestIdGuard -> RequestIdManager` and call this `RequestId -> RequestIdGuard`?!

jsdw: I like that idea!

Issue #538, opened 2021-10-23
Support `Map` parameters in proc macro 


DefinitelyNotHilbert:
Hi there!

I'm working with a third-party jsonrpc API which expects subscription parameters to be passed as maps, not arrays. Now, if I understand correctly, it is possible to pass parameters as maps when one uses the `subscribe` function. However, the proc macro always generates parameter arrays and as far as I can tell, there is no way of changing this in the current version of jsonrpsee.

Would there be any interest in a contribution that makes the parameter format configurable, for instance by adding another parameter to the `subscribe` attribute?

DefinitelyNotHilbert:
I've submitted the MR, couldn't find the time earlier this week.

PR #544, opened 2021-10-29
Proc mac support map param

DefinitelyNotHilbert:
This is my MR for #538.

- Added `param_format` parameter to the `method` and `subscription` attributes. It takes a string parameter, `"map"`to make the macro generate a function that sends the parameters as a map, everything else will fall back to the default `"array"`. I'm not too happy with the fact that it is a string, so if anybody has an alternative, be my friend :) 

- Updated `render_method` and `render_sub` to generate the necessary code. There's an additional function that extracts the parameter names from the function signature

I wasn't sure how / if you test your macros, so I didn't do anything related to unit test, but I can confirm that the code is generated correctly and works, judging from how it interacts with json_rpc APIs.

Looking forward to your comments :)

### DIFF
--- a/proc-macros/src/attributes.rs
+++ b/proc-macros/src/attributes.rs
@@ -43,7 +43,7 @@ pub(crate) struct Argument {
 #[derive(Debug, Clone)]
 pub enum ParamFormat {
 	Array,
-	Map
+	Map,
 }
 
 #[derive(Debug, Clone)]
@@ -198,8 +198,8 @@ where
 
 pub(crate) fn parse_param_format(arg: Result<Argument, MissingArgument>) -> ParamFormat {
 	match optional(arg, Argument::string).unwrap().as_deref() {
-			None | Some("array") => ParamFormat::Array,
-			Some("map") => ParamFormat::Map,
-			err => panic!("param_format must be either `map` or `array`, got {:?}", err)
+		None | Some("array") => ParamFormat::Array,
+		Some("map") => ParamFormat::Map,
+		err => panic!("param_format must be either `map` or `array`, got {:?}", err),
 	}
 }
--- a/proc-macros/src/attributes.rs
+++ b/proc-macros/src/attributes.rs
@@ -41,6 +41,12 @@ pub(crate) struct Argument {
 }
 
 #[derive(Debug, Clone)]
+pub enum ParamFormat {
+	Array,
+	Map
+}
+
+#[derive(Debug, Clone)]
 pub struct Resource {
 	pub name: syn::LitStr,
 	pub assign: Token![=],
@@ -188,4 +194,12 @@ where
 	F: Fn(Argument) -> syn::Result<T>,
 {
 	arg.ok().map(transform).transpose()
+}
+
+pub(crate) fn parse_param_format(arg: Result<Argument, MissingArgument>) -> ParamFormat {
+	match optional(arg, Argument::string).unwrap().as_deref() {
+			None | Some("array") => ParamFormat::Array,
+			Some("map") => ParamFormat::Map,
+			err => panic!("param_format must be either `map` or `array`, got {:?}", err)
+	}
 }
--- a/proc-macros/src/attributes.rs
+++ b/proc-macros/src/attributes.rs
@@ -196,10 +196,13 @@ where
 	arg.ok().map(transform).transpose()
 }
 
-pub(crate) fn parse_param_kind(arg: Result<Argument, MissingArgument>) -> ParamKind {
-	match optional(arg, Argument::string).unwrap().as_deref() {
-		None | Some("array") => ParamKind::Array,
-		Some("map") => ParamKind::Map,
-		err => panic!("param_kind must be either `map` or `array`, got {:?}", err),
+pub(crate) fn parse_param_kind(arg: Result<Argument, MissingArgument>) -> syn::Result<ParamKind> {
+	let kind: Option<syn::Ident> = optional(arg, Argument::value)?;
+
+	match kind {
+		None => Ok(ParamKind::Array),
+		Some(ident) if ident == "array" => Ok(ParamKind::Array),
+		Some(ident) if ident == "map" => Ok(ParamKind::Map),
+		ident => Err(Error::new(ident.span(), "param_kind must be either `map` or `array`")),
 	}
 }
--- a/proc-macros/src/attributes.rs
+++ b/proc-macros/src/attributes.rs
@@ -41,7 +41,7 @@ pub(crate) struct Argument {
 }
 
 #[derive(Debug, Clone)]
-pub enum ParamFormat {
+pub enum ParamKind {
 	Array,
 	Map,
 }
@@ -196,10 +196,10 @@ where
 	arg.ok().map(transform).transpose()
 }
 
-pub(crate) fn parse_param_format(arg: Result<Argument, MissingArgument>) -> ParamFormat {
+pub(crate) fn parse_param_kind(arg: Result<Argument, MissingArgument>) -> ParamKind {
 	match optional(arg, Argument::string).unwrap().as_deref() {
-		None | Some("array") => ParamFormat::Array,
-		Some("map") => ParamFormat::Map,
-		err => panic!("param_format must be either `map` or `array`, got {:?}", err),
+		None | Some("array") => ParamKind::Array,
+		Some("map") => ParamKind::Map,
+		err => panic!("param_kind must be either `map` or `array`, got {:?}", err),
 	}
 }
--- a/proc-macros/src/render_client.rs
+++ b/proc-macros/src/render_client.rs
@@ -28,7 +28,7 @@ use crate::helpers::generate_where_clause;
 use crate::rpc_macro::{RpcDescription, RpcMethod, RpcSubscription};
 use proc_macro2::TokenStream as TokenStream2;
 use quote::quote;
-use syn::TypeParam;
+use syn::{FnArg, Pat, PatIdent, PatType, TypeParam};
 
 impl RpcDescription {
 	pub(super) fn render_client(&self) -> Result<TokenStream2, syn::Error> {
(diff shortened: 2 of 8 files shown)


niklasad1 (commented):
great, 

please add a test in `jsonrpsee/proc-macros/tests/ui/correct/*` for this, to test that it actually works with the servers too.

you can look in `jsonrpsee/proc-macros/tests/ui/correct/basic` how to do it so up to you whether to add a new file for this or in existing one...

proc-macros/src/rpc_macro.rs
```diff
@@ -42,6 +42,7 @@ pub struct RpcMethod {
 	pub blocking: bool,
 	pub docs: TokenStream2,
 	pub params: Vec<(syn::PatIdent, syn::Type)>,
+	pub param_format: String,
```
niklasad1, line 45: I think I would prefer an `Option`,  to not regard invalid param format as `array`

such as
```
let params = match parms {
  None => Array,
  Some("array") => Array,
  Some("map") => Map
  err => compile_error!("Invalid param {:?} expected: `array` or `map`", err),
};
```

proc-macros/src/rpc_macro.rs
```diff
@@ -50,12 +51,16 @@ pub struct RpcMethod {
...
 	pub fn from_item(attr: Attribute, mut method: syn::TraitItemMethod) -> syn::Result<Self> {
-		let [aliases, blocking, name, resources] =
-			AttributeMeta::parse(attr)?.retain(["aliases", "blocking", "name", "resources"])?;
+		let [aliases, blocking, name, param_format, resources] =
+			AttributeMeta::parse(attr)?.retain(["aliases", "blocking", "name", "param_format", "resources"])?;
 
 		let aliases = parse_aliases(aliases)?;
 		let blocking = optional(blocking, Argument::flag)?.is_some();
 		let name = name?.string()?;
+		let param_format = match param_format {
```
niklasad1, line 60: see my comment above :point_up:

proc-macros/src/rpc_macro.rs
```diff
@@ -57,10 +57,7 @@ impl RpcMethod {
 		let aliases = parse_aliases(aliases)?;
 		let blocking = optional(blocking, Argument::flag)?.is_some();
 		let name = name?.string()?;
-		let param_format = match param_format {
-			Ok(param_format) => param_format.string()?,
-			Err(_) => String::from("array"),
-		};
+		let param_format = optional(param_format, Argument::string)?.or(Some("array".into()));
```
niklasad1, line 60: sorry if my suggestion was bad, but would be better to check `array` or `map` here then to introduce another type `Params` instead of `String`.

Such as:

```rust

enum Params {
   Array,
   Map,
}

let param_format = match optional(param_format, Argument::string)? {
  None | Some("array") => Params::Array,
  Some("map") => Params::Map,
  err => panic!("param_format must be either `map` or `array`, got {:?}", err),
};
```

niklasad1: this could be perhaps be a separate function as `fn parse_param_format` or something?!

to avoid duplicating the code...

DefinitelyNotHilbert: Ah, now I get you. I wasn't sure whether you would embrace an additional type, but I agree, it makes things much cleaner.

DefinitelyNotHilbert: > this could be perhaps be a separate function as fn parse_param_format or something?!

Agreed.

DefinitelyNotHilbert:
> great,
> 
> please add a test in `jsonrpsee/proc-macros/tests/ui/correct/*` for this, to test that it actually works with the servers too.
> 
> you can look in `jsonrpsee/proc-macros/tests/ui/correct/basic` how to do it so up to you whether to add a new file for this or in existing one...

I've added a test case in the spirit of `basic.rs` but I'm not 100% that it's exactly what you want.

niklasad1 (approved):
nice work, thank you

LGTM

niklasad1:
@DefinitelyNotHilbert just run `cargo fmt --all` and push it then the CI should be happy

proc-macros/src/render_client.rs
```diff
@@ -100,8 +100,31 @@ impl RpcDescription {
...
+				ParamFormat::Map => {
+					// Extract parameter names.
+					let param_names = extract_param_names(&method.signature.sig);
+					// Combine parameter names and values into tuples.
+					let params = param_names.iter().zip(params).map(|pair| {
+						let key = pair.0;
+						let value = pair.1;
+						quote! { (#key, #value) }
+					});
+					let jsonrpsee = self.jsonrpsee_client_path.as_ref().unwrap();
```
dvdplm, line 113: Could this use `jprs_client_item()` and do something like `let json_value = self.jprs_client_item(quote! { types::JsonValue });`? or even re-use the `serde_json` lvar from line 99?

proc-macros/src/attributes.rs
```diff
@@ -40,6 +40,12 @@ pub(crate) struct Argument {
 	pub tokens: TokenStream2,
 }
 
+#[derive(Debug, Clone)]
+pub enum ParamFormat {
```
dvdplm, line 44: This is a total nitpick, but what do you think about naming this `ParameterKind`? It's nor really a "format" but more akin to a type, or mode of operation of sorts.
Also: docs missing here.

proc-macros/src/render_client.rs
```diff
@@ -143,8 +166,31 @@ impl RpcDescription {
...
-			quote! {
-				Some(vec![ #(#params),* ].into())
+			match sub.param_format {
+				ParamFormat::Map => {
+					// Extract parameter names.
+					let param_names = extract_param_names(&sub.signature.sig);
+					// Combine parameter names and values into tuples.
+					let params = param_names.iter().zip(params).map(|pair| {
+						let key = pair.0;
+						let value = pair.1;
```
dvdplm, line 176:
```suggestion
						let param = pair.0;
						let value = pair.1;
```

dvdplm (commented):
Looks great, tyvm.

Some docs [here](https://github.com/paritytech/jsonrpsee/blob/master/proc-macros/src/lib.rs#L162) about the new attr would be good.

proc-macros/tests/ui/correct/param_format.rs
```diff
@@ -0,0 +1,63 @@
...
+
+#[tokio::main]
+async fn main() {
+	let server_addr = websocket_server().await;
+	let server_url = format!("ws://{}", server_addr);
+	let client = WsClientBuilder::default().build(&server_url).await.unwrap();
+
+	assert_eq!(client.method_with_array_param(0, "a".into()).await.unwrap(), 42);
+	assert_eq!(client.method_with_map_param(0, "a".into()).await.unwrap(), 42);
+	assert_eq!(client.method_with_array_param(0, "a".into()).await.unwrap(), 42);
```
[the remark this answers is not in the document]
DefinitelyNotHilbert: Yes, good catch.

proc-macros/src/attributes.rs
DefinitelyNotHilbert: I called it format because I was thinking of it in terms of parameter layout and not necessarily as types, but I've absolutely no hard feelings about this. I've renamed it to `ParamKind`

dvdplm (approved):

proc-macros/src/attributes.rs
```diff
@@ -189,3 +195,11 @@ where
...
 	arg.ok().map(transform).transpose()
 }
+
+pub(crate) fn parse_param_kind(arg: Result<Argument, MissingArgument>) -> ParamKind {
+	match optional(arg, Argument::string).unwrap().as_deref() {
+		None | Some("array") => ParamKind::Array,
+		Some("map") => ParamKind::Map,
+		err => panic!("param_kind must be either `map` or `array`, got {:?}", err),
+	}
+}
```
maciejhirsz, line 205: We should return an error here, unwrap is going to throw away span information on invalid parse of the argument value. Instead of a string we could parse to raw `Argument::value` using `syn::Ident` so you can skip quotes, plus then you have the span to use to emit an error instead of a panic.

No guarantees that this compiles:

```suggestion
pub(crate) fn parse_param_kind(arg: Result<Argument, MissingArgument>) -> syn::Result<ParamKind> {
	let kind: Option<syn::Ident> = optional(arg, Argument::value)?;

	match kind {
		None => Ok(ParamKind::Array),
		Some(ident) if ident == "array" => Ok(ParamKind::Array),
		Some(ident) if ident == "map" => Ok(ParamKind::Map),
		Some(ident) => Err(Error::new(ident.span(), "param_kind must be either `map` or `array`")),
	}
}
```

proc-macros/src/render_client.rs
```diff
@@ -100,8 +100,30 @@ impl RpcDescription {
...
+									)
+								)
+							)
+					}
+				}
+				ParamKind::Array => {
+					quote! {
+						Some(vec![ #(#params),* ].into())
+					}
+				}
```
maciejhirsz, line 126: Would be nice if this was DRY with lines 169..191.

Also would be nice if we allowed a flat `Box<RawValue>` variant here, but that's way beyond the scope of this PR :)

proc-macros/tests/ui/correct/param_kind.rs
```diff
@@ -0,0 +1,63 @@
...
+	types::{async_trait, RpcResult},
+	ws_client::*,
+	ws_server::WsServerBuilder,
+};
+
+use std::net::SocketAddr;
+
+#[rpc(client, server, namespace = "foo")]
+pub trait Rpc {
+	#[method(name="method_with_array_param", param_kind="array")]
```
maciejhirsz, line 12: Formatting checker won't pick these up since it's macro attributes, but we should be consistent with spacing, plus can drop quotes from `param_kind` now:

```suggestion
	#[method(name = "method_with_array_param", param_kind = array)]
```

maciejhirsz (changes requested):
I'm going to be a bit more pedantic here, sorry :D. Added suggestions on how to turn the `"array"|"map"` into `array|map` and one other nit.

In general I think that since this is a binary option, and that it's always going to be a binary option, we should probably use a flag argument. So instead of:

```rust
#[method(name = "foo", param_kind = map)]
```

You'd just write:

```rust
#[method(name = "foo", serialize_as_map)]
```

Or some such. I reckon that's easier to grok since you don't have to look up possible values, and code should also be easier since there are fewer error conditions.

proc-macros/src/attributes.rs
DefinitelyNotHilbert: I figure that this won't be necessary if we go with your suggestion of turning `parma_kind` into a flag. Personally, I prefer the more verbose version (that's why I've implemented it in this way), but I complete get your point. Since this is obviously not my library, I'm happy to turn it into a flag if you want, I just care about the feature as a whole.

maciejhirsz: Yeah, this is only if we want to keep the verbose variant.

I don't have a super strong opinion on this, and don't necessarily want to turn it into a bike shed, and we can still change the syntax of it later before 0.5 is stabilized.

maciejhirsz (approved):
@DefinitelyNotHilbert cheers!

dvdplm:
@DefinitelyNotHilbert Thank you for your work here. Please post a comment with a polkadot address to receive your tip. :)

DefinitelyNotHilbert:
polkadot address: 13bL3pENnNbFKYD6GPjibJHv4jUTnwNPoVjp9YpU4oJfFefS

dvdplm:
Hmm, stupid bot. We'll get this sorted.

Issue #241, opened 2021-03-12
[client]: not possible to disable TLS


niklasad1:
Ideally, it should be possible for users to choose whether they want to use `TLS` or not for the clients.

It brings in an additional dependencies: `hyper-rustls and async-tls`

niklasad1:
Fixed by https://github.com/paritytech/jsonrpsee/pull/545

PR #545, opened 2021-10-30
clients: feature gate `tls`

niklasad1:
Closing #241

### DIFF
--- a/http-client/Cargo.toml
+++ b/http-client/Cargo.toml
@@ -21,7 +21,6 @@ serde_json = "1.0"
 thiserror = "1.0"
 tokio = { version = "1.8", features = ["time"] }
 tracing = "0.1"
-url = "2.2"
 
 [dev-dependencies]
 jsonrpsee-test-utils = { path = "../test-utils" }
--- a/http-client/Cargo.toml
+++ b/http-client/Cargo.toml
@@ -11,7 +11,7 @@ documentation = "https://docs.rs/jsonrpsee-http-client"
 
 [dependencies]
 async-trait = "0.1"
-hyper-rustls = "0.22"
+hyper-rustls = { version = "0.22", optional = true }
 hyper = { version = "0.14.10", features = ["client", "http1", "http2", "tcp"] }
 jsonrpsee-types = { path = "../types", version = "0.4.1" }
 jsonrpsee-utils = { path = "../utils", version = "0.4.1", features = ["http-helpers"] }
@@ -26,3 +26,7 @@ fnv = "1"
 [dev-dependencies]
 jsonrpsee-test-utils = { path = "../test-utils" }
 tokio = { package = "tokio", version = "1", features = ["net", "rt-multi-thread", "macros"] }
+
+[features]
+default = ["tls"]
+tls = ["hyper-rustls"]
--- a/tests/tests/integration_tests.rs
+++ b/tests/tests/integration_tests.rs
@@ -245,7 +245,7 @@ async fn https_works() {
 #[tokio::test]
 #[ignore]
 async fn wss_works() {
-	let client = WsClientBuilder::default().build("wss://kusama-rpc.polkadot.io").await.unwrap();
+	let client = WsClientBuilder::default().build("wss://kusama-rpc.polkadot.io:443").await.unwrap();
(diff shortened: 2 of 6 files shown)


tests/tests/integration_tests.rs
```diff
@@ -245,7 +245,7 @@ async fn https_works() {
 #[tokio::test]
 #[ignore]
 async fn wss_works() {
-	let client = WsClientBuilder::default().build("wss://kusama-rpc.polkadot.io").await.unwrap();
+	let client = WsClientBuilder::default().build("wss://kusama-rpc.polkadot.io:443").await.unwrap();
```
niklasad1, line 248:
```suggestion
	let client = WsClientBuilder::default().build("wss://kusama-rpc.polkadot.io").await.unwrap();
```

ws-client/src/stream.rs
```diff
@@ -32,30 +32,29 @@ use futures::{
...
 use std::{io::Error as IoError, pin::Pin, task::Context, task::Poll};
+use tokio::net::TcpStream;
 use tokio_util::compat::{TokioAsyncReadCompatExt, TokioAsyncWriteCompatExt};
 
 /// Stream to represent either a unencrypted or encrypted socket stream.
 #[pin_project(project = EitherStreamProj)]
-#[derive(Debug, Copy, Clone)]
-pub enum EitherStream<S, T> {
+#[derive(Debug)]
+pub enum EitherStream {
```
niklasad1, line 41: I did this but I realized I can just used placeholder type such as `()` to keep the generics here not sure that we need the generics anyway.

Unless we plan to support any transport using `tokio rustls` :)

ws-client/src/transport.rs
```diff
@@ -169,38 +166,46 @@ impl Receiver {
...
+		};
+
 		for _ in 0..self.max_redirections {
 			tracing::debug!("Connecting to target: {:?}", target);
 
 			// The sockaddrs might get reused if the server replies with a relative URI.
 			let sockaddrs = std::mem::take(&mut target.sockaddrs);
 			for sockaddr in &sockaddrs {
-				let tcp_stream = match connect(*sockaddr, self.timeout, &target.host, &tls_connector).await {
+				#[cfg(feature = "tls")]
```
niklasad1, line 189: I had to do this to avoid feature gate the entire method.

http-client/Cargo.toml
```diff
@@ -26,3 +26,7 @@ url = "2.2"
 [dev-dependencies]
 jsonrpsee-test-utils = { path = "../test-utils" }
 tokio = { version = "1.8", features = ["net", "rt-multi-thread", "macros"] }
+
+[features]
+default = ["tls"]
+tls = ["hyper-rustls/webpki-tokio"]
```
maciejhirsz, line 32: TIL

maciejhirsz (approved):
Looks good, do we have CI that tests with different features on and off?

niklasad1:
> Looks good, do we have CI that tests with different features on and off?

yeah, we have [this](https://github.com/paritytech/jsonrpsee/blob/master/.github/workflows/ci.yml#L70) which should compile each crate with `default-features = false` and test each feature "standalone". However we just run the tests on the `default features` because it was so slow but I can double check the tests locally to be on the safe-side

http-client/Cargo.toml
```diff
@@ -13,16 +13,19 @@ documentation = "https://docs.rs/jsonrpsee-http-client"
...
-hyper-rustls = { version = "0.23", features = ["webpki-tokio"] }
+hyper-rustls = { version = "0.23", optional = true }
 jsonrpsee-types = { path = "../types", version = "0.6.0" }
 jsonrpsee-utils = { path = "../utils", version = "0.6.0", features = ["client", "http-helpers"] }
 serde = { version = "1.0", default-features = false, features = ["derive"] }
 serde_json = "1.0"
 thiserror = "1.0"
 tokio = { version = "1.8", features = ["time"] }
 tracing = "0.1"
-url = "2.2"
```
niklasad1, line 24: This will close https://github.com/paritytech/jsonrpsee/issues/554

dvdplm (approved):
lgtm

PR #550, opened 2021-11-04
Allow awaiting on server handles

### LINKED ISSUE #547: Feature request: expose an access to http server handle
Call to [`HttpServer::start`](https://docs.rs/jsonrpsee-http-server/0.4.1/jsonrpsee_http_server/struct.HttpServer.html#method.start) consumes the server, spawns a task and gives you a [`StopHandle`](https://docs.rs/jsonrpsee-http-server/0.4.1/jsonrpsee_http_server/struct.HttpStopHandle.html).
As far as I see it, there's no easy way to simply "run forever"? `tokio` join handle is private, `StopHandle` doesn't implement `Future`, probably I miss something, but this looks a little bit inconvenient.

I'd propose to rename `StopHandle` into `ServerHandle` and make it awaitable the same way it works in actix:
[`HttpServer::run`](https://docs.rs/actix-web/3.3.2/actix_web/struct.HttpServer.html#method.run) returns [`Server`](https://docs.rs/actix-web/3.3.2/actix_web/dev/struct.Server.html) which can be used to [`stop`](https://docs.rs/actix-web/3.3.2/actix_web/dev/struct.Server.html#method.stop) the server or can be awaited on to run the application forever (since it [implements](https://docs.rs/actix-web/3.3.2/actix_web/dev/struct.Server.html#impl-Future) `Future`).

slumber:
* Rename both `HttpStopHandle` and `WsStopHandle` to  `HttpServerHandle` and `WsServerHandle` respectively
* Implement `Future` for both allowing to do
```rust
let (addr, handle) = server.start(rpc).unwrap();
handle.await
```
to run the server forever.

The interfaces for HTTP and WS are slightly different in the sense that the latter handle can be cloned. This can be easily implemented via channels for HTTP but that's a subject for different PR and was not introduced here.

Resolves #547

### DIFF
--- a/http-server/src/server.rs
+++ b/http-server/src/server.rs
@@ -26,8 +26,7 @@
 
 use crate::{response, AccessControl};
 use futures_channel::mpsc;
-use futures_util::future::join_all;
-use futures_util::stream::StreamExt;
+use futures_util::{future::join_all, stream::StreamExt, FutureExt};
 use hyper::{
 	server::{conn::AddrIncoming, Builder as HyperBuilder},
 	service::{make_service_fn, service_fn},
@@ -49,7 +48,10 @@ use serde_json::value::RawValue;
 use socket2::{Domain, Socket, Type};
 use std::{
 	cmp,
+	future::Future,
 	net::{SocketAddr, TcpListener},
+	pin::Pin,
+	task::{Context, Poll},
 };
 
 /// Builder to create JSON-RPC HTTP server.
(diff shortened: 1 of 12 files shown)


http-server/src/tests.rs
```diff
@@ -380,6 +381,27 @@ async fn can_register_modules() {
...
+}
+
+#[tokio::test]
+async fn run_forever() {
+	const TIMEOUT: Duration = Duration::from_millis(200);
+
+	let _ = env_logger::try_init();
+	let (_addr, server_handle) = server().with_default_timeout().await.unwrap();
+	// Timed out.
+	server_handle.with_timeout(TIMEOUT).await.unwrap_err();
```
niklasad1, line 395: I would prefer to have an explicit assert here that is easier to read IMO then you can skip the comment :)

http-server/src/tests.rs
```diff
@@ -380,6 +381,27 @@ async fn can_register_modules() {
...
+
+	let (_addr, server_handle) = server().await;
+	server_handle.handle.as_ref().unwrap().abort();
```
niklasad1, line 398: ok, you did this instead of `stop` because it moves the `JoinHandle`?

ws-server/src/tests.rs
```diff
@@ -538,12 +540,27 @@ async fn can_register_modules() {
...
+}
+
+#[tokio::test]
+async fn run_forever() {
+	const TIMEOUT: Duration = Duration::from_millis(200);
+
+	let _ = env_logger::try_init();
+	let (_addr, server_handle) = server_with_handles().with_default_timeout().await.unwrap();
+	// Timed out.
+	server_handle.with_timeout(TIMEOUT).await.unwrap_err();
```
niklasad1, line 560: dito, I prefer to have an explicit assert here that is easier to read IMO.

niklasad1 (commented):
great, I like it

some minor questions to answered :)

http-server/src/tests.rs
slumber: Sure, makes sense

http-server/src/server.rs
```diff
@@ -143,24 +145,37 @@ impl Default for Builder {
...
 }
 
-/// Handle used to stop the running server.
+/// Handle used to run or stop the server.
 #[derive(Debug)]
-pub struct StopHandle {
+pub struct ServerHandle {
 	stop_sender: mpsc::Sender<()>,
-	stop_handle: Option<tokio::task::JoinHandle<()>>,
+	pub(crate) handle: Option<tokio::task::JoinHandle<()>>,
```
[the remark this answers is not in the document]
slumber: It's used for tests

http-server/src/tests.rs
slumber: Stop is a little bit different, here we want to emulate the actual server behavior -- if a task is cancelled or panics, the future handle resolves too.

niklasad1 (approved):

maciejhirsz (approved):
Solid work!

dvdplm:
@slumber Thank you for your work here! If you want to you can add your polkadot address to the PR description and I'll invoke the magic "Parity tip-bot" :)

Edit the description and add a line saying: "polkadot address: YOUR_ADDRESS_HERE".

PR #537, opened 2021-10-20
ws server: respect max limit for received messages

niklasad1:
This PR changes such that we use the `max limit` for the received messages in `soketto (WS)` to deny too large message and not allocate a huge buffer that we did before.

PR #553, opened 2021-11-09
ws server: reject too big response

niklasad1:
builds on top of #537


ws-server/src/server.rs
```diff
@@ -252,9 +257,31 @@ async fn background_task(
...
-					let _ = sender.send_text(response).await;
-					let _ = sender.flush().await;
+					if response.len() > max_request_body_size as usize {
+						tracing::warn!(
+							"WS Transport error: response to method call too large {}, max: {}",
+							response.len(),
+							max_request_body_size
+						);
+						// TODO(niklasad1): include `id` in the response and send back `response too big; id=id` here?!
+						// also we could terminate the connection...
```
niklasad1, line 267: /cc @maciejhirsz 

currently we have `UnboundedChannel::<String>` but we would like the extract the `ID` from there in case the response is too big and send `{"jsonrpc":"2.0", "result":"Response to call was too big", "id":<ID>}`

however, we could simply use `Id::Null` or `Error` here too these will likely be thrown by the jsonrpc clients anyway I guess...

maciejhirsz: It sounds like it could be something that's cought earlier in the pipeline, in the module itself perhaps?

We could also do a custom `Write` impl that takes care of the max buffer size [during serialization](https://docs.rs/serde_json/1.0.69/serde_json/fn.to_writer.html).

dvdplm: Why not make this into an error, `InvalidRequest` (code `-32600`)?

niklasad1: > It sounds like it could be something that's cought earlier in the pipeline, in the module itself perhaps?

> We could also do a custom Write impl that takes care of the max buffer size during serialization.

Good idea

> Why not make this into an error, InvalidRequest (code -32600)?

yeah, an error sounds good I forgot that the error also contains request ID.

utils/src/server/rpc_module.rs
```diff
@@ -421,11 +428,12 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 		F: Fn(Params, &Context) -> Result<R, Error> + Send + Sync + 'static,
 	{
 		let ctx = self.ctx.clone();
+		let max_call_size = self.max_call_size;
 		let callback = self.methods.verify_and_insert(
 			method_name,
 			MethodCallback::new_sync(Arc::new(move |id, params, tx, _| {
```
maciejhirsz, line 434: I think it might be better if we pass the `max_call_size` (shouldn't it be `max_response_size`?) as a param to closures here, so it can be set by either server when calling the method. This would need piping it through on every `execute`, but I reckon that's better than configuring it globally on the module itself.

maciejhirsz (changes requested):
Nice! Bunch of random nits and suggestions of varying severity :)

utils/src/server/rpc_module.rs
```diff
@@ -393,12 +393,19 @@ impl<Context> DerefMut for RpcModule<Context> {
 pub struct RpcModule<Context> {
 	ctx: Arc<Context>,
 	methods: Methods,
+	/// The maximum size of a executed call.
```
dvdplm, line 396:
```suggestion
	/// The maximum size of an executed call.
```

utils/src/server/rpc_module.rs
```diff
@@ -393,12 +393,19 @@ impl<Context> DerefMut for RpcModule<Context> {
...
 }
 
 impl<Context> RpcModule<Context> {
 	/// Create a new module with a given shared `Context`.
 	pub fn new(ctx: Context) -> Self {
-		Self { ctx: Arc::new(ctx), methods: Default::default() }
+		Self { ctx: Arc::new(ctx), methods: Default::default(), max_call_size: 10 * 1024 * 1024 }
+	}
+
+	/// Set the maximum size limit for a executed call.
```
dvdplm, line 406:
```suggestion
	/// Set the maximum size limit for the result of executing a call.
```

utils/src/server/rpc_module.rs
dvdplm: +1 for `max_response_size`

utils/src/server/helpers.rs
```diff
@@ -36,21 +36,31 @@ use jsonrpsee_types::v2::{
 use serde::Serialize;
 use std::io;
 
-struct BoundedWriter {
+/// Bounded writer that allow writing at most `max_len` bytes.
```
dvdplm, line 39:
```suggestion
/// Bounded writer that allows writing at most `max_len` bytes.
```

dvdplm (approved):

utils/src/server/rpc_module.rs
niklasad1: thanks yeah indeed much cleaner

utils/src/server/helpers.rs
```diff
@@ -28,20 +28,81 @@ use crate::server::rpc_module::MethodSink;
...
+		}
 		Err(err) => {
 			tracing::error!("Error serializing response: {:?}", err);
 
-			return send_error(id, tx, ErrorCode::InternalError.into());
+			let err = if err.is_io() {
+				ErrorObject {
+					code: ErrorCode::ServerError(OVERSIZED_RESPONSE_CODE),
+					message: OVERSIZED_RESPONSE_MSG,
+					data: None,
```
niklasad1, line 99: //cc @dvdplm 

I tried to add `data: to_raw_value(&format!("exceeded max limit {}", max_response_size)).ok().as_deref()`

but I had some issues w.r.t lifetimes because `pub data: Option<&'a RawValue>`

utils/src/server/helpers.rs
```diff
@@ -28,20 +28,81 @@ use crate::server::rpc_module::MethodSink;
...
+				}
```
maciejhirsz, line 100: It works if you move return into the block:

```suggestion
				let data = serde_json::value::to_raw_value(&format!("exceeded max limit {}", max_response_size)).ok();

				let err = ErrorObject {
					code: ErrorCode::ServerError(OVERSIZED_RESPONSE_CODE),
					message: OVERSIZED_RESPONSE_MSG,
					data: data.as_deref(),
				};

				return send_error(id, tx, err, max_response_size);
```

utils/src/server/helpers.rs
```diff
@@ -51,12 +112,12 @@ pub fn send_response(id: Id, tx: &MethodSink, result: impl Serialize) {
 }
 
 /// Helper for sending JSON-RPC errors to the client
-pub fn send_error(id: Id, tx: &MethodSink, error: ErrorObject) {
-	let json = match serde_json::to_string(&RpcError { jsonrpc: TwoPointZero, error, id }) {
-		Ok(json) => json,
+pub fn send_error(id: Id, tx: &MethodSink, error: ErrorObject, max_response_size: u32) {
```
maciejhirsz, line 115: Not sure this is necessary, it should be pretty safe to assume that errors never exceed whatever limit people have set, plus if they have set the limit to, like, `10`, it would be nice to ignore it to send the error that we've exceeded the limit of `10` :D.

niklasad1:
> Not sure this is necessary, it should be pretty safe to assume that errors never exceed whatever limit people have set, plus if they have set the limit to, like, 10, it would be nice to ignore it to send the error that we've exceeded the limit of 10 :D.

Yeah, I agree it's nice to ensure that we can propagate the errors maybe we could document somewhere that it might send big responses if their error messages are huge... but again I think this shouldn't be a practical concern I will revert it.

maciejhirsz (approved):

PR #556, opened 2021-11-10
Periodically wake `DriverSelect` so we can poll whether or not `stop` had been called.

### LINKED ISSUE #549: `WsServer` not shutting down if a connected client stays idle
Easy to reproduce by [removing the `drop` in this test](https://github.com/paritytech/jsonrpsee/blob/ff3337b107bf29bef6067164c20c6a0b0b5bdc75/tests/tests/integration_tests.rs#L359).

This might be related to the memory leak we see in substrate jsonrpsee branch. Needs a deep dive on the `StopMonitor` and when/how it's being polled.

maciejhirsz:
Fixes #549. This is all pretty esoteric, but TL;DR is:

+ The loop in the per-connection background task has been reworked to more closely resemble the loop in accepting incoming connections, making sure we check the `StopMonitor` every time the driver has been polled.
+ Added an `Interval` internally in `FutureDriver`, which is only polled in the `DriverSelect` instances. It effectively triggers a forced poll every second, which checks the stop flag, which will shut down all loops.

The tests are now properly terminating without having to drop the client (with a 1s delay):

```
$ cargo test --test resource_limiting
   Compiling jsonrpsee-integration-tests v0.4.1 (/home/maciej/rust/jsonrpsee/tests)
    Finished test [unoptimized + debuginfo] target(s) in 2.94s
     Running tests/resource_limiting.rs (target/debug/deps/resource_limiting-17aaec4a6e8ab5eb)

running 4 tests
test http_server_with_macro_module ... ok
test http_server_with_manual_module ... ok
test ws_server_with_manual_module ... ok
test ws_server_with_macro_module ... ok

test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.00s
```

### DIFF
--- a/ws-server/src/future.rs
+++ b/ws-server/src/future.rs
@@ -36,6 +36,10 @@ use std::sync::{
 	Arc, Weak,
 };
 use std::task::{Context, Poll};
+use tokio::time::{self, Duration, Interval};
+
+/// Polling for server stop monitor interval in milliseconds.
+const POLLING_HEARTBEAT: u64 = 1000;
 
 /// This is a flexible collection of futures that need to be driven to completion
 /// alongside some other future, such as connection handlers that need to be
@@ -45,11 +49,16 @@ use std::task::{Context, Poll};
 /// `select_with` providing some other future, the result of which you need.
 pub(crate) struct FutureDriver<F> {
 	futures: Vec<F>,
+	heartbeat: Interval,
 }
 
 impl<F> Default for FutureDriver<F> {
 	fn default() -> Self {
-		FutureDriver { futures: Vec::new() }
+		let mut heartbeat = time::interval(Duration::from_millis(POLLING_HEARTBEAT));
+
+		heartbeat.set_missed_tick_behavior(time::MissedTickBehavior::Skip);
+
+		FutureDriver { futures: Vec::new(), heartbeat }
 	}
 }
 
@@ -92,6 +101,12 @@ where
 			}
 		}
 	}
+
+	fn poll_heartbeat(&mut self, cx: &mut Context) {
+		// We don't care about the ticks of the heartbeat, it's here only
+		// to periodically wake the `Waker` on `cx`.
+		let _ = self.heartbeat.poll_tick(cx);
+	}
 }
 
 impl<F> Future for FutureDriver<F>
@@ -132,6 +147,7 @@ where
 		let this = Pin::into_inner(self);
 
 		this.driver.drive(cx);
+		this.driver.poll_heartbeat(cx);
 
 		this.selector.poll_unpin(cx)
 	}
--- a/ws-server/src/future.rs
+++ b/ws-server/src/future.rs
@@ -39,7 +39,7 @@ use std::task::{Context, Poll};
 use tokio::time::{self, Duration, Interval};
 
 /// Polling for server stop monitor interval in milliseconds.
-const POLLING_HEARTBEAT: u64 = 1000;
+const STOP_MONITOR_POLLING_INTERVAL: u64 = 1000;
 
 /// This is a flexible collection of futures that need to be driven to completion
 /// alongside some other future, such as connection handlers that need to be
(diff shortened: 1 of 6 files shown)


niklasad1 (approved):
LGTM,

Did you run benches by any chance?

ws-server/src/future.rs
```diff
@@ -36,6 +36,10 @@ use std::sync::{
 	Arc, Weak,
 };
 use std::task::{Context, Poll};
+use tokio::time::{self, Duration, Interval};
+
+/// Polling for server stop monitor interval in milliseconds.
+const POLLING_HEARTBEAT: u64 = 1000;
```
dvdplm, line 42:
```suggestion
const STOP_MONITOR_POLLING_INTERVAL: u64 = 1000;
```

ws-server/src/future.rs
```diff
@@ -45,11 +49,16 @@ use std::task::{Context, Poll};
...
 	futures: Vec<F>,
+	heartbeat: Interval,
 }
 
 impl<F> Default for FutureDriver<F> {
 	fn default() -> Self {
-		FutureDriver { futures: Vec::new() }
+		let mut heartbeat = time::interval(Duration::from_millis(POLLING_HEARTBEAT));
+
+		heartbeat.set_missed_tick_behavior(time::MissedTickBehavior::Skip);
```
dvdplm, line 59: TIL

ws-server/src/future.rs
```diff
@@ -92,6 +101,12 @@ where
 			}
 		}
 	}
+
+	fn poll_heartbeat(&mut self, cx: &mut Context) {
```
dvdplm, line 105:
```suggestion
	fn poll_stop_monitor(&mut self, cx: &mut Context) {
```

(Not an accurate name I know, but gives a hint in the right direction?)

ws-server/src/server.rs
```diff
@@ -122,46 +122,64 @@ impl Server {
...
 
 		connections.await
 	}
 }
 
-/// This is a glorified select listening to new connections, while also checking
-/// for `stop_receiver` signal.
-struct Incoming<'a> {
-	listener: TcpListener,
+/// This is a glorified select listening to new messages, while also checking the `stop_receiver` signal.
```
dvdplm, line 136:
```suggestion
/// This is a glorified select listening for new messages, while also checking the `stop_receiver` signal.
```

ws-server/src/server.rs
```diff
@@ -275,31 +293,39 @@ async fn background_task(
...
+			if let Err(err) = method_executors.select_with(Monitored::new(receive, &stop_server)).await {
+				match err {
+					MonitoredError::Selector(SokettoError::Closed) => {
+						tracing::debug!("Remote peer terminated the connection: {}", conn_id);
+						tx.close_channel();
+						return Ok(());
+					}
+					MonitoredError::Selector(SokettoError::MessageTooLarge { current, maximum }) => {
+						tracing::warn!(
+							"WS transport error: message is too big error ({} bytes, max is {})",
```
dvdplm, line 314:
```suggestion
							"WS transport error: outgoing message is too big error ({} bytes, max is {})",
```

ws-server/src/server.rs
```diff
@@ -275,31 +293,39 @@ async fn background_task(
...
+		{
+			// Need the extra scope to drop this pinned future and reclaim access to `data`
+			let receive = receiver.receive_data(&mut data);
+
+			tokio::pin!(receive);
+
+			if let Err(err) = method_executors.select_with(Monitored::new(receive, &stop_server)).await {
+				match err {
+					MonitoredError::Selector(SokettoError::Closed) => {
+						tracing::debug!("Remote peer terminated the connection: {}", conn_id);
```
dvdplm, line 308:
```suggestion
						tracing::debug!("WS transport error: remote peer terminated the connection: {}", conn_id);
```

(not to self, plug in close reason here one day)

ws-server/src/future.rs
```diff
@@ -45,11 +49,16 @@ use std::task::{Context, Poll};
 /// `select_with` providing some other future, the result of which you need.
 pub(crate) struct FutureDriver<F> {
 	futures: Vec<F>,
+	heartbeat: Interval,
```
dvdplm, line 52:
```suggestion
	stop_monitor_heartbeat: Interval,
```

maybe? Either way, some docs here would be good.

dvdplm (approved):

PR #557, opened 2021-11-11
Array syntax aliases

maciejhirsz:
This change requires aliases to be defined using array syntax instead of a string we have to split and separate by commas. I've removed all `trim`s in proc macro code, meaning method names are taken verbatim as-is without any magic.

### DIFF
--- a/proc-macros/src/rpc_macro.rs
+++ b/proc-macros/src/rpc_macro.rs
@@ -31,9 +31,9 @@ use crate::{
 	helpers::extract_doc_comments,
 };
 
-use std::borrow::Cow;
 use proc_macro2::TokenStream as TokenStream2;
 use quote::quote;
+use std::borrow::Cow;
 use syn::spanned::Spanned;
 use syn::{punctuated::Punctuated, Attribute, Token};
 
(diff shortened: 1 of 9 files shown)


proc-macros/src/rpc_macro.rs
```diff
@@ -276,32 +277,25 @@ impl RpcDescription {
...
-	let method = existing_method.trim();
-	let mut new_method = String::from("unsubscribe");
-	if method.starts_with("subscribe") {
-		new_method.extend(method.chars().skip(9));
-	} else {
-		new_method.push_str(method);
-	}
-	new_method
+fn build_unsubscribe_method(method: &str) -> String {
+	format!("unsubscribe{}", method.strip_prefix("subscribe").unwrap_or(method))
```
niklasad1, line 300: :+1:

niklasad1 (approved):

dvdplm (approved):

dvdplm:
@maciejhirsz given this is a breaking change, can you add it to the `0.5` tracking ticket so we don't forget to mention it in the changelog?

Issue #559, opened 2021-11-11
[ws server]: `unsubscribe_call` to non-active subscription should indicate error


niklasad1:
```
		// Subscribe call.
		let sub_resp = rpc
			.call_with("subscribeSomething", Vec::<()>::new())
			.await
			.unwrap();
		let sub_id: SubscriptionId = deser_call(sub_resp);

		// Unsubscribe
		assert_eq!(
			rpc.call_with("unsubscribeSomething", [sub_id.clone()]).await,
			Some(r#"{"jsonrpc":"2.0","result":"Unsubscribed","id":0}"#.into())
		);

		// Should indicate that subscription was already closed and not active anymore!
               // but doesn't
		assert_eq!(
			rpc.call_with("unsubscribeSomething", [sub_id]).await,
			Some("{\"jsonrpc\":\"2.0\",\"error\":{\"code\":-32602,\"message\":\"Invalid subscription id.\"},\"id\":1}".into()),
		);
```

niklasad1:
Closed by #561

PR #561, opened 2021-11-15
rpc module: report error on invalid subscription

niklasad1:
Closing #559

### DIFF
--- a/types/src/v2/error.rs
+++ b/types/src/v2/error.rs
@@ -44,6 +44,13 @@ pub struct RpcError<'a> {
 	pub id: Id<'a>,
 }
 
+impl<'a> RpcError<'a> {
+	/// Create a new `RpcError`.
+	pub fn new(error: ErrorObject<'a>, id: Id<'a>) -> Self {
+		Self { jsonrpc: TwoPointZero, error, id }
+	}
+}
+
 impl<'a> fmt::Display for RpcError<'a> {
 	fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
 		write!(f, "{}", serde_json::to_string(&self).expect("infallible; qed"))
@@ -62,6 +69,13 @@ pub struct ErrorObject<'a> {
 	#[serde(skip_serializing_if = "Option::is_none")]
 	#[serde(borrow)]
 	pub data: Option<&'a RawValue>,
+}
+
+impl<'a> ErrorObject<'a> {
+	/// Create a new `ErrorObject` with optional data.
+	pub fn new(code: ErrorCode, data: Option<&'a RawValue>) -> ErrorObject<'a> {
+		Self { code, message: code.message(), data }
+	}
 }
 
 impl<'a> From<ErrorCode> for ErrorObject<'a> {
@@ -214,6 +228,11 @@ impl serde::Serialize for ErrorCode {
 	{
 		serializer.serialize_i32(self.code())
 	}
+}
+
+/// Create a invalid subscription ID error.
+pub fn invalid_subscription_err(data: Option<&RawValue>) -> ErrorObject {
+	ErrorObject::new(ErrorCode::ServerError(INVALID_SUBSCRIPTION_CODE), data)
 }
 
 #[cfg(test)]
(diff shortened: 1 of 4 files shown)


types/src/v2/error.rs
```diff
@@ -73,7 +73,7 @@ impl<'a> From<ErrorCode> for ErrorObject<'a> {
 impl<'a> PartialEq for ErrorObject<'a> {
 	fn eq(&self, other: &Self) -> bool {
 		let this_raw = self.data.map(|r| r.get());
-		let other_raw = self.data.map(|r| r.get());
+		let other_raw = other.data.map(|r| r.get());
```
niklasad1, line 76: unrelated bug I found

ws-server/src/tests.rs
```diff
@@ -569,3 +587,58 @@ async fn run_forever() {
 	// Send the shutdown request from one handle and await the server on the second one.
 	join(server_handle.clone().stop().unwrap(), server_handle).with_timeout(TIMEOUT).await.unwrap();
 }
+
+#[tokio::test]
+async fn unsubscribe_twice_should_indicate_error() {
```
niklasad1, line 592: ideally these tests should be in the RPC module but I have to refactor the `call_with` for that to work because inner `channel` is dropped in `call_with/call/execute`...

types/src/v2/error.rs
dvdplm: Oops. No test for this eh?

ws-server/src/tests.rs
```diff
@@ -107,6 +115,16 @@ async fn server_with_handles() -> (SocketAddr, ServerHandle) {
 			Ok("Yawn!")
 		})
 		.unwrap();
+	module
+		.register_subscription("subscribe_hello", "unsubscribe_hello", |_, sink, _| {
+			std::thread::spawn(move || {
+				// prevent the sink for being dropped and keep the subscription alive for testing.
+				let _ = sink;
+				loop {}
```
dvdplm, line 123: Clippy is here all red in the face for the upset about this.

ws-server/src/tests.rs
```diff
@@ -569,3 +587,58 @@ async fn run_forever() {
...
+	let sub_id = to_json_raw_value(&sub_id).unwrap();
+	assert_eq!(
+		unsub_2_err,
+		RpcError {
+			jsonrpc: v2::TwoPointZero,
+			error: v2::ErrorObject {
+				code: v2::ErrorCode::ServerError(v2::error::INVALID_SUBSCRIPTION_CODE),
+				message: v2::error::INVALID_SUBSCRIPTION_MSG,
+				data: Some(&to_json_raw_value(&format!("ID={} is not active", sub_id)).unwrap()),
+			},
```
dvdplm, line 616: We should have a macro for converting a code to an ErrorObject, `to_error_object!(INVALID_SUBSCRIPTION_CODE)`

dvdplm (approved):

types/src/v2/error.rs
niklasad1: Indeed

ws-server/src/tests.rs
niklasad1: we already have a `From<ErrorCode> for ErrorObject` lemme see I can make this a little less ugly

types/src/v2/error.rs
```diff
@@ -84,6 +98,8 @@ pub const PARSE_ERROR_CODE: i32 = -32700;
 pub const OVERSIZED_REQUEST_CODE: i32 = -32701;
 /// Oversized response error code.
 pub const OVERSIZED_RESPONSE_CODE: i32 = -32702;
+/// Oversized response error code.
+pub const INVALID_SUBSCRIPTION_CODE: i32 = -32703;
```
niklasad1, line 102: BTW, technically these error codes should be used for application errors.

To be really nitpicky I guess we should use `-32000 to -32099 Server defined errors here`

dvdplm: Given this is a new code, do you think we should do the right thing for new error codes and stay within the right range?

niklasad1: yeah, I used `32002` for this

Issue #448, opened 2021-09-03
[proc macros]: client implementations with subscriptions doesn't work for HTTP client


niklasad1:
For example if a define a API like this:

```rust
#[rpc(server, client, namespace = "state")]
pub trait Rpc<Hash: Clone, StorageKey>
where
	Hash: std::fmt::Debug,
{
	/// Async method call example.
	#[method(name = "getKeys")]
	async fn storage_keys(&self, storage_key: StorageKey, hash: Option<Hash>) -> Result<Vec<StorageKey>, Error>;

	/// Subscription that takes a `StorageKey` as input and produces a `Vec<Hash>`.
	#[subscription(name = "subscribeStorage", item = Vec<Hash>)]
	fn subscribe_storage(&self, keys: Option<Vec<StorageKey>>);
}

#[tokio::main]
async fn main() {
    let client = HttpClientBuilder::default().build(url)?;
    Rpc::<(), ()>::storage_keys((), None).await.unwrap();
}
```

Then if I try to use for the `HTTP client` it won't work because the client implementation requires `SubscriptionClient trait bound` when it has at least one subscription in the trait.

For those we should just ignore the subscriptions and require the `Client trait bound` instead of possible, thoughts or suggestions?

dvdplm:
> For those we should just ignore the subscriptions and require the `Client bound` instead of possible, thought or suggestions?

Not sure I understand what this means, can you elaborate?

niklasad1:
Alright,

To elaborate a little more, for the clients we have two internal traits to model the capabilities of the client:

1. Client - basic functionality to perform JSON-RPC method calls and notifications.
2. SubscriptionClient - client functionality + subscription support.

If I define a trait `RPC trait` with only requests, for example as the following:

```rust
#[rpc(server, client, namespace = "state")]
pub trait Rpc<Hash: Clone, StorageKey>
where
	Hash: std::fmt::Debug,
{
	/// Async method call example.
	#[method(name = "getKeys")]
	async fn storage_keys(&self, storage_key: StorageKey, hash: Option<Hash>) -> Result<Vec<StorageKey>, Error>;
}
```

Expands to
```rust
pub trait RpcClient<Hash: Clone, StorageKey>: jsonrpsee::types::traits::Client
where
    Hash: Send + Sync + 'static + jsonrpsee::types::Serialize + std::fmt::Debug,
    StorageKey:
[...]

PR #563, opened 2021-11-15
Implement SubscriptionClient for HttpClient

dvdplm:
Closes https://github.com/paritytech/jsonrpsee/issues/448

This PR adds an implementation for `SubscriptionClient` to the `HttpClient` struct, which makes it possible for http clients to use macro-generated RPC servers. If an http client tries to set up a subscription it will fail with a `HttpNotImplemented` error.

### DIFF
--- a/http-client/src/client.rs
+++ b/http-client/src/client.rs
@@ -26,9 +26,9 @@
 
 use crate::transport::HttpTransportClient;
 use crate::types::{
-	traits::Client,
+	traits::{Client, SubscriptionClient},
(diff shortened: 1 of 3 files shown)


niklasad1 (approved):
I would prefer:

> ignore any subscriptions for http clients and only implement the Client trait for them

but I'm not sure if that's possible so let's go with this for now

dvdplm:
> but I'm not sure if that's possible so let's go with this for now

Yeah me neither, I don't know how that would work tbf.

PR #566, opened 2021-11-16
[rpc module]: improve `TestSubscription` to return `None` when closed

niklasad1:


utils/src/server/rpc_module.rs
```diff
@@ -928,14 +932,39 @@ mod tests {
...
+			})
+			.unwrap();
+
+		let mut my_sub: TestSubscription = module.test_subscription("my_sub", Vec::<()>::new()).await;
+		let (val, id) = my_sub.next::<String>().await.unwrap();
+		assert_eq!(&val, "lo");
+		assert_eq!(id, v2::params::SubscriptionId::Num(my_sub.subscription_id()));
+
+		// close the subscription to ensure it doesn't return any items.
+		my_sub.close();
```
niklasad1, line 967: technically this doesn't test that inner thread was terminated correctly and it's bit cheating because we terminate the `SubscriptionSink` directly there but we do the exact thing in `WsServer` so should be sufficient here.

niklasad1: I guess it's possible to inject a `channel/flag` to the RpcModule for testing but really annoying for other however the `SubscriptionSink` terminates the subscription once `SubscriptionSink::send` fails anyway...

dvdplm (approved):

maciejhirsz (approved):

TarikGul (commented):
👍

Issue #562, opened 2021-11-15
Unable to discern address in use error


maciejhirsz:
When the JSON-RPC address is in use, [Substrate retries with a random port](https://github.com/paritytech/substrate/blob/7db0768a85dc36a3f2a44d042b32f3715c00a90d/client/service/src/lib.rs#L353-L355). This work fine with `jsonrpc`, but we unfortunately erase std::io errors type with anyhow.

We should have a new error variant(s) just for `AddressInUse | PermissionDenied`. Bonus points if we can either not consume the builder, or return the builder in the error for easy retries. Alternatively support the strategy of trying random port directly into jsonrpsee.

niklasad1:
right, didn't think about this....

maciejhirsz:
Solved by #567.

PR #567, opened 2021-11-16
feat: make it possible to try several sockaddrs when starting server

niklasad1:
Closing #562

### DIFF
--- a/http-server/src/server.rs
+++ b/http-server/src/server.rs
@@ -114,8 +114,8 @@ impl Builder {
 	///       "127.0.0.1:10".parse().unwrap(),
 	///       "127.0.0.1:0".parse().unwrap(),
 	///   ];
-	///   assert!(jsonrpsee_http_server::HttpServerBuilder::default().build(addrs[0]).await.is_err());
-	///   assert!(jsonrpsee_http_server::HttpServerBuilder::default().build(addrs).await.is_ok());
+	///   assert!(jsonrpsee_http_server::HttpServerBuilder::default().build(addrs[0]).is_err());
+	///   assert!(jsonrpsee_http_server::HttpServerBuilder::default().build(addrs).is_ok());
 	/// }
 	/// ```
 	pub fn build(self, addrs: impl ToSocketAddrs) -> Result<Server, Error> {
--- a/http-server/src/server.rs
+++ b/http-server/src/server.rs
@@ -110,11 +110,13 @@ impl Builder {
 	/// ```rust
 	/// #[tokio::main]
 	/// async fn main() {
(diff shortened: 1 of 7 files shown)


http-server/src/server.rs
```diff
@@ -106,7 +106,49 @@ impl Builder {
...
+	///   let mut builder = jsonrpsee_http_server::HttpServerBuilder::default();
+	///   // try several addresses
+	///   let addrs: &[std::net::SocketAddr] = &[
+	///       "127.0.0.1:10".parse::<std::net::SocketAddr>().unwrap(),
+	///       "127.0.0.1:0".parse::<std::net::SocketAddr>().unwrap(),
+	///   ];
+	///   let server = builder.build(addrs).unwrap();
+	/// }
+	/// ```
+	pub fn build(self, addrs: impl ToSocketAddrs) -> Result<Server, Error> {
```
niklasad1, line 122: NOTE: this is `std::net::ToSocketAddrs` because `tokio::std::ToSocketAddrs` is sealed i.e, doesn't provide `to_socket_addrs` because we configure the Socket ourselves with some settings.

http-server/src/server.rs
```diff
@@ -106,7 +106,49 @@ impl Builder {
...
+		let mut err: Option<Error> = None;
+
+		for addr in addrs.to_socket_addrs()? {
+			let (listener, local_addr) = match self.inner_builder(addr) {
+				Ok(l) => l,
```
niklasad1, line 127:
```suggestion
				Ok(res) => res,
```

dvdplm (approved):
This lgtm. It would be nice to have a test for this though?

niklasad1:
> This lgtm. It would be nice to have a test for this though?

I added a test in the rust docs should be sufficient?
Port 10 doesn't work if you are not running in ordinary `user mode`

http-server/src/server.rs
```diff
@@ -106,7 +106,49 @@ impl Builder {
...
+	///
+	/// ```rust
+	/// #[tokio::main]
+	/// async fn main() {
+	///   let mut builder = jsonrpsee_http_server::HttpServerBuilder::default();
+	///   // try several addresses
+	///   let addrs: &[std::net::SocketAddr] = &[
+	///       "127.0.0.1:10".parse::<std::net::SocketAddr>().unwrap(),
+	///       "127.0.0.1:0".parse::<std::net::SocketAddr>().unwrap(),
+	///   ];
```
maciejhirsz, line 118: Either the type or the turbofish should be unnecessary here.

maciejhirsz (approved):
Just one tiny nit on doc comments.

PR #568, opened 2021-11-17
feat: make it possible to override `method name` in subscriptions

niklasad1:
Fixes https://github.com/paritytech/substrate/pull/8783#discussion_r749664072 to be compatible with current the behavior on `substrate master` to make it possible to configure custom method name for subscriptions (i.e, not the same as subscription method name) to avoid breakage for libraries that can't distinguish between `ordinary notifications` and `subscription notifications`

Example:
```
/// Use `override` as method name in the subscription notif
#[subscription(name = "subscribe_method" => "override", item = u32)]
fn sub_with_override(&self) -> RpcResult<()>;


/// Use default `subscription name` as method name in the subscription notif
#[subscription(name = "subscribeHello", item = u32)]
fn sub_with_override(&self) -> RpcResult<()>;

```

### DIFF
--- a/proc-macros/src/rpc_macro.rs
+++ b/proc-macros/src/rpc_macro.rs
@@ -97,9 +97,9 @@ impl RpcMethod {
 #[derive(Debug, Clone)]
 pub struct RpcSubscription {
 	pub name: String,
-	/// By default the server will send out subscriptions using `method` in `name` above
-	/// but it's possible to override it with another name that this field does.
-	pub override_notif_method: Option<String>,
+	/// By default the server will send data to subscribers using the `name` field above as the `method` field in the JSON payload
+	/// but it's possible to override it with another name using this field.
+	pub payload_name_override: Option<String>,
 	pub docs: TokenStream2,
 	pub unsubscribe: String,
 	pub params: Vec<(syn::PatIdent, syn::Type)>,
--- a/proc-macros/src/rpc_macro.rs
+++ b/proc-macros/src/rpc_macro.rs
@@ -95,6 +95,9 @@ impl RpcMethod {
 #[derive(Debug, Clone)]
 pub struct RpcSubscription {
 	pub name: String,
+	/// By default the server will send out subscriptions `method` in `name` above
+	/// but it's possible to override it with another name that this field does.
+	pub override_notif_method: Option<String>,
 	pub docs: TokenStream2,
 	pub unsubscribe: String,
 	pub params: Vec<(syn::PatIdent, syn::Type)>,
@@ -107,11 +110,15 @@ pub struct RpcSubscription {
 
 impl RpcSubscription {
 	pub fn from_item(attr: syn::Attribute, mut sub: syn::TraitItemMethod) -> syn::Result<Self> {
-		let [aliases, item, name, param_kind, unsubscribe_aliases] =
-			AttributeMeta::parse(attr)?.retain(["aliases", "item", "name", "param_kind", "unsubscribe_aliases"])?;
+		let [aliases, item, name, param_kind, unsubscribe_aliases, override_notif_method] = AttributeMeta::parse(attr)?
+			.retain(["aliases", "item", "name", "param_kind", "unsubscribe_aliases", "override_notif_method"])?;
 
 		let aliases = parse_aliases(aliases)?;
 		let name = name?.string()?;
+		let override_notif_method = match override_notif_method {
+			Ok(arg) => Some(arg.string()?),
+			Err(_) => None,
+		};
 		let item = item?.value()?;
 		let param_kind = parse_param_kind(param_kind)?;
 		let unsubscribe_aliases = parse_aliases(unsubscribe_aliases)?;
@@ -135,7 +142,18 @@ impl RpcSubscription {
 		// We've analyzed attributes and don't need them anymore.
 		sub.attrs.clear();
 
-		Ok(Self { name, unsubscribe, unsubscribe_aliases, params, param_kind, item, signature: sub, aliases, docs })
+		Ok(Self {
+			name,
+			override_notif_method,
+			unsubscribe,
+			unsubscribe_aliases,
+			params,
+			param_kind,
+			item,
+			signature: sub,
+			aliases,
+			docs,
+		})
 	}
 }
 
@@ -282,6 +300,18 @@ impl RpcDescription {
 			format!("{}_{}", ns, method).into()
 		} else {
 			Cow::Borrowed(method)
+		}
+	}
+
+	/// Based on the namespace, renders the full name of the RPC method/subscription.
+	/// Examples:
+	/// For namespace `foo` and method `makeSpam`, result will be `foo_makeSpam`.
+	/// For no namespace and method `makeSpam` it will be just `makeSpam.
+	pub(crate) fn optional_rpc_identifier<'a>(&self, method: Option<&'a str>) -> Option<Cow<'a, str>> {
+		if let Some(method) = method {
+			Some(self.rpc_identifier(method))
+		} else {
+			None
 		}
 	}
 }
--- a/proc-macros/src/rpc_macro.rs
+++ b/proc-macros/src/rpc_macro.rs
@@ -101,7 +101,7 @@ pub struct RpcSubscription {
 	/// in the JSON data sent to subscribers.
 	/// Each subscription thus has one method name to set up the subscription,
 	/// one to unsubscribe and, optionally, a third method name used to describe the
-	/// payload sent back from the server to subscribers.
+	/// payload (aka "notification") sent back from the server to subscribers.
 	/// If no override is provided, the subscription method name is used.
 	pub name_override: Option<String>,
 	pub docs: TokenStream2,
--- a/proc-macros/src/rpc_macro.rs
+++ b/proc-macros/src/rpc_macro.rs
@@ -97,10 +97,12 @@ impl RpcMethod {
 #[derive(Debug, Clone)]
 pub struct RpcSubscription {
 	pub name: String,
-	/// By default the server will send data to subscribers using the `name` field above as the `method` field in the JSON payload
-	/// but it's possible to override it with another name using this field.
-	///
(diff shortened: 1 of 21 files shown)


dvdplm:
> Thoughts about this such as naming?!

I absolutely hate it. 😆 

I would like to not have to support this, or at the very least understand why we need it.

utils/src/server/rpc_module.rs
```diff
@@ -406,12 +406,13 @@ impl<Context> DerefMut for RpcModule<Context> {
...
 	ctx: Arc<Context>,
 	methods: Methods,
+	notif_overrides: SubscriptionNotifOverride,
 }
 
 impl<Context> RpcModule<Context> {
 	/// Create a new module with a given shared `Context`.
 	pub fn new(ctx: Context) -> Self {
-		Self { ctx: Arc::new(ctx), methods: Default::default() }
+		Self { ctx: Arc::new(ctx), methods: Default::default(), notif_overrides: SubscriptionNotifOverride::default() }
```
maciejhirsz, line 415: This doesn't really prevent us from having the same notification name in two different modules (the set is local to the instance and is thrown away the moment you turn it into `Methods`).

If we can't easily guarantee that notification names are unique, and there isn't really much of a point in checking that they are in the server (notifications will be sent regardless, and we rely on sub id more than name anyway), I think it might actually be better to drop the check altogether.

utils/src/server/rpc_module.rs
```diff
@@ -544,6 +545,41 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
...
+		)
+	}
+
+	fn register_subscription_inner<F>(
+		&mut self,
+		subscribe_method_name: &'static str,
+		custom_notif_method_name: Option<&'static str>,
+		unsubscribe_method_name: &'static str,
+		callback: F,
+	) -> Result<(), Error>
```
maciejhirsz, line 582: Any reason why this can't be the actual one-and-only public method? I'd even drop the `Option`, having duplicate `&'static str`s doesn't cost anything (or at very least doesn't cost more than having an extra `Option<&'static str>`, rustc will dedupe them and we save ourselves the extra branching in code).

```suggestion
	pub fn register_subscription<F>(
		&mut self,
		subscribe_method_name: &'static str,
		notification_method_name: &'static str,
		unsubscribe_method_name: &'static str,
		callback: F,
	) -> Result<(), Error>
```

maciejhirsz (approved):
Some open questions, but looks good to me otherwise!

utils/src/server/rpc_module.rs
niklasad1: ty, can't we handle it in `RpcModule::merge`?

utils/src/server/rpc_module.rs
niklasad1: that should work but rather annoying API to use without the macros.

utils/src/server/rpc_module.rs
maciejhirsz: You can, but you'd have to move the hashset inside `Methods` because that's what merge operates on.

utils/src/server/rpc_module.rs
maciejhirsz: If you think it's better to have separate methods I'd still drop the `Option` and merge it with `register_subscription_with_custom_notif`, and then alias `register_subscription` so that it calls `register_subscription_with_custom_notif` with a copy of the str, because that's also virtually free and you get rid of the branch.

utils/src/server/rpc_module.rs
niklasad1: let's keep it simple I removed all of this but kept the check in the proc macros for now, ok?

proc-macros/src/rpc_macro.rs
```diff
@@ -95,6 +97,9 @@ impl RpcMethod {
 #[derive(Debug, Clone)]
 pub struct RpcSubscription {
 	pub name: String,
+	/// By default the server will send out subscriptions `method` in `name` above
```
niklasad1, line 100:
```suggestion
	/// By default the server will send out subscriptions using `method` in `name` above
```

utils/src/server/rpc_module.rs
niklasad1: I did what you wished for let's see what @dvdplm says :)

utils/src/server/rpc_module.rs
dvdplm: Good pt @maciejhirsz. I'm ok making this a "sharp" feature, but we need to document it properly. "NOTE: the author is expected to ensure the uniqueness of the override across **all** `RpcModule`s used on the server." or some such.

dvdplm (commented):
Sorry for being a trouble maker wrt naming; I'm concerned about mixing up the concept of "notification" with the spec's definition.
We need to update the macro docs with a clear explanation of what this change is and when it's useful (stressing it should never be used for new code; people should use the subscription ID to track things).

tests/tests/proc_macros.rs
```diff
@@ -312,7 +312,7 @@ async fn multiple_blocking_calls_overlap() {
 
 #[tokio::test]
 async fn subscriptions_do_not_work_for_http_servers() {
-	let htserver = HttpServerBuilder::default().build("127.0.0.1:0".parse().unwrap()).unwrap();
+	let htserver = HttpServerBuilder::default().build("127.0.0.1:0").unwrap();
```
TarikGul, line 315: 👍

TarikGul (approved):
lgtm, fwiw I agree with David that some clearer naming could be a +1.

maciejhirsz:
@dvdplm 

> I'm concerned about mixing up the concept of "notification" with the spec's definition.

There is nothing wrong with us using the term "Notification", and the spec explicitly allows / remains agnostic about server acting as a client:

> The Client is defined as the origin of Request objects and the handler of Response objects.
> The Server is defined as the origin of Response objects and the handler of Request objects.
>
> **One implementation of this specification could easily fill both of those roles, even at the same time, to other different clients or the same client. This specification does not address that layer of complexity.**

We are using JSON-RPC notifications in a very specific fashion, but I don't see how that stops them from being notifications. TCP packets don't stop being TCP packets if they carry HTTP paylods. Everything we do with subscriptions is fully compliant with the JSON-RPC spec, we don't change anything about the spec, nor do we add anything to the spec, subscriptions are just a meta protocol bolted on top of it.

niklasad1:
> lgtm, fwiw I agree with David that some clearer naming could be a +1.

Technically the naming is accurate because `subscriptions` is `notifications` according the the spec but `subscription payload` works too if that's more understandable.... 

Good to get your review still to get some eyes outside the `jsonrpc bubble` :)

TarikGul:
> Technically the naming is accurate because subscriptions is a notification according the the spec which has a special field subscription passed in the params where the subscription ID goes but subscription payload works too if that's more understandable....

@niklasad1 Just read @maciejhirsz comments as well. Totally understand where the convention is coming from. Happy to see it kept as is as well. With a third glance over I can see it makes sense to have `notif`. :)

utils/src/server/rpc_module.rs
niklasad1: Note, we have a check for this in the `proc macros` where each trait ensures uniqueness for each name.

Do you want me to remove it to somewhat consistent?

dvdplm:
@maciejhirsz You are right and I am wrong; the spec is indeed agnostic just the way you quote it. My concern was we'd introduce a naming convention that conflicts with a specific meaning defined in the spec. Instead there is no such conflict.

I'll go over my naming suggestions again and see if I know feel differently. Ty!

utils/src/server/rpc_module.rs
```diff
@@ -516,9 +516,21 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
...
+	/// # Arguments
+	///
+	/// * `subscription_method_name` - name of the method to call to initiate a subscription
+	/// * `notif_method_name` - name of method to be used in the subscription payload (technically a JSON-RPC notification)
+	/// * `unsubscription_method` - name of the method to call to terminate a subscription
+	/// *  `callback` - A callback to invoke on each subscrption, it takes three parameters
+	///     - [`Params`]: JSON-RPC parameters in the subscription call.
 	///     - [`SubscriptionSink`]: A sink to send messages to the subscriber.
 	///     - Context: Any type that can be embedded into the [`RpcModule`].
 	///
```
dvdplm, line 536:
```suggestion
	/// Register a new RPC subscription that invokes s callback on every subscription call.
	///
	/// This method ensures that the `subscription_method_name` and `unsubscription_method_name` are unique.
	/// The `notif_method_name` argument sets the content of the `method` field in the JSON document that
	/// the server sends back to the client. The uniqueness of this value is not machine checked and it's up to
	/// the user to ensure it is not used in any other [`RpcModule`] used in the server.
	///
	/// # Arguments
	///
	/// * `subscription_method_name` - name of the method to call to initiate a subscription
	/// * `notif_method_name` - name of method to be used in the subscription payload (technically a JSON-RPC notification)
	/// * `unsubscription_method` - name of the method to call to terminate a subscription
	/// *  `callback` - A callback to invoke on each subscription; it takes three parameters:
	///     - [`Params`]: JSON-RPC parameters in the subscription call.
	///     - [`SubscriptionSink`]: A sink to send messages to the subscriber.
	///     - Context: Any type that can be embedded into the [`RpcModule`].
	///
```

utils/src/server/rpc_module.rs
```diff
@@ -529,7 +541,7 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 	/// use jsonrpsee_utils::server::rpc_module::RpcModule;
 	///
 	/// let mut ctx = RpcModule::new(99_usize);
-	/// ctx.register_subscription("sub", "unsub", |params, mut sink, ctx| {
+	/// ctx.register_subscription("sub", "payload_name", "unsub", |params, mut sink, ctx| {
```
dvdplm, line 544:
```suggestion
	/// ctx.register_subscription("sub", "notif_name", "unsub", |params, mut sink, ctx| {
```

dvdplm (approved):

PR #569, opened 2021-11-18
Arrow syntax for overwrites

maciejhirsz:
Only applied to `name = ...`, but you could use `NameMapping` (feel free to name it something better) in aliases via `Bracketed<NameMapping>`.


niklasad1 (approved):
great, thanks

PR #570, opened 2021-11-21
proc-macros: Support deprecated methods for rpc client

lexnv:
Calling a deprecated method of the RPC client should warn
the user at compile-time.

Extract the `#[deprecated]` macro as is while parsing the
RpcMethod, and pass through the macro to the RPC client
rendering.

### Testing

To ensure that the test will fail during compilation,
warnings are denied.

Check that the deprecate macro will generate warnings
just for the methods that are utilized.

### DIFF
--- a/proc-macros/src/render_client.rs
+++ b/proc-macros/src/render_client.rs
@@ -98,9 +98,12 @@ impl RpcDescription {
 		let parameters = self.encode_params(&method.params, &method.param_kind, &method.signature);
 		// Doc-comment to be associated with the method.
 		let docs = &method.docs;
+		// Mark the method as deprecated, if previously declared as so.
+		let deprecated = &method.deprecated;
 
 		let method = quote! {
 			#docs
+			#deprecated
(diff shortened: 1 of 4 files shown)


maciejhirsz (approved):
Code looks super clean despite proc macros are notorious for being hard to navigate, and tests are spot on so this is definitely a pass from me! :)

I'm looking over the issue now and I'm a bit confused about what the intended deprecation message should be there (as in, should have a log on the server, or send something in the JSON response?), but this is definitely a valid interpretation! If there is anything we want to change we can do that outside of this PR.

Cheers, and have a great Sunday :)

niklasad1 (approved):
great, thanks

//cc @dvdplm tip plz :)

dvdplm:
@lexnv Good stuff here, tyvm. Can you please edit the description of the PR adding your polkadot address on a separate line like so `polkadot address: <SS58 Address>` and I'll invoke the tip-bot as a small sign of appreciation! :)

PR #576, opened 2021-11-23
Middleware for metrics

maciejhirsz:
Still a bunch of loose ends: 

+ [x] Add a method to set the metrics on the server builder.
+ [x] HTTP server side.
+ [x] TODO doc comments.
+ [x] Test / examples.

Other than that it should be functional and is backwards compatible. I needed to add an extra boxed future in async method call pipeline that I may be able to remove down the line (depending on how much I want to mess with the `FutureDriver`).

### DIFF
--- a/examples/middleware.rs
+++ b/examples/middleware.rs
@@ -1,0 +1,94 @@
+// Copyright 2019-2021 Parity Technologies (UK) Ltd.
+//
+// Permission is hereby granted, free of charge, to any
+// person obtaining a copy of this software and associated
+// documentation files (the "Software"), to deal in the
+// Software without restriction, including without
+// limitation the rights to use, copy, modify, merge,
+// publish, distribute, sublicense, and/or sell copies of
+// the Software, and to permit persons to whom the Software
+// is furnished to do so, subject to the following
+// conditions:
+//
+// The above copyright notice and this permission notice
+// shall be included in all copies or substantial portions
+// of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF
+// ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED
+// TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+// PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT
+// SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+// CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR
+// IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+// DEALINGS IN THE SOFTWARE.
+
+use jsonrpsee::{
+	types::traits::Client,
+	utils::server::middleware,
+	ws_client::WsClientBuilder,
+	ws_server::{RpcModule, WsServerBuilder},
+};
+use std::net::SocketAddr;
+use std::sync::atomic;
+
+#[derive(Default)]
+struct ManInTheMiddle {
+	when: atomic::AtomicU64,
+}
+
+impl Clone for ManInTheMiddle {
+	fn clone(&self) -> Self {
+		ManInTheMiddle { when: atomic::AtomicU64::new(self.when.load(atomic::Ordering::SeqCst)) }
+	}
+}
+
+impl middleware::Middleware for ManInTheMiddle {
+	type Instant = u64;
+	fn on_request(&self) -> Self::Instant {
+		self.when.fetch_add(1, atomic::Ordering::SeqCst)
+	}
+
+	fn on_call(&self, name: &str) {
+		println!("They called '{}'", name);
+	}
+
+	fn on_result(&self, name: &str, succeess: bool, started_at: Self::Instant) {
+		println!("call={}, worked? {}, when? {}", name, succeess, started_at);
+	}
+
+	fn on_response(&self, started_at: Self::Instant) {
+		println!("Response started_at={}", started_at);
+	}
+}
+
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+	tracing_subscriber::FmtSubscriber::builder()
+		.with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
+		.try_init()
+		.expect("setting default subscriber failed");
+
+	let addr = run_server().await?;
+	let url = format!("ws://{}", addr);
+
+	let client = WsClientBuilder::default().build(&url).await?;
+	let response: String = client.request("say_hello", None).await?;
+	tracing::info!("response: {:?}", response);
+	// TODO: This prints `They called 'blabla'` but nothing more. I expected the `on_response` callback to be called too?
+	let _response: Result<String, _> = client.request("blabla", None).await;
+	let _ = client.request::<String>("say_hello", None).await?;
+
+	Ok(())
+}
+
+async fn run_server() -> anyhow::Result<SocketAddr> {
+	let m = ManInTheMiddle::default();
+	let server = WsServerBuilder::with_middleware(m).build("127.0.0.1:0").await?;
+	let mut module = RpcModule::new(());
+	module.register_method("say_hello", |_, _| Ok("lo"))?;
+	let addr = server.local_addr()?;
+	server.start(module)?;
+	Ok(addr)
+}
--- a/examples/middleware.rs
+++ b/examples/middleware.rs
@@ -31,23 +31,16 @@ use jsonrpsee::{
 	ws_server::{RpcModule, WsServerBuilder},
 };
 use std::net::SocketAddr;
-use std::sync::atomic;
+use std::time::Instant;
 
-#[derive(Default)]
-struct ManInTheMiddle {
-	when: atomic::AtomicU64,
-}
-
-impl Clone for ManInTheMiddle {
-	fn clone(&self) -> Self {
-		ManInTheMiddle { when: atomic::AtomicU64::new(self.when.load(atomic::Ordering::SeqCst)) }
-	}
-}
+#[derive(Default, Clone)]
+struct ManInTheMiddle;
 
 impl middleware::Middleware for ManInTheMiddle {
-	type Instant = u64;
+	type Instant = Instant;
+
 	fn on_request(&self) -> Self::Instant {
-		self.when.fetch_add(1, atomic::Ordering::SeqCst)
+		Instant::now()
 	}
 
 	fn on_call(&self, name: &str) {
@@ -55,11 +48,12 @@ impl middleware::Middleware for ManInTheMiddle {
 	}
 
 	fn on_result(&self, name: &str, succeess: bool, started_at: Self::Instant) {
-		println!("call={}, worked? {}, when? {}", name, succeess, started_at);
+		// println!("call={}, worked? {}, when? {}", name, succeess, started_at);
+		println!("call={}, worked? {}, duration {:?}", name, succeess, started_at.elapsed());
 	}
 
 	fn on_response(&self, started_at: Self::Instant) {
-		println!("Response started_at={}", started_at);
+		println!("Response duration {:?}", started_at.elapsed());
 	}
 }
 
@@ -75,7 +69,7 @@ async fn main() -> anyhow::Result<()> {
 
 	let client = WsClientBuilder::default().build(&url).await?;
 	let response: String = client.request("say_hello", None).await?;
-	tracing::info!("response: {:?}", response);
+	println!("response: {:?}", response);
 	// TODO: This prints `They called 'blabla'` but nothing more. I expected the `on_response` callback to be called too?
 	let _response: Result<String, _> = client.request("blabla", None).await;
 	let _ = client.request::<String>("say_hello", None).await?;
--- a/jsonrpsee/src/lib.rs
+++ b/jsonrpsee/src/lib.rs
@@ -76,7 +76,6 @@ pub use jsonrpsee_types as types;
 #[cfg(any(feature = "http-server", feature = "ws-server"))]
 pub use jsonrpsee_utils::server::rpc_module::{RpcModule, SubscriptionSink};
 
-/// TODO: (dp) any reason not to export this? narrow the scope to `jsonrpsee_utils::server`?
 #[cfg(any(feature = "http-server", feature = "ws-server"))]
 pub use jsonrpsee_utils as utils;
 
--- a/jsonrpsee/src/lib.rs
+++ b/jsonrpsee/src/lib.rs
@@ -76,6 +76,10 @@ pub use jsonrpsee_types as types;
 #[cfg(any(feature = "http-server", feature = "ws-server"))]
 pub use jsonrpsee_utils::server::rpc_module::{RpcModule, SubscriptionSink};
 
+/// TODO: (dp) any reason not to export this? narrow the scope to `jsonrpsee_utils::server`?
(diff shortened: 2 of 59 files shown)


utils/src/server/middleware.rs
```diff
@@ -0,0 +1,85 @@
...
+// SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+// CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR
+// IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+// DEALINGS IN THE SOFTWARE.
+
+//! TODO
+
+/// TODO
+pub trait Middleware: Send + Sync + Clone + 'static {
```
dvdplm, line 30: Maybe this should live in `jsonrpsee::types::traits`?

utils/src/server/middleware.rs
```diff
@@ -0,0 +1,85 @@
...
+	type Instant: Send + Copy;
+
+	/// Called when a new JSON-RPC comes to the server.
+	fn on_request(&self) -> Self::Instant;
+
+	/// Called on each JSON-RPC method call, batch requests will trigger `on_call` multiple times.
+	fn on_call(&self, name: &str);
+
+	/// Called on each JSON-RPC method completion, batch requests will trigger `on_result` multiple times.
+	fn on_result(&self, name: &str, succeess: bool, started_at: Self::Instant);
```
dvdplm, line 42:
```suggestion
	fn on_result(&self, name: &str, success: bool, started_at: Self::Instant);
```

utils/src/server/middleware.rs
```diff
@@ -0,0 +1,85 @@
...
+//! TODO
+
+/// TODO
+pub trait Middleware: Send + Sync + Clone + 'static {
+	/// Intended to carry timestamp of a request, for example `std::time::Instant`. How the middleware
+	/// measures time, if at all, is entirely up to the implementation.
+	type Instant: Send + Copy;
+
+	/// Called when a new JSON-RPC comes to the server.
+	fn on_request(&self) -> Self::Instant;
```
niklasad1, line 36: I don't understand the difference between `on_request` and `on_call`.

is `on_request`  intended for `JSON-RPC notifications and method call`? When is this useful to have?

Compared to `on_call`, `on_notif`?

utils/src/server/helpers.rs
```diff
@@ -82,67 +81,108 @@ impl<'a> io::Write for &'a mut BoundedWriter {
...
-			} else {
-				return send_error(id, tx, ErrorCode::InternalError.into());
+impl MethodSink {
+	/// Create a new `MethodSink` with unlimited response size
+	pub fn new(tx: mpsc::UnboundedSender<String>) -> Self {
+		MethodSink { tx, max_response_size: u32::MAX }
+	}
+
+	/// Create a new `MethodSink` with a limited response size
+	pub fn new_with_limit(tx: mpsc::UnboundedSender<String>, max_response_size: u32) -> Self {
```
niklasad1, line 100: nice :+1

utils/src/server/middleware.rs
maciejhirsz: This is mirroring `jsonrpc` which I believe treats batch arrays as single requests. It does logically make sense to differentiate the two, and it makes batches easier to manage because you don't have to get a timestamp for all calls in the batch (which would be the same). All that said, I'm happy with changes to labeling here.

dvdplm: As I understood it:
`on_request`: all incoming requests, regardless of what kind (call, notification, subscription, unsubscription…); for instance a batch request would count as one request
`on_call`: actual method calls; each call in a batch would count as one call

niklasad1: that makes sense, the naming is alright but the documentation could be a little more detailed.

batch requests can contain notifications but that isn't a problem here right?

utils/src/server/middleware.rs
```diff
@@ -0,0 +1,85 @@
...
+	fn on_request(&self) -> Self::Instant {}
+
+	fn on_call(&self, _name: &str) {}
+
+	fn on_result(&self, _name: &str, _succeess: bool, _started_at: Self::Instant) {}
+
+	fn on_response(&self, _started_at: Self::Instant) {}
+}
+
+impl<A, B> Middleware for (A, B)
```
niklasad1, line 60: dq: why (A, B) here and not just A?

dvdplm: I think this is how you pile up moar middlewarez, `(M1, M2)` and then `(M1, (M2, M3))`, `(M1, (M2, (M3, M4)))` etc

maciejhirsz: This is a generic impl which lets you stack multiple middlewares if you need to use more than one. Only need a tuple of two to enable arbitrary number, e.g.: if you have 3 you can go `(mw1, (mw2, mw3))`.

examples/middleware.rs
```diff
@@ -0,0 +1,94 @@
...
+}
+
+impl middleware::Middleware for ManInTheMiddle {
+	type Instant = u64;
+	fn on_request(&self) -> Self::Instant {
+		self.when.fetch_add(1, atomic::Ordering::SeqCst)
+	}
+
+	fn on_call(&self, name: &str) {
+		println!("They called '{}'", name);
```
niklasad1, line 54: should this be `tracing::info`?!

utils/src/server/middleware.rs
niklasad1: ah, I see.

examples/middleware.rs
dvdplm: The example should be a bit more polished for sure. In general I think examples can use `println!` so casual readers can skip running with `RUST_LOG=debug`.

utils/src/server/middleware.rs
maciejhirsz: > batch requests can contain notifications but that isn't a problem here right?

We don't support client->server notifications yet so that's not really a thing :upside_down_face:. The notifications sent from the server as a part of subscription aren't tracked by middleware, there isn't really a way to make it so without sticking middleware into modules, and that creates a huge mess.

ws-server/src/server.rs
```diff
@@ -332,70 +345,93 @@ async fn background_task(
...
+										let success = fut.await;
+										middleware.on_result(name, success, request_start);
+									}),
+									None => None,
+								}
+							}))
+							.await;
+
+							rx_batch.close();
+							let results = collect_batch_response(rx_batch).await;
```
niklasad1, line 414: why are you not calling `on_response` if batch was successfully executed?

maciejhirsz: Good catch!

utils/src/server/middleware.rs
niklasad1: yupp, right we don't even try to decode `batches` with notifications good point.

niklasad1 (commented):
I like this, it's just matter of adding the remaining code

jsonrpsee/src/lib.rs
```diff
@@ -76,6 +76,10 @@ pub use jsonrpsee_types as types;
 #[cfg(any(feature = "http-server", feature = "ws-server"))]
 pub use jsonrpsee_utils::server::rpc_module::{RpcModule, SubscriptionSink};
 
+/// TODO: (dp) any reason not to export this? narrow the scope to `jsonrpsee_utils::server`?
+#[cfg(any(feature = "http-server", feature = "ws-server"))]
```
niklasad1, line 80: you can remove these feature guards, the `utils` by itself have conditional compilation that should be sufficient but I think you have to change `utils` to be a non-optional dependency for that to work.

examples/middleware.rs
```diff
@@ -0,0 +1,94 @@
...
+		.with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
+		.try_init()
+		.expect("setting default subscriber failed");
+
+	let addr = run_server().await?;
+	let url = format!("ws://{}", addr);
+
+	let client = WsClientBuilder::default().build(&url).await?;
+	let response: String = client.request("say_hello", None).await?;
+	tracing::info!("response: {:?}", response);
```
jsdw, line 78: Nit: since this is an example and to be consistent with the above; `println!` instead?

utils/src/server/middleware.rs
```diff
@@ -0,0 +1,85 @@
...
+// IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+// DEALINGS IN THE SOFTWARE.
+
+//! TODO
+
+/// TODO
+pub trait Middleware: Default + Send + Sync + Clone + 'static {
+	/// Intended to carry timestamp of a request, for example `std::time::Instant`. How the middleware
+	/// measures time, if at all, is entirely up to the implementation.
+	type Instant: Send + Copy;
```
jsdw, line 33: I wonder whether `Instant` is the right name; I'd rather assume very little about how it will be used and call it something like `OnRequest` perhaps. In the example, for instance, it's used as a counter, which seems like a perfectly reasonable use case too :)

I also wonder about allowing `Clone` rather than `Copy`; would this have any performance implications if the type is copyable anyway? It would perhaps open up some additional use cases. (on the other hand, we can relax restrictions later, and there aren't many use cases for an `on_request` that knows nothing about the request anyway, so I don't mind either way really!)

utils/src/server/middleware.rs
```diff
@@ -0,0 +1,85 @@
...
+
+	fn on_call(&self, _name: &str) {}
+
+	fn on_result(&self, _name: &str, _succeess: bool, _started_at: Self::Instant) {}
+
+	fn on_response(&self, _started_at: Self::Instant) {}
+}
+
+impl<A, B> Middleware for (A, B)
+where
```
jsdw, line 61: Is it worth a macro impl to cover tuples up to some certain size (just for better ergonomics fpr >2 middlewares)?

utils/src/server/middleware.rs
dvdplm: Naming this one is hard, like really hard. Maybe it should be `X` to communicate its "whatever"-nature... (only half-joking). 
Or we could take the opposite stance and remove it and force `std::time::Instant` – after all we're not committing to this being the final shape of middleware; all we need is enough to cover substrate's needs.

utils/src/server/middleware.rs
dvdplm: Probably, but not until we know we need it imo. :)

utils/src/server/middleware.rs
dvdplm: How about `Track`? `type Track = std::time::Instant` reads alright; `type Track = usize` is ok-ish. It probably violates all kinds of API guidelines, but this one is a bit special. :/

ws-server/src/server.rs
```diff
@@ -520,7 +520,30 @@ impl Builder {
...
+	///         Instant::now()
+	///     }
+	///
+	///     fn on_result(&self, method: &str, success: bool, started_at: Instant) {
+	///         println!("Call to '{}' took {:?}", started_at.elapsed());
+	///     }
+	/// }
+	///
+	/// let builder = WsServerBuilder::with_middleware(());
+	/// builder.set_allowed_origins(["https://example.com"]);
```
dvdplm: Wouldn't `let builder = WsServerBuilder::with_middleware(()).set_allowed_origins(["https://example.com"]);` be more idiomatic?

jsdw:
A strange thing from me; on my M1 mac, when I run `cargo clean` and then `cargo check` in the root folder, I run into:

```
% cargo check
    Checking jsonrpsee-types v0.5.1 (/Users/james/Work/jsonrpsee/types)
    Checking jsonrpsee-utils v0.5.1 (/Users/james/Work/jsonrpsee/utils)
    Checking jsonrpsee-ws-client v0.5.1 (/Users/james/Work/jsonrpsee/ws-client)
    Checking jsonrpsee-ws-server v0.5.1 (/Users/james/Work/jsonrpsee/ws-server)
    Checking jsonrpsee-http-client v0.5.1 (/Users/james/Work/jsonrpsee/http-client)
    Checking jsonrpsee-http-server v0.5.1 (/Users/james/Work/jsonrpsee/http-server)
error[E0433]: failed to resolve: could not find `MissedTickBehavior` in `time`
  --> ws-server/src/future.rs:59:44
   |
59 |         heartbeat.set_missed_tick_behavior(time::MissedTickBehavior::Skip);
   |                                                  ^^^^^^^^^^^^^^^^^^ could not find `MissedTickBehavior` in `time`

error[E0599]: no method named `set_missed_tick_behavior` found for struct `Interval` in the current scope
  --> ws-server/src/future.rs:59:13
   |
59 |         heartbeat.set_missed_tick_behavior(time::MissedTickBehavior::Skip);
   |                   ^^^^^^^^^^^^^^^^^^^^^^^^ method not found in `Interval`

Some errors have detailed explanations: E0433, E0599.
For more information about an error, try `rustc --explain E0433`.
error: could not compile `jsonrpsee-ws-server` due to 2 previous errors
``` 

Does anybody else run into this?

niklasad1:
> Edit 3: It looks like tokio 1.8.0 is the first to contain the MissedTickBehaviour stuff.

Let's pin `tokio to 1.8` then :)

utils/src/server/helpers.rs
```diff
@@ -82,67 +81,108 @@ impl<'a> io::Write for &'a mut BoundedWriter {
...
+				if err.is_io() {
+					let data = to_json_raw_value(&format!("Exceeded max limit {}", self.max_response_size)).ok();
+					let err = ErrorObject {
+						code: ErrorCode::ServerError(OVERSIZED_RESPONSE_CODE),
+						message: OVERSIZED_RESPONSE_MSG.into(),
+						data: data.as_deref(),
+					};
+					return self.send_error(id, err);
+				} else {
+					return self.send_error(id, ErrorCode::InternalError.into());
```
jsdw, line 127: I assume that the max size error is an error based on user input being too large?

I wonder whether the `tracing::error!("Error serializing response: {:?}", err);` line should instead be above this InternalError bit, so we only get output for the unexpected errors?

ws-server/src/server.rs
```diff
@@ -307,102 +324,135 @@ async fn background_task(
...
+				sink.send_error(Id::Null, ErrorCode::ParseError.into());
+			}
 		}
-	}
+	};
+
+	middleware.on_disconnect();
 
 	// Drive all running methods to completion
 	method_executors.await;
```
jsdw, line 453: Just a note that we should perhaps add a comment or something to make it clear to our future selves that we souldn't return early before this await runs?

jsdw (approved):
Looking great to me :) Love the examples and doc comments. Just a couple of small notes (the mains things being ensuring tokio ^1.8 is used, and a comment about early returning before that `await` if that seems sensible)

utils/src/server/helpers.rs
niklasad1: It means that `output` from an executed call exceeded the max limit but the "request" itself was below the limit.
Thus, it depends what you mean with the user but the `registered callback` created a response bigger than the max limit.

Yes, or downgrade the error above to warn or something but the `InternalError` "should" be infallible/bug IIRC.

ws-server/src/server.rs
```diff
@@ -307,102 +324,135 @@ async fn background_task(
...
+									}
+								}
+							}))
+							.await;
+
+							rx_batch.close();
+							let results = collect_batch_response(rx_batch).await;
+
+							if let Err(err) = sink.send_raw(results) {
 								tracing::error!("Error sending batch response to the client: {:?}", err)
```
niklasad1, line 427: hmm, if the `send_raw` will only fail once the server is closed, is that why we don't register `on_response` when that fails?

otherwise, we seem to register failed responses too.

niklasad1 (approved):
LGTM, clean

PR #577, opened 2021-11-23
Add a setter for middleware

dvdplm:
…and an example. 

Came about as way to review #576

### DIFF
--- a/examples/middleware.rs
+++ b/examples/middleware.rs
@@ -31,16 +31,23 @@ use jsonrpsee::{
 	ws_server::{RpcModule, WsServerBuilder},
 };
 use std::net::SocketAddr;
+use std::sync::atomic;
 
-#[derive(Clone, Default)]
+#[derive(Default)]
 struct ManInTheMiddle {
-	when: u64,
+	when: atomic::AtomicU64,
+}
+
+impl Clone for ManInTheMiddle {
+	fn clone(&self) -> Self {
+		ManInTheMiddle { when: atomic::AtomicU64::new(self.when.load(atomic::Ordering::SeqCst)) }
+	}
 }
 
 impl middleware::Middleware for ManInTheMiddle {
 	type Instant = u64;
 	fn on_request(&self) -> Self::Instant {
-		self.when + 1
+		self.when.fetch_add(1, atomic::Ordering::SeqCst)
 	}
 
 	fn on_call(&self, name: &str) {
--- a/examples/middleware.rs
+++ b/examples/middleware.rs
@@ -1,0 +1,97 @@
+// Copyright 2019-2021 Parity Technologies (UK) Ltd.
+//
+// Permission is hereby granted, free of charge, to any
+// person obtaining a copy of this software and associated
+// documentation files (the "Software"), to deal in the
+// Software without restriction, including without
+// limitation the rights to use, copy, modify, merge,
+// publish, distribute, sublicense, and/or sell copies of
+// the Software, and to permit persons to whom the Software
+// is furnished to do so, subject to the following
+// conditions:
+//
+// The above copyright notice and this permission notice
+// shall be included in all copies or substantial portions
+// of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF
+// ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED
+// TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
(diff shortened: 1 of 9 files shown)


examples/middleware.rs
```diff
@@ -0,0 +1,97 @@
...
+
+#[derive(Clone, Default)]
+struct ManInTheMiddle {
+	when: u64
+}
+
+impl middleware::Middleware for ManInTheMiddle {
+	type Instant = u64;
+	fn on_request(&self) -> Self::Instant {
+		self.when + 1
```
maciejhirsz, line 43:
```suggestion
		self.when += 1;
		self.when
```

ws-server/src/server.rs
```diff
@@ -56,7 +56,7 @@ const MAX_CONNECTIONS: u64 = 100;
 
 /// A WebSocket JSON RPC server.
 #[derive(Debug)]
-pub struct Server<M = ()> {
+pub struct Server<M> {
```
maciejhirsz, line 59: Why remove the default?

ws-server/src/server.rs
```diff
@@ -598,6 +610,11 @@ impl Builder {
 		self
 	}
 
+	/// TODO: (dp)
+	pub fn with_middleware(mut self, m: M) -> Self {
+		self.middleware = m;
+		self
+	}
```
maciejhirsz, line 617:
```suggestion
	pub fn with_middleware<T>(self, middleware: T) -> Builder<T> {
		Builder {
			settings: self.settings,
			resources: self.resources,
			middleware,
		}
	}
```

maciejhirsz (changes requested):

ws-server/src/server.rs
```diff
@@ -489,13 +489,25 @@ impl Default for Settings {
...
+
+impl<M: Default> Default for Builder<M> {
+	fn default() -> Self {
+		Self {
+			settings: Default::default(),
+			resources: Default::default(),
+			middleware: Default::default(),
+		}
+	}
 }
```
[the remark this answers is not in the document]
dvdplm: `impl Default for Builder<()>` works! Ty!

ws-server/src/server.rs
dvdplm: Ty!

I wonder if we should move away from `Builder::default().set_something().build()` and instead have `Builder::with_middleware(MMM).set_something().build()`/``Builder::new().set_something().build()`. Wdyt? This "default" is increasingly awkward imo.

ws-server/src/server.rs
dvdplm: Was trying to make `Builder::default()` work and this seemed to be in the way, but I don't think this was the problem. OTOH: do we need it?

maciejhirsz: I'd keep it, all it does is if you don't use middleware you can just write `WsServer` instead of `WsServer<()>`.

ws-server/src/server.rs
```diff
@@ -489,13 +489,31 @@ impl Default for Settings {
...
 
-impl Builder {
+impl<M> Builder<M> {
+	/// Build a default server.
+	pub fn new() -> Builder<()> {
+		Default::default()
+	}
+
```
maciejhirsz, line 510:
```suggestion
impl Builder {
	/// Build a default server.
	pub fn new() -> Self {
		Default::default()
	}
}

impl<M> Builder<M> {
```

ws-server/src/server.rs
```diff
@@ -489,13 +489,31 @@ impl Default for Settings {
...
+	/// Build a server with the specified [`Middleware`].
+	pub fn with_middleware(middleware: M) -> Builder<M> {
+		Builder { settings: Default::default(), resources: Default::default(), middleware }
+	}
+
+
```
maciejhirsz, line 516:
```suggestion
```

ws-server/src/server.rs
```diff
@@ -489,13 +489,32 @@ impl Default for Settings {
...
 impl Builder {
+	/// Build a default server.
+	pub fn new() -> Self {
+		Default::default()
+	}
+}
+
+impl<M> Builder<M> {
+	/// Build a server with the specified [`Middleware`].
+	pub fn with_middleware(middleware: M) -> Builder<M> {
```
maciejhirsz, line 514:
```suggestion
	pub fn with_middleware(middleware: M) -> Self {
```

:)

maciejhirsz (approved):
Feel free to merge at will, any kind of polish can be done on the upstream branch too.

PR #578, opened 2021-11-25
fix rpc error: support escaped strings

niklasad1:


jsdw:
Ooooh, I see! owned strings needed for when you can't just copy the message straight from the buffer, got it!

jsdw (approved):

dvdplm (approved):
Great catch!

types/src/v2/error.rs
```diff
@@ -64,7 +65,7 @@ pub struct ErrorObject<'a> {
 	/// Code
 	pub code: ErrorCode,
 	/// Message
-	pub message: &'a str,
+	pub message: Cow<'a, str>,
```
maciejhirsz, line 68:
```suggestion
	#[serde(borrow)]
	pub message: Cow<'a, str>,
```

Otherwise it always deserializes to owned variant.

niklasad1: danke :pray:

maciejhirsz (approved):

TarikGul (approved):

insipx (approved):

PR #580, opened 2021-11-29
extract async client abstraction.

niklasad1:
Let's have a round of reviews for this too, it extracts the `shareable` parts of the `ws client` to an `async client abstration`. 
I have called it `core-client` and it's not used by `HTTP` because it can't be extracted to a `sending` and `receiving` end.
That's we have two different transport traits for a `sender` and `receiver` such that they can be used in different tasks.

We can bikeshed on the naming....

Remaining is to expose the client transports from the wrapper crate `jsonrpsee`

I have tested it in [`subxt here`](https://github.com/paritytech/subxt/pull/341)

The benefit that it would be easy to add other transports such as `IPC`, `web-sys` and so on...


types/src/traits.rs
```diff
@@ -127,3 +127,29 @@ tuple_impls! {
...
 }
+
+/// Transport interface to send data asynchronous.
+#[async_trait]
+/// Transport interface for an asyncronous client.
+pub trait TransportSender: Send + 'static {
+	/// Error.
+	type Error: std::error::Error + Send + Sync;
+
+	/// Send..
```
niklasad1, line 138:
```suggestion
	/// Send.
```

client/core-client/src/lib.rs
```diff
@@ -1,57 +1,26 @@
...
-	manager::RequestManager,
+mod helpers;
+mod manager;
+
+use std::time::Duration;
+
+use futures_channel::{mpsc, oneshot};
+use futures_util::{future::Either, SinkExt, StreamExt};
+use jsonrpsee_types::traits::{
+	Client as ClientT, SubscriptionClient as SubscriptionClientT, TransportReceiver, TransportSender,
```
maciejhirsz, line 9: Should we just rename the trait to `ClientT`?

maciejhirsz (approved):
Looks good to me, one question.

client/core-client/src/lib.rs
niklasad1: yeah, sounds good to me.

client/ws-client/src/lib.rs
```diff
@@ -0,0 +1,178 @@
...
+		self.max_concurrent_requests = max;
+		self
+	}
+
+	/// Set max concurrent notification capacity for each subscription; when the capacity is exceeded the subscription
+	/// will be dropped.
+	///
+	/// You can also prevent the subscription being dropped by calling
+	/// [`Subscription::next()`](crate::types::Subscription) frequently enough such that the buffer capacity doesn't
+	/// exceeds.
```
TarikGul, line 139:
```suggestion
	/// exceed its limit.
```

Im a little confused by `frequently enough`. Does this mean a user should monitor the subscriptions and as it starts to get close to exceeding its capacity to call `[`Subscription::next()`]`, therefore the implementation on what `frequently` is, is based on the user?

niklasad1: Maybe it's poorly phrased but it means that if the server produces items much faster than you would call `Subscription::next()` that might end up with the filling buffer and once it exceeds the configured `max_limit`.

In that scenario you need to either call/poll `Subscription::next` more often or increase the buffer max limit, k?

TarikGul (approved):
lgtm

client/ws-client/src/lib.rs
TarikGul: Ahh, okay yea that totally makes sense. 

> more often or increase the buffer max limit

This is what i assumed, just needed some clarification :). Thanks for the explanation

client/core-client/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "jsonrpsee-core-client"
+version = "0.6.1"
+authors = ["Parity Technologies <admin@parity.io>", "Pierre Krieger <pierre.krieger1708@gmail.com>"]
+description = "JSON-RPC async core client "
+edition = "2018"
+license = "MIT"
+repository = "https://github.com/paritytech/jsonrpsee"
+homepage = "https://github.com/paritytech/jsonrpsee"
+documentation = "https://docs.rs/jsonrpsee-ws-client"
```
dvdplm, line 10:
```suggestion
documentation = "https://docs.rs/jsonrpsee-core-client"
```

client/core-client/src/lib.rs
dvdplm: If we do, we should do it to all of the traits I think.

client/core-client/src/lib.rs
```diff
@@ -1,57 +1,25 @@
...
-	Subscription, SubscriptionClient, SubscriptionKind, SubscriptionMessage,
-};
-use jsonrpsee_core::Error;
-use serde::de::DeserializeOwned;
-use tokio::sync::Mutex;
 
-pub use soketto::handshake::client::Header;
+pub use jsonrpsee_types as types;
 
 /// Wrapper over a [`oneshot::Receiver`](futures::channel::oneshot::Receiver) that reads
```
dvdplm, line 24:
```suggestion
/// Wrapper over a [`oneshot::Receiver`](futures_channel::oneshot::Receiver) that reads
```

…but `oneshot::Receiver` should be in scope without the link I think?

client/ws-client/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+// IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+// DEALINGS IN THE SOFTWARE.
+
+#![warn(missing_debug_implementations, missing_docs, unreachable_pub)]
+
+//! # jsonrpsee-ws-client
+//!
+//! `jsonrpsee-ws-client` is a [JSON RPC](https://www.jsonrpc.org/specification) WebSocket client library that's is built for `async/await`.
+//!
+//! ## Runtime support
```
dvdplm, line 33:
```suggestion
//! ## Async runtime support
```

client/ws-client/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+//!
+//! This library uses `tokio` as the runtime and does not support other kinds of runtimes.
```
dvdplm, line 35:
```suggestion
//! This library uses `tokio` as the runtime and does not support other runtimes.
```

client/ws-client/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+	/// exceeds.
+	///
+	/// **Note**: The actual capacity is `num_senders + max_subscription_capacity`
+	/// because it is passed to [`futures::channel::mpsc::channel`].
+	pub fn max_notifs_per_subscription(mut self, max: usize) -> Self {
+		self.max_notifs_per_subscription = max;
+		self
+	}
+
+	/// Set the max number of redirections to perform until a connection is regarded as failed.
```
dvdplm, line 149:
```suggestion
	/// Set the max number of redirects to follow during the WS handshake until a connection is considered failed.
```

Also: state the default in the docs.

dvdplm (approved):

client/ws-client/src/lib.rs
```diff
@@ -0,0 +1,169 @@
...
+///          .build("wss://localhost:443")
+///          .await
+///          .unwrap();
+///
+///     // use client....
+/// }
+///
+/// ```
+#[derive(Clone, Debug)]
+pub struct WsClientBuilder<'a> {
```
niklasad1, line 70: this crate could be removed but I kept for it may be easier to use than to use the core client and plug in the actual transport.

core/Cargo.toml
```diff
@@ -37,6 +37,14 @@ server = [
 	"tokio",
 ]
 client = ["futures-util"]
+async-client = [
```
niklasad1, line 40: I had to introduce a feature to avoid bring in `tokio` and some other deps for `http`

http-server/Cargo.toml
```diff
@@ -13,8 +13,8 @@ documentation = "https://docs.rs/jsonrpsee-http-server"
 hyper = { version = "0.14.10", features = ["server", "http1", "http2", "tcp"] }
 futures-channel = "0.3.14"
 futures-util = { version = "0.3.14", default-features = false }
-jsonrpsee-types = { path = "../types", version = "0.6.0" }
-jsonrpsee-core = { path = "../core", version = "0.6.0", features = ["server", "http-helpers"] }
+jsonrpsee-types = { path = "../types", version = "0.6.1" }
+jsonrpsee-core= { path = "../core", version = "0.6.1", features = ["server", "http-helpers"] }
```
niklasad1, line 17:
```suggestion
jsonrpsee-core = { path = "../core", version = "0.6.1", features = ["server", "http-helpers"] }
```

niklasad1:
@maciejhirsz @dvdplm @TarikGul 

can take you a quick look on this again after some changes?
if no one objects I'll merge it then :)

client/transport/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+    "futures", 
+    "http", 
+    "tokio", 
+    "tokio-util", 
+    "soketto", 
+    "pin-project",
+    "jsonrpsee-types",
+    "thiserror",
+    "tracing"
+]
```
TarikGul, line 41: new line

dvdplm, line 41:
```suggestion
]

```

PR #581, opened 2021-11-29
An example of adding multiple middlewares.

dvdplm:


examples/multi-middleware.rs
```diff
@@ -0,0 +1,116 @@
...
+
+	fn on_response(&self, started_at: Self::Instant) {
+		println!("[Timings] Response duration {:?}", started_at.elapsed());
+	}
+}
+
+/// Example middleware to keep a watch on the number of total threads started in the system.
+#[derive(Clone)]
+struct ThreadWatcher;
+impl middleware::Middleware for ThreadWatcher {
```
maciejhirsz, line 66:
```suggestion

impl middleware::Middleware for ThreadWatcher {
```

examples/multi-middleware.rs
```diff
@@ -0,0 +1,116 @@
...
+	type Instant = isize;
```
maciejhirsz, line 67: Can there be negative threads?

examples/Cargo.toml
```diff
@@ -26,6 +27,10 @@ path = "middleware_ws.rs"
 name = "middleware_http"
 path = "middleware_http.rs"
 
+[[example]]
+name = "multi-middleware"
+path = "multi-middleware.rs"
```
maciejhirsz, line 32: All other examples use underscores

```suggestion
name = "multi_middleware"
path = "multi_middleware.rs"
```

maciejhirsz (approved):

examples/multi-middleware.rs
dvdplm: Yeah, well, like the subtraction of "end - start" can be negative (these are system threads, so anything can happen while the program executes).

maciejhirsz: Yeah, I guess better safe than sorry, and it's just an example

PR #582, opened 2021-11-29
Move `Middleware` to jsonrpsee-types

dvdplm:
I could be swayed if you don't like this.


maciejhirsz:
Yeah, I think that's fine. Traits aren't technically types, but we already have a traits module there and I'm not going to be a joyless pedant about it :)

Issue #560, opened 2021-11-12
Make clients clonable


slumber:
It's a widespread pattern to implement `Clone` for clients, [jsonrpc](https://docs.rs/jsonrpc-core-client/18.0.0/jsonrpc_core_client/struct.TypedClient.html#impl-Clone), [reqwest](https://docs.rs/reqwest/0.11.6/reqwest/struct.Client.html#impl-Clone), [actix_web](https://docs.rs/actix-web/3.3.2/actix_web/client/struct.Client.html#impl-Clone), etc. Makes it easier to pass the client around.

There's another strong argument towards this: if you wrap your client into `Arc`, you lose all the generated traits implementations from jsonrpsee proc macros.

niklasad1:
> It's a widespread pattern to implement Clone for clients, jsonrpc, reqwest, actix_web, etc. Makes it easier to pass the client around.

We have been reasoned about it as follows: https://github.com/paritytech/jsonrpsee/pull/243#discussion_r594153898, to make it explicit for users `one client (WS) == one background task` then users can share it by a `Mutex, Arc, or similar` and have to deal with this logic because we didn't want to have complicated logic such as restarting. If this is super annoying I guess we could change it

> There's another strong argument towards this: if you wrap your client into Arc, you lose all the generated traits implementations from jsonrpsee proc macros.

You could use `Arc::Deref` and it should work AFAIK such as:

```RpcClient::subscribe_storage(&*client, None).await.unwrap();```?

**Comment by slumber** [comment]
> We have been reasoned about it as follows: [#243 (comment)](https://github.com/paritytech/jsonrpsee/pull/243#discussion_r594153898), to make it explicit for users `one client (WS) == one background task` then users can share it by a `Mutex, Arc, or similar` and have to deal with this logic because we didn't want to have complicated logic such as restarting. If this is super annoying I guess we could change it but not a restart API.
> 

Totally makes sense for websocket client, but it's different in nature from http, there's nothing wrong with cloning http client since we can think of it as being stateless. Also `reqwest` actually has a pool of connections which makes it even better.

> You could use `Arc::Deref` and it should work AFAIK such as:
> 
> `RpcClient::subscribe_storage(&*client, None).await.unwrap();`?

I mean suppose you have a struct
```rust
struct MyClientWrapper<C: MyRpcClient + Clone>  {
     client: C,
}
```
You can't pass `Arc<HttpClient>` because it doesn't automatically derive the trait.
[...]

PR #583, opened 2021-11-29
fix(http client): impl Clone

niklasad1:
Closing #560 

Heads up: request limit is shared by the client when it's cloned.

### DIFF
--- a/http-client/src/client.rs
+++ b/http-client/src/client.rs
@@ -33,7 +33,7 @@ use crate::types::{
 use async_trait::async_trait;
 use fnv::FnvHashMap;
 use serde::de::DeserializeOwned;
-use std::time::Duration;
+use std::{sync::Arc, time::Duration};
 
 /// Http Client Builder.
 #[derive(Debug)]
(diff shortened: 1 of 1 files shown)


insipx (approved):

TarikGul (approved):

http-client/src/client.rs
```diff
@@ -75,7 +75,7 @@ impl HttpClientBuilder {
 			.map_err(|e| Error::Transport(e.into()))?;
 		Ok(HttpClient {
 			transport,
-			id_manager: RequestIdManager::new(self.max_concurrent_requests),
+			id_manager: Arc::new(RequestIdManager::new(self.max_concurrent_requests)),
```
dvdplm, line 78: dq: could this be a `Box` too?

dvdplm: Maybe with:

```
impl Clone for RequestIdManager {
	fn clone(&self) -> Self {
		RequestIdManager {
			current_pending: self.current_pending.clone(),
			max_concurrent_requests: self.max_concurrent_requests.clone(),
			current_id: AtomicU64::new(self.current_id.load(std::sync::atomic::Ordering::SeqCst)),
		}
	}
}
```

niklasad1: Because we rely on the strong count on `current_pending` for how many concurrent requests cloning `current_pending` it increase the `strong count` and it will not work properly.

Thus, that's why I used  `Arc<RequestIdManager>`.

Why is `Box` better?

dvdplm: > Box only works when RequestIdManager is clone right?! I don't want it to be Clone because the reason ☝️

Ah, right, ok, I get it now. 

Yes, `RequestIdManager` would have to be `Clone` (and if it were you wouldn't need nor `Box` or `Arc`).

PR #588, opened 2021-12-02
rpc module: add `call_and_subscribe`

niklasad1:
Needed to support `in memory queries` in substrate with "optional subscriptions`, [example here](https://github.com/paritytech/substrate/blob/3e96b8ec9c863422889245e73516551cc56127c5/client/service/src/lib.rs#L88-#L99)

This might be useful elsewhere too but I think we should refactor this in `v0.7` many similar functions there now.

It won't work unless we expose `receiver` of the stream which this does.

### DIFF
--- a/types/src/v2/response.rs
+++ b/types/src/v2/response.rs
@@ -45,6 +45,13 @@ pub struct Response<'a, T> {
 	pub id: Id<'a>,
 }
 
+impl<'a, T> Response<'a, T> {
+	/// Create a new `Response`.
+	pub fn new(result: T, id: Id<'a>) -> Response<'a, T> {
+		Response { jsonrpc: TwoPointZero, result, id }
+	}
+}
+
 /// Return value for subscriptions.
 #[derive(Serialize, Deserialize, Debug)]
 pub struct SubscriptionPayload<T> {
--- a/types/src/v2/response.rs
+++ b/types/src/v2/response.rs
@@ -46,7 +46,7 @@ pub struct Response<'a, T> {
 }
 
 impl<'a, T> Response<'a, T> {
-	/// Create a new `Response`.
+	/// Create a new [`Response`].
 	pub fn new(result: T, id: Id<'a>) -> Response<'a, T> {
 		Response { jsonrpc: TwoPointZero, result, id }
 	}
--- a/utils/src/server/rpc_module.rs
+++ b/utils/src/server/rpc_module.rs
@@ -369,7 +369,15 @@ impl Methods {
 		rx.next().await
 	}
 
-	/// Perform a method call and receive further subscriptions.
+	/// Perform a "in memory JSON-RPC method call" and receive further subscriptions.
+	/// This is useful if you want to support both `method calls` and `subscriptions`
+	/// in the same API.
+	///
+	/// There are better variants than this method if you only want
+	/// method calls or subscriptions.
+	///
+	/// See [`RpcModule::test_subscription`] and [`RpcModule::call_with`] for
+	/// for further documentation.
 	///
 	/// Returns a response to actual method call and a stream to process
 	/// futher subscriptions if a subcription was registered on the call.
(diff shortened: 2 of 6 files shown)


maciejhirsz (approved):

types/src/v2/response.rs
```diff
@@ -45,6 +45,13 @@ pub struct Response<'a, T> {
 	pub id: Id<'a>,
 }
 
+impl<'a, T> Response<'a, T> {
+	/// Create a new `Response`.
```
dvdplm, line 49:
```suggestion
	/// Create a new [`Response`].
```

Same for the others.

utils/src/server/rpc_module.rs
```diff
@@ -374,13 +369,56 @@ impl Methods {
...
+	///     use jsonrpsee::RpcModule;
+	///     use jsonrpsee::types::{
+	///         EmptyParams,
+	///         v2::{Response, SubscriptionResponse}
+	///     };
+	///     use futures_util::StreamExt;
+	///
+	///     let mut module = RpcModule::new(());
+	///     module.register_subscription("hi", "hi", "goodbye", |_, mut sink, _| {
+	///     sink.send(&"one answer").unwrap();
```
dvdplm, line 389: indentation is off one level I believe

utils/src/server/rpc_module.rs
```diff
@@ -374,13 +369,56 @@ impl Methods {
 		rx.next().await
 	}
 
+	/// Perform a method call and receive further subscriptions.
+	///
+	/// Returns a response to actual method call and a stream to process
+	/// futher subscriptions if a subcription was registered on the call.
```
dvdplm, line 375:
```suggestion
	/// Returns a response to the actual method call and a stream to process
	/// further notifications if a subscription was registered by the call.
```

utils/src/server/rpc_module.rs
```diff
@@ -374,13 +369,56 @@ impl Methods {
...
+	///
```
dvdplm, line 376: Worth mentioning what the intended use case for this is? "In-memory RPC queries, e.g. in tests"?

dvdplm (approved):
Not 100% sold on this. Couldn't the code in substrate (and cumulus I believe) could be refactored to use `test_subscription` instead?

Not going to make a fuss, but the API is growing larger than I'd like.

niklasad1:
> Not 100% sold on this. Couldn't the code in substrate (and cumulus I believe) could be refactored to use test_subscription instead?

Nah, it would panic if the response wasn't a subscription response but I get your point it's getting messy.

niklasad1:
@dvdplm can you take another look at this?

I think we should remove `call` in public API, rename `call_with` to `call` and perhaps unify the inner the implementation.

utils/src/server/rpc_module.rs
```diff
@@ -369,10 +369,18 @@ impl Methods {
 		rx.next().await
 	}
 
-	/// Perform a method call and receive further subscriptions.
+	/// Perform a "in memory JSON-RPC method call" and receive further subscriptions.
+	/// This is useful if you want to support both `method calls` and `subscriptions`
+	/// in the same API.
```
dvdplm, line 374:
```suggestion
	/// Perform an in memory JSON-RPC method call or subscription without passing through the http/ws layer. 
	/// This is useful if you want to support both `method calls` and `subscriptions`
	/// in the same testing API.
```

utils/src/server/rpc_module.rs
```diff
@@ -369,10 +369,18 @@ impl Methods {
...
 	///
-	/// Returns a response to actual method call and a stream to process
-	/// futher subscriptions if a subcription was registered on the call.
+	/// There are better variants than this method if you only want
+	/// method calls or subscriptions.
```
dvdplm, line 377:
```suggestion
	/// There are better variants than this method if you only want
	/// method calls or only subscriptions.
```

utils/src/server/rpc_module.rs
```diff
@@ -369,10 +369,18 @@ impl Methods {
...
+	///
+	/// See [`Methods::test_subscription`] and [`Methods::call_with`] for
+	/// for further documentation.
+	///
+	/// Returns a response to the actual method call and a stream to process
+	/// further notifications if a subscription was registered by the call.
```
dvdplm, line 383:
```suggestion
	/// Returns a response to the actual method call and a stream to process
	/// for further notifications if a subscription was registered by the call.
```

dvdplm (commented):
A few changes to the docs.

dvdplm:
> I think we should remove call in public API, rename call_with to call and perhaps unify the inner the implementation.

I agree and I think I tried it but ran into some kind of issue that made me stop trying.

PR #590, opened 2021-12-06
Crate restructuring

maciejhirsz:
I need to start making smaller PRs. This one is pretty straight forward, removed the v2 module and moved things around to hopefully sensible locations.

Rationale: We added the `v2` module when doing some internal refactoring that ultimately lead to different versions of types with similar names (`Request` most notably), but have since unified the apis to use same types, and with the "v1" absent there is no need for the module to exist.

Edit: [see comment](https://github.com/paritytech/jsonrpsee/pull/590#issuecomment-993425547) for ongoing changes.


types/src/lib.rs
```diff
@@ -49,10 +46,20 @@ pub mod traits;
 /// Middleware trait and implementation.
 pub mod middleware;
 
+/// JSON_RPC params related types.
```
niklasad1, line 49:
```suggestion
/// JSON-RPC params related types.
```

niklasad1 (approved):
:+1:

dvdplm:
@maciejhirsz The PR description is a bit lacking here. Can you elaborate on why removing the `v2` module is good? Not saying I disagree, but it is there for a reason (separate spec-mandated types from the rest) so it'd be good to write down why that reason is invalid/superseded.

utils/src/server/helpers.rs
```diff
@@ -26,13 +26,11 @@
...
-use jsonrpsee_types::v2::error::{OVERSIZED_RESPONSE_CODE, OVERSIZED_RESPONSE_MSG};
-use jsonrpsee_types::v2::{
-	error::{CALL_EXECUTION_FAILED_CODE, UNKNOWN_ERROR_CODE},
-	ErrorCode, ErrorObject, Id, InvalidRequest, Response, RpcError, TwoPointZero,
-};
+use jsonrpsee_types::error::rpc::{CALL_EXECUTION_FAILED_CODE, UNKNOWN_ERROR_CODE};
+use jsonrpsee_types::error::rpc::{OVERSIZED_RESPONSE_CODE, OVERSIZED_RESPONSE_MSG};
+use jsonrpsee_types::error::{CallError, Error, ErrorCode, ErrorObject, RpcError};
+use jsonrpsee_types::{to_json_raw_value, Id, TwoPointZero};
+use jsonrpsee_types::{InvalidRequest, Response};
```
dvdplm, line 33: Let's use some braces here. :)

maciejhirsz:
@dvdplm Should be all tidy now, thoughts on having a linter for forcing braces on all the imports?

niklasad1:
> Rationale: We added the v2 module when doing some internal refactoring that ultimately lead to different versions of types with similar names (Request most notably), but have since unified the apis to use same types, and with the "v1" absent there is no need for the module to exist.

..., I thought we used `v2` to indicate that these types are `JSON-RPC v2 spec related types` and not just `common` types after we removed the `JSON-RPC spec related from the jsonrpc crate (way back)`. I like this change because it's a bit annoying to have to write `jsonrpsee_types::v2::Response` instead of `jsonrpsee_types::Response`...

> thoughts on having a linter for forcing braces on all the imports

I don't like the braces for single imports if that's what you are suggesting but.... feel free to convince me :) 
Then we need some special rustfmt rule...

maciejhirsz:
> ..., I thought we used `v2` to indicate that these types are `JSON-RPC v2 spec related types` (...)

We did? I thought it was just to keep stuff separate and it always bugged me. Given that 2.0 is the only version we support, and I don't think it's likely to change in the future I also don't see why we would need that differentiation.

> I don't like the braces for single imports if that's what you are suggesting but.... feel free to convince me :)
> Then we need some special rustfmt rule...

```
use foo::{a, b, c};
use foo::{d, e, f};
```

Or

```
use foo::{
    a, b, c,
    d, e, f,
};
```
(these names are short enough to fit in a single line, but you get the idea)

maciejhirsz:
> I don't care that much about that but I like point_down better

Yeah, that's my preference too, I can revert some of the imports here.

dvdplm:
> We did? I thought it was just to keep stuff separate and it always bugged me. Given that 2.0 is the only version we support, and I don't think it's likely to change in the future I also don't see why we would need that differentiation.

It was never so much about versioning as it was about separating **spec mandated types** from the rest. The types from the spec must be kept intact forever while other public types can change with a breaking release. I think that is a worthy motivation for keeping things separate, and `v2` seemed like a decent enough name for it. 
However, given that wasn't clear to you @maciejhirsz I bet it isn't to many others, so we should perhaps consider a different name, maybe `types::spec::StuffFromTheSpec`?

maciejhirsz:
> The types from the spec must be kept intact forever while other public types can change with a breaking release.

I don't think that's true, we can easily make a breaking change to a spec type by changing stuff like `String` to `Cow<'a, str>`, or getting rid of the `TwoPointZero` type and just building a manual `Serialize` impl on request etc., without breaking spec compatibility.

dvdplm:
> I don't think that's true, we can easily make a breaking change to a spec type by changing stuff like `String` to `Cow<'a, str>`, or getting rid of the `TwoPointZero` type and just building a manual `Serialize` impl on request etc., without breaking spec compatibility.

You are correct, but you're missing the point: these types do require some extra care outside of the normal semver compatibility narrative. Question is: is this something that is worth signalling by keeping them in their own namespace? I think so, and that's what lies behind the `v2` module. 
I'm not opposed to rename the module and I'm willing to listen to arguments against keeping them separate from the rest but so far I remain unconvinced.

dvdplm:
After discussion here's the plan:

1. Rename `jsonrpsee_utils` to `jsonrpsee_core`
1. Rename `jsonrpsee_types::RpcError` to `jsonrpsee_types::ErrorResponse`.
1. Move everything that's not a JSON-RCP transport type from `jsonrpsee_types` to `jsonrpsee_core`. This includes:
3.1 Middleware and traits.
3.2 Serialize, DeserializeOwned, beef::Cow, to_json_value, to_json_raw_value re-exports.
3.3 All the error stuff that's not ErrorResponse, enums and other internal utils.

ws-client/src/client.rs
```diff
@@ -590,7 +593,7 @@ async fn background_task(
 					}
 				}
 				// Error response
```
dvdplm, line 595: Clippy is angry about line 195

core/src/error.rs
```diff
@@ -0,0 +1,212 @@
...
+	/// Failed to parse the data.
+	#[error("Parse error: {0}")]
+	ParseError(#[from] serde_json::Error),
+	/// Invalid subscription ID.
+	#[error("Invalid subscription ID")]
+	InvalidSubscriptionId,
+	/// Invalid request ID.
+	#[error("Invalid request ID")]
+	InvalidRequestId,
+	/// Client received a notification with an unregistered method
```
TarikGul, line 85:
```suggestion
	/// Client received a notification with an unregistered method.
```

TarikGul: I don't want to fill the PR with unnecessary noise, so I just made one suggestion, but not sure if you want the docs to all have periods after each description. Just an observation

TarikGul: Yea makes sense :). Will close this one.

core/src/server/mod.rs
```diff
@@ -26,6 +26,9 @@
 
 //! Shared modules for the JSON-RPC servers.
 
+/// Client types.
```
niklasad1, line 29:
```suggestion
```

core/src/server/mod.rs
```diff
@@ -26,6 +26,9 @@
...
+// mod client;
```
niklasad1, line 30:
```suggestion
```

core/src/lib.rs
```diff
@@ -39,3 +48,25 @@ pub mod server;
...
+/// dependencies to be explicitly added on the client side.
+#[doc(hidden)]
+pub mod __reexports {
+	pub use async_trait::async_trait;
+	pub use serde;
+	pub use serde_json;
+}
+
+pub use beef::Cow;
+pub use serde::{de::DeserializeOwned, Serialize};
```
niklasad1, line 68: maybe we should re-export `Deserialize` here too..

core/src/traits.rs
```diff
@@ -54,6 +56,7 @@ pub trait Client {
 }
 
 /// [JSON-RPC](https://www.jsonrpc.org/specification) client interface that can make requests, notifications and subscriptions.
+#[cfg(feature = "client")]
```
niklasad1, line 59: I don't mind but then the `Client trait` should be feature gated too.

Also, I think it's fine to not feature gate these traits.

proc-macros/src/rpc_macro.rs
```diff
@@ -26,16 +26,14 @@
 
 //! Declaration of the JSON RPC generator procedural macros.
 
-use crate::{
-	attributes::{
-		optional, parse_param_kind, Aliases, Argument, AttributeMeta, MissingArgument, NameMapping, ParamKind, Resource,
-	},
-	helpers::extract_doc_comments,
-};
+use std::borrow::Cow;
```
niklasad1, line 29: use `beef::Cow`? :)

core/Cargo.toml
```diff
@@ -1,44 +1,42 @@
...
 beef = { version = "0.5.1", features = ["impl_serde"] }
-thiserror = { version = "1", optional = true }
-futures-channel = { version = "0.3.14", default-features = false, optional = true }
+thiserror = "1"
+futures-channel = { version = "0.3.14", default-features = false }
 futures-util = { version = "0.3.14", default-features = false, optional = true }
-hyper = { version = "0.14.10", default-features = false, features = ["stream"], optional = true }
-jsonrpsee-types = { path = "../types", version = "0.6.0", optional = true }
+hyper = { version = "0.14.10", default-features = false, features = ["stream"] }
+jsonrpsee-types = { path = "../types", version = "0.6.0" }
```
niklasad1, line 18: this crate is now what the `v2` was?

niklasad1 (approved):
LGTM, a few small questions....

This makes the repo structure more clear all good.

core/Cargo.toml
maciejhirsz: Correct

proc-macros/src/rpc_macro.rs
maciejhirsz: It's a bit annoying with `beef::Cow` because it doesn't have `ToTokens` impl, and re-borrowing everything to `&str` on stack after call adds boilerplate.

Could add a feature flag adding `ToTokens` to `beef` maybe later.

core/src/lib.rs
maciejhirsz: Mayhaps, we are also super inconsistent about reexports, sometimes we put them in `__reexports` module, and sometimes not, I'd leave that for a future PR.

proc-macros/src/rpc_macro.rs
niklasad1: ah right it's macros, nvm

TarikGul (approved):
lgtm, It generally looks way cleaner to me (ie. the imports are way easier to follow).

The Macro export for `rpc_params` in `core/src/client.rs` took me a hot second to understand because github formatted it very weird, but overall 👍  from me.

PR #591, opened 2021-12-06
rpc module: refactor calls/subs without a server

niklasad1:
This PR does the following:

- Remove `call_with` and `call_with_subscription` (these are confusing)
- Change `call` to return a typed response instead of json string.
- Change `test_subscription` to `subscription`
- Introduce `raw_json_request` to just accept as `JSON String` that can be either subscription or call then just return raw json text as response.

Note, we don't have anything that supports calls with `parameters as JSON Object` besides doing raw json...

Any better ideas?!


utils/src/server/rpc_module.rs
```diff
@@ -349,34 +352,23 @@ impl Methods {
...
 		}
-
-		rx.next().await
+		Err(resp)
 	}
 
-	/// Perform a "in memory JSON-RPC method call" and receive further subscriptions.
-	/// This is useful if you want to support both `method calls` and `subscriptions`
-	/// in the same API.
+	/// Perform a raw "in memory JSON-RPC method call".
```
niklasad1, line 364:
```suggestion
	/// Perform one or more raw "in memory JSON-RPC method calls".
```

utils/src/server/rpc_module.rs
```diff
@@ -349,34 +352,23 @@ impl Methods {
...
 		}
-
-		rx.next().await
+		Err(resp)
 	}
 
-	/// Perform a "in memory JSON-RPC method call" and receive further subscriptions.
-	/// This is useful if you want to support both `method calls` and `subscriptions`
-	/// in the same API.
+	/// Perform one or more raw "in memory JSON-RPC method calls".
```
niklasad1, line 364:
```suggestion
	/// Perform a "in memory JSON-RPC method call" and receive further subscriptions.
	/// This is useful if you want to support both `method calls` and `subscriptions`
	/// in the same API.
```

utils/src/server/rpc_module.rs
```diff
@@ -349,34 +352,23 @@ impl Methods {
...
+	/// 
+	/// You can use this to support method calls and subscriptions
```
niklasad1, line 368:
```suggestion
```

utils/src/server/rpc_module.rs
```diff
@@ -397,48 +391,37 @@ impl Methods {
...
 	///     let sub_resp: SubscriptionResponse<String> = serde_json::from_str(&raw_sub_resp).unwrap();
 	///     assert_eq!(&sub_resp.params.result, "one answer");
 	/// }
 	/// ```
-	pub async fn call_and_subscribe<Params: ToRpcParams>(
-		&self,
-		method: &str,
-		params: Params,
-	) -> Option<(String, mpsc::UnboundedReceiver<String>)> {
+	pub async fn raw_call(&self, method: &str, params: Box<RawValue>) -> RawRpcResponse {
```
niklasad1, line 401: this is still confusing I think...

utils/src/server/rpc_module.rs
```diff
@@ -811,7 +794,7 @@ pub struct TestSubscription {
 }
 
 impl TestSubscription {
-	/// Close the subscription channel.
+	/// Close the subscription channel by doing a unsubscribe call.
```
dvdplm, line 797:
```suggestion
	/// Close the subscription channel by making an unsubscribe call.
```

utils/src/server/rpc_module.rs
dvdplm: How about this reasoning: instead of struggling to find a good name for these, how about we give them crystal clear but ugly names that clearly signals what they are: building blocks for tests. So in the vein of `test_subscription`, we can call this `test_method_call` or `method_call_helper`.

niklasad1: hehe, this could be either of those :)

if you look at the code now, both `call` and `test_subscription` uses `raw_call` under the hood thus more or less similar to `handle_request` in the `jsonrpc` crate.

however, the best is if we ditch this in the public API and just have `call` and `subscribe` I guess but won't to send raw json.

James has a point that these might be useful else than only in tests, https://github.com/paritytech/jsonrpsee/pull/458#discussion_r707126871

dvdplm: How about:
`test_subscription` –> `subscription_helper`
`raw_call` –> `method_call_helper`

@maciejhirsz wdyt?

niklasad1: can you read the PR description and skim the code again?

I made some changes and updated the PR description right now.

utils/src/server/rpc_module.rs
```diff
@@ -349,96 +350,58 @@ impl Methods {
...
 
-	/// Test helper that sets up a subscription using the given `method`. Returns a tuple of the
-	/// [`SubscriptionId`] and a channel on which subscription JSON payloads can be received.
-	pub async fn test_subscription(&self, method: &str, params: impl ToRpcParams) -> TestSubscription {
-		let params = params.to_rpc_params().expect("valid JSON-RPC params");
-		tracing::trace!("[Methods::test_subscription] Calling subscription method: {:?}, params: {:?}", method, params);
-		let req = Request::new(method.into(), Some(&params), Id::Number(0));
+	/// Make a request (JSON-RPC method call or subscription) by using raw JSON.
+	///
+	/// Returns the raw JSON response to the call and a stream to receive notitications if was a subscription.
```
dvdplm, line 372:
```suggestion
	/// Returns the raw JSON response to the call and a stream to receive notifications from if the call was a subscription.
```

utils/src/server/rpc_module.rs
```diff
@@ -349,96 +350,58 @@ impl Methods {
...
-	///         sink.send(&"one answer").unwrap();
-	///         Ok(())
-	///     }).unwrap();
-	///     let (resp, mut stream) = module.call_and_subscribe("hi", EmptyParams::new()).await.unwrap();
-	///     assert!(serde_json::from_str::<Response<u64>>(&resp).is_ok());
-	///     let raw_sub_resp = stream.next().await.unwrap();
-	///     let sub_resp: SubscriptionResponse<String> = serde_json::from_str(&raw_sub_resp).unwrap();
-	///     assert_eq!(&sub_resp.params.result, "one answer");
-	/// }
-	/// ```
```
dvdplm, line 406: This example was good to have there I think. Can we port it over to the new API.

dvdplm (approved):
This is better than what we had, sorry for making confused comments. :)

utils/src/server/rpc_module.rs
niklasad1: sure

utils/src/server/rpc_module.rs
```diff
@@ -349,96 +350,113 @@ impl Methods {
...
-
-		let (tx, mut rx) = mpsc::unbounded();
-		let sink = MethodSink::new(tx);
-
-		if let MethodResult::Async(fut) = self.execute(&sink, req, 0) {
-			fut.await;
+	///
+	/// Returns the decoded value of the `result field` in JSON-RPC response if succesful.
+	///
+	/// ```
```
jsdw, line 356: Love the example addition! To keep with how Rust docs these things, how about adding an

```
# Example
```

header before it?

jsdw: (I guess this comment would apply to other examples, too)

jsdw (approved):
Looks great, and I love the added docs :)

utils/src/server/rpc_module.rs
niklasad1: sure why not :)

utils/src/server/rpc_module.rs
```diff
@@ -977,27 +1003,23 @@ mod tests {
...
 
 		// Call sync method with no params
-		let result = module.call("rebel_without_cause", None).await.unwrap();
-		assert_eq!(result, r#"{"jsonrpc":"2.0","result":false,"id":0}"#);
-
-		// Call sync method with no params, alternative way.
-		let result = module.call_with::<[u8; 0]>("rebel_without_cause", []).await.unwrap();
-		assert_eq!(result, r#"{"jsonrpc":"2.0","result":false,"id":0}"#);
+		let res: bool = module.call("rebel_without_cause", EmptyParams::new()).await.unwrap();
+		assert_eq!(res, false);
```
niklasad1, line 1007:
```suggestion
		assert!(!res);
```

niklasad1:
```suggestion
		assert_eq!(!res);
```

PR #592, opened 2021-12-07
clients: use `FxHashMap` instead `FnvHashMap`

niklasad1:
`FxHashMap` is already used in the `utils crate` then let's use it in the clients too because it's faster and to reduce the dependencies a bit.


insipx (approved):
always nice to be consistent!

TarikGul (approved):

dvdplm:
Include in 0.6.1?

niklasad1:
already released :(

PR #593, opened 2021-12-07
chore: release v0.6.1

niklasad1:


TarikGul (approved):
lgtm

dvdplm (approved):
I assume you need this somewhere?

At some point we should have a think about whether we really need all crates to be versioned in lock-step or independently.

maciejhirsz (approved):

niklasad1:
> I assume you need this somewhere?

> At some point we should have a think about whether we really need all crates to be versioned in lock-step or independently.

I need it in substrate to avoid introducing breaking changes.

PR #594, opened 2021-12-08
types: make subscription ID String Cow.

niklasad1:


insipx (approved):

types/src/v2/params.rs
```diff
@@ -288,18 +288,49 @@ impl<'a> From<&'a [JsonValue]> for ParamsSer<'a> {
...
 		match sub_id {
 			SubscriptionId::Num(n) => n.into(),
-			SubscriptionId::Str(s) => s.into(),
+			SubscriptionId::Str(s) => s.into_owned().into(),
+		}
+	}
+}
+
+impl<'a> TryFrom<JsonValue> for SubscriptionId<'a> {
+	type Error = ();
```
jsdw, line 309: Would there be any use in making this more specific so that we can catch the "wrong value type" vs "value too large" errors below?

jsdw:
Offhand it looks like we just use `SubscriptionId<'static>` around the place; what's the advantage of using `Cow`? (I'm sure I'm missing something!)

dvdplm:
A bit of motivation for this change would be good, the code looks fine but I'm not sure why we need this.

maciejhirsz (approved):
LGTM, not sure it's entirely necessary given we should be using numbers anyway :)

niklasad1:
> LGTM, not sure it's entirely necessary given we should be using numbers anyway :)

Yeah, it's compat stuff for other libraries (e.g. jsonrpc crate) but not sure if we need anymore so maybe better just to use `u64`?!
I would love to remove that code :)

niklasad1:
> Offhand it looks like we just use SubscriptionId<'static> around the place; what's the advantage of using Cow? (I'm sure I'm missing something!)

Yeah you are correct, I didn't manage to get rid of `static lifetimes` but because we the send these over the channel and store them in a HashMap so won't work without further refactoring. So needless currently but I was annoyed that we have `Id` which is `Cow` and `SubscriptionID` is not `Cow` :)

however, this is not useful without further refactoring I guess besides to `TryFrom` instead of de-serializing SubscriptionID from `serde::Value`

types/src/v2/params.rs
niklasad1: yupp,

PR #595, opened 2021-12-08
fix: duplicate env logger deps

niklasad1:


insipx (approved):

jsdw (approved):

tests/Cargo.toml
```diff
@@ -8,7 +8,7 @@ license = "MIT"
 publish = false
 
 [dev-dependencies]
-env_logger = "0.8"
```
dvdplm, line 11: oops. I wish there was a CI check for dupes...

dvdplm (approved):

maciejhirsz (approved):

tests/Cargo.toml
niklasad1: we could use `cargo deny` for but it will generate a bunch of errors/warnings because of the `benches with jsonrpc crate` and `your added example  but maybe we could feature gate it instead.

PR #596, opened 2021-12-08
benches: make jsonrpc crates optional

niklasad1:
Make the dependency tree smaller with the default bench dependencies.

### DIFF
--- a/benches/Cargo.toml
+++ b/benches/Cargo.toml
@@ -7,17 +7,17 @@ edition = "2018"
 license = "MIT"
 publish = false
 
-[dev-dependencies]
+[dependencies]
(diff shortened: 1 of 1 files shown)


dvdplm (approved):
Doesn't this require a change to CI too, so we run the `jsonrpc` benches too?

niklasad1:
> Doesn't this require a change to CI too, so we run the jsonrpc benches too?

We're not running the daily benches on jsonrpc crate (hidden behind the feature flag `jsonrpc-crate`), why should we?
Isn't sufficient to run against it on releases?

However, would probably good to automate it otherwise it's easy to forget... so maybe not a bad idea do run it daily/weekly or something.

jsdw (approved):

PR #601, opened 2021-12-10
impl Stream on Subscription and tweak built-in next() method to align

jsdw:
Adds a `Stream` impl on the `Subscription` type, and tweaks the type signatuer of the built-in `next()` method to be identical (and make it use the Stream impl).

Really, the only reason to have a built-in `next()` method now is to avoid needing to import `StreamExt`, but I think that's a perfectly good reason to keep it, and if the type sig is identical it hopefully won't lead to much confusion :)

### DIFF
--- a/types/src/client.rs
+++ b/types/src/client.rs
@@ -167,8 +167,8 @@ where
 	Notif: DeserializeOwned,
 {
 	/// Returns the next notification from the stream.
-	/// This may return `None` if the subscription has been terminated.
-	/// This may happen if the channel becomes full or is dropped.
+	/// This may return `None` if the subscription has been terminated,
+	/// which may happen if the channel becomes full or is dropped.
 	pub async fn next(&mut self) -> Option<Result<Notif, Error>> {
 		StreamExt::next(self).await
 	}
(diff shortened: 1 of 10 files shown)


types/src/client.rs
```diff
@@ -157,17 +167,26 @@ where
...
+}
+
+impl<Notif> Stream for Subscription<Notif>
+where
+	Notif: DeserializeOwned,
+{
+	type Item = Result<Notif, Error>;
+	fn poll_next(mut self: Pin<&mut Self>, cx: &mut task::Context<'_>) -> task::Poll<Option<Self::Item>> {
+		let n = futures_util::ready!(self.notifs_rx.poll_next_unpin(cx));
+		let res = n.and_then(|n| match serde_json::from_value::<NotifResponse<Notif>>(n) {
```
niklasad1, line 184: does `map` work here?

I don't mind but clippy doesn't like `and_then` here...

niklasad1 (approved):

maciejhirsz (approved):
Nice, looks pretty clean too!

types/src/client.rs
jsdw: Ooh good point! I thought I needed `and_then` but then didn't realise that I then surrounded all of the result arms in `Some` :)

Issue #599, opened 2021-12-09
Add a way to customize `SubscriptionId` generation


librelois:
I started to migrate the codebase of [frontier](https://github.com/paritytech/frontier) on jsonrpsee based on the branch `polkadot-jsonrpsee-v0.9.13` and I notice that jsonrpsee lacks an essential feature that makes it impossible to use it in frontier: the possibility to customize the way to generate `SubscriptionId`.

Indeed, frontier must reproduce the ethereum RPC api with accuracy, this includes providing SubscriptionId in string form with 16 bytes in hexadecimal representation prefixed by `"0x"`.

Frontier currently uses jsonrpc, which allows custom generation of subscription ids with the `with_id_provider` function: https://github.com/paritytech/jsonrpc/blob/609d7a6cc160742d035510fa89fb424ccf077660/pubsub/src/manager.rs#L154

niklasad1:
fair enough then, we should support this.

//cc @maciejhirsz 

Should we simply have another param to `register_subscription` that takes a closure that generates a `String` or `u64`?
However, it will be hard to solve "nice" in the proc macro... I guess

Thoughts?

maciejhirsz:
Do we need to touch proc macros with this? I'd think this is more of a server config stuff. And yeah, a closure (generator) that yields a new `SubscriptionId` should suffice.

niklasad1:
`SubscriptionId` generator is in the `RpcModule` so it's not possible to make this configuration based on the server itself. However, we could include it the `callback closure` similar to what we did for `MaxPayloadSize a.k.a MethodSink` that should work then no proc macro stuff needs changing....

any other ideas?

niklasad1:
Closed by #604

PR #604, opened 2021-12-13
feat: server configurable subscriptionID

niklasad1:
Closing #599


types/src/error.rs
```diff
@@ -183,13 +185,14 @@ impl Error {
 #[derive(Deserialize, Serialize, Debug, PartialEq)]
 pub struct SubscriptionClosedError {
 	subscription_closed: String,
-	id: u64,
+	// TODO: #[serde(borrow)]..
```
niklasad1, line 188: //cc @maciejhirsz 

I have lifetimes issues with serde, ideally we should have `SubscriptionId<'a>` here but that forces us to change the NotifResponse to 

```
#[derive(Debug, Deserialize, Serialize)]
#[serde(untagged)]
enum NotifResponse<'a, Notif> {
	Ok(Notif),
	#[serde(borrow)]..
	Err(SubscriptionClosedError<'a>),
}
```
which brakes this:

```rust
impl<Notif> Stream for Subscription<Notif>
where
	Notif: DeserializeOwned,
{
	type Item = Result<Notif, Error>;
	fn poll_next(mut self: Pin<&mut Self>, cx: &mut task::Context<'_>) -> task::Poll<Option<Self::Item>> {
		let n: Option<serde_json::Value> = futures_util::ready!(self.notifs_rx.poll_next_unpin(cx));
		let res = n.map(|n| match serde_json::from_value::<NotifResponse<Notif>>(n) {
			Ok(NotifResponse::Ok(parsed)) => Ok(parsed),
			Ok(NotifResponse::Err(e)) => Err(Error::SubscriptionClosed(e)),
			Err(e) => Err(Error::ParseError(e)),
		});
		task::Poll::Ready(res)
	}
}
```

most likely because of `DeserializedOwned` because the clients send `serde_json::Value` back and forth it doesn't work, any ideas?

ws-server/src/server.rs
```diff
@@ -504,17 +529,29 @@ impl Default for Settings {
 	}
 }
 
+//#[derive(Debug)]
```
niklasad1, line 532:
```suggestion
```

dvdplm (approved):

PR #611, opened 2021-12-14
Better error messages for method arguments ignored with a `_`

### LINKED ISSUE #609: [proc_macros]: `_` as ident doesn't work
```
use jsonrpsee_proc_macros::rpc;
use bytes::Bytes;

#[rpc(server)]
pub trait ARpc {

    #[method(name = "a")]
    async fn a(&self, _: Bytes);
}
```
will report:
```
error: custom attribute panicked
   |
14 | #[rpc(server)]
   | ^^^^^^^^^^^^^^
   |
   = help: message: Identifier in signature must be an ident
```

dvdplm:
Fixes #609

### DIFF
--- a/proc-macros/src/rpc_macro.rs
+++ b/proc-macros/src/rpc_macro.rs
@@ -71,9 +71,8 @@ impl RpcMethod {
 			None => quote!(),
 		};
 
-		let span = sig.span();
 		if blocking && sig.asyncness.is_some() {
-			return Err(syn::Error::new(span, "Blocking method must be synchronous"));
+			return Err(syn::Error::new(sig.span(), "Blocking method must be synchronous"));
 		}
 
 		let params: Vec<_> = sig
@@ -88,7 +87,7 @@ impl RpcMethod {
 						"Method argument names must be valid Rust identifiers; got `_` instead",
 					))),
 					_ => Some(Err(syn::Error::new(
-						span,
+						arg.span(),
 						format!("Unexpected method signature input; got {:?} ", *arg.pat),
 					))),
 				},
--- a/proc-macros/src/rpc_macro.rs
+++ b/proc-macros/src/rpc_macro.rs
@@ -93,7 +93,7 @@ impl RpcMethod {
 					))),
 				},
 			})
-			.collect::<Result<Vec<_>, _>>()?;
+			.collect::<Result<_, _>>()?;
 
 		let returns = match sig.output {
 			syn::ReturnType::Default => None,
--- a/proc-macros/src/rpc_macro.rs
+++ b/proc-macros/src/rpc_macro.rs
@@ -71,8 +71,9 @@ impl RpcMethod {
 			None => quote!(),
 		};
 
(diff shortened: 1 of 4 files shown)


proc-macros/src/rpc_macro.rs
```diff
@@ -81,11 +82,18 @@ impl RpcMethod {
 			.filter_map(|arg| match arg {
 				syn::FnArg::Receiver(_) => None,
 				syn::FnArg::Typed(arg) => match *arg.pat {
-					syn::Pat::Ident(name) => Some((name, *arg.ty)),
-					_ => panic!("Identifier in signature must be an ident"),
+					syn::Pat::Ident(name) => Some(Ok((name, *arg.ty))),
+					syn::Pat::Wild(_) => Some(Err(syn::Error::new(
+						span,
+						"Method argument names must be a valid Rust identifiers; got `_` instead",
```
TarikGul, line 88: Not sure what the correct verbage would be in order to correctly express the error behind the syntax, but it sounds like to me it would be either:

"be a valid Rust identifier"

or

"be valid Rust identifiers"

dvdplm: ty! :)

niklasad1 (approved):
LGTM,

I had no idea that `_` isn't a valid rust identifier.

proc-macros/src/rpc_macro.rs
```diff
@@ -81,11 +82,18 @@ impl RpcMethod {
...
+						"Method argument names must be valid Rust identifiers; got `_` instead",
+					))),
+					_ => Some(Err(syn::Error::new(
+						span,
+						format!("Unexpected method signature input; got {:?} ", *arg.pat),
+					))),
 				},
 			})
-			.collect();
+			.collect::<Result<Vec<_>, _>>()?;
```
maciejhirsz, line 96:
```suggestion
			.collect::<Result<_, _>>()?;
```

Or remove `Vec<_>` on line 79.

proc-macros/src/rpc_macro.rs
```diff
@@ -81,11 +82,18 @@ impl RpcMethod {
 			.filter_map(|arg| match arg {
 				syn::FnArg::Receiver(_) => None,
 				syn::FnArg::Typed(arg) => match *arg.pat {
-					syn::Pat::Ident(name) => Some((name, *arg.ty)),
-					_ => panic!("Identifier in signature must be an ident"),
+					syn::Pat::Ident(name) => Some(Ok((name, *arg.ty))),
+					syn::Pat::Wild(_) => Some(Err(syn::Error::new(
+						span,
```
maciejhirsz, line 87: Couldn't we get the span for the token itself here so the highlight actually shows what the problem is?

maciejhirsz (approved):
Looks good, minor nits.

proc-macros/src/rpc_macro.rs
dvdplm: Absolutely, fixed.

PR #612, opened 2021-12-14
refactor `SubscriptionClosed`

niklasad1:
I noticed that we "trying" to sending "close messages" on subscriptions
that already has been terminated by the client which is unnecessary.

Moreover, this introduces `fn is_closed` to check whether a given
channel is closed without having to try send an actual message on the channel.

Also it would help me to fix #604

### DIFF
--- a/types/src/error.rs
+++ b/types/src/error.rs
@@ -129,7 +129,7 @@ pub enum Error {
 	SubscriptionNameConflict(String),
 	/// Subscription got closed.
 	#[error("Subscription closed: {0:?}")]
-	SubscriptionClosed(SubscriptionClosedError),
+	SubscriptionClosed(SubscriptionClosed),
 	/// Request timeout
 	#[error("Request timeout")]
 	RequestTimeout,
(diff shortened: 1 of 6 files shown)


types/src/error.rs
```diff
@@ -177,32 +177,44 @@ impl Error {
...
+	/// Get the close reason.
+	pub fn close_reason(&self) -> &SubscriptionClosedReason {
+		&self.subscription_closed
 	}
 }
 
+/// A type to represent when a subscription gets closed
+/// by either the server or client side.
+#[derive(Deserialize, Serialize, Debug, PartialEq)]
+pub enum SubscriptionClosedReason {
```
niklasad1, line 209: I removed the subscription ID from here because `SubscriptionSink` inserts in the subid in the message anyway.

maciejhirsz (approved):
Good catch!

types/src/error.rs
```diff
@@ -177,32 +177,44 @@ impl Error {
...
+	fn from(kind: SubscriptionClosedReason) -> Self {
+		Self::new(kind)
 	}
+}
 
-	/// Get the reason why the subscription was closed.
-	pub fn close_reason(&self) -> &str {
-		&self.subscription_closed
+impl SubscriptionClosed {
+	/// Create a new subscription message.
```
dvdplm, line 195: This seems like it belongs somewhere else?

types/src/error.rs
```diff
@@ -177,32 +177,44 @@ impl Error {
...
-/// Error type with a special `subscription_closed` field to detect that
+/// A type with a special `subscription_closed` field to detect that
 /// a subscription has been closed to distinguish valid items produced
 /// by the server on the subscription stream from an error.
 #[derive(Deserialize, Serialize, Debug, PartialEq)]
-pub struct SubscriptionClosedError {
-	subscription_closed: String,
-	id: u64,
+pub struct SubscriptionClosed {
+	subscription_closed: SubscriptionClosedReason,
```
dvdplm, line 185:
```suggestion
	reason: SubscriptionClosedReason,
```

…maybe?

types/src/error.rs
dvdplm: That's fine; do we have a test somewhere that asserts that the ID is still available?

types/src/error.rs
```diff
@@ -177,32 +177,44 @@ impl Error {
...
+/// A type to represent when a subscription gets closed
+/// by either the server or client side.
+#[derive(Deserialize, Serialize, Debug, PartialEq)]
+pub enum SubscriptionClosedReason {
+	/// The subscription was closed by calling the unsubscribe method.
+	Unsubscribed,
+	/// The client closed the connection.
+	ConnectionReset,
+	/// The server closed the subscription.
+	Server(String),
```
dvdplm, line 215: Would be good to document what information the `String` carries here.

dvdplm (approved):
Nice

types/src/error.rs
niklasad1: Sure, we got the [SubscriptionSink::close](https://github.com/paritytech/jsonrpsee/blob/master/utils/src/server/rpc_module.rs#L807) where one can add the reason why a subscription was closed on the server-side.

types/src/error.rs
niklasad1: it will break compatibility with clients on older versions, worth it?

types/src/error.rs
```diff
@@ -177,32 +177,44 @@ impl Error {
...
 
+/// A type to represent when a subscription gets closed
+/// by either the server or client side.
+#[derive(Deserialize, Serialize, Debug, PartialEq)]
+pub enum SubscriptionClosedReason {
+	/// The subscription was closed by calling the unsubscribe method.
+	Unsubscribed,
+	/// The client closed the connection.
+	ConnectionReset,
+	/// The server closed the subscription.
```
dvdplm, line 214:
```suggestion
	/// The server closed the subscription, providing a description of the reason as a `String`.
```

types/src/error.rs
niklasad1: we do test this in the `integration tests` and `RpcModule::subscribe` tests because it's required that the response is a `SubscriptionResponse` i.e, has a field `subscription` where the `sub id` is provided.

If that is missing such as "ordinary notification without subscription field` it would not be decoded by the client

types/src/error.rs
niklasad1: yeah, let's do it then :)

types/src/error.rs
```diff
@@ -177,32 +177,47 @@ impl Error {
...
-		self.id
+	/// Get the close reason.
+	pub fn close_reason(&self) -> &SubscriptionClosedReason {
+		&self.reason
 	}
 }
 
+/// A type to represent when a subscription gets closed
+/// by either the server or client side.
+#[derive(Deserialize, Serialize, Debug, PartialEq)]
```
niklasad1, line 211: //cc @maciejhirsz @dvdplm  as we already changed `close_reason` -> `reason`

should have some obscure `#[serde(tag = "_secret_tag")]` such that the type get mixed up with a something else?

niklasad1: right, I think `tag doesn't` work on new type pattern nvm i.e, Server(String)

PR #617, opened 2021-12-21
[ci] add check-benchmark job

alvicsam:
The pr adds additional job, that checks current benchmark result and compares is with the previous one. If difference between results is more than threshold then a GitHub issue is created ([Example](https://github.com/tripleightech/gl-gh-test/issues/24))

Threshold can be set in pipeline, default is 20%

The checking script can be found [here](https://github.com/paritytech/scripts/tree/master/dockerfiles/benchmarks)

Closes https://github.com/paritytech/ci_cd/issues/275

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -11,6 +11,7 @@ stages:
   - lint
   - test
   - benchmark
+  - check_benchmark
   - publish
 
 variables:                         &default-vars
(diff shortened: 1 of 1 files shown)


niklasad1 (approved):
great, LGTM

radupopa2010 (approved):

PR #625, opened 2021-12-22
fix jsonrpsee: feature renamed `client` -> `async-client`

niklasad1:


jsonrpsee/Cargo.toml
```diff
@@ -24,7 +24,7 @@ jsonrpsee-types = { path = "../types", version = "0.6.1", package = "jsonrpsee-t
 [features]
 client-ws-transport = ["jsonrpsee-client-transport/ws", "jsonrpsee-client-transport/tls"]
 client-ws-transport-no-tls = ["jsonrpsee-client-transport/ws"]
-core-client = ["jsonrpsee-core/client"]
+core-client = ["jsonrpsee-core/async-client"]
```
dvdplm, line 27: Maybe `async-client` is a better name for the feature, so it's consistent and we avoid the ambiguous "core"?

niklasad1: agree.

jsonrpsee/src/lib.rs
```diff
@@ -79,7 +82,13 @@ pub use jsonrpsee_types as types;
 #[cfg(any(feature = "http-server", feature = "ws-server"))]
 pub use jsonrpsee_core::server::rpc_module::{RpcModule, SubscriptionSink};
 
-#[cfg(any(feature = "http-server", feature = "ws-server", feature = "core-client"))]
+#[cfg(any(
+	feature = "http-server",
+	feature = "ws-server",
+	feature = "async-client",
```
dvdplm, line 88: do we need `feature = "client"` as well?

dvdplm (approved):

jsonrpsee/src/lib.rs
niklasad1: I don't think so but added it too the be on the safe side :)

PR #639, opened 2022-01-05
feat(rpc module): `stream API` for SubscriptionSink

### LINKED ISSUE #627: [rpc module]: SubscriptionSink more user-friendly API for sending items from stream.
We have noticed the API for `SubscriptionSink` is a bit messy to use if you want to await for stream to be sent via the `SubscriptionSink` if the connection has already been closed it might leak until the stream/fut "resolves" because the stream itself doesn't know anything of the connection state.

For example given that we have the following impl:

```rust
impl MyServerTrait for () 
{
	fn subscribe_to_something(&self, mut sink: SubscriptionSink) -> RpcResult<()> {
                // assume this stream lives forever and produces items "rarely"...
		let stream = create_stream();

		let fut = async move {
			stream
				.take_while(|sc| future::ready(sink.send(sc).map_or(false, |_| true)))
				.for_each(|_| future::ready(()))
				.await
		};

	       tokio::spawn(fut);
	}
}
```

If connection has already this case the `tokio task` will live until `sink.send` fails.
To be on the safe-side you would need to something like:

```rust
impl MyServerTrait for () {
    fn subscribe_to_something(&self, mut sink: SubscriptionSink) -> RpcResult<()> {
        // assume this stream lives forever and produces items "rarely"...
        let stream = create_stream();

        let fut = async move {
            loop {
                let timeout = tokio::time::sleep(std::time::Duration::from_secs(60));
                tokio::pin!(timeout);

                tokio::select! {
                    Some(item) = stream.next() => {
                        if let Err(e) = sink.send(&item) {
                          break;
                        }
                    },
                    _ = &mut timeout => {
                        if sink.is_closed() {
                            break;
                        }
                    }
                    else => break,
                };
            }
        };

        tokio::spawn(fut);
    }
}
```

niklasad1:
Fixes https://github.com/paritytech/jsonrpsee/issues/627

This PR introduces a new API for the `SubscriptionSink` to add a stream to send items over the subscription when "new items" get produced, if the connection gets closed it will terminate the task because it has additional channel linked to actual connection.

### DIFF
--- a/core/src/server/rpc_module.rs
+++ b/core/src/server/rpc_module.rs
@@ -746,12 +746,26 @@ impl SubscriptionSink {
 		self.inner_send(msg).map_err(Into::into)
 	}
 
-	/// Read data from the stream and send back data on the subscription when items gets produced by the stream.
+	/// Consumes the `SubscriptionSink` and reads data from the `stream` and sends back data on the subscription
+	/// when items gets produced by the stream.
 	///
 	/// Returns `Ok(())` if the stream or connection was terminated.
 	/// Returns `Err(_)` if one of the items couldn't be serialized.
 	///
-	pub async fn streamify<S, T>(mut self, mut stream: S) -> Result<(), Error>
+	/// # Examples
+	///
+	/// ```no_run
+	///
+	/// use jsonrpsee_core::server::rpc_module::RpcModule;
+	///
+	/// let mut m = RpcModule::new(());
+	/// m.register_subscription("sub", "_", "unsub", |params, mut sink, _| {
+	///     let stream = futures_util::stream::iter(vec![1_u32, 2, 3]);
+	///     tokio::spawn(sink.consume_and_streamify(stream));
+	///     Ok(())
+	/// });
+	/// ```
+	pub async fn consume_and_streamify<S, T>(mut self, mut stream: S) -> Result<(), Error>
 	where
 		S: Stream<Item = T> + Unpin,
 		T: Serialize,
@@ -769,7 +783,10 @@ impl SubscriptionSink {
 							self.close(&close_reason);
 							break Ok(());
 						}
-						Err(err) => break Err(err),
+						Err(err) => {
+							tracing::warn!("Subscription got error: {:?} terminating task", err);
+							break Err(err)
+						}
 					};
 					close = c;
 					item = stream.next();
--- a/core/src/server/rpc_module.rs
+++ b/core/src/server/rpc_module.rs
@@ -746,8 +746,12 @@ impl SubscriptionSink {
 		self.inner_send(msg).map_err(Into::into)
 	}
 
-	/// Consume the sink by passing a stream to be sent via the sink.
-	pub async fn add_stream<S, T>(mut self, mut stream: S)
+	/// Read data from the stream and send back data on the subscription when items gets produced by the stream.
+	///
+	/// Returns `Ok(())` if the stream or connection was terminated.
+	/// Returns `Err(_)` if one of the items couldn't be serialized.
+	///
+	pub async fn read_stream_and_send<S, T>(mut self, mut stream: S) -> Result<(), Error>
 	where
 		S: Stream<Item = T> + Unpin,
 		T: Serialize,
@@ -759,9 +763,11 @@ impl SubscriptionSink {
 		loop {
 			match futures_util::future::select(item, close).await {
 				Either::Left((Some(i), c)) => {
-					if let Err(Error::SubscriptionClosed(_)) = self.send(&i) {
-						break;
-					}
+					match self.send(&i) {
+						Ok => (),
+						Err(Error::SubscriptionClosed(_)) => return Ok(()),
+						err => return err,
+					};
 					close = c;
 					item = stream.next();
 				}
@@ -772,7 +778,7 @@ impl SubscriptionSink {
 					close = close_stream.next();
 				}
 				// Stream or connection has been terminated.
-				Either::Right((None, _)) | Either::Left((None, _)) => break,
+				Either::Right((None, _)) | Either::Left((None, _)) => return Ok(()),
 			}
 		}
 	}
--- a/core/src/server/rpc_module.rs
+++ b/core/src/server/rpc_module.rs
@@ -51,7 +51,7 @@ use serde::{de::DeserializeOwned, Serialize};
 /// implemented as a function pointer to a `Fn` function taking four arguments:
 /// the `id`, `params`, a channel the function uses to communicate the result (or error)
 /// back to `jsonrpsee`, and the connection ID (useful for the websocket transport).
-pub type SyncMethod = Arc<dyn Send + Sync + Fn(Id, Params, &MethodSink, MaybeConnState) -> bool>;
+pub type SyncMethod = Arc<dyn Send + Sync + Fn(Id, Params, &MethodSink, Option<ConnState>) -> bool>;
 /// Similar to [`SyncMethod`], but represents an asynchronous handler and takes an additional argument containing a [`ResourceGuard`] if configured.
 pub type AsyncMethod<'a> = Arc<
 	dyn Send + Sync + Fn(Id<'a>, Params<'a>, MethodSink, ConnectionId, Option<ResourceGuard>) -> BoxFuture<'a, bool>,
@@ -61,8 +61,6 @@ pub type AsyncMethod<'a> = Arc<
 pub type ConnectionId = usize;
 /// Raw RPC response.
 pub type RawRpcResponse = (String, mpsc::UnboundedReceiver<String>, mpsc::UnboundedSender<String>);
-/// Connection state for stateful protocols such as WebSocket
-pub type MaybeConnState<'a> = Option<ConnState<'a>>;
 
 /// Data for stateful connections.
 pub struct ConnState<'a> {
@@ -76,7 +74,7 @@ pub struct ConnState<'a> {
 
 impl<'a> std::fmt::Debug for ConnState<'a> {
 	fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-		f.debug_struct("Server").field("conn_id", &self.conn_id).field("close", &self.close).finish()
+		f.debug_struct("ConnState").field("conn_id", &self.conn_id).field("close", &self.close).finish()
 	}
 }
 
@@ -175,7 +173,7 @@ impl MethodCallback {
 	pub fn execute(
 		&self,
 		sink: &MethodSink,
-		conn_state: MaybeConnState<'_>,
+		conn_state: Option<ConnState>,
 		req: Request<'_>,
 		claimed: Option<ResourceGuard>,
 	) -> MethodResult<bool> {
@@ -325,7 +323,7 @@ impl Methods {
 	}
 
 	/// Attempt to execute a callback, sending the resulting JSON (success or error) to the specified sink.
-	pub fn execute(&self, sink: &MethodSink, conn_state: MaybeConnState, req: Request) -> MethodResult<bool> {
+	pub fn execute(&self, sink: &MethodSink, conn_state: Option<ConnState>, req: Request) -> MethodResult<bool> {
 		tracing::trace!("[Methods::execute] Executing request: {:?}", req);
 		match self.callbacks.get(&*req.method) {
 			Some(callback) => callback.execute(sink, conn_state, req, None),
@@ -341,7 +339,7 @@ impl Methods {
 	pub fn execute_with_resources<'r>(
 		&self,
 		sink: &MethodSink,
-		conn_state: MaybeConnState<'r>,
+		conn_state: Option<ConnState<'r>>,
 		req: Request<'r>,
 		resources: &Resources,
 	) -> Result<(&'static str, MethodResult<bool>), Cow<'r, str>> {
@@ -727,7 +725,7 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 			self.methods.mut_callbacks().insert(
 				unsubscribe_method_name,
 				MethodCallback::new_sync(Arc::new(move |id, params, sink, conn_state| {
-					let conn = conn_state.expect("conn must be Some; this is bug");
+					let c = conn_state.expect("conn must be Some; this is bug");
 
 					let sub_id = match params.one::<RpcSubscriptionId>() {
 						Ok(sub_id) => sub_id,
@@ -747,7 +745,7 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 
 					if subscribers
 						.lock()
-						.remove(&SubscriptionKey { conn_id: conn.conn_id, sub_id: sub_id.clone() })
+						.remove(&SubscriptionKey { conn_id: c.conn_id, sub_id: sub_id.clone() })
 						.is_some()
 					{
 						sink.send_response(id, "Unsubscribed")
--- a/core/src/server/rpc_module.rs
+++ b/core/src/server/rpc_module.rs
@@ -776,15 +776,15 @@ impl SubscriptionSink {
 
 		loop {
 			match futures_util::future::select(item, close).await {
-				Either::Left((Some(i), c)) => {
-					match self.send(&i) {
+				Either::Left((Some(result), c)) => {
+					match self.send(&result) {
 						Ok(_) => (),
 						Err(Error::SubscriptionClosed(close_reason)) => {
 							self.close(&close_reason);
 							break Ok(());
 						}
 						Err(err) => {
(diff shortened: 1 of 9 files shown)


core/src/server/rpc_module.rs
```diff
@@ -51,16 +51,35 @@ use serde::{de::DeserializeOwned, Serialize};
...
-	dyn Send + Sync + Fn(Id<'a>, Params<'a>, MethodSink, Option<ResourceGuard>, &dyn IdProvider) -> BoxFuture<'a, bool>,
+	dyn Send + Sync + Fn(Id<'a>, Params<'a>, MethodSink, ConnectionId, Option<ResourceGuard>) -> BoxFuture<'a, bool>,
 >;
 /// Connection ID, used for stateful protocol such as WebSockets.
 /// For stateless protocols such as http it's unused, so feel free to set it some hardcoded value.
 pub type ConnectionId = usize;
 /// Raw RPC response.
 pub type RawRpcResponse = (String, mpsc::UnboundedReceiver<String>, mpsc::UnboundedSender<String>);
+/// Connection state for stateful protocols such as WebSocket
+/// This is used to keep track whether the connection has been closed.
```
niklasad1, line 65:
```suggestion
```

core/src/server/rpc_module.rs
```diff
@@ -51,16 +51,34 @@ use serde::{de::DeserializeOwned, Serialize};
...
 pub type RawRpcResponse = (String, mpsc::UnboundedReceiver<String>, mpsc::UnboundedSender<String>);
+/// Connection state for stateful protocols such as WebSocket
+pub type MaybeConnState<'a> = Option<ConnState<'a>>;
+
+/// Data for stateful connections.
+pub struct ConnState<'a> {
+	/// Connection ID
+	pub conn_id: ConnectionId,
+	/// Channel to know whether the connection is closed or not.
+	pub close: async_channel::Receiver<()>,
```
niklasad1, line 72: NOTE: [this panics if count > usize::MAX / 2](https://github.com/smol-rs/async-channel/blob/master/src/lib.rs#L692-#L696)

but if we reach that we likely have other problems such as OOM :)

tests/tests/integration_tests.rs
```diff
@@ -359,6 +359,44 @@ async fn ws_server_should_stop_subscription_after_client_drop() {
 	assert!(matches!(close_err.close_reason(), &SubscriptionClosedReason::ConnectionReset));
 }
 
+#[tokio::test]
+async fn ws_server_cancel_stream_after_reset_conn() {
```
dvdplm, line 363:
```suggestion
async fn ws_server_cancels_stream_after_reset_conn() {
```

tests/tests/integration_tests.rs
```diff
@@ -359,6 +359,44 @@ async fn ws_server_should_stop_subscription_after_client_drop() {
...
+	let client = WsClientBuilder::default().build(&server_url).await.unwrap();
+	let _sub1: Subscription<usize> =
+		client.subscribe("subscribe_never_produce", None, "unsubscribe_never_produce").await.unwrap();
+	let _sub2: Subscription<usize> =
+		client.subscribe("subscribe_never_produce", None, "unsubscribe_never_produce").await.unwrap();
+
+	// terminate connection.
+	drop(client);
+	assert_eq!(Some(()), rx.next().await, "subscription stream should be terminated after the client was dropped");
+	assert_eq!(Some(()), rx.next().await, "subscription stream should be terminated after the client was dropped");
```
dvdplm, line 397: I felt like I understood the test until I got here. I thought `rx` would produce `None` after the client was dropped and so it'd be `assert!(rx.next().await.is_none())`. What am I missing? :/

dvdplm (commented):
This looks pretty good to me, what is it you want to do more here?

tests/tests/integration_tests.rs
niklasad1: haha, maybe it's more clear as you described it but the test actually sends a message on the channel when the subscription terminated.

niklasad1:
> This looks pretty good to me, what is it you want to do more here?

It should be complete but the grumbles is maybe the API (naming) and cloning a bunch `async_channel::Receiver`...

core/src/server/rpc_module.rs
```diff
@@ -426,8 +435,11 @@ impl Methods {
 	async fn inner_call(&self, req: Request<'_>) -> RawRpcResponse {
 		let (tx, mut rx) = mpsc::unbounded();
 		let sink = MethodSink::new(tx.clone());
+		let (_tx, rx_2) = async_channel::unbounded();
 
-		if let MethodResult::Async(fut) = self.execute(&sink, req, 0, &RandomIntegerIdProvider) {
+		let conn_state = Some(ConnState { conn_id: 0, close: rx_2, id_provider: &RandomIntegerIdProvider });
+
+		if let MethodResult::Async(fut) = self.execute(&sink, conn_state, req) {
```
niklasad1, line 442: TODO: fix this to test the subscription closed properly which now works :)

core/src/server/rpc_module.rs
```diff
@@ -786,9 +808,30 @@ impl SubscriptionSink {
...
+				// Stream or connection was dropped => close stream.
+				else => break,
+			}
+		}
+	}
+
 	/// Returns whether this channel is closed without needing a context.
 	pub fn is_closed(&self) -> bool {
-		self.inner.is_closed()
+		self.inner.is_closed() || self.close.is_closed()
```
niklasad1, line 834: this is probably needless both should be dropped at the same time...

core/src/server/rpc_module.rs
```diff
@@ -671,12 +683,13 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 			let subscribers = subscribers.clone();
 			self.methods.mut_callbacks().insert(
 				subscribe_method_name,
-				MethodCallback::new_sync(Arc::new(move |id, params, method_sink, conn_id, id_provider| {
+				MethodCallback::new_sync(Arc::new(move |id, params, method_sink, conn| {
 					let (conn_tx, conn_rx) = oneshot::channel::<()>();
+					let c = conn.expect("conn must be Some; this is bug");
```
maciejhirsz, line 688: If it bust be `Some`, why don't we restrict it on parameter level?

core/src/server/rpc_module.rs
```diff
@@ -157,24 +173,23 @@ impl MethodCallback {
...
 
 		let result = match &self.callback {
 			MethodKind::Sync(callback) => {
 				tracing::trace!(
-					"[MethodCallback::execute] Executing sync callback, params={:?}, req.id={:?}, conn_id={:?}",
+					"[MethodCallback::execute] Executing sync callback, params={:?}, req.id={:?}, conn_state={:?}",
 					params,
 					id,
-					conn_id
+					conn_state
```
maciejhirsz, line 189: I'd rather unwrap `conn_state` here than later (as per other comment), but it's still not clear to me why it's mandatory here but optional for async calls.

niklasad1: I moved `conn_id, id_provider, close` to `ConnState` before we required to pass `conn_id` explictly here which is unused.

Thus, for HTTP (with connection state) you would just pass `execute(&sink, req, None, resources)` however I get your point that this is error prone and you can't really `unwrap` here but `subscriptions` must be initialized with `Some(conn_state)` if that makes sense.

What do say about `execute_with_state` and `execute`? However I guess we would need additional callback types then.... :grimacing:

core/src/server/rpc_module.rs
niklasad1: it's way better but I think it requires additional callback types right? see my comment below.

thoughts?

niklasad1: Ok, I had a look at the code to solve this properly at `parameter level` one need to lookup at the actual callback to avoid to pass down a bunch of unused parameters.

```rust
  // server code
  let result = match RpcModule::as_callback(&req.method) {
     None => {
        send_error(sink, msg),
        continue
     }
     Some(MethodKind::Sync(cb)) => (cb)(id, params, sink, conn_id),
     Some(MethodKind::Async(cb)) => (cb)(id, params, sink, resources).await,
     // should be used for subscriptions...
     // servers that don't support subscriptions should throw an error here...
     Some(MethodKind::Subscription) => (cb)(id, params, sink, resources, conn_state),
  };

  // modify register 
  pub fn register_subscription<F>(
     &mut self,
     notif_method_name: &'static str,
     callback: F,
 ) -> Result<(), Error> {
   .... 
   ....
   self.methods.mut_callbacks().insert(
      subscribe_method_name,
      MethodCallback::new_subscription(Arc::new(move |id, params, method_sink, conn| {
        ... 
     }
  );
 }


```

niklasad1: one benefit is not to have to clone this "introduced channel connection_state" for every method instead only for subscriptions where it's actually used... not sure I liked how the old code abstracted this away

maciejhirsz: I already thought about splitting async calls from subscriptions when doing my refactoring earlier, and for regular method calls returning values instead of passing in the sink as a parameter, I reckon that would make things much more readable and straight forward, and potentially make the final binary smaller. So if you want to go that route and add another enum variant I think that's cool, and I can do a PR that switches method calls to have a return value later on :).

niklasad1: ah, I already added the enum variant in this PR but just a hacky draft to check that it works and to show you and David what I had in mind :)

niklasad1: sounds great to get rid of the sink for the synchronous calls.

core/src/server/rpc_module.rs
```diff
@@ -437,11 +459,18 @@ impl Methods {
...
 
-		let conn_state = Some(ConnState { conn_id: 0, close: close_rx, id_provider: &RandomIntegerIdProvider });
+		let id = req.id.clone();
+		let params = Params::new(req.params.map(|params| params.get()));
 
-		if let MethodResult::Async(fut) = self.execute(&sink, conn_state, req) {
-			fut.await;
-		}
+		let _result = match self.as_callback(&req.method).map(|c| &c.callback) {
+			None => todo!(),
```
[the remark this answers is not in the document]
niklasad1: exactly, forgot this good catch

ws-server/src/server.rs
```diff
@@ -301,21 +306,23 @@ async fn background_task(
...
 		}
-		// terminate connection.
+
+		// Terminate connection and send close message.
 		let _ = sender.close().await;
-		// NOTE(niklasad1): when the receiver is dropped no further requests or subscriptions
-		// will be possible.
+
+		// Force `conn_tx` to this async block and close it down
+		// when the connection closes to be on safe side.
```
niklasad1, line 324: //cc @dvdplm this is what I meant by closing down the channel.

once this tasks terminates we drop `conn_tx` before I dropped it close/drop shouldn't matter just that it's moved into this block.

core/src/server/rpc_module.rs
```diff
@@ -786,9 +746,40 @@ impl SubscriptionSink {
 		self.inner_send(msg).map_err(Into::into)
 	}
 
+	/// Consume the sink by passing a stream to be sent via the sink.
+	pub async fn add_stream<S, T>(mut self, mut stream: S)
+	where
+		S: Stream<Item = T> + Unpin,
+		T: Serialize,
+	{
+		let mut close_stream = self.close.clone();
```
niklasad1, line 755: this is hack; to workaround borrowing issues when we call `self.send` below.

we may refactor this to split this up to several types or something....

core/src/server/rpc_module.rs
```diff
@@ -786,9 +746,40 @@ impl SubscriptionSink {
 		self.inner_send(msg).map_err(Into::into)
 	}
 
+	/// Consume the sink by passing a stream to be sent via the sink.
+	pub async fn add_stream<S, T>(mut self, mut stream: S)
```
dvdplm, line 750: What do you think about calling this `into_stream`? I think "add" implies there could be more than one and that it doesn't quite relay the information about the important changes that this call makes to the sink.

core/src/server/rpc_module.rs
dvdplm: I think we need plenty of comments in this method, it requires some pretty deep understanding of the architecture and I think we should capture it here.

ws-server/src/server.rs
dvdplm: ok, ty

dvdplm (approved):

core/src/server/rpc_module.rs
dvdplm: Fair; I'd have liked `as_stream` but `as_` is "taken" with different semantics so can't do that.

Of your suggestions I like `from_stream` the best.

tests/tests/integration_tests.rs
```diff
@@ -359,6 +359,44 @@ async fn ws_server_should_stop_subscription_after_client_drop() {
...
+	let mut module = RpcModule::new(tx);
+
+	module
+		.register_subscription("subscribe_never_produce", "n", "unsubscribe_never_produce", |_, sink, mut tx| {
+			// create stream that doesn't produce items.
+			let stream = futures::stream::empty::<usize>();
+			tokio::spawn(async move {
+				sink.add_stream(stream).await;
+				let send_back = Arc::make_mut(&mut tx);
+				send_back.feed(()).await.unwrap();
```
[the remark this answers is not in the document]
niklasad1: No, it's not the `SubscriptionSink` that is using `feed()`, it's the `mpsc::channel` to send messages back for testing purposes.

`feed` is another method on the `SinkExt trait` that doesn't require flushing as `send` does, https://docs.rs/futures/0.3.19/futures/sink/trait.SinkExt.html#method.feed. 

TLDR;  it's doesn't matter....

core/src/server/rpc_module.rs
jsdw: So, this method consumes a stream, feeding the items into the subscription?

I guess I'd go with something like `consume_stream` or `read_from_stream`. `into_`, `as_`, and `from_` all sortof feel like I should expect some result back from this call to me!

niklasad1: I guess we could return a type that impls `Sink/SinkExt` instead here to make it more readable and flexible i.e, to deal with errors and so on.

dvdplm: `streamify()`? 
I think `consume_stream` is so-so. Yes, we do consume it, but that's not really the point. Rather we're "hooking up" the stream to the sink and leave it there for the duration of the subscription.
`with_stream`?

core/src/server/rpc_module.rs
```diff
@@ -786,9 +746,46 @@ impl SubscriptionSink {
...
+					match self.send(&i) {
+						Ok(_) => (),
+						Err(Error::SubscriptionClosed(_)) => break Ok(()),
+						Err(err) => break Err(err),
+					};
+					close = c;
+					item = stream.next();
+				}
+				// No messages should be sent over this channel
+				// if that's occur just continue.
```
niklasad1, line 775:
```suggestion
				// if that occurs just continue.
```

core/src/server/rpc_module.rs
jsdw: `pipe`, maybe? we're piping a stream into the subscription

dvdplm: `sink.pipe_from_stream`? I quite like `pipe`!

core/src/server/rpc_module.rs
```diff
@@ -786,9 +746,71 @@ impl SubscriptionSink {
...
+				Either::Left((Some(i), c)) => {
+					match self.send(&i) {
+						Ok(_) => (),
+						Err(Error::SubscriptionClosed(close_reason)) => {
+							self.close(&close_reason);
+							break Ok(());
+						}
+						Err(err) => {
+							tracing::warn!("Subscription got error: {:?} terminating task", err);
+							break Err(err);
```
niklasad1, line 788: this is most likely a bug when serializing T fails, should we have a flag for this when the exit, just log an continue, or something else?

The error is not really possible to handle but if you require each element to be processed in order, ignore might break expectations so I just terminated the "task" when an error is detected.

core/src/server/rpc_module.rs
niklasad1: I like `pipe_from_stream`, let's settle for that?

core/src/server/rpc_module.rs
dvdplm: I think you made the right choice. We could also panic I guess but let's terminate for now.

jsdw (approved):
Looks good to me overall, and I like the simplification in substrate!

maciejhirsz (approved):
Nice, good job separating subscriptions!

Issue #638, opened 2022-01-05
WS to HTTP redirection doesn't seem to be handle correctly


paulormart:
@paulormart 

please open an issue in `jsonrpsee` but the issue is that you got a HTTP redirection there... to a HTTP endpoint

```
target: Target { sockaddrs: [104.26.8.125:80, 104.26.9.125:80, 172.67.71.26:80, [2606:4700:20::681a:97d]:80, [2606:4700:20::ac43:471a]:80, [2606:4700:20::681a:87d]:80], host: "rpc.polkadot.io", host_header: "rpc.polkadot.io:80", _mode: Plain, path_and_query: "/" }
2022-01-05T15:32:34.132404Z DEBUG jsonrpsee_client_transport::ws: Connecting to target: Target { sockaddrs: [104.26.8.125:80, 104.26.9.125:80, 172.67.71.26:80, [2606:4700:20::681a:97d]:80, [2606:4700:20::ac43:471a]:80, [2606:4700:20::681a:87d]:80], host: "rpc.polkadot.io", host_header: "rpc.polkadot.io:80", _mode: Plain, path_and_query: "/" }
2022-01-05T15:32:34.266608Z DEBUG jsonrpsee_client_transport::ws: Redirection: status_code: 301, location: https://rpc.polkadot.io/
thread 'main' panicked at 'Some("https")', client/transport/src/ws/mod.rs:440:17
```

Thus, meanwhile use `wss://rpc.polkadot.io:443` if that's alright for you.

_Originally posted by @niklasad1 in https://github.com/paritytech/subxt/issues/384#issuecomment-1005824575_

paulormart:
Thanks for reviewing this @niklasad1 

If `wss://rpc.polkadot.io:443` is used from the `subxt` side we get this error `Error: Rpc(Transport(Invalid URL: URL scheme not supported, expects 'ws'))`

paulormart:
Ok got it. `client-ws-transport` feature was missing from the client library.

Cargo file should specify it like `jsonrpsee = { version = "0.7.0", features = ["macros", "ws-client", "http-client", "client-ws-transport"] }`

niklasad1:
The redirection is still an issue but I don't think there is anything we can do beside print a better error message in the client.

> Ok got it. client-ws-transport feature was missing from the client library.

good catch, it's a bug the ws-client crate should expose that feature but you can solve it like that too :)

PR #642, opened 2022-01-06
fix(ws client): improve error message bad URL

niklasad1:
Fixes/improves the error message when a bad URL is detected

Partly addresses https://github.com/paritytech/jsonrpsee/issues/638 to print an error when an absolute URI 
in a redirection couldn't be parsed as a WS target.

### DIFF
--- a/client/transport/src/ws/mod.rs
+++ b/client/transport/src/ws/mod.rs
@@ -286,7 +286,7 @@ impl<'a> WsTransportClientBuilder<'a> {
 							Ok(uri) => {
 								// Absolute URI.
 								if uri.scheme().is_some() {
-									let uri: Target = uri.try_into().map_err(|e| {
+									let target: Target = uri.try_into().map_err(|e| {
 										tracing::error!("Redirection failed: {:?}", e);
 										e
 									})?;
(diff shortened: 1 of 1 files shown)


client/transport/src/ws/mod.rs
```diff
@@ -434,11 +437,11 @@ impl TryFrom<Uri> for Target {
...
 			#[cfg(feature = "tls")]
 			Some("wss") => Mode::Tls,
-			_ => {
+			invalid_scheme => {
 				#[cfg(feature = "tls")]
-				let err = "URL scheme not supported, expects 'ws' or 'wss'";
+				let err = format!("{:?} not supported, expects 'ws' or 'wss'", invalid_scheme);
 				#[cfg(not(feature = "tls"))]
-				let err = "URL scheme not supported, expects 'ws'";
+				let err = format!("{:?} not supported, expects 'ws' ('wss' requires the tls feature)", invalid_scheme);
```
jsdw, line 444: This will print something like `Some("http") not supported, expects 'ws' or 'wss'`; are we happy with that or would it be better to maybe do something like `invalid_scheme.unwrap_or("no scheme")`?

jsdw (approved):
LGTM; minor question on error message format!

client/transport/src/ws/mod.rs
niklasad1: Right, I think you suggestion is better.

maciejhirsz (approved):

Issue #480, opened 2021-09-20
Serde_json arbitrary precision


717a56e1:
`jsonrpc` had a feature flag to allow `arbitrary_precision`, this library currently breaks if it is built with it. Could be we open to adding a similar `serde_from_str` override?

dvdplm:
Yes, I think we need to add that. It's not pretty but I don't see a better way at the moment.

For other readers, here's some context:
- https://github.com/serde-rs/json/issues/505 describes the problem and suggests [the work-around that is used in `jsonrpc`](https://docs.rs/jsonrpc-core/18.0.0/src/jsonrpc_core/lib.rs.html#69-79).
- the likely underlying cause in serde: https://github.com/serde-rs/serde/issues/1183
- at the deep end of the rabbit hole we have the "Associated type defaults" [RFC](https://github.com/rust-lang/rust/issues/29661) and all the fall-out from that

@717a56e1 Are you open to contributing a PR?

niklasad1:
I agree but IIRC this affects mainly `u128/i128` and floats but comes with extra costs.

717a56e1:
Yup, got it running with a patched `from_slice`, should get up PR up. Any place we'd like to put the function? I've got it in `types/lib.rs`.

dvdplm:
Probably better off in `jsonrpsee-utils` I think. @niklasad1 wdyt?

niklasad1:
I don't know really because we re-export serde stuff from types so a utility function in types probably makes most sense but I need to see to code first before saying yes or no :)

Either should be fine I think...

niklasad1:
Hey, I have investigated `arbitrary precision` feature in `serde` defeats our `zero-alloc` parsing of `RpcParams` and it's really inefficient in practice one would have to do `json_str -> serde_json::Value -> T`

Further, as the Request/Response types in `jsonrpsee` doesn't use `#[serde(untagged/flatten)]` those will work anyway but there are two cases where `arbitrary precision` are needed at to moment:

1. custom types as "JSON-RPC params" or "JSON-RPC result" that have `#[serde(untagged/flatten)] with u128/u128 as field ` (no way to mitigate those)
2. [ParamsSer](https://github.com/paritytech/jsonrpsee/blob/master/types/src/v2/params.rs#L258-#L267) that is currently only used by the client but could be fixed.

[...]

PR #644, opened 2022-01-09
Check string Id in response

whalelephant:
Hi, 

I would like to add support checking `Repsonse.id` as `Id::Str(x)` when client is parsing responses.

Not sure if this is something already the works so please let me know. thanks!

### DIFF
--- a/client/http-client/src/client.rs
+++ b/client/http-client/src/client.rs
@@ -192,12 +192,26 @@ impl ClientT for HttpClient {
 		// NOTE: `R::default` is placeholder and will be replaced in loop below.
 		let mut responses = vec![R::default(); ordered_requests.len()];
 		for rp in rps {
-			let response_id = rp.id.as_number().copied().ok_or(Error::InvalidRequestId)?;
-			let pos = match request_set.get(&response_id) {
-				Some(pos) => *pos,
-				None => return Err(Error::InvalidRequestId),
+			match rp.id.as_number().copied() {
+				Some(id) => {
+					let pos = match request_set.get(&id) {
+						Some(pos) => *pos,
+						None => return Err(Error::InvalidRequestId),
+					};
+					responses[pos] = rp.result
+				}
+				None => match rp.id.as_str() {
+					Some(s) => {
+						let id: u64 = s.trim().parse().map_err(|_| Error::InvalidRequestId)?;
+						let pos = match request_set.get(&id) {
+							Some(pos) => *pos,
+							None => return Err(Error::InvalidRequestId),
+						};
+						responses[pos] = rp.result
+					}
+					None => return Err(Error::InvalidRequestId),
+				},
 			};
-			responses[pos] = rp.result
 		}
 		Ok(responses)
 	}
--- a/client/http-client/src/client.rs
+++ b/client/http-client/src/client.rs
@@ -142,10 +142,18 @@ impl ClientT for HttpClient {
 			}
 		};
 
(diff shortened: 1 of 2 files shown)


niklasad1:
Hey @whalelephant 

Thanks for your PR, the reason why we assert that the response ID must be a `Number` is because the library itself only transmit method calls with `ID` as `Number`.

Can you elaborate why you need `IDs` as `String`? If that's desired it's should be configurable to choose request ID type or allow users to provide their own IDs such as unique IDs or something..

whalelephant:
Hi @niklasad1 

> the reason why we assert that the response ID must be a Number is because the library itself only transmit method calls with ID as Number.

makes sense! 

> Can you elaborate why you need IDs as String? If that's desired it's should be configurable to choose request ID type or allow users to provide their own IDs such as unique IDs or something..

Sure! I think the main reason is that I cannot guarentee the server parses / deserializes to the same Id type in the response since Id supports both str and num. This all comes from the investigation related to deserializing the `u64` type (mentioned #480 and [here](https://github.com/AleoHQ/snarkOS/issues/1369)). A way to parse any type is to read the Id as `serde_json::value::RawValue`, which mean it may be returned as a string. 

Thanks!

niklasad1:
https://github.com/AleoHQ/snarkOS/issues/1369 is not related this library but I see that you need to workaround this because it suffers from serde bug that it can't deduce the number type properly probably because of the `arbitrary precision feature`

FWIW, we are not supporting `arbitrary precision` in `jsonrpsee servers` and doesn't suffer from this bug so #480 is not an issue for this case but won't work for `u128` or `untagged serde enums` (unless someone enables that feature on serde_json). We are not planning to support it, I should explain it in the issue.

From the "clients" point-of-view the server must be reply with the same `id` for the request to be completed (1 == 1 &&  1 != "1"), the example https://github.com/AleoHQ/snarkOS/issues/1369 when replies with `null` the client can't know which request that failed (expect HTTP).

With that said if you could implement a configuration on the clients in the builders to support `numeric IDs` or `Strings (maybe random generated controlled by jsonrpsee)` I would accept it and solve your problems right?

client/http-client/src/client.rs
```diff
@@ -142,10 +142,18 @@ impl ClientT for HttpClient {
...
+			}
+		} else if let Some(response_id) = response.id.as_str() {
+			if response_id == id.inner().to_string() {
+				Ok(response.result)
+			} else {
+				Err(Error::InvalidRequestId)
+			}
 		} else {
 			Err(Error::InvalidRequestId)
 		}
```
maciejhirsz, line 159: This would have been much cleaner if it was a `match` on the enum variants.

client/http-client/src/client.rs
```diff
@@ -184,12 +192,26 @@ impl ClientT for HttpClient {
...
+					Some(s) => {
+						let id: u64 = s.trim().parse().map_err(|_| Error::InvalidRequestId)?;
+						let pos = match request_set.get(&id) {
+							Some(pos) => *pos,
+							None => return Err(Error::InvalidRequestId),
+						};
+						responses[pos] = rp.result
+					}
+					None => return Err(Error::InvalidRequestId),
+				},
```
maciejhirsz, line 213: ditto

maciejhirsz (approved):
~~Looks good, minor grumbles.~~

Nevermind, I see there is another PR that replaces this one.

dvdplm:
Closing in favor of https://github.com/paritytech/jsonrpsee/pull/659

Issue #606, opened 2021-12-13
[ws-client] Make wasm compatible


athei:
There is #227 which covers making the http client wasm compatible. However, I think the ws client is much more important because it is the only one allowing subscriptions. It would allow for creating web frontends written in Rust through `subxt`.

With wasm compatible I mean "runs in a browser". This means making use of `web-sys` to interact with the browser.

niklasad1:
> However, I think the ws client is much more important because it is the only one allowing subscription

yeah, indeed.

It's been a while since I looked at the https://github.com/najamelan/ws_stream_wasm but it won't because the `"async/ws" client` assumes it assumes that it can be split into a sender and receiver because we using them in different tasks.

However, https://github.com/nash-io/nash-ws looks quite promising. I think it should be straight-forward to port over.

athei:
Maybe its worthwhile to use [`web-sys`](https://rustwasm.github.io/wasm-bindgen/api/web_sys/struct.WebSocket.html) directly. This crate looks kind of abandoned.

niklasad1:
Closed by #648

PR #648, opened 2022-01-12
feat: WASM client via web-sys transport

niklasad1:
This PR adds `web-sys` transport to the jsonrpsee client in the browser (via WebSocket), it actually uses "std" currently but I'm not sure whether this will be an issue or not for users.

I added another feature flag `async-wasm-client` to `core` to use a client to that spawns a background task using `wasm_bindgen_futures::spawn_local` instead of `tokio` similar to how `async-client` works. We can probably bikeshed on the naming there.

Another option is to go by `target.'cfg(target_arch = wasm32 )' instead to reduce the number of complicated feature flags to understand in this repo...

Closing #606


client/transport/src/web_sys.rs
```diff
@@ -0,0 +1,98 @@
...
+	websocket.set_onopen(Some(on_open_callback.as_ref().unchecked_ref()));
+	websocket.set_onmessage(Some(on_msg_callback.as_ref().unchecked_ref()));
+	websocket.set_onclose(Some(on_close_callback.as_ref().unchecked_ref()));
+
+	// Prevent for being dropped (this will be leaked intentionally).
+	on_msg_callback.forget();
+	on_open_callback.forget();
+	on_close_callback.forget();
+
+	conn_rx.next().await;
```
niklasad1, line 95: this is a channel because the `FnMut closures` then `oneshot` didn't work

core/src/client/async_client/mod.rs
```diff
@@ -130,9 +138,33 @@ impl ClientBuilder {
...
+		S: TransportSenderT,
+		<S as TransportSenderT>::Error: Sync,
+		R: TransportReceiverT,
+		<R as TransportReceiverT>::Error: Sync,
+	{
+		let (to_back, from_front) = mpsc::channel(self.max_concurrent_requests);
+		let (err_tx, err_rx) = oneshot::channel();
+		let max_notifs_per_subscription = self.max_notifs_per_subscription;
+
+		wasm_bindgen_futures::spawn_local(async move {
```
niklasad1, line 155: https://crates.io/crates/wasm-futures-executor is a candidate but doesn't compile for stable rust yet....

client/transport/src/web_sys.rs
```diff
@@ -0,0 +1,98 @@
...
+	let (tx, rx) = mpsc::unbounded();
+
+	let websocket = WebSocket::new(url.as_ref()).map_err(|_| ())?;
+	websocket.set_binary_type(web_sys::BinaryType::Arraybuffer);
+
+	let tx1 = tx.clone();
+
+	let on_msg_callback = Closure::wrap(Box::new(move |e: MessageEvent| {
+		// Binary message.
+		if let Ok(abuf) = e.data().dyn_into::<js_sys::ArrayBuffer>() {
```
niklasad1, line 60: decode with serde?!

client/transport/src/web_sys.rs
```diff
@@ -0,0 +1,98 @@
...
+use wasm_bindgen::closure::Closure;
+use wasm_bindgen::{JsCast, JsValue};
+use web_sys::{CloseEvent, MessageEvent, WebSocket};
+
+/// Sender.
+#[derive(Debug)]
+pub struct Sender(WebSocket);
+
+// TODO: safety.
+unsafe impl Send for Sender {}
```
niklasad1, line 14: I could fix it with another channel...

client/transport/src/web_sys.rs
```diff
@@ -0,0 +1,124 @@
...
+
+#[derive(Debug)]
+enum WebSocketMessage {
+	Data(String),
+	Close,
+}
+
+/// Sender.
+#[derive(Debug)]
+pub struct Sender(mpsc::UnboundedSender<WebSocketMessage>);
```
niklasad1, line 22: https://github.com/rustwasm/wasm-bindgen/pull/955

The `JsValue` is !Send and I had to use another channel here instead of having `WebSocket` here directly.

core/src/client/async_client/manager.rs
```diff
@@ -32,14 +32,16 @@
...
 use crate::Error;
+use core::hash::BuildHasherDefault;
 use futures_channel::{mpsc, oneshot};
+use hashbrown::hash_map::{Entry, HashMap};
 use jsonrpsee_types::{Id, SubscriptionId};
-use rustc_hash::FxHashMap;
+use rustc_hash::FxHasher;
 use serde_json::value::Value as JsonValue;
 
+type FxHashMap<K, V> = HashMap<K, V, BuildHasherDefault<FxHasher>>;
```
dvdplm, line 43: Is `FxHashMap` not wasm-compatible? I would have thought it was defined pretty much like you did here?

core/src/client/async_client/mod.rs
```diff
@@ -115,7 +116,12 @@ impl ClientBuilder {
 	/// ## Panics
 	///
 	/// Panics if being called outside of `tokio` runtime context.
```
dvdplm, line 118:
```suggestion
	/// Panics if called outside of `tokio` runtime context.
```

core/src/client/async_client/mod.rs
dvdplm: I guess we could call the WASM client "experimental" and require nightly until the `wasm-futures-executor` stabilizes?

core/src/client/async_client/mod.rs
```diff
@@ -186,16 +208,10 @@ impl ClientT for Client {
...
-			_ = timeout => return Err(Error::RequestTimeout)
-		};
-
-		match res {
-			Ok(()) => Ok(()),
-			Err(_) => Err(self.read_error_from_backend().await),
+		match future::select(fut, Delay::new(self.request_timeout)).await {
+			Either::Left((Ok(()), _)) => Ok(()),
+			Either::Left((Err(_), _)) => Err(self.read_error_from_backend().await),
+			Either::Right((_, _)) => Err(Error::RequestTimeout),
```
dvdplm, line 214: To me this looks like it does exactly the same thing as before but this looks more readable. Is that the only reason you switched away from `select!`?

core/src/client/mod.rs
```diff
@@ -128,7 +129,7 @@ pub trait TransportSenderT: Send + 'static {
 /// Transport interface to receive data asynchronous.
 #[async_trait]
 pub trait TransportReceiverT: Send + 'static {
-	/// Error that occur during send or receiving a message.
+	/// Error.
```
dvdplm, line 132: Is that not true anymore?

core/src/client/async_client/mod.rs
```diff
@@ -130,9 +136,31 @@ impl ClientBuilder {
 			id_manager: RequestIdManager::new(self.max_concurrent_requests, self.id_kind),
 		}
 	}
+
+	/// Build the client with given transport.
+	#[cfg(feature = "async-wasm-client")]
+	pub fn build_with_wasm<S, R>(self, sender: S, receiver: R) -> Client
```
dvdplm, line 142: `build_for_wasm`/`build_for_tokio`?

dvdplm: Another option is to keep a single `build` method and then use `cfg_if` to conditionally compile just the spawning?

core/src/client/async_client/mod.rs
niklasad1: it seems be fast enough for now, let's open a issue and investigate the performance gain for it instead....

core/src/client/async_client/manager.rs
niklasad1: you are probably right, I don't remember.

core/src/client/async_client/mod.rs
niklasad1: yes because `tokio` doesn't compile for wasm AFAIU but maybe the macros does?!

core/src/client/async_client/mod.rs
niklasad1: The reason I did like this is to avoid to have `mutually exclusive features` for `async-client` and `async-wasm-client` this way it becomes much cleaner and it should rather `intuitive` because one have to explicit call `build_with_wasm/build_with_tokio`..

opinions?

dvdplm: You're right, this is better.

core/src/client/async_client/helpers.rs
```diff
@@ -225,12 +225,11 @@ pub(crate) fn process_error_response(manager: &mut RequestManager, err: ErrorRes
...
-	timeout: Duration,
-	rx: oneshot::Receiver<Result<T, Error>>,
-) -> Result<Result<T, Error>, oneshot::Canceled> {
-	let timeout = tokio::time::sleep(timeout);
-	tokio::select! {
-		res = rx => res,
-		_ = timeout => Ok(Err(Error::RequestTimeout))
+	timeout: std::time::Duration,
+	rx: futures_channel::oneshot::Receiver<Result<T, Error>>,
+) -> Result<Result<T, Error>, futures_channel::oneshot::Canceled> {
```
niklasad1, line 230: changed to `future::select` to make it work for wasm

jsdw:
> Another option is to go by `target.'cfg(target_arch = wasm32 )' instead to reduce the number of complicated feature flags to understand in this repo...

I wonder whether it's better to have a feature flag like "js"; just because it's compiled to wasm doesn't mean it'll be executed in a browser with access to the web-sys JS APIs (and I think some other crates use "js" but I'd have to go looking to check!) :)

athei:
> I wonder whether it's better to have a feature flag like "js"; just because it's compiled to wasm doesn't mean it'll be executed in a browser with access to the web-sys JS APIs

Does this have pluggable transpart (as in replaceable by a user)? If yes it makes sense to have a feature. If not: What is the point in compiling it for wasm without having access to the web APIs if there is no way to communicate? 

I would propose  using `web` as the feature flag name if there will be a feature flag.

niklasad1:
> Does this have pluggable transpart (as in replaceable by a user)?

That's a good point it's quite complicated to configure the client: the transport needs `--feature web-sys/wasm` for the and then the abstract client needs `--feature wasm-client` to spawn the event loop with `wasm_bindgen_futures::spawn_local`.

We might want to provide a thin wrapper crate that does that mess instead of the users.

But yeah `web` or `web-sys` sounds good to me for the transport then maybe we can get rid of the feature `async-client-wasm` with the thin wrapper crate :point_up:

core/src/client/async_client/mod.rs
```diff
@@ -114,8 +115,13 @@ impl ClientBuilder {
 	///
 	/// ## Panics
 	///
-	/// Panics if being called outside of `tokio` runtime context.
-	pub fn build<S: TransportSenderT, R: TransportReceiverT>(self, sender: S, receiver: R) -> Client {
+	/// Panics if called outside of `tokio` runtime context.
+	#[cfg(feature = "async-client")]
+	pub fn build_with_tokio<S, R>(self, sender: S, receiver: R) -> Client
```
niklasad1, line 120: ideally I would get rid of these but that would require some abstraction over the executor which I haven't found

core/Cargo.toml
```diff
@@ -8,44 +8,58 @@ license = "MIT"
...
 	"rustc-hash",
-	"tokio/sync",
-	"tokio/macros",
-	"tokio/time",
-	"tracing"
+	"tokio",
+	"tracing",
+	"futures-timer",
+]
+async-wasm-client = [
```
niklasad1, line 56: we could reduce the features the `async client` was extracted to a separate crate...

.github/workflows/ci.yml
```diff
@@ -149,3 +154,27 @@ jobs:
...
+
+ test_wasm:
+    name: Test wasm
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@master
+
+    - name: Install
+      run: curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
```
jsdw, line 166: I've done `cargo install wasm-pack` in the past; is there an advantage to doing it this way?

jsdw: Aah; one advantage of the script by the looks of it is that it'll download pre-compiled binaries if they exist; that's good to know!

client/transport/src/web_sys.rs
```diff
@@ -0,0 +1,90 @@
+#![cfg(target_arch = "wasm32")]
```
jsdw, line 1: I guess the cfg in lib.rs makes this unnecessary?

.github/workflows/ci.yml
```diff
@@ -149,3 +154,27 @@ jobs:
 
       - name: Cargo nextest
         run: cargo nextest run --workspace
+
+ test_wasm:
```
jsdw, line 158: super pedantic; needs an extra space in front (I can see only 1). Also how about `tests_wasm` (added "s") to be consistent with the other tests_x steps?

core/src/client/mod.rs
```diff
@@ -110,10 +111,28 @@ pub trait SubscriptionClientT: ClientT {
 }
 
 /// Transport interface to send data asynchronous.
+#[cfg(target_arch = "wasm32")]
+#[async_trait(?Send)]
+/// Transport interface for an asynchronous client.
+pub trait TransportSenderT: 'static {
```
jsdw, line 117: So I think the only difference between these wasm and "normal" traits are that wasm doesn't require `Send`. What would be the issue if `Send` existed here?

I wonder whether some duplication could be avoided by eg having a trait like `MaybeSend` that either implies `Send` or does not based on the feature flag, and then use that here instead?

jsdw (approved):
This looks fantastic!

core/src/client/mod.rs
niklasad1: This is `gloo-net` related those types are wrapped in `Rc` and other non-send types so won't work. If you have a look you can see that we split the `WebSocket type` to a sender and receiver which impls `TransportSenderT` and `TransportReceiverT`

First I added extra channel for this i.e, another to use in the `Sender` and `Receiver` types but I got so annoyed by the extra overhead.

Given our use-case i.e. `wasm-bindgen-futures` only can produce !Send futures it might be safe to implement `Send` for our types so I did this hack instead.

With that said `MaybeSend` would be really neat but not sure how to do it with `async-trait`

jsdw: I've no idea whether this would work, but I wondered about something like:

```
#[cfg(target_arch = "wasm32")]
trait MaybeSend: Send {}

#[cfg(not(target_arch = "wasm32"))]
trait MaybeSend {}

impl <T> MaybeSend for T {}
```

and then for the attr on the wasm async_trait:

```
#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
```

lexnv (approved):

PR #659, opened 2022-01-19
feat(client): support request id as Strings.

niklasad1:
Supersedes #644

It makes it possible to configure the clients to generate `request object ID` as a strings instead of numbers, I did the most simple way and converting the `numeric ID` to strings from the `RequestIdManager`. 

We might want to use some more optimized `itoa` instead of format in future or generate random IDs.

However, `numeric ids` is still default because it avoids some allocations.

The rationale behind this is that some `JSON-RPC servers` suffers from the issue where they can't decode `Id` as numbers because of the `serde-json arbritary-precision` bug, so this is a workaround for that.

//cc @whalelephant I think this should be what you need right?

### DIFF
--- a/client/http-client/src/client.rs
+++ b/client/http-client/src/client.rs
@@ -130,7 +130,7 @@ impl ClientT for HttpClient {
 	{
 		let guard = self.id_manager.next_request_id()?;
 		let id = guard.inner();
-		let request = RequestSer::new(id.clone(), method, params);
+		let request = RequestSer::new(&id, method, params);
 
 		let fut = self.transport.send_and_read_body(serde_json::to_string(&request).map_err(Error::ParseError)?);
 		let body = match tokio::time::timeout(self.request_timeout, fut).await {
@@ -171,7 +171,7 @@ impl ClientT for HttpClient {
 		let ids: Vec<Id> = guard.inner();
 
 		for (pos, (method, params)) in batch.into_iter().enumerate() {
-			batch_request.push(RequestSer::new(ids[pos].clone(), method, params));
+			batch_request.push(RequestSer::new(&ids[pos], method, params));
 			ordered_requests.push(ids[pos].clone());
 			request_set.insert(ids[pos].clone(), pos);
 		}
--- a/client/http-client/src/client.rs
+++ b/client/http-client/src/client.rs
@@ -30,7 +30,7 @@ use std::time::Duration;
 use crate::transport::HttpTransportClient;
 use crate::types::{ErrorResponse, Id, NotificationSer, ParamsSer, RequestSer, Response};
 use async_trait::async_trait;
-use jsonrpsee_core::client::{CertificateStore, ClientT, RequestIdManager, Subscription, SubscriptionClientT};
+use jsonrpsee_core::client::{CertificateStore, ClientT, IdKind, RequestIdManager, Subscription, SubscriptionClientT};
 use jsonrpsee_core::{Error, TEN_MB_SIZE_BYTES};
 use rustc_hash::FxHashMap;
 use serde::de::DeserializeOwned;
@@ -42,6 +42,7 @@ pub struct HttpClientBuilder {
 	request_timeout: Duration,
 	max_concurrent_requests: usize,
 	certificate_store: CertificateStore,
+	id_kind: IdKind,
 }
 
 impl HttpClientBuilder {
@@ -69,6 +70,12 @@ impl HttpClientBuilder {
 		self
 	}
 
+	/// Configure the format of the request object id.
+	pub fn id_format(mut self, id_kind: IdKind) -> Self {
+		self.id_kind = id_kind;
+		self
+	}
+
 	/// Build the HTTP client with target to connect to.
 	pub fn build(self, target: impl AsRef<str>) -> Result<HttpClient, Error> {
 		let transport = HttpTransportClient::new(target, self.max_request_body_size, self.certificate_store)
@@ -77,6 +84,7 @@ impl HttpClientBuilder {
 			transport,
 			id_manager: Arc::new(RequestIdManager::new(self.max_concurrent_requests)),
 			request_timeout: self.request_timeout,
+			id_kind: self.id_kind,
 		})
 	}
 }
@@ -88,6 +96,7 @@ impl Default for HttpClientBuilder {
 			request_timeout: Duration::from_secs(60),
 			max_concurrent_requests: 256,
 			certificate_store: CertificateStore::Native,
+			id_kind: IdKind::Number,
 		}
 	}
 }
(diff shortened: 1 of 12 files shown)


client/http-client/src/client.rs
```diff
@@ -101,6 +110,8 @@ pub struct HttpClient {
 	request_timeout: Duration,
 	/// Request ID manager.
 	id_manager: Arc<RequestIdManager>,
+	/// Configure the format of the request object id.
```
dvdplm, line 113:
```suggestion
	/// Configure the data type of the request object ID (strings or numbers). Default is number.
```

client/http-client/src/client.rs
```diff
@@ -160,11 +176,21 @@ impl ClientT for HttpClient {
...
-		let ids = self.id_manager.next_request_ids(batch.len())?;
+		let guard = self.id_manager.next_request_ids(batch.len())?;
+
+		let ids: Vec<Id> = guard
+			.inner()
+			.into_iter()
+			.map(|&id| match self.id_kind {
+				IdKind::Number => Id::Number(id),
+				IdKind::String => Id::Str(format!("{}", id).into()),
+			})
```
dvdplm, line 187: It's a bit annoying to have to loop over and transform each id one by one. If `self.id_manager` knew what `IdKind` the user has configured, then the `next_request_ids()` call could returns the IDs of the right kind. Do you think such a refactoring is possible?

client/ws-client/src/lib.rs
```diff
@@ -143,6 +145,12 @@ impl<'a> WsClientBuilder<'a> {
 		self
 	}
 
+	/// See documentation [`ClientBuilder::id_format`] (default is Number).
```
dvdplm, line 148:
```suggestion
	/// See documentation for [`ClientBuilder::id_format`] (default is Number).
```

core/src/client/async_client/mod.rs
```diff
@@ -102,6 +104,12 @@ impl ClientBuilder {
 		self
 	}
 
+	/// Configure the format of the request object id.
```
dvdplm, line 107:
```suggestion
	/// Configure the data type of the request object ID. Default is number.
```

dvdplm (commented):
Feels like `RequestIdManager` could be a bit smarter about the IDs, and be configured to use the right `IdKind` so that other code doesn't have to do the transformation itself. Or is that very annoying to do?

niklasad1:
> Feels like RequestIdManager could be a bit smarter about the IDs, and be configured to use the right IdKind so that other code doesn't have to do the transformation itself. Or is that very annoying to do?

Sounds reasonable to me, I did the most naive way I could think of to test that it worked so sure shouldn't be that hard I think

client/http-client/src/client.rs
```diff
@@ -69,13 +70,19 @@ impl HttpClientBuilder {
 		self
 	}
 
+	/// Configure the format of the request object id.
```
niklasad1, line 73:
```suggestion
	/// Configure the data type of the request object ID (default is number).
```

core/src/client/mod.rs
```diff
@@ -399,13 +402,31 @@ pub enum CertificateStore {
...
+	String,
+	/// Number.
+	Number,
+}
+
+impl IdKind {
+	fn into_id(self, id: u64) -> Id<'static> {
+		match self {
+			IdKind::Number => Id::Number(id),
+			IdKind::String => Id::Str(format!("{}", id).into()),
```
niklasad1, line 418: we could probably use some optimized `itoa` or something but should fast enough I guess for now...

whalelephant:
Hi @niklasad1, 

Yes, this includes what I was looking for. Thank you!

client/http-client/src/tests.rs
```diff
@@ -42,7 +41,33 @@ async fn method_call_works() {
...
 		.unwrap();
-	assert_eq!(JsonValue::String("hello".into()), result);
+	assert_eq!("hello", &result);
+}
+
+#[tokio::test]
+async fn method_call_with_wrong_id() {
+	let exp = "id as string";
+	let server_addr =
+		http_server_with_hardcoded_response(ok_response(exp.into(), Id::Num(0))).with_default_timeout().await.unwrap();
```
jsdw, line 51: To clarify, this fails because the server is configured to return numeric IDs and the client expects string IDs?

(would "wrong_id_kind" perhaps be a more precise test name?)

types/src/params.rs
```diff
@@ -350,7 +350,7 @@ impl<'a> SubscriptionId<'a> {
 }
 
 /// Request Id
-#[derive(Debug, PartialEq, Clone, Hash, Eq, Deserialize, Serialize)]
+#[derive(Debug, PartialEq, Clone, Hash, Eq, Deserialize, Serialize, PartialOrd, Ord)]
 #[serde(deny_unknown_fields)]
 #[serde(untagged)]
 pub enum Id<'a> {
```
jsdw, line 356: I didn't notice offhand; is a borrowed version of an ID ever used here? I pretty much assume every new ID will need to be an owned variant anyway. so I wonder how useful the lifetime/`Cow` is?

(thinking down that line, in general any `serde_json::Value` _could_ be returned as an ID (whether or not certain values make sense, who knows) so I was wondering whether using that here would simplify anything)

jsdw (approved):
LGTM!

types/src/params.rs
niklasad1: I think you are right at least in the point of view for the client.

However, I think we actually can use the lifetime and do zero-alloc deserialization in the servers with `#[serde](borrow)]`

client/http-client/src/tests.rs
niklasad1: exactly, thanks good point

core/src/client/mod.rs
```diff
@@ -399,13 +402,31 @@ pub enum CertificateStore {
...
+
 #[cfg(test)]
 mod tests {
-	use super::RequestIdManager;
+	use super::{IdKind, RequestIdManager};
 
 	#[test]
 	fn request_id_guard_works() {
-		let manager = RequestIdManager::new(2);
+		let manager = RequestIdManager::new(2, IdKind::Number);
```
TarikGul, line 429: Would it make sense to have a test where `IdKind::String`, not sure if this is covered elsewhere residually though.

TarikGul (approved):

maciejhirsz (approved):

core/src/client/mod.rs
maciejhirsz: Could also subslice byte array of all numbers from `00` to `99` as a static `Cow`, but yeah, this is good enough :).

core/src/client/mod.rs
niklasad1: I think that's covered by some unit tests for the whole `ws client` but wouldn't hurt with a unit test for `String` also I suppose :)

PR #662, opened 2022-01-20
fix: nit in WsError

niklasad1:


client/transport/src/ws/mod.rs
```diff
@@ -178,7 +178,7 @@ pub enum WsHandshakeError {
 #[derive(Debug, Error)]
 pub enum WsError {
 	/// Error in the WebSocket connection.
-	#[error("WebSocket connection error: {}", 0)]
+	#[error("WebSocket connection error: {0}")]
```
niklasad1, line 181: the old code will print `0` and not the actual error :shushing_face:

TarikGul (approved):

client/transport/src/ws/mod.rs
dvdplm: oops!

dvdplm (approved):

PR #671, opened 2022-01-24
update readme: crate is "stable" now

niklasad1:
The `crate` is rather stable now, so I think it makes sense to remove that from the README


jsdw (approved):
Seems fair to me!

dvdplm (approved):

PR #673, opened 2022-01-24
[Do not merge] benchmark debug

alvicsam:
The goal of this PR is to debug benchmark job. 
The CI runs only this job and can be rerun either with empty commit or via gitlab interface.

Part of https://github.com/paritytech/ci_cd/issues/302


lexnv:
Hi! Do we still need this PR for benchmarking?

niklasad1:
No, we are using the GHA for now

PR #684, opened 2022-02-01
refactor(ws server): impl `IdProvider` for Box<T> 

niklasad1:
Implement `IdProvider` for `Box<T>` to support `Box<dyn IdProvider>`, in addition I also added `std::fmt::Debug` to the IdProvider.

//cc @maciejhirsz 
I had a look how difficult it would be to make this purely static dispatch because currently we pass `dyn IdProvider` to the RpcModule if the call is a subscription,

Then we would have to add a type param on `Methods`, `RpcModule` and related types and the only user-facing would be the the `RpcModule` which is a bit annoying but it would kill the extra indirection's. The benefit of this is that we could get rid of the `Arc<dyn IdProvider>` to `I: IdProvider` and require `Clone` on IdProvider.

### DIFF
--- a/ws-server/src/server.rs
+++ b/ws-server/src/server.rs
@@ -27,6 +27,7 @@
 use std::future::Future;
 use std::net::SocketAddr;
 use std::pin::Pin;
+use std::sync::Arc;
 use std::task::{Context, Poll};
 
 use crate::future::{FutureDriver, ServerHandle, StopMonitor};
@@ -60,7 +61,7 @@ pub struct Server<M> {
 	stop_monitor: StopMonitor,
 	resources: Resources,
 	middleware: M,
-	id_provider: Box<dyn IdProvider>,
+	id_provider: Arc<dyn IdProvider>,
 }
 
 impl<M> std::fmt::Debug for Server<M> {
@@ -213,7 +214,7 @@ enum HandshakeResponse<'a, M> {
 		cfg: &'a Settings,
 		stop_monitor: &'a StopMonitor,
 		middleware: M,
-		id_provider: Box<dyn IdProvider>,
+		id_provider: Arc<dyn IdProvider>,
 	},
 }
 
@@ -288,7 +289,7 @@ async fn background_task(
 	max_request_body_size: u32,
 	stop_server: StopMonitor,
 	middleware: impl Middleware,
-	id_provider: Box<dyn IdProvider>,
+	id_provider: Arc<dyn IdProvider>,
 ) -> Result<(), Error> {
 	// And we can finally transition to a websocket background_task.
 	let mut builder = server.into_builder();
@@ -647,7 +648,7 @@ pub struct Builder<M = ()> {
 	settings: Settings,
 	resources: Resources,
 	middleware: M,
-	id_provider: Box<dyn IdProvider>,
+	id_provider: Arc<dyn IdProvider>,
 }
 
 impl<M> std::fmt::Debug for Builder<M> {
@@ -662,7 +663,7 @@ impl Default for Builder {
 			settings: Settings::default(),
 			resources: Resources::default(),
 			middleware: (),
-			id_provider: Box::new(RandomIntegerIdProvider),
+			id_provider: Arc::new(RandomIntegerIdProvider),
 		}
 	}
 }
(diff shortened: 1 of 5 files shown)


ws-server/src/tests.rs
```diff
@@ -642,7 +643,7 @@ async fn custom_subscription_id_works() {
 
 	init_logger();
 	let server = WsServerBuilder::default()
-		.set_id_provider(HardcodedSubscriptionId)
+		.set_id_provider(Box::new(HardcodedSubscriptionId))
```
jsdw, line 646: Just curious really; what's the motivation for exposing the `Box`-ing here, rather than boxing it internally?

ws-server/src/server.rs
```diff
@@ -648,7 +647,7 @@ pub struct Builder<M = ()> {
 	settings: Settings,
 	resources: Resources,
 	middleware: M,
-	id_provider: Arc<dyn IdProvider>,
+	id_provider: Box<dyn IdProvider>,
```
jsdw, line 650: So, when the `id_provider` was `Arc`, cloning it just incremented the pointer count and handed out a new pointer. Now, Cloning actually clones the underlying `IdProvider` thing (and allocates a new box each time).

Are we happy with that allocation-on-clone over an increment-atomic-integer-on-clone? I'd have to peruse the code a little more to get a feeling for how often this was cloned, but perhaps the Arc works out better?

ws-server/src/tests.rs
niklasad1: the motivation/rationale is that if you have something like:

`Option<Box<IdProvider>>` then having `impl IdProvider` won't work, thus I changed it to `Box<dyn IdProvider>`

ws-server/src/server.rs
niklasad1: actually we could kill the `Arc` too just if we require `trait IdProvider: Clone` which I believe is reasonable.

it will be cloned once for every connection, should be ok I guess?!

ws-server/src/server.rs
```diff
@@ -819,8 +818,8 @@ impl<M> Builder<M> {
 	///
 	/// let builder = WsServerBuilder::default().set_id_provider(RandomStringIdProvider::new(16));
 	/// ```
-	pub fn set_id_provider(mut self, id_provider: impl IdProvider + 'static) -> Self {
```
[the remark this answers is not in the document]
niklasad1: > If I understand right, the main motivation for the change was to remove this generic, which may otherwise have been propagated as some type T: IdProvider type at higher levels?

exactly, yes

I guess somebody could have passed a Box<dyn IdProvider> to this as it stood (maybe making sure we had an impl <T: IdProvider> IdProvider for Box<T> if needed). if the reason for the changes here that doing that would have meant we had a double boxing (a Box inside an Arc) and after the change we have a single Boxing?

that sounds better, lemme check if that works

ws-server/src/tests.rs
jsdw: Oh, I just meant that instead of `.set_id_provider(Box::new(HardcodedSubscriptionId))`, we could keep `.set_id_provider(HardcodedSubscriptionId)` and do the Boxing inside the function?

ws-server/src/server.rs
jsdw: yeah I think so; better than cloning the Box every time :)

ws-server/src/tests.rs
niklasad1: I changed it back with your comment to implement `IdProvider for Box<T>`, so all good now anyway.

dvdplm: Late to the conversation here, apologies. So the point of the test is now to prove that one *can* use a `Box<T>` (where `T: IdProvider`) if desired? It would be good imo to have a test that shows both ways, boxed and plain, and perhaps a comment that motivates why one might want to use the longer boxed version? And/or some docs on [the impl for `Box<T>`](https://github.com/paritytech/jsonrpsee/blob/1e00d8b2331bf34b46c704577505b0c4781f59eb/core/src/traits.rs#L79)?

dvdplm (approved):
Works for me.

ws-server/src/tests.rs
```diff
@@ -631,7 +631,8 @@ async fn unsubscribe_wrong_sub_id_type() {
 }
 
 #[tokio::test]
-async fn custom_subscription_id_works() {
+async fn custom_subscription_boxed_id_works() {
```
niklasad1, line 634:
```suggestion
async fn custom_boxed_subscription_id_works() {
```

maciejhirsz (approved):

PR #687, opened 2022-02-02
fix(http client): set reuseaddr and nodelay.

niklasad1:


dvdplm (approved):

dvdplm:
@niklasad1 Do we still want to do this, even though it doesn't fix the benches on macos? I didn't look into the windows CI failure, maybe we need some `cfg`-fu to make it work?

niklasad1:
> @niklasad1 Do we still want to do this, even though it doesn't fix the benches on macos? I didn't look into the windows CI failure, maybe we need some cfg-fu to make it work?

Good catch I forgot about this one.
Yes, I think we should include this anyway.

lexnv (approved):

jsdw (approved):
Looks sane to me; I'm not really sure of the ramifications of reuseport and reuseaddress being set though!

niklasad1:
> Looks sane to me; I'm not really sure of the ramifications of reuseport and reuseaddress being set though!

Yeah, I'm not sure about reuseport at least I'll remove it.

TarikGul (approved):

PR #693, opened 2022-02-08
fix(rpc module): unsubscribe according ethereum pubsub spec

niklasad1:
As we already have adopted the [ethereum pubsub spec](https://geth.ethereum.org/docs/rpc/pubsub) for subscriptions let's stick to it completely.


core/src/server/rpc_module.rs
```diff
@@ -681,27 +680,17 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
...
-					if subscribers
+					let result = subscribers
 						.lock()
 						.remove(&SubscriptionKey { conn_id: conn.conn_id, sub_id: sub_id.clone() })
-						.is_some()
-					{
-						sink.send_response(id, "Unsubscribed")
-					} else {
-						let err = to_json_raw_value(&format!(
-							"Invalid subscription ID={}",
```
dvdplm, line 699: The only downside to this PR is that we loose this information (I think?). 

With this PR, when an unsubscribe call fails, what do users get back?

dvdplm (approved):

core/src/server/rpc_module.rs
niklasad1: yeah, unfortunately they will just receive a `bool` set to false and then the client(s) have to figure it on their own :(

dvdplm: Let's document this properly in the macro docs (and on this method): "When attempting to unsubscribe with an unknown subscription ID this call will return `false`" or something along those lines?

niklasad1: I guess it's better to document that we implement the `ethereum pubsub specification` when we refer to subscriptions and it's possible to provide custom subscription IDs but that's it.

Sure, we could add some documentation to both `RpcModule::register_subscription` and the macros.

jsdw (approved):

PR #701, opened 2022-02-15
CI: try nextest

TriplEight:
An attempt to switch to [`cargo-nextest`](https://github.com/nextest-rs/nextest).

Note it [doesn't support doctests](https://github.com/nextest-rs/nextest/issues/16) so they should be run separately.
@dvdplm @niklasad1 could you suggest a good way of getting the list of doctests please?

---

Oh lovely.
So `nextest` with some cache (on the second run) performed like this
image

instead of a regular `cargo test`
image

it's worth mentioning though that without cache it takes another ~3m30s to install the tool. Thankfully, there's a cached [`cargo-install`](https://github.com/marketplace/actions/cargo-install).
image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,6 +92,32 @@ jobs:
           command: test
           args: --workspace
 
+  nextests:
+    name: Run nextests on Ubuntu
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2.4.0
+
+      - name: Install Rust stable toolchain
+        uses: actions-rs/toolchain@v1.0.7
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
+
+      - name: Install cargo-nextest
+        uses: baptiste0928/cargo-install@v1
+        with:
+          crate: cargo-nextest
+          version: 0.9
+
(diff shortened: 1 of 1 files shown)


niklasad1:
cool, it's 1 minute faster than `cargo test`

rcny (approved):

alvicsam (approved):

niklasad1:
@TriplEight 

hrm, I would like at least one approval from the `tools team` before merging any PRs to this repo in future.

We rely heavily on doc tests and we need have to `doc tests` enabled to much sure that we don't miss any tests. 

I'll create a follow--up PR to add that back

TriplEight:
@niklasad1 sorry for that, noted.
https://github.com/paritytech/jsonrpsee/blob/master/.github/workflows/ci.yml#L46 or
https://github.com/paritytech/jsonrpsee/blob/master/.gitlab-ci.yml#L99 I missed that these are not running all the doctests.
Anyway, doctests usually do not require compilation and can be safely separated to another job.

The follow-up issue to work on this repo is here https://github.com/paritytech/ci_cd/issues/333
And this is how we'll prepare to the nextest deployment next time https://github.com/paritytech/ci_cd/issues/334
Please feel free to add your ideas.

Issue #698, opened 2022-02-14
[proc macros]: generated unsubscribe methods with `subscribe` in method name doesn't work.


niklasad1:
```
#[rpc(client)]
pub trait RpcApi {
       #[subscription(
		name = "author_submitAndWatchExtrinsic" => "author_extrinsicUpdate",
		item = TransactionStatus<Hash, Hash>,
	)]
	fn watch_extrinsic(&self, bytes: &Bytes) -> RpcResult<()>;
}
```

This will generate -> `"unsubscribeauthor_submitAndWatchExtrinsic"`, see https://github.com/paritytech/jsonrpsee/blob/master/proc-macros/src/rpc_macro.rs#L342-#L344

For the servers that shouldn't matter much but for the clients it's a big concern i.e, the unsubscribe won't work properly.
So maybe it could make sense return `Option<String>` instead and require users to provide `unsubscribe` if not the method starts with `subscribe`

niklasad1:
Closed by #702

PR #702, opened 2022-02-15
[proc macros]: only generate unsub method if not provided

niklasad1:
Closing #698 

It will only try to generate an unsubscribe method if it starts with `subscribe`.
Also, if `unsubscribe` is part of the RPC API it will override the `generated unsubscribe method name`

### DIFF
--- a/proc-macros/src/lib.rs
+++ b/proc-macros/src/lib.rs
@@ -182,7 +182,7 @@ pub(crate) mod visitor;
 ///
 /// - `name` (mandatory): name of the RPC method. Does not have to be the same as the Rust method name.
 /// - `unsubscribe` (optional): name of the RPC method to unsubscribe from the subscription. Must not be the same as `name`.
-///                             This is generated for you if the subscription name starts with `subscribe`-
+///                             This is generated for you if the subscription name starts with `subscribe`.
 /// - `aliases` (optional): aliases for `name`. Aliases are processed ignoring the namespace,
 ///                         so add the complete name, including the namespace.
 /// - `unsubscribe_aliases` (optional): Similar to `aliases` but for `unsubscribe`.
--- a/proc-macros/src/lib.rs
+++ b/proc-macros/src/lib.rs
@@ -163,6 +163,8 @@ pub(crate) mod visitor;
 ///
 /// - `name` (mandatory): name of the RPC method. Does not have to be the same as the Rust method name.
 /// - `aliases`: list of name aliases for the RPC method as a comma separated string.
+///              The aliases are kept outside namespace, so you need add that if you want aliases in the current
+///              namespace.
 /// - `blocking`: when set method execution will always spawn on a dedicated thread. Only usable with non-`async` methods.
 /// - `param_kind`: kind of structure to use for parameter passing. Can be "array" or "map", defaults to "array".
 ///
@@ -180,8 +182,10 @@ pub(crate) mod visitor;
 ///
 /// - `name` (mandatory): name of the RPC method. Does not have to be the same as the Rust method name.
 /// - `unsubscribe` (optional): name of the RPC method to unsubscribe from the subscription. Must not be the same as `name`.
-/// - `aliases` (optional): aliases for `name`.
-/// - `unsubscribe_aliases` (optional): aliases for `unsubscribe`.
+///                             This is generated for you if the subscription name starts with `subscribe`-
+/// - `aliases` (optional): aliases for `name`. The aliases are kept outside namespace,
+///                         so you need add that if you want aliases in the current namespace.
+/// - `unsubscribe_aliases` (optional): Similar to `aliases` but for `unsubscribe`.
 /// - `item` (mandatory): type of items yielded by the subscription. Note that it must be the type, not string.
 /// - `param_kind`: kind of structure to use for parameter passing. Can be "array" or "map", defaults to "array".
 ///
--- a/proc-macros/src/lib.rs
+++ b/proc-macros/src/lib.rs
@@ -179,7 +179,7 @@ pub(crate) mod visitor;
 /// **Arguments:**
 ///
 /// - `name` (mandatory): name of the RPC method. Does not have to be the same as the Rust method name.
-/// - `unsubscribe` (optional): name of the RPC method to unsubscribe from the subscription. Must not be the same as `name`.
+/// - `unsubscribe` (optional): name of the RPC method used to unsubscribe from the subscription. Must not be the same as `name`.
 /// - `aliases` (optional): aliases for `name`.
 /// - `unsubscribe_aliases` (optional): aliases for `unsubscribe`.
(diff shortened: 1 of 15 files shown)


proc-macros/src/rpc_macro.rs
```diff
@@ -150,7 +150,12 @@ impl RpcSubscription {
 
 		let sig = sub.sig.clone();
 		let docs = extract_doc_comments(&sub.attrs);
-		let unsubscribe = build_unsubscribe_method(&name);
+		let unsubscribe = match parse_subscribe(unsubscribe)? {
+			Some(unsub) => unsub,
+			None => build_unsubscribe_method(&name).expect(
+				"Could not generate `unsubscribe method`, you need to provide the `unsubscribe` manually in your API",
```
dvdplm, line 156:
```suggestion
				format!("Could not generate the unsubscribe method with name '{name}'.  You need to provide the name manually using the `unsubscribe` attribute in your RPC API definition", name),
```

proc-macros/tests/ui/correct/alias_doesnt_use_namespace.rs
```diff
@@ -1,12 +1,12 @@
-use jsonrpsee::{proc_macros::rpc, core::RpcResult};
+use jsonrpsee::{core::RpcResult, proc_macros::rpc};
 
 #[rpc(client, server, namespace = "myapi")]
 pub trait Rpc {
 	/// Alias doesn't use the namespace so not duplicated.
 	#[method(name = "getTemp", aliases = ["getTemp"])]
```
dvdplm, line 6: I think we should clarify this comment a bit. This macro invocation generates two methods: `myapi_getTemp` and `getTemp` right?

proc-macros/tests/ui/correct/alias_doesnt_use_namespace.rs
```diff
@@ -1,12 +1,12 @@
...
 	async fn async_method(&self, param_a: u8, param_b: String) -> RpcResult<u16>;
 
-	#[subscription(name = "getFood", item = String, aliases = ["getFood"], unsubscribe_aliases = ["unsubscribegetFood"])]
+	#[subscription(name = "subscribeGetFood", item = String, aliases = ["getFood"], unsubscribe_aliases = ["unsubscribegetFood"])]
```
dvdplm, line 9: Let's see if I have this right. The methods generated here are:
- `myapi_subscribeGetFood` (auto)
- `getFood` (alias)
- `myapi_unsubscribeGetFood` (auto)
- `unsubscribegetFood` (alias)

Did I get that right?

proc-macros/src/lib.rs
```diff
@@ -84,7 +84,7 @@ pub(crate) mod visitor;
 ///     #[method(name = "bar")]
 ///     fn sync_method(&self) -> String;
 ///
-///     #[subscription(name = "sub", unsub = "unsub", item = "String")]
+///     #[subscription(name = "subscribe", item = "String")]
```
dvdplm, line 87: I think you need to update the generated code sample below as well right? On line 102

proc-macros/src/lib.rs
```diff
@@ -179,7 +179,9 @@ pub(crate) mod visitor;
 /// **Arguments:**
 ///
 /// - `name` (mandatory): name of the RPC method. Does not have to be the same as the Rust method name.
-/// - `unsub` (mandatory): name of the RPC method to unsubscribe from the subscription. Must not be the same as `name`.
+/// - `unsubscribe` (optional): name of the RPC method to unsubscribe from the subscription. Must not be the same as `name`.
```
dvdplm, line 182:
```suggestion
/// - `unsubscribe` (optional): name of the RPC method used to unsubscribe from the subscription. Must not be the same as `name`.
```

proc-macros/src/lib.rs
```diff
@@ -179,7 +179,9 @@ pub(crate) mod visitor;
...
+/// - `aliases` (optional): aliases for `name`.
```
dvdplm, line 183: We should mention how this interacts with the namespace setting as well (or rather: how it **doesn't** interact with it).

dvdplm (approved):
lgtm, minor grumbles

proc-macros/src/lib.rs
niklasad1: nah, it's the actual impl so it's safe.

proc-macros/src/lib.rs
```diff
@@ -163,6 +163,8 @@ pub(crate) mod visitor;
 ///
 /// - `name` (mandatory): name of the RPC method. Does not have to be the same as the Rust method name.
 /// - `aliases`: list of name aliases for the RPC method as a comma separated string.
+///              The aliases are kept outside namespace, so you need add that if you want aliases in the current
+///              namespace.
```
dvdplm, line 167:
```suggestion
///              Aliases are processed ignoring the namespace, so add the complete name, including the
///              namespace.
```

proc-macros/src/lib.rs
```diff
@@ -179,9 +181,11 @@ pub(crate) mod visitor;
 /// **Arguments:**
 ///
 /// - `name` (mandatory): name of the RPC method. Does not have to be the same as the Rust method name.
-/// - `unsubscribe` (optional): name of the RPC method used to unsubscribe from the subscription. Must not be the same as `name`.
-/// - `aliases` (optional): aliases for `name`.
-/// - `unsubscribe_aliases` (optional): aliases for `unsubscribe`.
+/// - `unsubscribe` (optional): name of the RPC method to unsubscribe from the subscription. Must not be the same as `name`.
+///                             This is generated for you if the subscription name starts with `subscribe`-
```
dvdplm, line 185:
```suggestion
///                             This is generated for you if the subscription name starts with `subscribe`.
```

proc-macros/src/lib.rs
```diff
@@ -179,9 +181,11 @@ pub(crate) mod visitor;
...
+/// - `aliases` (optional): aliases for `name`. The aliases are kept outside namespace,
+///                         so you need add that if you want aliases in the current namespace.
```
dvdplm, line 187:
```suggestion
/// - `aliases` (optional): aliases for `name`. Aliases are processed ignoring the namespace,
///                         so add the complete name, including the namespace.
```

proc-macros/tests/ui/correct/alias_doesnt_use_namespace.rs
```diff
@@ -2,7 +2,8 @@ use jsonrpsee::{core::RpcResult, proc_macros::rpc};
 
 #[rpc(client, server, namespace = "myapi")]
 pub trait Rpc {
-	/// Alias doesn't use the namespace so not duplicated.
+	/// Aliases doesn't use the namespace.
```
dvdplm, line 5:
```suggestion
	/// Aliases don't use the namespace.
```

Issue #697, opened 2022-02-14
The subscription of websocket does not have an example of broadcasting to all subscribers, hoping to give guidance


keienWang:
For example, I want to publish a message to all subscribers, how can I collect all subscribers at this time? In the example, I only see, publish an i:= 4, but I want to change the value of i and then broadcast to each subscriber, how to change it
![image](https://user-images.githubusercontent.com/42377006/153810820-182be353-2d68-4075-89cc-d599318f51e4.png)

keienWang:
![image](https://user-images.githubusercontent.com/42377006/153811227-ffb3c6b2-64d2-4374-b3d7-a7fcc8676d7e.png)
The jump here on the homepage is wrong, jump to http-server
span

niklasad1:
Hey @keienWang 

The result should published to each subscriber however your example is confusing, you should really remove the first `thread::spawn` and `move i` to the `register_subscription` closure. Otherwise wrap `i` in a `Arc<Mutex> or something similar` if you want a synchronized value....

Can you share the entire example otherwise I can't re-produce thus? It could be that you only poll the subscription once or something?!

Also, we should update the `pubsub examples` to use [the new subscription API](https://github.com/paritytech/jsonrpsee/blob/master/core/src/server/rpc_module.rs#L756-#L775). We happily accept any PRs regarding the examples :)

keienWang:
[source example](https://github.com/paritytech/jsonrpsee/blob/master/examples/ws_subscription.rs)，I just want to implement an observer pattern and send an external changing value，For example, there are two subscribers, and I want to send new blocks to them at the same time

niklasad1:
@keienWang 

Ok, I see the easiest would be use message passing i.e, once a new block/item is produced/changed you send a message to the receiver that implements `Stream`

The idea that something like should work:

```rust
async fn run_server() -> anyhow::Result<SocketAddr> {
	let server = WsServerBuilder::default().build("127.0.0.1:0").await?;

	let (tx, rx) = async_channel::unbounded();
	let mut module = RpcModule::new(());

	tokio::spawn(async move {
		let mut i = 4;
		loop {
			i += 1;
			tx.send(i).await.unwrap();
			tokio::time::sleep(std::time::Duration::from_secs(5)).await;
		}
	});

	module.register_subscription("subscribe_hello", "s_hello", "unsubscribe_hello", move |_, mut sink, _| {
		// required because of the `Fn closure above`
		let rx = rx.clone();
[...]

PR #705, opened 2022-02-17
[examples]: update pubsub examples

niklasad1:
Closing #697


examples/ws_subscription.rs
```diff
@@ -24,16 +24,25 @@
 // IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 // DEALINGS IN THE SOFTWARE.
 
-use std::net::SocketAddr;
+//! Example that shows how to broadcasts the produced values to active all subscriptions using `mpsc channels`.
+//!
+//! It's possible to use `tokio::sync::broadcast` too but because the Receiver doesn't implement
+//! stream we chose this `mpsc channels` in this example.
```
niklasad1, line 30:
```suggestion
//! stream thus `mpsc channels` were picked in this example.
```

examples/ws_subscription.rs
```diff
@@ -44,33 +50,51 @@ async fn main() -> anyhow::Result<()> {
...
+	let (tx, _) = broadcast::channel(1024);
+
+	tokio::spawn(produce_items(tx.clone()));
+
+	module.register_subscription("subscribe_hello", "s_hello", "unsubscribe_hello", move |_, sink, _| {
+		let rx = tx.subscribe();
+
+		// Convert stream from `Item = Result<T: Serialize, Error>` to `Item = T::Serialize`.
+		let stream =
+			BroadcastStream::new(rx).take_while(|r| future::ready(r.is_ok())).filter_map(|r| future::ready(r.ok()));
```
dvdplm, line 78: This is a bit unfortunate, a lot of line-noise just to massage the stream to do the right thing. Any chance it can be simpler?

dvdplm (approved):

examples/ws_subscription.rs
niklasad1: yeah, we need to change to API for stream to take `impl Stream<Item = Result<T::Serialize, Error>>` instead of `impl Stream<Item = T: Serialize` then this goes away but other streams such `mpsc::Receiver` need to wrap it in a Result which more flexible.

niklasad1: switched to the `async-broadcast` the only reason for `tokio::sync::Broadcast` to fail is if the receiver can't keep up with sender. 

similar to what we do in the `ws client` but for readability let's use async-broadcast

jsdw (approved):

examples/ws_pubsub_broadcast.rs
```diff
@@ -24,13 +24,18 @@
 // IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 // DEALINGS IN THE SOFTWARE.
 
+//! Example that shows how to broadcasts the produced values to all active subscriptions using `tokio::sync::broadcast`.
```
niklasad1, line 27:
```suggestion
//! Example that shows how to broadcast to all active subscriptions using `tokio::sync::broadcast`.
```

PR #708, opened 2022-02-22
Replace async-channel

### LINKED ISSUE #691: [core/server]: replace `async-channel` with `tokio::sync::watch` in RpcModule
We already depend heavily on `tokio` and it doesn't make sense to bring `async-channel` into the dependency tree just for this.
Tokio has similar primitives such as `watch` and `broadcast`

dvdplm:
Replace `async-channel` with `tokio::sync::Notify` to use a dependency less and (possibly?) be a bit faster by using a lighter primitive (maybe?). The problem the `async-channel` is solving is: how does the `SubscriptionSink` know about rude subscribers that leaves without calling `unsubscribe_*`? Using a full-on channel for this might be overkill and this is an attempt at fixing that.

Fixes https://github.com/paritytech/jsonrpsee/issues/691

### DIFF
--- a/core/src/server/rpc_module.rs
+++ b/core/src/server/rpc_module.rs
@@ -904,7 +904,9 @@ impl Subscription {
 	/// Close the subscription channel.
 	pub fn close(&mut self) {
 		tracing::trace!("[Subscription::close] Notifying");
-		self.close_notify.take().and_then(|n| {
+if let Some(n) = self.close_notify.take() { 
+     n.notify_one() 
+}
 			n.notify_one();
 			None::<Arc<Notify>>
 		});
--- a/core/src/server/rpc_module.rs
+++ b/core/src/server/rpc_module.rs
@@ -24,6 +24,7 @@
 // IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 // DEALINGS IN THE SOFTWARE.
 
+use std::borrow::BorrowMut;
 use std::collections::hash_map::Entry;
 use std::fmt::{self, Debug};
 use std::future::Future;
@@ -46,6 +47,7 @@ use jsonrpsee_types::{
 use parking_lot::Mutex;
 use rustc_hash::FxHashMap;
 use serde::{de::DeserializeOwned, Serialize};
+use tokio::sync::Notify;
 
 /// A `MethodCallback` is an RPC endpoint, callable with a standard JSON-RPC request,
 /// implemented as a function pointer to a `Fn` function taking four arguments:
@@ -63,21 +65,25 @@ pub type SubscriptionMethod = Arc<dyn Send + Sync + Fn(Id, Params, &MethodSink, 
 /// For stateless protocols such as http it's unused, so feel free to set it some hardcoded value.
 pub type ConnectionId = usize;
 /// Raw RPC response.
-pub type RawRpcResponse = (String, mpsc::UnboundedReceiver<String>, async_channel::Sender<()>);
+pub type RawRpcResponse = (String, mpsc::UnboundedReceiver<String>, Arc<Notify>);
+// pub type RawRpcResponse = (String, mpsc::UnboundedReceiver<String>, watch::Receiver<()>);
+// pub type RawRpcResponse = (String, mpsc::UnboundedReceiver<String>, async_channel::Sender<()>);
 
 /// Data for stateful connections.
 pub struct ConnState<'a> {
 	/// Connection ID
 	pub conn_id: ConnectionId,
 	/// Channel to know whether the connection is closed or not.
-	pub close: async_channel::Receiver<()>,
+	pub close_notify: Arc<Notify>,
+	// pub close_tx: watch::Sender<()>,
+	// pub close: async_channel::Receiver<()>,
 	/// ID provider.
 	pub id_provider: &'a dyn IdProvider,
 }
 
 impl<'a> std::fmt::Debug for ConnState<'a> {
 	fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-		f.debug_struct("ConnState").field("conn_id", &self.conn_id).field("close", &self.close).finish()
+		f.debug_struct("ConnState").field("conn_id", &self.conn_id).field("close", &self.close_notify).finish()
 	}
 }
 
@@ -369,7 +375,11 @@ impl Methods {
 	async fn inner_call(&self, req: Request<'_>) -> RawRpcResponse {
 		let (tx, mut rx) = mpsc::unbounded();
 		let sink = MethodSink::new(tx);
-		let (close_tx, close_rx) = async_channel::unbounded();
+		// TODO: (dp) pretty annoying having to do this for all `inner_call`s. We only need it for subscriptions yeah?
+		let notify1 = Arc::new(Notify::new());
+		let notify2 = notify1.clone();
+		// let (close_tx, close_rx) = watch::channel(());
+		// let (close_tx, close_rx) = async_channel::unbounded();
 
 		let id = req.id.clone();
 		let params = Params::new(req.params.map(|params| params.get()));
@@ -379,13 +389,14 @@ impl Methods {
 			Some(MethodKind::Sync(cb)) => (cb)(id, params, &sink),
 			Some(MethodKind::Async(cb)) => (cb)(id.into_owned(), params.into_owned(), sink, 0, None).await,
 			Some(MethodKind::Subscription(cb)) => {
-				let conn_state = ConnState { conn_id: 0, close: close_rx, id_provider: &RandomIntegerIdProvider };
+				let conn_state = ConnState { conn_id: 0, close_notify: notify1, id_provider: &RandomIntegerIdProvider };
+				// let conn_state = ConnState { conn_id: 0, close_tx, id_provider: &RandomIntegerIdProvider };
 				(cb)(id, params, &sink, conn_state)
 			}
 		};
 
 		let resp = rx.next().await.expect("tx and rx still alive; qed");
-		(resp, rx, close_tx)
+		(resp, rx, notify2)
 	}
 
 	/// Helper to create a subscription on the `RPC module` without having to spin up a server.
@@ -417,10 +428,11 @@ impl Methods {
 		let params = params.to_rpc_params()?;
 		let req = Request::new(sub_method.into(), Some(&params), Id::Number(0));
 		tracing::trace!("[Methods::subscribe] Calling subscription method: {:?}, params: {:?}", sub_method, params);
-		let (response, rx, tx) = self.inner_call(req).await;
+		let (response, rx, close_notify) = self.inner_call(req).await;
 		let subscription_response = serde_json::from_str::<Response<RpcSubscriptionId>>(&response)?;
 		let sub_id = subscription_response.result.into_owned();
-		Ok(Subscription { sub_id, rx, tx })
+		let close_notify = Some(close_notify);
+		Ok(Subscription { sub_id, rx, close_notify })
 	}
 
 	/// Returns an `Iterator` with all the method names registered on this server.
@@ -627,6 +639,7 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 		let ctx = self.ctx.clone();
 		let subscribers = Subscribers::default();
 
+		// Subscribe
 		{
 			let subscribers = subscribers.clone();
 			self.methods.mut_callbacks().insert(
@@ -647,7 +660,7 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 
 					let sink = SubscriptionSink {
 						inner: method_sink.clone(),
-						close: conn.close,
+						close_notify: Some(conn.close_notify),
 						method: notif_method_name,
 						subscribers: subscribers.clone(),
 						uniq_sub: SubscriptionKey { conn_id: conn.conn_id, sub_id },
@@ -668,6 +681,7 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 			);
 		}
 
+		// Unsubscribe
 		{
 			self.methods.mut_callbacks().insert(
 				unsubscribe_method_name,
@@ -730,7 +744,8 @@ pub struct SubscriptionSink {
 	/// Sink.
 	inner: MethodSink,
 	/// Close
-	close: async_channel::Receiver<()>,
+	close_notify: Option<Arc<Notify>>,
+	// close_gg: async_channel::Receiver<()>,
 	/// MethodCallback.
 	method: &'static str,
 	/// Unique subscription.
@@ -777,13 +792,27 @@ impl SubscriptionSink {
 		S: Stream<Item = T> + Unpin,
 		T: Serialize,
 	{
-		let mut close_stream = self.close.clone();
+		// let mut close_stream = self.close_gg.clone();
 		let mut item = stream.next();
-		let mut close = close_stream.next();
-
+		// let mut close = close_stream.next();
+		// let mut closed = Box::pin(self.close_tx.closed());
+		// tokio::pin!(closed);
+		// let mut closed = self.close_tx.take();
+		// let mut myclose = watch::Sender<()>::new();
+		// let n = Arc::new(Notify::new());
+		// let mut close = self.close_notify.take().unwrap().clone();
+		if self.close_notify.is_none() {
+			tracing::warn!("[SubscriptionSink::pipe_from_stream] We're closed.");
+			return Ok(())
+		}
+		let close = self.close_notify.clone().unwrap().clone();
+		tracing::trace!("[SubscriptionSink::pipe_from_stream] Entering loop");
 		loop {
-			match futures_util::future::select(item, close).await {
-				Either::Left((Some(result), c)) => {
+			// match futures_util::future::select(item, close).await {
+			match futures_util::future::select(item, Box::pin(close.notified())).await {
+				// Either::Left((Some(result), c)) => {
+				Either::Left((Some(result), _)) => {
+					tracing::trace!("[SubscriptionSink::pipe_from_stream] Left - sending a result back");
 					match self.send(&result) {
 						Ok(_) => (),
 						Err(Error::SubscriptionClosed(close_reason)) => {
@@ -791,24 +820,30 @@ impl SubscriptionSink {
 							break Ok(());
 						}
 						Err(err) => {
-							tracing::error!("subscription `{}` failed to send item got error: {:?}", self.method, err);
+							tracing::error!("subscription `{}` failed to send item. Error: {:?}", self.method, err);
 							break Err(err);
 						}
 					};
-					close = c;
+					// close = c;
 					item = stream.next();
 				}
-				// No messages should be sent over this channel
-				// if that occurred just ignore and continue.
-				Either::Right((Some(_), i)) => {
-					item = i;
-					close = close_stream.next();
-				}
-				// Connection terminated.
-				Either::Right((None, _)) => {
+				// Either::Right(((), _)) => {
+				Either::Right((n, _x)) => {
+					tracing::trace!("[SubscriptionSink::pipe_from_stream] Right - closing");
 					self.close(&SubscriptionClosed::new(SubscriptionClosedReason::ConnectionReset));
 					break Ok(());
 				}
+				// // No messages should be sent over this channel
+				// // if that occurred just ignore and continue.
+				// Either::Right((Some(_), i)) => {
+				// 	item = i;
+				// 	close = close_stream.next();
+				// }
+				// // Connection terminated.
+				// Either::Right((None, _)) => {
+				// 	self.close(&SubscriptionClosed::new(SubscriptionClosedReason::ConnectionReset));
+				// 	break Ok(());
+				// }
 				// Stream terminated.
 				Either::Left((None, _)) => break Ok(()),
 			}
@@ -817,7 +852,13 @@ impl SubscriptionSink {
 
 	/// Returns whether this channel is closed without needing a context.
 	pub fn is_closed(&self) -> bool {
-		self.inner.is_closed() || self.close.is_closed()
+		tracing::trace!(
+			"[SubscriptionSink::is_closed] self.inner.is_closed? {}, self.close_notify.is_none()? {}",
+			self.inner.is_closed(),
+			self.close_notify.is_none()
+		);
+		self.inner.is_closed() || self.close_notify.is_none()
+		// self.inner.is_closed() || self.close_notify.is_closed()
 	}
 
 	fn build_message<T: Serialize>(&self, result: &T) -> Result<String, Error> {
@@ -857,7 +898,7 @@ impl SubscriptionSink {
 		self.inner_close(Some(&close_reason));
 	}
 
-	/// Provide close from `SubscriptionClosed`.
+	/// Close the subscription sink with the provided [`SubscriptionClosed`].
 	pub fn close(&mut self, close_reason: &SubscriptionClosed) {
 		self.inner_close(Some(close_reason));
 	}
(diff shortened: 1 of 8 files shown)


core/src/server/rpc_module.rs
```diff
@@ -367,25 +373,27 @@ impl Methods {
...
-		let (tx, mut rx) = mpsc::unbounded();
-		let sink = MethodSink::new(tx);
-		let (close_tx, close_rx) = async_channel::unbounded();
-
+		let (tx_sink, mut rx_sink) = mpsc::unbounded();
+		let sink = MethodSink::new(tx_sink);
 		let id = req.id.clone();
 		let params = Params::new(req.params.map(|params| params.get()));
+		// TODO: (dp) pretty annoying having to do this for all `inner_call`s. We only need it for subscriptions yeah?
+		let notify = Arc::new(Notify::new());
```
niklasad1, line 381: this is just used for tests but you are right

ws-server/src/server.rs
```diff
@@ -466,8 +471,9 @@ async fn background_task(
 				let methods = &methods;
 				let sink = sink.clone();
 				let id_provider = id_provider.clone();
+				let close_notify2 = close_notify.clone();
 
-				let conn_rx2 = conn_rx.clone();
+				// let conn_rx2 = conn_rx.clone();
```
niklasad1, line 476:
```suggestion
```

core/src/server/rpc_module.rs
```diff
@@ -884,17 +895,20 @@ impl Drop for SubscriptionSink {
...
 	rx: mpsc::UnboundedReceiver<String>,
 	sub_id: RpcSubscriptionId<'static>,
 }
 
 impl Subscription {
 	/// Close the subscription channel.
 	pub fn close(&mut self) {
-		self.tx.close();
+		tracing::trace!("[Subscription::close] Notifying");
+		self.close_notify.take().and_then(|n| {
```
niklasad1, line 907: I think this is cleaner

```suggestion
if let Some(n) = self.close_notify.take() { 
     n.notify_one() 
}
```

core/src/server/rpc_module.rs
```diff
@@ -729,8 +740,8 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 pub struct SubscriptionSink {
 	/// Sink.
 	inner: MethodSink,
-	/// Close
-	close: async_channel::Receiver<()>,
+	/// Get notified when subscribers leave so we can exit
+	close_notify: Option<Arc<Notify>>,
```
niklasad1, line 744: does this have to be an `Option`?

might get rid of `Arc::clone` but would make the code must harder to reason about...

core/src/server/rpc_module.rs
```diff
@@ -777,47 +788,47 @@ impl SubscriptionSink {
...
+			tracing::warn!("[SubscriptionSink::pipe_from_stream] We're closed.");
+			// TODO: (dp) Is this right? Should return `Err`?
+			return Ok(());
 		}
 	}
 
 	/// Returns whether this channel is closed without needing a context.
 	pub fn is_closed(&self) -> bool {
-		self.inner.is_closed() || self.close.is_closed()
+		self.inner.is_closed() || self.close_notify.is_none()
```
niklasad1, line 831: ah I see why an `Option` is needed

core/src/server/rpc_module.rs
dvdplm: The only reason it's an `Option` is for the `SubscriptionSink` to be able to [check if we've been notified](https://github.com/paritytech/jsonrpsee/blob/fa2072fe28e66385d3fb981d50363ed76b77b741/core/src/server/rpc_module.rs#L831) about the Subscriber leaving. A `Notify` doesn't have a `was_notified()` thing. Can you think of a better way?

core/src/server/rpc_module.rs
dvdplm: Good point.

core/src/server/rpc_module.rs
niklasad1: No :)

tests/tests/integration_tests.rs
```diff
@@ -414,6 +419,71 @@ async fn ws_server_cancels_stream_after_reset_conn() {
...
+				out
+			});
+
+			tokio::spawn(async move {
+				sink.pipe_from_stream(stream).await.unwrap();
+				let send_back = Arc::make_mut(&mut tx);
+				send_back.feed(()).await.unwrap();
+			});
+			Ok(())
+		})
```
dvdplm, line 459: @niklasad1 I think something's wrong here but I'm not sure if it's my test code (likely) or our code. What I see when I run `RUST_LOG=jsonrpsee_core=trace cargo t ws_server_subscribe_with_stream` is this:

```
running 1 test
2022-02-23T12:20:55.022002Z TRACE jsonrpsee_core::client::async_client: [frontend]: subscribe: "subscribe_10_ints", unsubscribe: "unsubscribe_10_ints"
2022-02-23T12:20:55.023045Z TRACE jsonrpsee_core::server::rpc_module: [SubscriptionSink::pipe_from_stream] Entering loop
2022-02-23T12:20:55.128179Z TRACE jsonrpsee_core::server::rpc_module: [SubscriptionSink::pipe_from_stream] Left - sending a result back
2022-02-23T12:20:55.233057Z TRACE jsonrpsee_core::server::rpc_module: [SubscriptionSink::pipe_from_stream] Left - sending a result back
2022-02-23T12:20:55.333694Z TRACE jsonrpsee_core::server::rpc_module: [SubscriptionSink::pipe_from_stream] Left - sending a result back
2022-02-23T12:20:55.439302Z TRACE jsonrpsee_core::server::rpc_module: [SubscriptionSink::pipe_from_stream] Left - sending a result back
2022-02-23T12:20:55.542570Z TRACE jsonrpsee_core::server::rpc_module: [SubscriptionSink::pipe_from_stream] Left - sending a result back
2022-02-23T12:20:55.648021Z TRACE jsonrpsee_core::server::rpc_module: [SubscriptionSink::pipe_from_stream] Left - sending a result back
2022-02-23T12:20:55.752710Z TRACE jsonrpsee_core::server::rpc_module: [SubscriptionSink::pipe_from_stream] Left - sending a result back
2022-02-23T12:20:55.857984Z TRACE jsonrpsee_core::server::rpc_module: [SubscriptionSink::pipe_from_stream] Left - sending a result back
2022-02-23T12:20:55.958600Z TRACE jsonrpsee_core::server::rpc_module: [SubscriptionSink::pipe_from_stream] Left - sending a result back
2022-02-23T12:20:55.958987Z DEBUG jsonrpsee_core::server::rpc_module: Closing subscription: Num(7768645738722848) reason: Some(SubscriptionClosed { reason: Server("No close reason provided") })
2022-02-23T12:20:55.961238Z DEBUG jsonrpsee_core::client::async_client: [backend]: recv method_call Response { jsonrpc: TwoPointZero, result: Number(7768645738722848), id: Number(0) }
2022-02-23T12:20:55.961628Z DEBUG jsonrpsee_core::client::async_client: [backend]: recv subscription Notification { jsonrpc: TwoPointZero, method: "n", params: SubscriptionPayload { subscription: Num(7768645738722848), result: Number(1) } }
2022-02-23T12:20:55.961767Z DEBUG jsonrpsee_core::client::async_client: [backend]: recv subscription Notification { jsonrpc: TwoPointZero, method: "n", params: SubscriptionPayload { subscription: Num(7768645738722848), result: Number(2) } }
2022-02-23T12:20:55.961878Z DEBUG jsonrpsee_core::client::async_client: [backend]: recv subscription Notification { jsonrpc: TwoPointZero, method: "n", params: SubscriptionPayload { subscription: Num(7768645738722848), result: Number(3) } }
2022-02-23T12:20:55.961988Z DEBUG jsonrpsee_core::client::async_client: [backend]: recv subscription Notification { jsonrpc: TwoPointZero, method: "n", params: SubscriptionPayload { subscription: Num(7768645738722848), result: Number(4) } }
2022-02-23T12:20:55.962182Z DEBUG jsonrpsee_core::client::async_client: [backend]: recv subscription Notification { jsonrpc: TwoPointZero, method: "n", params: SubscriptionPayload { subscription: Num(7768645738722848), result: Number(5) } }
2022-02-23T12:20:55.962431Z DEBUG jsonrpsee_core::client::async_client: [backend]: recv subscription Notification { jsonrpc: TwoPointZero, method: "n", params: SubscriptionPayload { subscription: Num(7768645738722848), result: Number(6) } }
2022-02-23T12:20:55.962684Z DEBUG jsonrpsee_core::client::async_client: [backend]: recv subscription Notification { jsonrpc: TwoPointZero, method: "n", params: SubscriptionPayload { subscription: Num(7768645738722848), result: Number(7) } }
2022-02-23T12:20:55.962785Z DEBUG jsonrpsee_core::client::async_client: [backend]: recv subscription Notification { jsonrpc: TwoPointZero, method: "n", params: SubscriptionPayload { subscription: Num(7768645738722848), result: Number(8) } }
2022-02-23T12:20:55.962875Z DEBUG jsonrpsee_core::client::async_client: [backend]: recv subscription Notification { jsonrpc: TwoPointZero, method: "n", params: SubscriptionPayload { subscription: Num(7768645738722848), result: Number(9) } }
2022-02-23T12:20:55.962997Z DEBUG jsonrpsee_core::client::async_client: [backend]: recv subscription Notification { jsonrpc: TwoPointZero, method: "n", params: SubscriptionPayload { subscription: Num(7768645738722848), result: Object({"reason": Object({"Server": String("No close reason provided")})}) } }
2022-02-23T12:20:55.963208Z TRACE jsonrpsee_core::client::async_client: [frontend]: subscribe: "subscribe_10_ints", unsubscribe: "unsubscribe_10_ints"
2022-02-23T12:20:55.963844Z TRACE jsonrpsee_core::server::rpc_module: [SubscriptionSink::pipe_from_stream] Entering loop
2022-02-23T12:20:56.068428Z TRACE jsonrpsee_core::server::rpc_module: [SubscriptionSink::pipe_from_stream] Left - sending a result back
2022-02-23T12:20:56.173776Z TRACE jsonrpsee_core::server::rpc_module: [SubscriptionSink::pipe_from_stream] Left - sending a result back
2022-02-23T12:20:56.274950Z TRACE jsonrpsee_core::server::rpc_module: [SubscriptionSink::pipe_from_stream] Left - sending a result back
2022-02-23T12:20:56.380225Z TRACE jsonrpsee_core::server::rpc_module: [SubscriptionSink::pipe_from_stream] Left - sending a result back
2022-02-23T12:20:56.485690Z TRACE jsonrpsee_core::server::rpc_module: [SubscriptionSink::pipe_from_stream] Left - sending a result back
2022-02-23T12:20:56.587462Z TRACE jsonrpsee_core::server::rpc_module: [SubscriptionSink::pipe_from_stream] Left - sending a result back
2022-02-23T12:20:56.689253Z TRACE jsonrpsee_core::server::rpc_module: [SubscriptionSink::pipe_from_stream] Left - sending a result back
2022-02-23T12:20:56.793708Z TRACE jsonrpsee_core::server::rpc_module: [SubscriptionSink::pipe_from_stream] Left - sending a result back
2022-02-23T12:20:56.897540Z TRACE jsonrpsee_core::server::rpc_module: [SubscriptionSink::pipe_from_stream] Left - sending a result back
2022-02-23T12:20:56.897986Z DEBUG jsonrpsee_core::server::rpc_module: Closing subscription: Num(8581788347571076) reason: Some(SubscriptionClosed { reason: Server("No close reason provided") })
2022-02-23T12:20:56.900452Z DEBUG jsonrpsee_core::client::async_client: [backend]: recv method_call Response { jsonrpc: TwoPointZero, result: Number(8581788347571076), id: Number(2) }
2022-02-23T12:20:56.900874Z DEBUG jsonrpsee_core::client::async_client: [backend]: recv subscription Notification { jsonrpc: TwoPointZero, method: "n", params: SubscriptionPayload { subscription: Num(8581788347571076), result: Number(1) } }
2022-02-23T12:20:56.900991Z DEBUG jsonrpsee_core::client::async_client: [backend]: recv subscription Notification { jsonrpc: TwoPointZero, method: "n", params: SubscriptionPayload { subscription: Num(8581788347571076), result: Number(2) } }
2022-02-23T12:20:56.901083Z DEBUG jsonrpsee_core::client::async_client: [backend]: recv subscription Notification { jsonrpc: TwoPointZero, method: "n", params: SubscriptionPayload { subscription: Num(8581788347571076), result: Number(3) } }
2022-02-23T12:20:56.901174Z DEBUG jsonrpsee_core::client::async_client: [backend]: recv subscription Notification { jsonrpc: TwoPointZero, method: "n", params: SubscriptionPayload { subscription: Num(8581788347571076), result: Number(4) } }
2022-02-23T12:20:56.901448Z DEBUG jsonrpsee_core::client::async_client: [backend]: recv subscription Notification { jsonrpc: TwoPointZero, method: "n", params: SubscriptionPayload { subscription: Num(8581788347571076), result: Number(5) } }
2022-02-23T12:20:56.901541Z DEBUG jsonrpsee_core::client::async_client: [backend]: recv subscription Notification { jsonrpc: TwoPointZero, method: "n", params: SubscriptionPayload { subscription: Num(8581788347571076), result: Number(6) } }
2022-02-23T12:20:56.901655Z DEBUG jsonrpsee_core::client::async_client: [backend]: recv subscription Notification { jsonrpc: TwoPointZero, method: "n", params: SubscriptionPayload { subscription: Num(8581788347571076), result: Number(7) } }
2022-02-23T12:20:56.901746Z DEBUG jsonrpsee_core::client::async_client: [backend]: recv subscription Notification { jsonrpc: TwoPointZero, method: "n", params: SubscriptionPayload { subscription: Num(8581788347571076), result: Number(8) } }
2022-02-23T12:20:56.902038Z DEBUG jsonrpsee_core::client::async_client: [backend]: recv subscription Notification { jsonrpc: TwoPointZero, method: "n", params: SubscriptionPayload { subscription: Num(8581788347571076), result: Number(9) } }
2022-02-23T12:20:56.902565Z DEBUG jsonrpsee_core::client::async_client: [backend]: recv subscription Notification { jsonrpc: TwoPointZero, method: "n", params: SubscriptionPayload { subscription: Num(8581788347571076), result: Object({"reason": Object({"Server": String("No close reason provided")})}) } }
test ws_server_subscribe_with_stream ... ok
```

Notice how the two subscriptions seem to each trigger a full run-through of the stream and that this happens sequentially and before the subscriptions on the client side are even read. I wonder if the stream items are buffered up somewhere awaiting subscribers? That would be crazy though, so my test stream must be "wrong" (in the sense that it's not a useful example of a real world stream). What would a better example be?

Also a bit concerning is that when I `mem::forget(sub1)` I would expect to trigger [this](https://github.com/paritytech/jsonrpsee/blob/7351dadf12d5645652290e37164fb037027e6870/core/src/server/rpc_module.rs#L905) but I never see it. How can I write a test that drops the server side `Subscription` and triggers the `Notify` stuff?

tests/tests/integration_tests.rs
```diff
@@ -456,21 +465,22 @@ async fn ws_server_subscribe_with_stream() {
...
+	assert_eq!(sub2.next().await.unwrap().unwrap(), 2);
+	assert_eq!(sub2.next().await.unwrap().unwrap(), 3);
+
+	// Be rude, don't run the destructor
+	std::mem::forget(sub2);
+	// Sub1 is still in business
+	assert_eq!(sub1.next().await.unwrap().unwrap(), 3);
+
 	// terminate connection.
+	// TODO: (dp) removing the drop changes nothing. Why is that?
```
niklasad1, line 483: because some messages might already have been received and "buffered" so you will have to read the 10 items to be on the safe-side.

depending on execution time

tests/tests/integration_tests.rs
```diff
@@ -414,6 +419,71 @@ async fn ws_server_cancels_stream_after_reset_conn() {
...
+	let server_url = format!("ws://{}", server.local_addr().unwrap());
+
+	let (tx, mut rx) = mpsc::channel(1);
+	let mut module = RpcModule::new(tx);
+
+	module
+		.register_subscription("subscribe_10_ints", "n", "unsubscribe_10_ints", |_, sink, mut tx| {
+			use futures::task::Poll;
+			let mut int_counter = 1usize;
+			let stream = futures::stream::poll_fn(move |_| -> Poll<Option<usize>> {
```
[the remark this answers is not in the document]
dvdplm: It's not: your suggestion actually works the way I wanted. 🌈

tests/tests/integration_tests.rs
dvdplm: I'm still confused. I understand what you're saying here and it makes sense. The client has already received the items from the server, just that my test code hasn't read all of them yet. But I don't understand what this last assert actually does. We should be getting 10 `()`s sent on `rx` (and we do; if I read one more value the test hangs ofc) but how is that tied to dropping the client? It seems to act the same with and without dropping it?

ws-server/src/server.rs
```diff
@@ -324,7 +327,8 @@ async fn background_task(
 
 		// Force `conn_tx` to this async block and close it down
 		// when the connection closes to be on safe side.
-		conn_tx.close();
+		// conn_tx.close();
+		close_notify_server_stop.notify_one();
```
dvdplm, line 331: @niklasad1 I don't understand the comment here fully and I'm not sure if this is actually needed anymore. Should I keep this?

tests/tests/integration_tests.rs
niklasad1: the client may or may have received the items we can't know for user without a mutex or sending them over the channel.

"the last assert" below asserts ensure that server terminated the "sub" after the "the subscription was closed that".
For every subscription that the server closes a message is sent on this channel.
The entire point of that channel is to indicate that the subscription(s) was closed by the server, thus I think don't think it's necessary in this test because we have another test for that.

Thus, if you want that know that you should read the channel twice.

ws-server/src/server.rs
niklasad1: We want to ensure that `close_notify` gets dropped/notified when connection task is terminated.

So I think what you did is correct

tests/tests/integration_tests.rs
niklasad1: Fixed by now

tests/tests/integration_tests.rs
```diff
@@ -477,10 +468,11 @@ async fn ws_server_subscribe_with_stream() {
 	// Sub1 is still in business
 	assert_eq!(sub1.next().await.unwrap().unwrap(), 3);
 
-	// terminate connection.
-	// TODO: (dp) removing the drop changes nothing. Why is that?
-	drop(client);
+	// We expect the stream to run five times and should be getting 5 values on `rx`.
```
niklasad1, line 471:
```suggestion
	// We expect two subscription streams complete to and should be getting 2 values on `rx`.
```

tests/tests/integration_tests.rs
```diff
@@ -477,10 +468,11 @@ async fn ws_server_subscribe_with_stream() {
...
+	// We've read three, so two more expected and then the channel is closed.
+	assert_eq!(Some(()), rx.next().await, "subscription stream should be terminated after the client was dropped");
```
niklasad1, line 473: sorry David, `rx` is not coupled to the actual client subscription just to for us to know if test context that the subscription was terminated if that makes sense.

perhaps we should do something different if the code is so damn hard to understand

tests/tests/integration_tests.rs
```diff
@@ -477,10 +468,11 @@ async fn ws_server_subscribe_with_stream() {
 	// Sub1 is still in business
 	assert_eq!(sub1.next().await.unwrap().unwrap(), 3);
```
niklasad1, line 469:
```suggestion
	assert_eq!(sub1.collect::<Vec<usize>().await.unwrap().unwrap(), vec![3, 4, 5]);
```

tests/tests/integration_tests.rs
```diff
@@ -477,10 +468,11 @@ async fn ws_server_subscribe_with_stream() {
...
 
-	// terminate connection.
-	// TODO: (dp) removing the drop changes nothing. Why is that?
-	drop(client);
+	// We expect the stream to run five times and should be getting 5 values on `rx`.
+	// We've read three, so two more expected and then the channel is closed.
+	assert_eq!(Some(()), rx.next().await, "subscription stream should be terminated after the client was dropped");
 	assert_eq!(Some(()), rx.next().await, "subscription stream should be terminated after the client was dropped");
+	assert!(rx.try_next().is_err());
```
niklasad1, line 475: remove this

niklasad1 (approved):
LGTM

core/src/server/rpc_module.rs
```diff
@@ -777,47 +787,42 @@ impl SubscriptionSink {
...
-					close = close_stream.next();
-				}
-				// Connection terminated.
-				Either::Right((None, _)) => {
-					self.close(&SubscriptionClosed::new(SubscriptionClosedReason::ConnectionReset));
-					break Ok(());
+		if let Some(close_notify) = self.close_notify.clone() {
+			let mut item = stream.next();
+			loop {
+				match futures_util::future::select(item, Box::pin(close_notify.notified())).await {
```
jsdw, line 793: To save some allocatin,g how about moving the `Box::pin(close_notify.notified())` up next to the `let mut item`, and "re-set" it on each loop like we do the item on line 803?

jsdw: (also I wonder whether we could use https://docs.rs/futures/latest/futures/macro.pin_mut.html instead of `Box::pin`?)

niklasad1: good catch missed that

core/src/server/rpc_module.rs
```diff
@@ -777,47 +788,46 @@ impl SubscriptionSink {
...
+						self.close(&SubscriptionClosed::new(SubscriptionClosedReason::ConnectionReset));
+						break Ok(());
+					}
 				}
-				// Stream terminated.
-				Either::Left((None, _)) => break Ok(()),
 			}
+		} else {
+			// The sink is closed.
+			return Ok(());
```
niklasad1, line 824:
```suggestion
			Ok(())
```

core/src/server/rpc_module.rs
```diff
@@ -777,47 +788,46 @@ impl SubscriptionSink {
...
-				// Connection terminated.
-				Either::Right((None, _)) => {
-					self.close(&SubscriptionClosed::new(SubscriptionClosedReason::ConnectionReset));
-					break Ok(());
+		if let Some(close_notify) = self.close_notify.clone() {
+			let mut stream_item = stream.next();
+			let closed_fut = close_notify.notified();
+			pin_mut!(closed_fut);
+			loop {
+				// match futures_util::future::select(item, Box::pin(close_notify.notified())).await {
```
niklasad1, line 796:
```suggestion
```

core/src/server/rpc_module.rs
```diff
@@ -777,47 +788,46 @@ impl SubscriptionSink {
...
-				// if that occurred just ignore and continue.
-				Either::Right((Some(_), i)) => {
-					item = i;
-					close = close_stream.next();
-				}
-				// Connection terminated.
-				Either::Right((None, _)) => {
-					self.close(&SubscriptionClosed::new(SubscriptionClosedReason::ConnectionReset));
-					break Ok(());
+		if let Some(close_notify) = self.close_notify.clone() {
```
niklasad1, line 791: why do we need the clone here @dvdplm?!

dvdplm: I can't move out the `Arc<Notify>` and I can't use a reference either (because `send` takes a mutable reference below). :/

niklasad1: should be possible with `take` here but it didn't work when I tried so I just wonder why :)

dvdplm: `take()` compiles but then the tests fail. :/

niklasad1:
@dvdplm can you remove the `async-channel` in `core` and `ws-server` plz before merging this?

core/src/server/rpc_module.rs
maciejhirsz: Played a bit with this, and not sure there is a super clean solution. You either clone it here, or somehow decouple the `closed_fut` from `self` lifetime wise, which is really hard with pinned futures.

maciejhirsz (approved):
Looks good!

PR #711, opened 2022-03-19
feat(servers): add `max_response_size`

quake:
I wanted to limit max_request_body_size to a relatively small value, say 10 kb, but I found that this setting affected max_response_size, and after reading the code I think the parameter passed here is incorrect:

https://github.com/paritytech/jsonrpsee/blob/9bd212748be033d8461a88988adfc279b1e4fd79/http-server/src/server.rs#L449

to fix this problem, this PR add a max_response_size to the builder,  the default value is u32::MAX.

### DIFF
--- a/http-server/src/server.rs
+++ b/http-server/src/server.rs
@@ -56,6 +56,7 @@ pub struct Builder<M = ()> {
 	access_control: AccessControl,
 	resources: Resources,
 	max_request_body_size: u32,
+	max_response_size: u32,
 	keep_alive: bool,
 	/// Custom tokio runtime to run the server on.
 	tokio_runtime: Option<tokio::runtime::Handle>,
@@ -66,6 +67,7 @@ impl Default for Builder {
 	fn default() -> Self {
 		Self {
 			max_request_body_size: TEN_MB_SIZE_BYTES,
+			max_response_size: u32::MAX,
 			resources: Resources::default(),
 			access_control: AccessControl::default(),
 			keep_alive: true,
@@ -111,6 +113,7 @@ impl<M> Builder<M> {
 	pub fn set_middleware<T: Middleware>(self, middleware: T) -> Builder<T> {
 		Builder {
 			max_request_body_size: self.max_request_body_size,
+			max_response_size: self.max_response_size,
 			resources: self.resources,
 			access_control: self.access_control,
 			keep_alive: self.keep_alive,
@@ -122,6 +125,12 @@ impl<M> Builder<M> {
 	/// Sets the maximum size of a request body in bytes (default is 10 MiB).
 	pub fn max_request_body_size(mut self, size: u32) -> Self {
 		self.max_request_body_size = size;
+		self
+	}
+
+	/// Set the maximum size of a response in bytes. Default is u32::MAX.
+	pub fn max_response_size(mut self, size: u32) -> Self {
+		self.max_response_size = size;
 		self
 	}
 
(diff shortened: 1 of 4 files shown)


niklasad1 (changes requested):
LGTM except the default limit for `max_response_size`, it should be same default for both `requests` and `responses`. I think 10MB is reasonable but I'm open for suggestions...

FWIW, the old code was not wrong simply that we didn't allow that kind of granularity and applied `max_request_size` to responses as well.

However, this is good thank you

http-server/src/server.rs
```diff
@@ -66,6 +67,7 @@ impl Default for Builder {
 	fn default() -> Self {
 		Self {
 			max_request_body_size: TEN_MB_SIZE_BYTES,
+			max_response_size: u32::MAX,
```
niklasad1, line 70: I think 10MB is a better default value for this

ws-server/src/server.rs
```diff
@@ -638,6 +642,7 @@ impl Default for Settings {
 	fn default() -> Self {
 		Self {
 			max_request_body_size: TEN_MB_SIZE_BYTES,
```
niklasad1, line 644: If we agree to name this `max_reponse_size` then we should rename `max_request_body_size` to `max_request_size`

quake: I renamed `max_response_size` to `max_response_body_size`  and kept the `max_request_body_size` unchanged for code compatibility.

niklasad1 (approved):
Thank you, good job

0xslipk (approved):
LGTM

http-server/src/tests.rs
```diff
@@ -405,3 +405,46 @@ async fn run_forever() {
...
+	let addr = "127.0.0.1:0";
+	// Set the max response size to 100 bytes
+	let server = HttpServerBuilder::default().max_response_body_size(100).build(addr).unwrap();
+	let mut module = RpcModule::new(());
+	module.register_method("anything", |_p, _cx| Ok("a".repeat(101))).unwrap();
+	let addr = server.local_addr().unwrap();
+	let uri = to_http_uri(addr);
+	let handle = server.start(module).unwrap();
+
+	// Oversized response.
```
jsdw, line 444: as above was done above test, is it worth also checking that the request size is not an issue when only the response size is limited?

ws-server/src/tests.rs
```diff
@@ -197,6 +197,28 @@ async fn can_set_the_max_request_body_size() {
...
+	let addr = "127.0.0.1:0";
+	// Set the max response body size to 100 bytes
+	let server = WsServerBuilder::default().max_response_body_size(100).build(addr).await.unwrap();
+	let mut module = RpcModule::new(());
+	module.register_method("anything", |_p, _cx| Ok("a".repeat(101))).unwrap();
+	let addr = server.local_addr().unwrap();
+	let handle = server.start(module).unwrap();
+
+	let mut client = WebSocketTestClient::new(addr).await.unwrap();
+
```
jsdw, line 221: Same as my above comment; should we check that setting one hasn't impacted the other here too?

jsdw (approved):
LGTM; comments are minor

lexnv (approved):

http-server/src/server.rs
```diff
@@ -66,6 +67,7 @@ impl Default for Builder {
 	fn default() -> Self {
 		Self {
 			max_request_body_size: TEN_MB_SIZE_BYTES,
+			max_response_body_size: TEN_MB_SIZE_BYTES,
```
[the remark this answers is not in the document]
niklasad1: yeah, could make sense :)

http-server/src/tests.rs
niklasad1: it won't work unless you register another method for that.

PR #712, opened 2022-03-21
perf: use async_lock instead of parking_lot

librelois:
`parkint_lot` uses a blocking mutex, i.e. if the lock cannot be obtained immediately, the whole thread is put on hold.
It is a bad practice to use a "classic" mutex in the context of an asynchronous executor like tokio for example.
The threads of the tokio threadpool dedicated to the execution of async tasks must never be put on hold by the operating system, it is up to tokio to manage this directly.
The `parkint_lot` mutex yields to the operating system, while the `async_lock` mutex yields to the chosen asynchronous executor (usually tokio).

There are other asynchronous implementations of mutexes, I chose async_lock because it is the most proven and "light" (it does not impose a particular asynchronous executor to the user).

I left parking_lot at the level of opening subscriptions, because it requires a more complex refactoring, and it is less impacting for the performances (there are in practice much more requests than openings of subscriptions).
But it would be nice to migrate also the opening of subscriptions in a next PR.

### DIFF
--- a/core/src/server/resource_limiting.rs
+++ b/core/src/server/resource_limiting.rs
@@ -93,7 +93,7 @@ use std::sync::Arc;
 
 use crate::Error;
 use arrayvec::ArrayVec;
-use parking_lot::Mutex;
+use async_lock::Mutex;
 
 // The number of kinds of resources that can be used for limiting.
 const RESOURCE_COUNT: usize = 8;
@@ -139,8 +139,8 @@ impl Resources {
 	/// Attempt to claim `units` units for each resource, incrementing current totals.
 	/// If successful, returns a [`ResourceGuard`] which decrements the totals by the same
 	/// amounts once dropped.
-	pub fn claim(&self, units: ResourceTable) -> Result<ResourceGuard, Error> {
-		let mut totals = self.totals.lock();
+	pub async fn claim(&self, units: ResourceTable) -> Result<ResourceGuard, Error> {
+		let mut totals = self.totals.lock().await;
 		let mut sum = *totals;
 
 		for (idx, sum) in sum.iter_mut().enumerate() {
@@ -169,8 +169,12 @@ pub struct ResourceGuard {
 
 impl Drop for ResourceGuard {
 	fn drop(&mut self) {
-		for (sum, claimed) in self.totals.lock().iter_mut().zip(self.units) {
-			*sum -= claimed;
-		}
+		let totals = self.totals.clone();
+		let units = self.units;
+		tokio::spawn(async move {
+			for (sum, claimed) in totals.lock().await.iter_mut().zip(units) {
+				*sum -= claimed;
+			}
+		});
 	}
 }
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -10,6 +10,7 @@ license = "MIT"
 anyhow = "1"
 arrayvec = "0.7.1"
 async-trait = "0.1"
+async-lock = { version = "2.4.0", optional = true }
 beef = { version = "0.5.1", features = ["impl_serde"] }
 thiserror = "1"
 futures-channel = { version = "0.3.14", default-features = false }
@@ -29,6 +30,7 @@ tokio = { version = "1.8", optional = true }
 default = []
 http-helpers = ["futures-util"]
 server = [
+	"async-lock",
 	"futures-util",
 	"rustc-hash",
 	"tracing",
--- a/core/src/server/rpc_module.rs
+++ b/core/src/server/rpc_module.rs
@@ -180,10 +180,10 @@ impl MethodCallback {
 
 	/// Attempt to claim resources prior to executing a method. On success returns a guard that releases
 	/// claimed resources when dropped.
-	pub fn claim(&self, name: &str, resources: &Resources) -> Result<ResourceGuard, Error> {
+	pub async fn claim(&self, name: &str, resources: &Resources) -> Result<ResourceGuard, Error> {
 		match self.resources {
 			MethodResources::Uninitialized(_) => Err(Error::UninitializedMethod(name.into())),
-			MethodResources::Initialized(units) => resources.claim(units),
+			MethodResources::Initialized(units) => resources.claim(units).await,
 		}
 	}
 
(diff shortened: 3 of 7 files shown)


niklasad1:
Hey @librelois

Do you have any benchmarks results that proves that this makes things better?
We could also use `tokio::sync::Mutex` it's already in the dependency tree.

dvdplm:
> We could also use `tokio::sync::Mutex` it's already in the dependency tree.

Hmm. An async lock of any kind is likely slower than a sync lock (see [the tokio article on shared state](https://tokio.rs/tokio/tutorial/shared-state)). If we can confirm that the resources mutex is indeed a performance problem with a benchmark then I think we should look into trying to be smarter about when we lock. :/

core/src/server/resource_limiting.rs
```diff
@@ -169,8 +169,12 @@ pub struct ResourceGuard {
 
 impl Drop for ResourceGuard {
 	fn drop(&mut self) {
-		for (sum, claimed) in self.totals.lock().iter_mut().zip(self.units) {
-			*sum -= claimed;
-		}
```
dvdplm, line 174: To my eyes this does not look slow; the lock should be held for a **really** short time here.

core/src/server/resource_limiting.rs
```diff
@@ -169,8 +169,12 @@ pub struct ResourceGuard {
...
+		let totals = self.totals.clone();
+		let units = self.units;
+		tokio::spawn(async move {
```
dvdplm, line 174: Instinctively spawning a task in a `Drop` impl seems like a bad idea.

librelois:
> Hmm. An async lock of any kind is likely slower than a sync lock (see [the tokio article on shared state](https://tokio.rs/tokio/tutorial/shared-state)). 

The article you point out states an assumption that is not true for async_lock: "However, switching to tokio::sync::Mutex usually does not help as the asynchronous mutex uses a synchronous mutex internally."
If this is true for the tokio mutex, it is not the case for async_lock, which is based on event-listener, the latter using only atomic processor instructions, thus not calling the operating system.

> If we can confirm that the resources mutex is indeed a performance problem with a benchmark then I think we should look into trying to be smarter about when we lock. :/

From our first tests, it seems that we observe performance regressions from jsonrpc to josnrpsee in case there are many requests at the same time. And after the change of this PR, we did not observe any regression.

We will try to reproduce this in reproducible tests and then we will give you more details.

niklasad1:
@librelois 

On which transport did you experience these regressions? On both or only HTTP?

I tried your PR on https://github.com/paritytech/jsonrpsee/pull/713

```$ cargo bench ws_concurrent_conn_calls```

```$ cargo bench http_concurrent_conn_calls```

I couldn't find that this was faster on `ws` for example than current master but removing the mutex entirely was ~10-50% faster.

On HTTP jsonrpsee is slower than the `jsonrpc crate` according the benchmarks for `conns >= 512`.

Also it possible to run these benchmarks against `jsonrpc crate` by ```$ cargo bench http_concurrent_conn_calls --features jsonrpc-crate```

maciejhirsz (commented):
To add to what @dvdplm already listed, [Mutex documentation in tokio is pretty clear](https://docs.rs/tokio/1.17.0/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use):

> Contrary to popular belief, it is ok and often preferred to use the ordinary [Mutex](https://doc.rust-lang.org/nightly/std/sync/mutex/struct.Mutex.html) from the standard library in asynchronous code.
>
> The feature that the async mutex offers over the blocking mutex is the ability to keep it locked across an .await point. This makes the async mutex more expensive than the blocking mutex, so the blocking mutex should be preferred in the cases where it can be used. The primary use case for the async mutex is to provide shared mutable access to IO resources such as a database connection. If the value behind the mutex is just data, it’s usually appropriate to use a blocking mutex such as the one in the standard library or [parking_lot](https://docs.rs/parking_lot).

---

Referring to `tokio` since that's the runtime we're using, but to be prudent:

> If this is true for the tokio mutex, it is not the case for async_lock, which is based on event-listener, the latter using only atomic processor instructions, thus not calling the operating system.

`event-listener` [uses a sync `Mutex`](https://github.com/smol-rs/event-listener/blob/c73ad619f6f9a76d6dd87bfcac02231958a4bbad/src/lib.rs#L80-L92) internally, so `async_lock` does too. Async code isn't magic, if you need to synchronize multiple waiters, you must, at some point, use synchronous primitives to do it.

I'm not philosophically opposed to using async `Mutex` here (be it the one from `async_lock` or `tokio`), but nothing I've seen would suggest this to be an improvment. All of our locks are extremely short, the math is done on a fixed-size arrays and likely already auto-vectorized to SIMD by rustc. If this PR fixes regressions, it might be a side effect of some other change, like removing the `join_all` on batches or some such.

librelois:
Thank you @niklasad1 for your benchmarks, they confirm the performance regression in HTTP that we observe and that is blocking us.

Indeed, our rpc clients are the tools that consume the ETH RPC api, and most of these tools use HTTP.
We already have a lot of trouble handling the huge HTTP traffic to our RPC nodes, we can't afford any regression at this level.

I tested the `http_concurrent_conn_calls` benchmark on a machine with 32 cores, and I see indeed a huge regression with jsonrpsee, about 98%!

I ran it in this order:

1. jsonrpsee master 78055feb19dc95f53f6145aba606a253efe9f82a
2. This PR
3. jsonrpc

In 2. I observe a gain of 11% for the `async/1024` case only, the performance is similar in the other cases.

In 3. I observe a performance gain of 98% in `sync/512`, `sync/1024`, `async/512` and `async/1024`, similar performance in the other cases.

Conclusions:

- The regression is very significant.
- My PR does not solve this regression.

maciejhirsz:
@librelois 

> I tested the `http_concurrent_conn_calls` benchmark on a machine with 32 cores, and I see indeed a huge regression with jsonrpsee, about 98%!
> 
> ...
>
> In 3. I observe a performance gain of 98% in `sync/512`, `sync/1024`, `async/512` and `async/1024`, similar performance in the other cases.

I assume that's the same 98%, as in using jsonrpc is twice as fast as jsonrpsee, not that jsonrpsee is 50x slower (which is what 98% _regression_ would be).

I'll close this PR for now given this is unlikely to be the main culprit. Can you make a separate issue for this? WS has been our priority thus far, I reckon there should be ample room to improve things on HTTP side too.

librelois:
The regression issue: https://github.com/paritytech/jsonrpsee/issues/716

Issue #685, opened 2022-02-01
[benchmarks] Panic while running benchmarks


lexnv:
### Details
- `cargo bench -p jsonrpsee-benchmarks -- --output-format bencher` results in a repeating panic
- commit: `2864756 - dependabot[bot] - chore(deps): update parking_lot requirement from 0.11 to 0.12 (#682)`
- os | hardware: `macOS Monterey v12.2` | apple M1 pro

### Panic Details

```
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: Transport(Error while performing the HTTP request)', benches/bench.rs:286:107
stack backtrace:
   0: rust_begin_unwind
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/std/src/panicking.rs:498:5
   1: core::panicking::panic_fmt
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/panicking.rs:107:14
   2: core::result::unwrap_failed
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/result.rs:1613:5
   3: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
   4: tokio::runtime::task::core::CoreStage<T>::poll
   5: tokio::runtime::task::harness::Harness<T,S>::poll
   6: std::thread::local::LocalKey<T>::with
   7: tokio::runtime::thread_pool::worker::Context::run_task
   8: tokio::runtime::thread_pool::worker::Context::run
   9: tokio::macros::scoped_tls::ScopedKey<T>::set
  10: tokio::runtime::thread_pool::worker::run
  11: tokio::loom::std::unsafe_cell::UnsafeCell<T>::with_mut
  12: tokio::runtime::task::harness::Harness<T,S>::poll
  13: tokio::runtime::blocking::pool::Inner::run
```

dvdplm:
> * `cargo bench -p jsonrpsee-benchmarks -- --output-format bencher` results in a repeating panic

Is it the use of `--output-format` that triggers the panic?

niklasad1:
nah, it most likely [this](https://github.com/paritytech/jsonrpsee/blob/master/benches/bench.rs#L279) that opens too many connections

similar to what you experienced before @dvdplm with your `ulimit attempts` on `macOS`

@lexnv 

perhaps you could try modify the number of connections and report back if you still get the same panic

[...]

PR #713, opened 2022-03-22
benches: improve concurrent benches with 10 requests per conn

niklasad1:
Improve current benchmarks to run up 1024 concurrent connections,

websocket 10 calls are sent per connections
http 1 call per connection (doesn't make sense make several calls using the http client)


benches/README.md
```diff
@@ -0,0 +1,28 @@
+# jsonrpsee benchmarks
+
+This crate contains benchmarks mainly to test the server implementations of some common scenarios such as concurrent connections.
+Further, running these will open lots of sockets and file descriptors, it doesn't work well on macOS for instance.
```
jsdw, line 4: Is it worth suggesting some commands that can be run to help with this? (eg https://github.com/paritytech/substrate-telemetry/blob/d6ed878cb41a97f7b7b546e996f46ac5dd6215aa/backend/telemetry_core/tests/soak_tests.rs#L22 .. not sure how many of those are actually useful but I just paste them all each time!)

dvdplm (approved):

jsdw (approved):
Alas I was still unable to run on MacOS (my guess is it's trying to open too many connections at once, and benchmarking telemetry I found on MacOS that I needed to open connections one at a time and then run the benchmark), but the code looks good, so happy to approve assuming they run fine on linux!

We could potentially add a followup issue to see how easily we could get them running on MacOS too.

niklasad1:
ah, I was expecting that but thanks for testing that anyway.
We already got an issue for this https://github.com/paritytech/jsonrpsee/issues/685

Last I checked it worked fine running it right now again after the recent changes.

Issue #716, opened 2022-03-23
Huge regression in case of too many HTTP requests


librelois:
I tested the `http_concurrent_conn_calls` benchmark on a machine with 32 cores, and I see indeed a huge regression with jsonrpsee, about 98%!

I ran it in this order:

1. jsonrpsee master https://github.com/paritytech/jsonrpsee/commit/78055feb19dc95f53f6145aba606a253efe9f82a
2. jsonrpc

In 2. I observe a performance gain of 98% in `sync/512`, `sync/1024`, `async/512` and `async/1024`, similar performance in the other cases.

This means that jsonrpsee is twice as slow as jsonrpc in cases `sync/512`, `sync/1024`, `async/512` and `async/1024`.

niklasad1:
Closed by #718, feel to try it out again @librelois 

A followup PR is coming soon where it will possible to configure to `backlog` in the HTTP server.

PR #718, opened 2022-03-26
http server: fix regression set backlog to 1024

niklasad1:
Closing #716

### DIFF
--- a/http-server/src/server.rs
+++ b/http-server/src/server.rs
@@ -223,7 +223,7 @@ impl<M> Builder<M> {
 		let address = addr.into();
 		socket.bind(&address)?;
 
-		socket.listen(128)?;
+		socket.listen(1024)?;
 		let listener: TcpListener = socket.into();
 		let local_addr = listener.local_addr().ok();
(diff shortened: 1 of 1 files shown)


jsdw (approved):
LGTM!

Any idea what the negative effects of a higher backlog is? As a followup I wonder whether it's worth making this configurable?

niklasad1:
> Any idea what the negative effects of a higher backlog is? As a followup I wonder whether it's worth making this configurable?

I don't there is any downsides with a bigger backlog rather than that the kernel might use more memory and potential DOS for big values (also the OS might truncate this value). For instance the max size for Linux is 4096 and this is highly platform dependent so yes we should make this configurable for use but 1024 seems like more reasonable default value for instance `tokio/mio` and `paritytech/jsonrpc` uses 1024 already.

Let's take care of the configuration in another PR.

lexnv:
I wonder if this is capped by `kern.ipc.somaxconn`, currently the MacOS default is `kern.ipc.somaxconn=128`.

lexnv (approved):

TarikGul (approved):

PR #719, opened 2022-03-29
feat(http server): add new builder APIs `build_from_tcp` and `build_from_hyper`

niklasad1:
Before the HTTP server configured the socket explictly but this PR changes to use `tokio::net::TcpListener` similar to what the `WsServer does`

In addition this PR adds another API for users to provide their own TcpListener, see the inline example I added with socket2.

Example:

 ```rust
	/// use jsonrpsee_http_server::HttpServerBuilder;
	/// use socket2::{Domain, Socket, Type};
	///
	/// #[tokio::main]
	/// async fn main() {
	///   let addr = "127.0.0.1:0".parse().unwrap();
	///   let domain = Domain::for_address(addr);
	///   let socket = Socket::new(domain, Type::STREAM, None).unwrap();
	///   socket.set_nodelay(true).unwrap();
	///   socket.set_reuse_address(true).unwrap();
	///   socket.set_nonblocking(true).unwrap();
	///   socket.set_keepalive(true).unwrap();
	///
	///   let address = addr.into();
	///   socket.bind(&address).unwrap();
	///
	///   socket.listen(4096).unwrap();
	///
	///   let server = HttpServerBuilder::new().build_from_tcp(socket).await.unwrap();
	/// }
	/// ```

```

### DIFF
--- a/http-server/src/server.rs
+++ b/http-server/src/server.rs
@@ -157,8 +157,7 @@ impl<M> Builder<M> {
 		self
 	}
 
-	/// Finalizes the configuration of the server with customized TCP settings on the socket.
-	///
+	/// Finalizes the configuration of the server with customized TCP settings on the socket and on hyper.
 	///
 	/// ```rust
 	/// use jsonrpsee_http_server::{HttpServerBuilder, HyperTcpConfig};
@@ -170,10 +169,7 @@ impl<M> Builder<M> {
 	///   let addr = "127.0.0.1:0".parse().unwrap();
 	///   let domain = Domain::for_address(addr);
 	///   let socket = Socket::new(domain, Type::STREAM, None).unwrap();
-	///   socket.set_nodelay(true).unwrap();
-	///   socket.set_reuse_address(true).unwrap();
 	///   socket.set_nonblocking(true).unwrap();
-	///   socket.set_keepalive(true).unwrap();
 	///
 	///   let address = addr.into();
 	///   socket.bind(&address).unwrap();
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

**Comment by jsdw** [comment]
Ooh offhand this looks very neat, I like it! I also like that it removes socket2 as a dependency and leaves it to the user to provide their own socket thing (if I read it right)!

**Comment by niklasad1** [comment]
yeah, you are right and that is the idea :)

AFAIU, most socket libraries implements `Into<StdTcpListener>` and `hyper` has a similar API

File: http-server/src/server.rs [hunk 1]
```diff
@@ -178,56 +218,26 @@ impl<M> Builder<M> {
...
-		socket.set_keepalive(self.keep_alive)?;
-		let address = addr.into();
-		socket.bind(&address)?;
-
-		socket.listen(1024)?;
-		let listener: TcpListener = socket.into();
 		let local_addr = listener.local_addr().ok();
-		let listener = hyper::Server::from_tcp(listener)?;
-		Ok((listener, local_addr))
+		let listener = hyper::Server::from_tcp(listener)?.tcp_nodelay(true);
```
jsdw, line 229: in the comment above, you say that hyper sets tcp_nodelay == true, so is this `nodelay(true)` going to be ignored anyway?

jsdw (approved):
Loooks good to me; I really like letting somebody provide (and configure) their own socket rather than handling it internally :)

http-server/src/server.rs
niklasad1: yeah, I was thinking that we might want to expose these settings in the builder.

It's like a warning that even if folk configure these on their own sockets it might get overwritten but let's revert this then.

lexnv (approved):

http-server/src/server.rs
niklasad1: nah, my comment was wrong hyper sets it to false

http-server/src/server.rs
```diff
@@ -167,6 +157,56 @@ impl<M> Builder<M> {
...
+	///   socket.set_reuse_address(true).unwrap();
+	///   socket.set_nonblocking(true).unwrap();
+	///   socket.set_keepalive(true).unwrap();
+	///
+	///   let address = addr.into();
+	///   socket.bind(&address).unwrap();
+	///
+	///   socket.listen(4096).unwrap();
+	///
+	///   // hyper does some settings on the provided socket, ensure that nothing breaks the our settings.
```
niklasad1, line 183:
```suggestion
	///   // hyper does some settings on the provided socket, ensure that nothing breaks our "expected settings".
```

http-server/src/server.rs
```diff
@@ -167,6 +157,56 @@ impl<M> Builder<M> {
...
+	///
+	///   socket.listen(4096).unwrap();
+	///
+	///   // hyper does some settings on the provided socket, ensure that nothing breaks our "expected settings".
+	///   let hyper_cfg = HyperTcpConfig { sleep_on_accept_errors: true, keepalive_timeout: Some(Duration::from_secs(1)), no_delay: true };
+	///
+	///   let server = HttpServerBuilder::new().build_from_tcp(socket, hyper_cfg).unwrap();
+	/// }
+	/// ```
+	pub fn build_from_tcp(self, listener: impl Into<StdTcpListener>, cfg: HyperTcpConfig) -> Result<Server<M>, Error> {
```
niklasad1, line 189: @jsdw is this API reasonable?

Basically the only thing I want to ensure is that no settings gets overwritten without anything informing the users of this API.

It's a bit awkward to some settings twice but no way around that....

niklasad1: fair enough, I think your suggestion is less awkward.

PR #720, opened 2022-03-30
feat: add  `SubscriptionSink::pipe_from_try_stream` to support streams that returns `Result`

niklasad1:
The rationale for this is that it is more flexible for use cases when `Stream<Item = Result<T, Error>>`.

Take for example `tokio_stream::Broadcast` then one would have to something like:

```rust
   let stream = BroadcastStream::new(rx).take_while(|r| future::ready(r.is_ok())).filter_map(|r| future::ready(r.ok()));
```

Note, I realized that `pipe_from_stream` will work if Error: Serialize bound is satisfied but that's unlikely for the error types but I  added a comment about that

### DIFF
--- a/core/src/server/rpc_module.rs
+++ b/core/src/server/rpc_module.rs
@@ -764,7 +764,7 @@ impl SubscriptionSink {
 	/// when items gets produced by the stream.
 	///
 	/// Returns `Ok(())` if the stream or connection was terminated.
-	/// Returns `Err(_)` if the underlying stream returns an error or if an item from the stream could not be serialized.
+	/// Returns `Err(_)` immediately if the underlying stream returns an error or if an item from the stream could not be serialized.
 	///
 	/// # Examples
 	///
@@ -775,7 +775,9 @@ impl SubscriptionSink {
 	///
 	/// let mut m = RpcModule::new(());
 	/// m.register_subscription("sub", "_", "unsub", |params, mut sink, _| {
-	///     let stream = futures_util::stream::iter(vec![Ok(1_u32), Ok(2), Ok(3), Err("error on the stream")]);
+	///     let stream = futures_util::stream::iter(vec![Ok(1_u32), Ok(2), Err("error on the stream")]);
+	///     // This will return send `[Ok(1_u32), Ok(2_u32)]` to the subscriber
+	///     // because after the `Err(_)` the stream is terminated.
 	///     tokio::spawn(sink.pipe_from_stream(stream));
 	///     Ok(())
 	/// });
@@ -808,7 +810,9 @@ impl SubscriptionSink {
 						closed_fut = next_closed_fut;
 					}
 					Either::Left((Some(Err(e)), _)) => {
-						break Err(Error::SubscriptionClosed(SubscriptionClosedReason::Server(e.to_string()).into()));
+						let close_reason = SubscriptionClosedReason::Server(e.to_string()).into();
+						self.close(&close_reason);
+						break Err(Error::SubscriptionClosed(close_reason));
 					}
 					// Stream terminated.
 					Either::Left((None, _)) => break Ok(()),
--- a/core/src/server/rpc_module.rs
+++ b/core/src/server/rpc_module.rs
@@ -764,25 +764,27 @@ impl SubscriptionSink {
 	/// when items gets produced by the stream.
 	///
 	/// Returns `Ok(())` if the stream or connection was terminated.
-	/// Returns `Err(_)` if one of the items couldn't be serialized.
+	/// Returns `Err(_)` if the underlying stream return an error or if an item from the stream could not be serialized.
 	///
 	/// # Examples
 	///
 	/// ```no_run
 	///
 	/// use jsonrpsee_core::server::rpc_module::RpcModule;
+	/// use anyhow::anyhow;
 	///
 	/// let mut m = RpcModule::new(());
 	/// m.register_subscription("sub", "_", "unsub", |params, mut sink, _| {
-	///     let stream = futures_util::stream::iter(vec![1_u32, 2, 3]);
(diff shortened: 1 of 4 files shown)


core/src/server/rpc_module.rs
```diff
@@ -764,25 +764,27 @@ impl SubscriptionSink {
 	/// when items gets produced by the stream.
 	///
 	/// Returns `Ok(())` if the stream or connection was terminated.
-	/// Returns `Err(_)` if one of the items couldn't be serialized.
+	/// Returns `Err(_)` if the underlying stream return an error or if an item from the stream could not be serialized.
```
TarikGul, line 767:
```suggestion
	/// Returns `Err(_)` if the underlying stream returns an error or if an item from the stream could not be serialized.
```

TarikGul (approved):
Overall, lgtm 👍 .

Only nit is the failing gitlab test:

```
tokio::spawn(sink.pipe_from_stream(stream));
   |                       ---------------- ^^^^^^ the trait `StdError` is not implemented for `anyhow::Error`
```

tests/tests/integration_tests.rs
```diff
@@ -399,7 +399,7 @@ async fn ws_server_cancels_stream_after_reset_conn() {
 	module
 		.register_subscription("subscribe_never_produce", "n", "unsubscribe_never_produce", |_, sink, mut tx| {
 			// create stream that doesn't produce items.
-			let stream = futures::stream::empty::<usize>();
+			let stream = futures::stream::empty::<usize>().map(|i| Ok::<_, Error>(i));
```
jsdw, line 402: perhaps `empty::<Result<usize, Error>>()` and no closure might be cleaner?

core/src/server/rpc_module.rs
```diff
@@ -764,25 +764,29 @@ impl SubscriptionSink {
...
 	/// ```no_run
 	///
 	/// use jsonrpsee_core::server::rpc_module::RpcModule;
+	/// use anyhow::anyhow;
 	///
 	/// let mut m = RpcModule::new(());
 	/// m.register_subscription("sub", "_", "unsub", |params, mut sink, _| {
-	///     let stream = futures_util::stream::iter(vec![1_u32, 2, 3]);
+	///     let stream = futures_util::stream::iter(vec![Ok(1_u32), Ok(2), Err("error on the stream")]);
+	///     // This will return send `[Ok(1_u32), Ok(2_u32)]` to the subscriber
```
jsdw, line 779: Won't it send [Ok(..), Ok(..), Err(..)] to the subscriber, (and nothing else after the error)?

core/src/server/rpc_module.rs
```diff
@@ -764,25 +764,29 @@ impl SubscriptionSink {
...
+	///     // because after the `Err(_)` the stream is terminated.
 	///     tokio::spawn(sink.pipe_from_stream(stream));
 	///     Ok(())
 	/// });
 	/// ```
-	pub async fn pipe_from_stream<S, T>(mut self, mut stream: S) -> Result<(), Error>
+	pub async fn pipe_from_stream<S, T, E>(mut self, mut stream: S) -> Result<(), Error>
```
jsdw, line 785: It definitely makes sense to me to support streams of results (so that we can handle the errors better and such)!

I wonder whether it's worth supporting both `pipe_from_stream` and `pipe_result_from_stream` or something (the former just converting everything to `Ok(..)` and sending it on to `pipe_result_from_stream`) just for the added convenience?

core/src/server/rpc_module.rs
niklasad1: yes, nice catch

core/src/server/rpc_module.rs
niklasad1: good idea, yepp I like that

tests/tests/integration_tests.rs
```diff
@@ -423,6 +423,44 @@ async fn ws_server_cancels_stream_after_reset_conn() {
...
+	server.start(module).unwrap();
+
+	let client = WsClientBuilder::default().build(&server_url).await.unwrap();
+	let mut sub: Subscription<usize> =
+		client.subscribe("subscribe_with_err_on_stream", None, "unsubscribe_with_err_on_stream").await.unwrap();
+
+	assert_eq!(sub.next().await.unwrap().unwrap(), 1);
+	let exp = SubscriptionClosed::new(SubscriptionClosedReason::Server(err.to_string()));
+	// The server closed down the subscription with the underlying error from the stream.
+	assert!(matches!(sub.next().await, Some(Err(Error::SubscriptionClosed(close_reason))) if close_reason == exp));
```
jsdw, line 461: Is it worth just calling `sub.next().await` once more after this to confirm that it's `None` and that the thing is closed?

jsdw (approved):
LGTM! A couple of thoughts but nothing blocking; do as you wish with them :)

tests/tests/integration_tests.rs
niklasad1: yeah, just found a bug let's tackle it another PR.

tests/tests/integration_tests.rs
```diff
@@ -423,6 +423,45 @@ async fn ws_server_cancels_stream_after_reset_conn() {
...
+	sub.next().await.unwrap();
```
niklasad1, line 462:
```suggestion
```

proc-macros/src/lib.rs
```diff
@@ -288,7 +288,7 @@ pub(crate) mod visitor;
 ///         // as subscription responses.
 ///         fn sub_override_notif_method(&self, mut sink: SubscriptionSink) -> RpcResult<()> {
 ///             tokio::spawn(async move {
-///                 let stream = futures_util::stream::iter(["one", "two", "three"]);
+///                 let stream = futures_util::stream::iter([Ok::<_, &str>("one"), Ok("two"), Ok("three")]);
```
niklasad1, line 291:
```suggestion
///                 let stream = futures_util::stream::iter(["one", "two", "three"]);
```

PR #725, opened 2022-04-05
fix(ws server): close all subscription when the connection is closed

niklasad1:


tests/tests/integration_tests.rs
```diff
@@ -393,34 +393,40 @@ async fn ws_server_cancels_stream_after_reset_conn() {
...
 	let server_url = format!("ws://{}", server.local_addr().unwrap());
 
-	let (tx, mut rx) = mpsc::channel(1);
+	let (tx, rx) = mpsc::channel(1);
 	let mut module = RpcModule::new(tx);
 
 	module
-		.register_subscription("subscribe_never_produce", "n", "unsubscribe_never_produce", |_, sink, mut tx| {
-			// create stream that doesn't produce items.
-			let stream = futures::stream::empty::<usize>();
```
niklasad1, line 402: this test was flaky I don't know what I was thinking.

niklasad1: Further, this could complete for `notify_one` successfully if the `receiver` from the connection is closed once the a new item is produced on the stream.

so really tricky test but I added sleep for one hour so should be "okeyisch" really tricky to test this.

ws-server/src/server.rs
```diff
@@ -326,9 +326,8 @@ async fn background_task(
 		// Terminate connection and send close message.
 		let _ = sender.close().await;
 
-		// Force `conn_tx` to this async block and close it down
-		// when the connection closes to be on safe side.
-		close_notify_server_stop.notify_one();
+		// Notify all listeners and close down associated tasks.
+		close_notify_server_stop.notify_waiters();
```
niklasad1, line 330: this is the fix

jsdw (approved):
LGTM from what I understand (and well done for finding this!!)! I assume we need notify_waiters because there is more than one thing that might be waiting on that to resolve in order to properly clean up (and it def seems like the "safe" option regardless)

tests/tests/integration_tests.rs
```diff
@@ -393,34 +393,40 @@ async fn ws_server_cancels_stream_after_reset_conn() {
...
+	for _ in 0..10 {
+		subs.push(client.subscribe::<usize>("subscribe_for_ever", None, "unsubscribe_for_ever").await.unwrap());
+	}
 
 	// terminate connection.
 	drop(client);
-	assert_eq!(Some(()), rx.next().await, "subscription stream should be terminated after the client was dropped");
-	assert_eq!(Some(()), rx.next().await, "subscription stream should be terminated after the client was dropped");
+
+	let rx_len = rx.take(10).fold(0, |acc, _| async move { acc + 1 }).await;
```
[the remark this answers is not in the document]
niklasad1: it could a few milliseconds to send the WS close message but so not really is my understanding `rx` here is another channel where `tx` is encapsulated in the `RpcModule` so it's not the `subscription stream`

Then after connection is closed the actual subscriptions should be terminated using `Notify::notify_waiters` after that this messages and sent to channel above...

ws-server/src/server.rs
TarikGul: Ahh okay makes sense. So listeners were being launched from `pipe_from_try_stream`, and in this case because it was `notify_one` only a single task was stopped. 

Good catch 👍

TarikGul (approved):
lgtm, good catch on the notifications.

In terms of the integration tests, I see how it's difficult to test. Will keep this one in the back on my mind.

ws-server/src/server.rs
niklasad1: yeah, basically every connection can have n number of subscriptions and we want to cancel them all when the connection is killed.

the subscriptions will/may be canceled as soon some message is sent via them but that might take a while so we want to do it as soon as the connection is closed

PR #726, opened 2022-04-05
refactor(log): downgrade send errors to `warn`

niklasad1:
These logs are most likely related to that the client terminated the connection
and they come with significant overhead because it occurs almost on every connection reset.


TarikGul (approved):
lgtm, the simplication of the error messages are also nice.

jsdw (approved):

PR #728, opened 2022-04-07
fix(rpc module): fail subscription calls with bad params

niklasad1:
To summarize, with this PR:
- We can reject a pending subscription with a `{ jsonrpc: "2.0", error: {...}, "id":<ID> }` style error of our choice using `pending.reject(error)`.
- We can end a subscription at any time with `{ jsonrpsee: "2.0", method: "subscribe_foo", params: { subscription: "123abc", error: { ... } } }` style error using `sink.close(error)`.
- Subscriptions that are dropped are now not silently closed with an error anymore (for compatibility reasons, incase subscriptions want to signal that they got closed by: "completed successfully" or "error" they must either call `close` on SubscriptionSink explictly). In other words, we enforce no opinion on what happens when a subscription is closed, and leave it up to the subscription itself to decide whether to emit an "error" or communicate closure via some other "ordinary notification".
- The clients have been changed to require the new error format for closing subscriptions `{ jsonrpsee: "2.0", method: "subscribe_foo", params: { subscription: "123abc", error: { ... } } }`
- `Subscription::next` in the client no longer parses `SubscriptionClosed` instead just returns `None` when a closed message is received.

Tested it on: https://github.com/paritytech/substrate/compare/dp-jsonrpsee-integration-2...na-jsonrpsee-comp-bad-params?expand=1

### DIFF
--- a/core/src/client/async_client/helpers.rs
+++ b/core/src/client/async_client/helpers.rs
@@ -31,6 +31,7 @@ use crate::client::{RequestMessage, TransportSenderT};
 use crate::Error;
 
 use futures_channel::{mpsc, oneshot};
+use jsonrpsee_types::error::CallError;
 use jsonrpsee_types::response::SubscriptionError;
 use jsonrpsee_types::{
 	ErrorResponse, Id, Notification, ParamsSer, RequestSer, Response, SubscriptionId, SubscriptionResponse,
@@ -243,12 +244,12 @@ pub(crate) fn process_error_response(manager: &mut RequestManager, err: ErrorRes
 	match manager.request_status(&id) {
 		RequestStatus::PendingMethodCall => {
 			let send_back = manager.complete_pending_call(id).expect("State checked above; qed");
-			let _ = send_back.map(|s| s.send(Err(Error::Call(err.error.to_call_error()))));
+			let _ = send_back.map(|s| s.send(Err(Error::Call(CallError::Custom(err.error.into_owned())))));
 			Ok(())
 		}
 		RequestStatus::PendingSubscription => {
 			let (_, send_back, _) = manager.complete_pending_subscription(id).expect("State checked above; qed");
-			let _ = send_back.send(Err(Error::Call(err.error.to_call_error())));
+			let _ = send_back.send(Err(Error::Call(CallError::Custom(err.error.into_owned()))));
 			Ok(())
 		}
 		_ => Err(Error::InvalidRequestId),
--- a/core/src/client/async_client/helpers.rs
+++ b/core/src/client/async_client/helpers.rs
@@ -106,30 +106,28 @@ pub(crate) fn process_subscription_response(
 	}
 }
 
-/// Attempts to process a subscription response.
+/// Attempts to close a subscription when a [`SubscriptionError`] is received.
 ///
-/// Returns `Ok()` if the response was successfully sent to the frontend.
-/// Return `Err(None)` if the subscription was not found.
-/// Returns `Err(Some(msg))` if the channel to the `Subscription` was full.
+/// Returns `Ok(())` if the subscription was removed
+/// Return `Err(e)` if the subscription was not found.
 pub(crate) fn process_subscription_close_response(
 	manager: &mut RequestManager,
 	response: SubscriptionError<JsonValue>,
-) -> Result<(), Option<RequestMessage>> {
+) -> Result<(), Error> {
 	let sub_id = response.params.subscription.into_owned();
 	let request_id = match manager.get_request_id_by_subscription_id(&sub_id) {
 		Some(request_id) => request_id,
 		None => {
 			tracing::error!("The server tried to close down an invalid subscription: {:?}", sub_id);
-			return Err(None);
+			return Err(Error::InvalidSubscriptionId);
 		}
 	};
 
-	if manager.remove_subscription(request_id, sub_id.clone()).is_some() {
-		Ok(())
-	} else {
-		tracing::error!("The server tried to close down an invalid subscription: {:?}", sub_id);
-		Err(None)
-	}
+	debug_assert!(
+		manager.remove_subscription(request_id, sub_id).is_some(),
+		"Both request ID and sub ID in RequestManager; qed"
+	);
+	Ok(())
 }
 
 /// Attempts to process an incoming notification
--- a/core/src/client/async_client/helpers.rs
+++ b/core/src/client/async_client/helpers.rs
@@ -237,17 +237,18 @@ pub(crate) fn build_unsubscribe_message(
 /// Returns `Ok` if the response was successfully sent.
 /// Returns `Err(_)` if the response ID was not found.
 pub(crate) fn process_error_response(manager: &mut RequestManager, err: ErrorResponse) -> Result<(), Error> {
-	let id = err.id.clone().into_owned();
+	let id = err.id().clone().into_owned();
 
 	match manager.request_status(&id) {
 		RequestStatus::PendingMethodCall => {
 			let send_back = manager.complete_pending_call(id).expect("State checked above; qed");
-			let _ = send_back.map(|s| s.send(Err(Error::Call(CallError::Custom(err.error.into_owned())))));
+			let _ =
+				send_back.map(|s| s.send(Err(Error::Call(CallError::Custom(err.error_object().clone().into_owned())))));
 			Ok(())
 		}
 		RequestStatus::PendingSubscription => {
 			let (_, send_back, _) = manager.complete_pending_subscription(id).expect("State checked above; qed");
-			let _ = send_back.send(Err(Error::Call(CallError::Custom(err.error.into_owned()))));
+			let _ = send_back.send(Err(Error::Call(CallError::Custom(err.error_object().clone().into_owned()))));
 			Ok(())
 		}
 		_ => Err(Error::InvalidRequestId),
--- a/core/src/client/async_client/helpers.rs
+++ b/core/src/client/async_client/helpers.rs
@@ -122,10 +122,7 @@ pub(crate) fn process_subscription_close_response(
 		}
 	};
 
-	debug_assert!(
-		manager.remove_subscription(request_id, sub_id).is_some(),
-		"Both request ID and sub ID in RequestManager; qed"
-	);
+	manager.remove_subscription(request_id, sub_id).expect("Both request ID and sub ID in RequestManager; qed");
 	Ok(())
 }
 
--- a/core/src/client/async_client/helpers.rs
+++ b/core/src/client/async_client/helpers.rs
@@ -28,10 +28,10 @@ use std::time::Duration;
 
 use crate::client::async_client::manager::{RequestManager, RequestStatus};
 use crate::client::{RequestMessage, TransportSenderT};
-use crate::error::SubscriptionClosed;
 use crate::Error;
 
 use futures_channel::{mpsc, oneshot};
+use jsonrpsee_types::response::SubscriptionError;
 use jsonrpsee_types::{
 	ErrorResponse, Id, Notification, ParamsSer, RequestSer, Response, SubscriptionId, SubscriptionResponse,
 };
@@ -81,20 +81,15 @@ pub(crate) fn process_subscription_response(
 	let sub_id = response.params.subscription.into_owned();
 	let request_id = match manager.get_request_id_by_subscription_id(&sub_id) {
 		Some(request_id) => request_id,
-		None => return Err(None),
+		None => {
+			tracing::error!("Subscription ID: {:?} is not an active subscription", sub_id);
+			return Err(None);
+		}
 	};
 
 	match manager.as_subscription_mut(&request_id) {
-		Some(send_back_sink) => match send_back_sink.try_send(response.params.result.clone()) {
+		Some(send_back_sink) => match send_back_sink.try_send(response.params.result) {
 			// The server sent a subscription closed notification, then close down the subscription.
-			Ok(()) if serde_json::from_value::<SubscriptionClosed>(response.params.result).is_ok() => {
-				if manager.remove_subscription(request_id, sub_id.clone()).is_some() {
-					Ok(())
-				} else {
-					tracing::error!("The server tried to close down an invalid subscription: {:?}", sub_id);
-					Err(None)
-				}
-			}
 			Ok(()) => Ok(()),
 			Err(err) => {
 				tracing::error!("Dropping subscription {:?} error: {:?}", sub_id, err);
@@ -107,6 +102,32 @@ pub(crate) fn process_subscription_response(
 			tracing::error!("Subscription ID: {:?} is not an active subscription", sub_id);
 			Err(None)
 		}
+	}
+}
+
+/// Attempts to process a subscription response.
+///
+/// Returns `Ok()` if the response was successfully sent to the frontend.
+/// Return `Err(None)` if the subscription was not found.
+/// Returns `Err(Some(msg))` if the channel to the `Subscription` was full.
+pub(crate) fn process_subscription_close_response(
+	manager: &mut RequestManager,
+	response: SubscriptionError<JsonValue>,
+) -> Result<(), Option<RequestMessage>> {
+	let sub_id = response.params.subscription.into_owned();
+	let request_id = match manager.get_request_id_by_subscription_id(&sub_id) {
+		Some(request_id) => request_id,
+		None => {
+			tracing::error!("The server tried to close down an invalid subscription: {:?}", sub_id);
+			return Err(None);
+		}
+	};
+
+	if manager.remove_subscription(request_id, sub_id.clone()).is_some() {
+		Ok(())
+	} else {
+		tracing::error!("The server tried to close down an invalid subscription: {:?}", sub_id);
+		Err(None)
 	}
 }
 
--- a/core/src/client/async_client/helpers.rs
+++ b/core/src/client/async_client/helpers.rs
@@ -90,7 +90,6 @@ pub(crate) fn process_subscription_response(
 
 	match manager.as_subscription_mut(&request_id) {
 		Some(send_back_sink) => match send_back_sink.try_send(response.params.result) {
-			// The server sent a subscription closed notification, then close down the subscription.
 			Ok(()) => Ok(()),
 			Err(err) => {
 				tracing::error!("Dropping subscription {:?} error: {:?}", sub_id, err);
--- a/core/src/client/async_client/mod.rs
+++ b/core/src/client/async_client/mod.rs
@@ -4,8 +4,9 @@ mod manager;
 use std::time::Duration;
 
 use crate::client::{
-	BatchMessage, ClientT, RegisterNotificationMessage, RequestMessage, Subscription, SubscriptionClientT,
-	SubscriptionKind, SubscriptionMessage, TransportReceiverT, TransportSenderT,
+	async_client::helpers::process_subscription_close_response, BatchMessage, ClientT, RegisterNotificationMessage,
+	RequestMessage, Subscription, SubscriptionClientT, SubscriptionKind, SubscriptionMessage, TransportReceiverT,
+	TransportSenderT,
 };
 use helpers::{
 	build_unsubscribe_message, call_with_timeout, process_batch_response, process_error_response, process_notification,
@@ -20,7 +21,8 @@ use futures_util::future::Either;
 use futures_util::sink::SinkExt;
 use futures_util::stream::StreamExt;
 use jsonrpsee_types::{
-	ErrorResponse, Id, Notification, NotificationSer, ParamsSer, RequestSer, Response, SubscriptionResponse,
+	response::SubscriptionError, ErrorResponse, Id, Notification, NotificationSer, ParamsSer, RequestSer, Response,
+	SubscriptionResponse,
 };
 use serde::de::DeserializeOwned;
 use tokio::sync::Mutex;
@@ -488,6 +490,11 @@ async fn background_task<S: TransportSenderT, R: TransportReceiverT>(
 					if let Err(Some(unsub)) = process_subscription_response(&mut manager, response) {
 						let _ = stop_subscription(&mut sender, &mut manager, unsub).await;
 					}
+				}
+				// Subscription error response.
+				else if let Ok(response) = serde_json::from_str::<SubscriptionError<_>>(&raw) {
+					tracing::debug!("[backend]: recv subscription err {:?}", response);
+					let _ = process_subscription_close_response(&mut manager, response);
 				}
 				// Incoming Notification
 				else if let Ok(notif) = serde_json::from_str::<Notification<_>>(&raw) {
--- a/core/src/client/async_client/mod.rs
+++ b/core/src/client/async_client/mod.rs
@@ -493,7 +493,7 @@ async fn background_task<S: TransportSenderT, R: TransportReceiverT>(
 				}
 				// Subscription error response.
 				else if let Ok(response) = serde_json::from_str::<SubscriptionError<_>>(&raw) {
-					tracing::debug!("[backend]: recv subscription err {:?}", response);
+					tracing::debug!("[backend]: recv subscription closed {:?}", response);
 					let _ = process_subscription_close_response(&mut manager, response);
 				}
 				// Incoming Notification
--- a/core/src/error.rs
+++ b/core/src/error.rs
@@ -156,6 +156,17 @@ pub enum SubscriptionClosed {
 	Server(String),
 }
 
+impl std::fmt::Display for SubscriptionClosed {
+	/// Get close reason as str.
+	fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> fmt::Result {
+		match self {
+			Self::Unsubscribed => write!(f, "Subscription was closed by a unsubscribe call"),
+			Self::ConnectionReset => write!(f, "Subscription was closed by connection reset"),
+			Self::Server(msg) => write!(f, "Subscription was closed by server: {}", msg.as_str()),
+		}
+	}
+}
+
 /// Generic transport error.
 #[derive(Debug, thiserror::Error)]
 pub enum GenericTransportError<T: std::error::Error + Send + Sync> {
--- a/core/src/error.rs
+++ b/core/src/error.rs
@@ -27,7 +27,7 @@
 use std::fmt;
 
 use jsonrpsee_types::error::{
-	CallError, ErrorObject, CALL_EXECUTION_FAILED_CODE, INVALID_PARAMS_CODE, SUBSCRIPTION_CLOSED,
+	CallError, ErrorObject, ErrorObjectOwned, CALL_EXECUTION_FAILED_CODE, INVALID_PARAMS_CODE, SUBSCRIPTION_CLOSED,
 	SUBSCRIPTION_CLOSED_WITH_ERROR,
 };
 use serde::{Deserialize, Serialize};
@@ -149,8 +149,8 @@ impl Error {
 	/// Get the JSON-RPC error object representation of the error.
 	pub fn as_error_object<'a, 'b: 'a>(&'b self) -> ErrorObject<'a> {
 		match self {
-			Error::Call(CallError::Custom { message, code, data }) => {
-				ErrorObject { code: (*code).into(), message: message.as_str().into(), data: data.as_deref() }
+			Error::Call(CallError::Custom(ErrorObjectOwned { code, message, data })) => {
+				ErrorObject { code: *code, message: message.as_str().into(), data: data.as_deref() }
 			}
 			Error::Call(CallError::InvalidParams(e)) => {
 				ErrorObject::code_and_message(INVALID_PARAMS_CODE, e.to_string().into())
--- a/core/src/error.rs
+++ b/core/src/error.rs
@@ -26,7 +26,10 @@
 
 use std::fmt;
 
-use jsonrpsee_types::error::CallError;
+use jsonrpsee_types::error::{
+	CallError, ErrorObject, CALL_EXECUTION_FAILED_CODE, INVALID_PARAMS_CODE, SUBSCRIPTION_CLOSED,
+	SUBSCRIPTION_CLOSED_WITH_ERROR,
+};
 use serde::{Deserialize, Serialize};
 
 /// Convenience type for displaying errors.
@@ -142,6 +145,33 @@ impl Error {
 	{
 		Error::Call(CallError::from_std_error(err))
 	}
+
+	/// Get the JSON-RPC error object representation of the error.
+	pub fn as_error_object<'a, 'b: 'a>(&'b self) -> ErrorObject<'a> {
+		match self {
+			Error::Call(CallError::Custom { message, code, data }) => {
+				ErrorObject { code: (*code).into(), message: message.as_str().into(), data: data.as_deref() }
+			}
+			Error::Call(CallError::InvalidParams(e)) => {
+				ErrorObject::code_and_message(INVALID_PARAMS_CODE, e.to_string().into())
+			}
+			Error::SubscriptionClosed(SubscriptionClosed::Server(CloseReason::Failed(e))) => {
+				ErrorObject::code_and_message(SUBSCRIPTION_CLOSED_WITH_ERROR, e.as_str().into())
+			}
+			Error::SubscriptionClosed(SubscriptionClosed::Server(CloseReason::Unknown)) => {
+				ErrorObject::code_and_message(
+					SUBSCRIPTION_CLOSED_WITH_ERROR,
+					"Subscription closed with unknown reason".into(),
+				)
+			}
+			// Not really errors see [`SubscriptionClosed`] for further info.
+			// Such as if the subscription was closed by the client and similar.
+			Error::SubscriptionClosed(other) => {
+				ErrorObject::code_and_message(SUBSCRIPTION_CLOSED, other.to_string().into())
+			}
+			_ => ErrorObject::code_and_message(CALL_EXECUTION_FAILED_CODE, self.to_string().into()),
+		}
+	}
 }
 
 /// A type to represent when a subscription gets closed
@@ -153,7 +183,24 @@ pub enum SubscriptionClosed {
 	/// The client closed the connection.
 	ConnectionReset,
 	/// The server closed the subscription, providing a description of the reason as a `String`.
-	Server(String),
+	Server(CloseReason),
+}
+
+impl From<CloseReason> for SubscriptionClosed {
+	fn from(reason: CloseReason) -> Self {
+		Self::Server(reason)
+	}
+}
+
+/// Represent why a subscription was closed by the server.
+#[derive(Deserialize, Serialize, Clone, Debug, PartialEq)]
+pub enum CloseReason {
+	/// The subscription was completed successfully.
+	Success,
+	/// The subscription was closed with unknown reason.
+	Unknown,
+	/// The subscription failed because of some error.
+	Failed(String),
 }
 
 impl std::fmt::Display for SubscriptionClosed {
@@ -162,7 +209,7 @@ impl std::fmt::Display for SubscriptionClosed {
 		match self {
 			Self::Unsubscribed => write!(f, "Subscription was closed by a unsubscribe call"),
 			Self::ConnectionReset => write!(f, "Subscription was closed by connection reset"),
-			Self::Server(msg) => write!(f, "Subscription was closed by server: {}", msg.as_str()),
+			Self::Server(reason) => write!(f, "Subscription was closed by server: {:?}", reason),
 		}
 	}
 }
--- a/core/src/error.rs
+++ b/core/src/error.rs
@@ -28,6 +28,7 @@ use std::fmt;
 
 use jsonrpsee_types::error::{
 	CallError, ErrorObject, ErrorObjectOwned, CALL_EXECUTION_FAILED_CODE, INVALID_PARAMS_CODE, SUBSCRIPTION_CLOSED,
+	UNKNOWN_ERROR_CODE,
 };
 
 /// Convenience type for displaying errors.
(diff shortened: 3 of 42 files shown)


core/src/server/rpc_module.rs
```diff
@@ -655,38 +657,27 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
...
-					let sub_id = {
-						let sub_id: RpcSubscriptionId = conn.id_provider.next_id().into_owned();
-						let uniq_sub = SubscriptionKey { conn_id: conn.conn_id, sub_id: sub_id.clone() };
-
-						subscribers.lock().insert(uniq_sub, (method_sink.clone(), conn_rx));
-
-						sub_id
-					};
-
-					method_sink.send_response(id.clone(), &sub_id);
```
niklasad1, line 669: unfortunately this will a subscription ID whether to call was actually successful or not

and we can't do this after the `callback` because the subscription sink might already have started sending out stuff which is no go.

so indeed we need the users manually to call `accept` or `reject` the subscription whether call was ok after parsing the params. we can't do it because it depends on their callback.

so I introduced `PendingSubscription` that you need to call `reject` or `accept` on which should work.

jsdw: I'm trying to understand; under which circumstances would the callback _not_ run `PendingSubscription::accept()`? Is this so that a provided subscription callback can do arbitrary processing in order to then decide that theere is some issue and we don't want to "begin" the subscription but instead want to return an error (eg parameters aren't what we expected)?

jsdw: (that being the case; I had a skim over and I like the PendingSubscription stuff; it's a nice way to ensure that the thing is either rejected or accepted before any messages are sent :))

niklasad1: > I'm trying to understand; under which circumstances would the callback not run PendingSubscription::accept()

It's a way to ensure to `params.parse()` would fail the in `proc macro code` which does that under the hood and additionally do let the user decide when to accept the subscription as you said.

core/src/server/rpc_module.rs
```diff
@@ -806,27 +885,41 @@ impl SubscriptionSink {
...
+								self.close(&ErrorObject::code_and_message(
+									SUBSCRIPTION_CLOSED_WITH_ERROR,
+									&close.to_string(),
+								));
 								break Ok(());
 							}
 							Err(err) => {
-								break Err(err);
+								let reason = err.to_string();
+								self.close(&ErrorObject::code_and_message(SUBSCRIPTION_CLOSED_WITH_ERROR, &reason));
```
niklasad1, line 897: There is no way to opt from these close messages but I argued that one could use `manual send` if that breaks things.

However, we could make this `&mut self` and let the user close it manually if it wants it because this might break compat with some libraries works with polkadot JS at least.

core/src/server/rpc_module.rs
```diff
@@ -877,63 +971,64 @@ impl SubscriptionSink {
...
-			}
+			let msg = self.build_error_message(&close).expect("valid json infallible; qed");
+			sink.send_raw(msg).is_ok()
+		} else {
+			false
 		}
 	}
 }
 
-impl Drop for SubscriptionSink {
```
niklasad1, line 930: removed this so people can opt out from the close messages.

core/src/server/rpc_module.rs
```diff
@@ -877,63 +971,64 @@ impl SubscriptionSink {
...
+	///  "method": "<method>",
+	///  "params": {
+	///    "subscription": "<subscriptionID>",
+	///    "error": { "code": <your code>, "message": <your message>, "data": <your data but might be omitted> }
+	///    }
+	///  }
+	/// }
+	/// ```
+	///
+	pub fn close(&mut self, close: &ErrorObject) -> bool {
```
niklasad1, line 1019: ideally we should take `impl: Serialize` here because `ErrorObject` is a bit annoying to use but I think it might be a footgun with `msg: Serialize`.

core/src/server/rpc_module.rs
```diff
@@ -877,63 +966,64 @@ impl SubscriptionSink {
...
+	///  "method": "<method>",
+	///  "params": {
+	///    "subscription": "<subscriptionID>",
+	///    "error": { "code": <your code>, "message": <your message>, "data": <your data but might be omitted> }
+	///    }
+	///  }
+	/// }
+	/// ```
+	///
+	pub fn close(&mut self, err: Error) -> bool {
```
jsdw, line 1014: Out of interest, is there a reason that this needs to take `&mut self` and not `mut self` (so that once it's been closed, people can't try using the sink again)?

You've added code so that an `Error` can be converted into an `ErrorObject`, but I don't think one can convert from an `ErrorObject` to an `Error` as easily. When somebody has an error, do we want people to create a Jsonrpsee `Error` (eg a `CallError`) normally, or would we prefer them to reach for `ErrorObject`?

If we want them to use the jsonrpsee `Error` type (eg `CallError`) then maybe we shouldn't really expose functions that take `ErrorObject`, and just let them construct `Error`s instead, which we internally convert (if people want full control, they can use a `CallError::Custom` by the looks of it)?

core/src/server/rpc_module.rs
niklasad1: > So, if I understand it right, there are basically 2 places an error can happen with a subscription;

Exactly.

> Before it is accepted (ie the params aren't valid, or some other thing that means it's not valid to ask for said subscription (eg server not configured to allow it or something)). Here, we never want to give back a subscription ID etc and want to bail with something like an { jsonrpc: "2.0", error: { code: 123, message: "Invalid params" }} style message, I guess?

Yes, before a subscription call with invalid params was replied to twice because of some of bug in the logic. Because the actual callback (user defined) we can't know whether the params were valid or not unless we could pause the mpsc channel but that seems like a bad idea. So instead we have this `accept/reject API` that user has to use. Additionally I have a drop impl for this `PendingSubscription` so the actual call will be rejected if this is dropped. 

> During the subscription. Some error could occur mid-subscription that, I assume, means we want to end the subscription and report the error back (would we always want to end the subscription on an error?). I guess in this case we want to return a message like { jsonrpsee: "2.0", method: "subscribe_foo", params: { subscription: "123abc", error: { code: 124, message: "Something went wrong" } } }.

> I guess this change allows for (1). Is there a way for people to do (2)?

Yes, there is the `close API` where folks can close down the subscription using the `jsonrpsee::Error`.
Before, jsonrpsee closed down subscription and sent out this under the hood by I have realized that this may not be needed for some use cases such as if subscription method itself has defined some state machine or something for this themselves then we force this overhead on them. Also not break comparability with some libraries "may" not support this.

> We also want a way to signal that the subscription has closed without an error. I guess the options for this are:

> Return an error (like in (2) above) with a code/message that signals the thing has closed.
    Invent more syntax like params: { subscription: "123abc", closed: true } to signal this.
    Leave it up to users to interpret a certain non-error response as meaning that no further messages will be yielded.

Currently, we treat "closed" as an error on send out the same format with some other status code that we pre-defined:

```
Notification { jsonrpc: TwoPointZero, method: "n", params: SubscriptionPayloadError { subscription: Num(60482361782361), error: Object({"code": Number(-32003), "message": String("Subscription was closed by server: Success")}) }
```

So that's is a bit annoying but I think that's reasonable workaround not to break comparability with polkadot JS for now, however this now possible to opt in/out from one needs to call `close` explictly now:  

`let _ = sink.pipe_from_try_stream(stream).await.map_err(|e| sink.close(e));`   

> How did JSONRPC handle this sort of thing? Are we leaning towards doing (1)?

JSON-RPC doesn't support `close notifications` AFAIU but it (1) similar to what is implemented in this PR.
So I think it relies on the impl to have definition when to close the subscriptions using ordinary notifications.

jsdw: Thank you, that's really useful info!

To summarize, with this PR:
- We can reject a pending subscription with a `{ jsonrpc: "2.0", error: {...} }` style error of our choice using `pending.reject(error)`.
- We can end a subscription at any time with `{ jsonrpsee: "2.0", method: "subscribe_foo", params: { subscription: "123abc", error: { ... } } }` style error using `sink.close(error)`.
- Subscriptions that are dropped are now silently closed (for compatibility reasons, incase subscriptions want to signal that they are closed via the "success" messages and not via a "subscription closed" error). In other words, we enforce no opinion on what happens when a subscription is closed, and leave it up to the subscription itself to decide whether to emit an "error" or communicate closure via some other message.

I think if I've understood all of that right, it sounds good to me! The more we stay compatible with how jsonrpc works, the easier!

core/src/client/async_client/helpers.rs
```diff
@@ -81,20 +81,15 @@ pub(crate) fn process_subscription_response(
...
+		None => {
+			tracing::error!("Subscription ID: {:?} is not an active subscription", sub_id);
+			return Err(None);
+		}
 	};
 
 	match manager.as_subscription_mut(&request_id) {
-		Some(send_back_sink) => match send_back_sink.try_send(response.params.result.clone()) {
+		Some(send_back_sink) => match send_back_sink.try_send(response.params.result) {
 			// The server sent a subscription closed notification, then close down the subscription.
```
jsdw, line 92: I can't quite grasp the significance of removing this code (which was added fairly recently iirc to properly close a subscription)? Is this removed because we have removed any assumptions about how a subscription will be closed, and so it's up to the client to decide?

jsdw: (also I guess the comment can be removed too now?)

core/src/client/async_client/helpers.rs
```diff
@@ -110,6 +105,32 @@ pub(crate) fn process_subscription_response(
 	}
 }
 
+/// Attempts to process a subscription response.
+///
+/// Returns `Ok()` if the response was successfully sent to the frontend.
+/// Return `Err(None)` if the subscription was not found.
+/// Returns `Err(Some(msg))` if the channel to the `Subscription` was full.
```
jsdw, line 112: This is identical to the above doc comment; should it be?

core/src/server/rpc_module.rs
niklasad1: > Out of interest, is there a reason that this needs to take &mut self and not mut self (so that once it's been closed, people can't try using the sink again)?

good catch, it makes sense to take ownership there.

> You've added code so that an Error can be converted into an ErrorObject, but I don't think one can convert from an ErrorObject to an Error as easily. When somebody has an error, do we want people to create a Jsonrpsee Error (eg a CallError) normally, or would we prefer them to reach for ErrorObject?

ErrorObject(borrowed) and CallError::Custom(owned) is the same thing, but `CallError::Custom` take ownership of `data.
We could could extract a take `ErrorObjectOwned` and add it in `CallError` maybe this more understandable?

`CallError` is used by the servers and `ErrorObject` by clients.

I guess if we do want I said :point_up: then it should make sense right?

core/src/server/rpc_module.rs
```diff
@@ -782,15 +868,19 @@ impl SubscriptionSink {
...
-	///     tokio::spawn(sink.pipe_from_try_stream(stream));
+	///     tokio::spawn(async move {
+	///         // jsonrpsee doesn't send an error notification if the stream was closed unless `close` is called.
+	///         let _ = sink.pipe_from_try_stream(stream).await.map_err(|e| sink.close(e));
+	///     });
 	///     Ok(())
 	/// });
 	/// ```
-	pub async fn pipe_from_try_stream<S, T, E>(mut self, mut stream: S) -> Result<(), Error>
+	pub async fn pipe_from_try_stream<S, T, E>(&mut self, mut stream: S) -> Result<(), Error>
```
niklasad1, line 883: we should update the docs but I'm not really sure about the API here i.e, return `SubscriptionClosed` as an error.

I was thinking of doing something like `Result<bool, Error>` or `Result<Option<()>, Error` but I found that to be even harder to understand

core/src/server/rpc_module.rs
```diff
@@ -738,6 +740,90 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
...
+pub struct PendingSubscription(Option<InnerPendingSubscription>);
+
+impl PendingSubscription {
+	/// Reject the subscription call.
+	pub fn reject(self, err: Error) -> Result<(), Error> {
+		self.reject_from_error_object(err.as_error_object())
+	}
+
+	/// Reject the subscription call from [`ErrorObject`].
+	pub fn reject_from_error_object(mut self, err: ErrorObject) -> Result<(), Error> {
```
jsdw, line 776: since `subscription.close()` doesn't expose a way to work with `ErrorObject`, I wonder whether we should be consistent and not expose it here either?

core/src/server/rpc_module.rs
```diff
@@ -738,6 +740,90 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 	}
 }
 
+/// Represent a pending subscription which waits being accepted or rejected.
+///
+/// Warning: you need to call either `PendingSubscription::accept` or `PendingSubscription::reject` otherwise
+/// the subscription will not make any progress.
```
jsdw, line 746:
```suggestion
/// the subscription will be dropped with an `InvalidParams` error.
```

Or something like that?

core/src/server/rpc_module.rs
jsdw: So, I think the stance we're taking is to let the subscription decide how to communicate that it's closed (ie to use an error code/message or some "ok" value)? And if a subscription has already been closed, then we return a `SubscriptionClosed` error if somebody tries to use it again (wonder whether we could be consuming `self` to prevent that?).

Is that true?

If so, I guess the subscription going away is always "Ok(())" (it might be that the "ok" responses have communicated this OK already?), and then any errors that are emmitted are just forwarded as `Error` in the response, so the type signature feels ok to me (if I'm not way off the mark with my understanding :))

core/src/server/rpc_module.rs
```diff
@@ -806,27 +896,25 @@ impl SubscriptionSink {
...
 					Either::Left((Err(e), _)) => {
-						let close_reason = SubscriptionClosedReason::Server(e.to_string()).into();
-						self.close(&close_reason);
+						let close_reason: SubscriptionClosed = CloseReason::Failed(e.to_string()).into();
+						break Err(Error::SubscriptionClosed(close_reason));
+					}
+					// Stream completed (this is not really an error)
+					Either::Left((Ok(None), _)) => {
+						let close_reason: SubscriptionClosed = CloseReason::Success.into();
 						break Err(Error::SubscriptionClosed(close_reason));
```
jsdw, line 914: If my understanding in the above comment is correct though, should we just return `Ok(())` here, since we've made no assumptions about how the subscription communicates that it's been closed (ie it may have already sent something via the "ok" channel)?

core/src/server/rpc_module.rs
niklasad1: yeah, I think that should correct after the latest changes I pushed for a short while ago.

The only thing is that the subscription API is bit verbose now to work with but it should work.
It might even make sense to do the jsonrpc approach and implement `SinkExt`  on the SubscriptionSink to not have to deal with all this.

core/src/server/rpc_module.rs
niklasad1: it is because the `Error type` is not clone I used it. It was needed by the proc macro code

core/src/server/rpc_module.rs
niklasad1: https://github.com/paritytech/jsonrpsee/pull/728/commits/17ef15b8296f789c6dd7cba6e5f3a9afaf0ff2b2

core/src/server/rpc_module.rs
niklasad1: I changed this look at the latest changes, should be according to the PR description now.

Changed from it  to Result<SubscriptionResult, Error>` but we just have have `SubscriptionClosed` there and remove the Result entirely for simplicity.

types/src/error.rs
```diff
@@ -104,6 +105,36 @@ impl<'a> PartialEq for ErrorObject<'a> {
...
+#[serde(deny_unknown_fields)]
+pub struct ErrorObjectOwned {
+	/// Code
+	pub code: ErrorCode,
+	/// Message
+	pub message: String,
+	/// Optional data
+	#[serde(skip_serializing_if = "Option::is_none")]
+	pub data: Option<Box<RawValue>>,
+}
```
jsdw, line 119: This may be a bad idea, but I wonder whether we can unify `ErrorObject` and `ErrorObjectOwned` with some `Cow`s, and pthen erhaps provide `borrowed(code: i32, message: &str, data: Option<&'a RawValue>) -> ErrorObject<'a>` and `owned(code: i32, message: String, data: Option<Box<RawValue>>) -> ErrorObject<'static>` constructors for it?

jsdw: (I'll have a quick play and see whether it's actually possible)

jsdw: Ok, so how about this (ignore the `ErrorObject2` naming; just playing around in the existing codebase so I needed a name that didn't overlap!):

```rust
use std::borrow::{ Cow as StdCow };
use std::borrow::Borrow;

#[derive(Debug, Deserialize, Serialize, Clone)]
#[serde(deny_unknown_fields)]
pub struct ErrorObject2<'a> {
	/// Code
	code: ErrorCode,
	/// Message
	message: StdCow<'a, str>,
	/// Optional data
	#[serde(skip_serializing_if = "Option::is_none")]
	data: Option<StdCow<'a, RawValue>>,
}

pub type ErrorObject2Owned = ErrorObject2<'static>;

impl ErrorObject2Owned {
	/// Return the error code
	pub fn code(&self) -> i32 {
		self.code.code()
	}

	/// Return the message
	pub fn message(&self) -> &str {
		self.message.borrow()
	}

	/// Return the data associated with this error, if any
	pub fn data(&self) -> Option<&RawValue> {
		self.data.as_ref().map(|d| d.borrow())
	}

	/// Create a new `ErrorObjectOwned` with optional data.
	pub fn owned(code: i32, message: impl Into<String>, data: Option<impl Serialize>) -> Self {
		let data = data.and_then(|d| serde_json::value::to_raw_value(&d).ok());
		ErrorObject2 {
			code: code.into(),
			message: message.into().into(),
			data: data.map(|d| StdCow::Owned(d))
		}
	}

	/// Create a new `ErrorObject` with optional data.
	pub fn borrowed<'a>(code: i32, message: &'a str, data: Option<&'a RawValue>) -> ErrorObject2<'a> {
		ErrorObject2 {
			code: code.into(),
			message: StdCow::Borrowed(message),
			data: data.map(|d| StdCow::Borrowed(d))
		}
	}

	/// Get the borrowed variant [`ErrorObject`].
	pub fn into_owned(self) -> ErrorObject2<'static> {
		ErrorObject2 {
			code: self.code,
			message: StdCow::Owned(self.message.into_owned()),
			data: self.data.map(|d| StdCow::Owned(d.into_owned()))
		}
	}

	/// Get the borrowed variant [`ErrorObject`].
	pub fn borrow<'a>(&'a self) -> ErrorObject2<'a> {
		ErrorObject2 {
			code: self.code,
			message: StdCow::Borrowed(self.message.borrow()),
			data: self.data.as_ref().map(|d| StdCow::Borrowed(d.borrow()))
		}
	}
}
```

jsdw: (I made both owned and borrowed functions accept `Option<data>` for consistency; with the separate borrowed and owned variants, one accepts Option and one does not)

niklasad1: looks good to me the reason why it's like is currently is because `beef::Cow` doesn't work with RawValue.

jsdw: That makes sense; `beef::Cow` will only ever support certain types I guess, so we should feel free to use the std `Cow` where useful! (perhaps I could have used the beef variant for strings above; I can't recall!)

core/src/server/rpc_module.rs
```diff
@@ -738,6 +748,78 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
...
+			}
+		} else {
+			Err(Error::Custom("Subscription is already claimed".to_string()))
+		}
+	}
+
+	/// Attempt to accept the subscription and respond the subscription method call.
+	///
+	/// Fails if the connection was closed
+	pub fn accept(mut self) -> Result<SubscriptionSink, Error> {
```
jsdw, line 796: Just some bikeshedding:

Is it worth returning an `Error` here? Since subscription closures no longer return a result (which makes sense), maybe returning `Option<SubscriptionSink>` would be a little easier on people (they don't need to worry about handling any error); what do you think?

likewise with `reject`, maybe returning a `bool` is easier (again, no real need to worry about handling a specific error that way, and probably people can just ignore the value of the bool)?

core/src/client/async_client/helpers.rs
jsdw: So I had a gander at this code and I think it all makes sense to me; the backend sends a message, and we handle it in `process_subscription_response`, and here we are sending that message on to a channel held by somebody, anmd if the send is Ok then we just keep going, and if the send failed then the frontend has dropped the channel and we should end the subscription!

The only thing left here is to remove that comment then; we are no longer doing what it says :)

core/src/client/async_client/mod.rs
```diff
@@ -489,6 +491,11 @@ async fn background_task<S: TransportSenderT, R: TransportReceiverT>(
 						let _ = stop_subscription(&mut sender, &mut manager, unsub).await;
 					}
 				}
+				// Subscription error response.
+				else if let Ok(response) = serde_json::from_str::<SubscriptionError<_>>(&raw) {
+					tracing::debug!("[backend]: recv subscription closed {:?}", response);
+					let _ = process_subscription_close_response(&mut manager, response);
```
jsdw, line 497: So here, the backend has sent back something like this I think?:

```
{ jsonrpc: "2.0", params: { subscription: "abc123", error: { code, message, data } } }
```

As a result, we close the subscription on the client. Right now, we don't pass on that error to the client though? (this may have been what you were getting at in a previous comment and I'm just catching up now :))

core/src/server/rpc_module.rs
```diff
@@ -82,13 +83,22 @@ pub struct ConnState<'a> {
 	pub id_provider: &'a dyn IdProvider,
 }
 
+/// Outcome of a successful terminated subscription.
+#[derive(Debug)]
+pub enum SubscriptionResult {
+	/// The subscription stream was executed successful.
```
jsdw, line 89:
```suggestion
	/// The subscription stream was executed successfully.
```

core/src/server/rpc_module.rs
```diff
@@ -82,13 +83,22 @@ pub struct ConnState<'a> {
...
+	Success,
+	/// The subscription was aborted the remote peer.
```
jsdw, line 91:
```suggestion
	/// The subscription was aborted by the remote peer.
```

core/src/server/rpc_module.rs
```diff
@@ -738,6 +748,78 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
...
+	subscribers: Subscribers,
+	/// Request ID.
+	id: Id<'static>,
+}
+
+/// Represent a pending subscription which waits for being accepted or rejected.
+///
+/// This type implement drop for ease of use, such as dropped in error short circuiting via `map_err()?`.
+#[derive(Debug)]
+pub struct PendingSubscription(Option<InnerPendingSubscription>);
```
jsdw, line 775: Perhaps we don't need an `Option<InnerPendingSubscription>` now, since we can only reject or accept it once, and each function takes ownership?

core/src/server/rpc_module.rs
```diff
@@ -738,6 +748,78 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
...
+	uniq_sub: SubscriptionKey,
+	/// Shared Mutex of subscriptions
+	subscribers: Subscribers,
+	/// Request ID.
+	id: Id<'static>,
+}
+
+/// Represent a pending subscription which waits for being accepted or rejected.
+///
+/// This type implement drop for ease of use, such as dropped in error short circuiting via `map_err()?`.
```
jsdw, line 773: Maybe worth noting what happens on drop too; something like:

> If this type is dropped, it will return an `InvalidParams` error to the subscriber

core/src/server/rpc_module.rs
```diff
@@ -738,6 +748,78 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
...
+			Ok(SubscriptionSink { inner: sink, close_notify, method, uniq_sub, subscribers, active_sub })
+		} else {
+			Err(Error::Custom("Connection is closed".into()))
+		}
+	}
+}
+
+impl Drop for PendingSubscription {
+	fn drop(&mut self) {
+		if let Some(inner) = self.0.take() {
```
jsdw, line 816: Ah, maybe we still need the `Option<InnerPendingSubscriptiuon>` so that the Drop impl knows whether to send anything or not; I see now!

core/src/server/rpc_module.rs
```diff
@@ -751,23 +833,27 @@ pub struct SubscriptionSink {
...
-	is_connected: Option<oneshot::Sender<()>>,
+	active_sub: Arc<()>,
 }
 
 impl SubscriptionSink {
 	/// Send a message back to subscribers.
-	pub fn send<T: Serialize>(&mut self, result: &T) -> Result<(), Error> {
+	///
+	/// Returns `Ok(true)` if the message could be send
+	/// Returns `Ok(false)` if the sink was closed (this is not an error because the subscription got closed or connection was terminated)
```
jsdw, line 843:
```suggestion
	/// Returns `Ok(false)` if the sink was closed (either because the subscription was closed or the connection was terminated)
```

core/src/server/rpc_module.rs
niklasad1: yupp, I agree.

before the we required the subscriptions to return `Result` then it was quite ergonomic to use but that is not true anymore

core/src/client/async_client/mod.rs
niklasad1: yes, because I was lazy because o the lifetime issues with serde.

core/src/server/rpc_module.rs
niklasad1: it won't work with the `drop impl` any better suggestions?

core/src/server/rpc_module.rs
```diff
@@ -778,19 +864,38 @@ impl SubscriptionSink {
...
 	///     // This will return send `[Ok(1_u32), Ok(2_u32), Err(Error::SubscriptionClosed))]` to the subscriber
 	///     // because after the `Err(_)` the stream is terminated.
-	///     tokio::spawn(sink.pipe_from_try_stream(stream));
-	///     Ok(())
+	///     tokio::spawn(async move {
+	///         // jsonrpsee doesn't send an error notification if the stream was closed unless `close` is called.
+	///         // jsonrpsee doesn't send a notification if the stream was terminated or canceled.
+	///         //
+	///         // But it's possible to get reason why subscription was terminated
+	///         //
```
jsdw, line 883:
```suggestion
	///         // jsonrpsee doesn't send an error notification unless `close` is explicitly called.
	///         // If we pipe messages to the sink, we can inspect why it ended:
```

core/src/server/rpc_module.rs
```diff
@@ -805,35 +910,32 @@ impl SubscriptionSink {
...
-						self.close(&SubscriptionClosed::new(SubscriptionClosedReason::ConnectionReset));
-						break Ok(());
+						break SubscriptionClosed::RemotePeerAborted;
 					}
 				}
 			}
 		} else {
-			// The sink is closed.
-			Ok(())
+			SubscriptionClosed::RemotePeerAborted
```
jsdw, line 938: Nit: We could remove this if/else and save a level of indentation with:

```rust
let close_notify = match self.close_notify.clone() {
    Some(close_notify) => close_notify,
    None => return SubscriptionClosed::RemotePeerAborted
};
```

core/src/server/rpc_module.rs
```diff
@@ -869,68 +971,72 @@ impl SubscriptionSink {
...
-
-			let closed_by_server =
-				maybe_close.map_or(false, |c| matches!(c.close_reason(), SubscriptionClosedReason::Server(_)));
+	/// Send an error notification on the subscription with a special field `error`.
+	///
+	/// This may be used for indicating to connected clients that a given subscription has been canceled
+	/// as well so not necessarily an error but this is used for compat reason with existing JavaScript libraries.
+	///
+	/// The APIs [`SubscriptionSink::pipe_from_stream`] and [`SubscriptionSink::pipe_from_stream`] uses this
+	/// when the stream is terminated.
```
jsdw, line 1008: Is this still true? I thought that they didn't send any explicit errors any more?

core/src/server/rpc_module.rs
```diff
@@ -869,68 +971,72 @@ impl SubscriptionSink {
...
+	///
```
jsdw, line 1009: Might also be worth adding to this comment:

> If you'd like to to close the subscription without sending an error, just drop it.

core/src/server/rpc_module.rs
```diff
@@ -869,68 +971,72 @@ impl SubscriptionSink {
...
-		self.is_connected.take();
-		if let Some((sink, _)) = self.subscribers.lock().remove(&self.uniq_sub) {
-			tracing::debug!("Closing subscription: {:?} reason: {:?}", self.uniq_sub.sub_id, maybe_close);
-
-			let closed_by_server =
-				maybe_close.map_or(false, |c| matches!(c.close_reason(), SubscriptionClosedReason::Server(_)));
+	/// Send an error notification on the subscription with a special field `error`.
+	///
+	/// This may be used for indicating to connected clients that a given subscription has been canceled
+	/// as well so not necessarily an error but this is used for compat reason with existing JavaScript libraries.
```
jsdw, line 1005:
```suggestion
	/// Close the subscription, sending a notification with a special `error` field containing the provided error.
	///
	/// This can be used to signal an actual error, or just to signal that the subscription has been closed,
	/// depending on your preference. 
```

proc-macros/src/lib.rs
```diff
@@ -55,7 +55,7 @@ pub(crate) mod visitor;
 /// - The trait will have one additional (already implemented) method, `into_rpc`, which turns any object that
 ///   implements the server trait into an `RpcModule`.
 /// - For subscription methods, there will be one additional argument inserted right after `&self`: `subscription_sink:
-///   SubscriptionSink`. It should be used to actually maintain the subscription.
+///   PendingSubscription`. It should be used to actually maintain the subscription.
```
jsdw, line 58:
```suggestion
///   PendingSubscription`. It should be used to accept and publish messages to the subscription.
```

core/src/server/rpc_module.rs
niklasad1: no, good catch.

core/src/client/async_client/mod.rs
jsdw: OK that makes sense; if it's a bit of a pain to do, let's add an issue to address this separately if it's not important to Substrate!

core/src/server/rpc_module.rs
jsdw: Nope, that makes sense! I realised wh ythe Option was here when I got to looking at that drop impl :)

core/src/client/async_client/mod.rs
niklasad1: no this is just the client code so shouldn't matter

core/src/server/rpc_module.rs
niklasad1: done

jsdw (approved):
Looking good to me now!

I like that users have a choice on how to signal subscription closing and when/how to report errors. It does make some code a bit more verbose, but if that gets annoying, we can improve on it in the future.

jsdw:
Just a random thought that came to mind (as a future ergonomic improvement and not something that needs to be worked on now):

Currently, subscriptions accept a `pending` thing to `accept() or `reject()`, and then once accepted they get a sink to send messages or an error back on. Closures don't return a `Result` any more because it isn't obvious what that would mean (does it send an error, and if so, what sort of error?)

One alternate API might be to have the old API of giving subscription closures a `sink` immediately (rather than a `pending`), and allowing them to return `Result<(), ErrorObject>`s, but to keep track of whether the `sink` has been used or not internally, so that:
- If the closure returns with an error but the sink hasn't been used yet, close the subscription and return that ErrorObject to the client  as a normal jsonrpc error (eg `{ jsonrpc: 2.0, error: { message, code, data } }`)
- If the closure returns with an error and the sink _has_ been used, then close the subscription and return the ErrorObject as a `{ jsonrpc: 2.0, params: { subscription: "abc", error: { message, code, data } }}`.

With that API, it might be a bit more ergonomic handling errors (no need to match on each one and send it through the `sink.close` method), and no explicit `pending.accept()` needed; just bail with an error before sending a message if params are invalid or whatever. The downside might also be that it is less explicit about how exactly an error will be communicated back, and that it is easier to "accidentally" send an error back to the client because `?` was used somewhere. Just thinking out loud really :)

niklasad1:
So, I think we still need a way for the user to tell when to generate the subscriptionID and reply to the `subscribe_call` either way (I guess we could technically generate for every call and use the error notification to indicate errors but seems like wrong thing to do)

so we could actually kill the `PendingSubscription type` and have method `accept/assign_id/reply to subscribe call` that internally has `Option<SubscriptionId>.

If :point_up: is None => return ErrorObject with ID else => return ErrorObject notification

jsdw:
Ah yup, it does make sense that the subscription can decide when to generate the ID and send off that initial subscription ID message separately from when it'll want to send any subscription notifications, and the `PendingSubscription` API is really nice for making it explicit when each of these things happens.

So yup, I'm up for merging this anyway, and we can ponder ways to make the API a little more ergonomic in the future :)

core/src/server/rpc_module.rs
```diff
@@ -738,6 +748,71 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 	}
 }
 
+/// Represent a pending subscription which waits being accepted or rejected.
```
dvdplm, line 751:
```suggestion
/// Represent a pending subscription which waits to be accepted or rejected.
```

core/src/server/rpc_module.rs
```diff
@@ -738,6 +748,71 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
...
+///
+/// Warning: you need to call either `PendingSubscription::accept` or `PendingSubscription::reject` otherwise
```
dvdplm, line 753:
```suggestion
/// Note: you need to call either `PendingSubscription::accept` or `PendingSubscription::reject` otherwise
```

core/src/server/rpc_module.rs
```diff
@@ -738,6 +748,71 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
...
+	method: &'static str,
+	/// Unique subscription.
+	uniq_sub: SubscriptionKey,
+	/// Shared Mutex of subscriptions
+	subscribers: Subscribers,
+	/// Request ID.
+	id: Id<'static>,
+}
+
+/// Represent a pending subscription which waits for being accepted or rejected.
```
dvdplm, line 771:
```suggestion
/// Represent a pending subscription which waits until it's either accepted or rejected.
```

core/src/server/rpc_module.rs
```diff
@@ -738,6 +748,71 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
...
+///
+/// This type implement drop for ease of use, such as dropped in error short circuiting via `map_err()?`.
```
dvdplm, line 773:
```suggestion
/// This type implements `Drop` for ease of use, e.g. when dropped in error short circuiting via `map_err()?`.
```

core/src/server/rpc_module.rs
```diff
@@ -738,6 +748,71 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
...
+			let active_sub = Arc::new(());
+			subscribers.lock().insert(uniq_sub.clone(), (sink.clone(), active_sub.clone()));
+			Some(SubscriptionSink { inner: sink, close_notify, method, uniq_sub, subscribers, active_sub })
+		} else {
+			None
+		}
+	}
+}
+
+// If this type is dropped, it will return an InvalidParams error to the subscriber
```
dvdplm, line 806:
```suggestion
// When dropped it returns an [`InvalidParams`] error to the subscriber
```

core/src/error.rs
```diff
@@ -144,46 +142,44 @@ impl Error {
...
-	/// The server closed the subscription, providing a description of the reason as a `String`.
-	Server(String),
+#[derive(Clone, Debug)]
+pub enum SubscriptionClosed {
+	/// The remote peer closed the connection or called the unsubscribe method.
+	RemotePeerAborted,
+	/// The subscription was completed successfully by the server.
+	Success,
+	/// The subscription failed during execution by the server.
+	Failed(ErrorObject<'static>),
```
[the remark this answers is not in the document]
niklasad1: no, `Success` is only used when the `stream was completed` i.e, returned `Ok(None)` all other server related errors will be treated as `Failed`.

However, if the `send` fails we might regard it as `AbortedByRemotePeer` when it gets `TrySendError(Disconnected)` technically the server could have been closed entirely but at that point nothing can sent anyway so should ok.

core/src/client/async_client/helpers.rs
```diff
@@ -110,6 +105,30 @@ pub(crate) fn process_subscription_response(
...
+	let request_id = match manager.get_request_id_by_subscription_id(&sub_id) {
+		Some(request_id) => request_id,
+		None => {
+			tracing::error!("The server tried to close down an invalid subscription: {:?}", sub_id);
+			return Err(Error::InvalidSubscriptionId);
+		}
+	};
+
+	debug_assert!(
+		manager.remove_subscription(request_id, sub_id).is_some(),
```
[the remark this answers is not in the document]
niklasad1: let's remove it, it just enabled for tests and so on.

lexnv (approved):

Issue #729, opened 2022-04-11
[ws server]: add a way to limit the number of subscriptions per connection


niklasad1:
It would be good to a way to limit the number of subscriptions per connection.

A suggested approach is to [wrap close_notify](https://github.com/paritytech/jsonrpsee/blob/master/ws-server/src/server.rs#L304) in a new type to get the strong count as it is increased by one for each subscription.

niklasad1:
Closed by #739

PR #739, opened 2022-04-26
feat: add a way to limit the number of subscriptions per connection

niklasad1:
Closing #729

### DIFF
--- a/core/src/server/helpers.rs
+++ b/core/src/server/helpers.rs
@@ -25,6 +25,7 @@
 // DEALINGS IN THE SOFTWARE.
 
 use std::io;
+use std::sync::Arc;
 
 use crate::{to_json_raw_value, Error};
 use futures_channel::mpsc;
@@ -32,6 +33,7 @@ use futures_util::StreamExt;
 use jsonrpsee_types::error::{ErrorCode, ErrorObject, ErrorResponse, OVERSIZED_RESPONSE_CODE, OVERSIZED_RESPONSE_MSG};
 use jsonrpsee_types::{Id, InvalidRequest, Response};
 use serde::Serialize;
+use tokio::sync::Notify;
 
 /// Bounded writer that allows writing at most `max_len` bytes.
 ///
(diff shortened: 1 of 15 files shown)


core/src/server/helpers.rs
```diff
@@ -196,8 +198,41 @@ pub async fn collect_batch_response(rx: mpsc::UnboundedReceiver<String>) -> Stri
 	buf
 }
 
+/// Wrapper over [`tokio::sync::Notify`] with bounds check.
+#[derive(Debug)]
+pub struct BoundedSubscriptions {
+	inner: Arc<Notify>,
```
jsdw, line 204: I'm wondering whether we need the `Arc` here? What would happen if we ditched it?

core/src/server/helpers.rs
```diff
@@ -196,8 +198,41 @@ pub async fn collect_batch_response(rx: mpsc::UnboundedReceiver<String>) -> Stri
 	buf
 }
 
+/// Wrapper over [`tokio::sync::Notify`] with bounds check.
+#[derive(Debug)]
+pub struct BoundedSubscriptions {
```
jsdw, line 203: I was just sortof wodnering whether this could be replaced with a Semaphore (https://docs.rs/tokio/latest/tokio/sync/struct.Semaphore.html), which would start at `max_subscriptions` and hand out permits for subscriptions until it runs out (and starts returning errors).

jsdw (approved):
Looks good to me!

Would be good to have an integration test that it works :)

core/src/server/helpers.rs
niklasad1: yeah, that would be nice the hard to part is where to store the semaphore after acquiring it.

I guess we could move it to the callbacks but that's quite annoying.

jsdw: I'm happy to not worry about it for now; we can always have a pass later to see whether we can simplify/optimise things!

core/src/server/rpc_module.rs
```diff
@@ -70,15 +72,15 @@ pub type ConnectionId = usize;
...
+///   - a [`crate::server::helpers::SubscriptionPermit`] to allow subscribers to notify their [`SubscriptionSink`] when they disconnect.
+pub type RawRpcResponse = (String, mpsc::UnboundedReceiver<String>, SubscriptionPermit);
 
 /// Helper struct to manage subscriptions.
 pub struct ConnState<'a> {
 	/// Connection ID
 	pub conn_id: ConnectionId,
-	/// Get notified when the connection to subscribers is closed.
-	pub close_notify: Arc<Notify>,
+	/// Get notified when the connection to subscribers is closed.1
```
lexnv, line 82:
```suggestion
	/// Get notified when the connection to subscribers is closed.
```

lexnv (approved):

core/src/server/rpc_module.rs
```diff
@@ -393,17 +405,19 @@ impl Methods {
 		let sink = MethodSink::new(tx_sink);
 		let id = req.id.clone();
 		let params = Params::new(req.params.map(|params| params.get()));
-		let notify = Arc::new(Notify::new());
+		let bounded_subs = BoundedSubscriptions::new(u32::MAX);
+		let close_notify = bounded_subs.acquire().unwrap();
+		let notify = bounded_subs.acquire().unwrap();
```
niklasad1, line 410:
```suggestion
		let notify = bounded_subs.acquire().expect("u32::MAX permits is sufficient; qed");
```

Issue #668, opened 2022-01-23
[rust docs]: feature gated docs doesn't show up on `docs.rs`


niklasad1:
The rustdocs doesn't document the code hidden behind feature flags i.e, it can't find for example the client module in https://docs.rs/jsonrpsee-core/0.8.0/jsonrpsee_core/

niklasad1:
Closed by #755

PR #741, opened 2022-04-27
fix: generate docs behind features

niklasad1:
Closing https://github.com/paritytech/jsonrpsee/issues/668

### DIFF
--- a/jsonrpsee/src/macros.rs
+++ b/jsonrpsee/src/macros.rs
@@ -1,3 +1,14 @@
+#[macro_export]
+macro_rules! cfg_feature {
+    ($feature:literal, $($item:item)*) => {
+        $(
+            #[cfg(feature = $feature)]
+            #[cfg_attr(docsrs, doc(cfg(feature = $feature)))]
+            $item
+        )*
+    }
+}
+
 macro_rules! cfg_client {
     ($($item:item)*) => {
         $(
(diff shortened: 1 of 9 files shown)


TarikGul (approved):
LGTM, its nice and cleaner

dvdplm (approved):

lexnv (approved):

jsonrpsee/src/macros.rs
```diff
@@ -0,0 +1,116 @@
...
+
+macro_rules! cfg_proc_macros {
+    ($($item:item)*) => {
+        $(
+            #[cfg(feature = "jsonrpsee-proc-macros")]
+            #[cfg_attr(docsrs, doc(cfg(feature = "jsonrpsee-proc-macros")))]
+            $item
+        )*
+    }
+}
```
jsdw, line 97: Would it make sense to write a macro like:

```rust
macro_rules! cfg_feature {
    ($feature:literal, $($item:item)*) => {
        $(
            #[cfg(feature = $feature)]
            #[cfg_attr(docsrs, doc(cfg(feature = $feature)))]
            $item
        )*
    }
}
```

And then all of the other macros across these crates can just call that internally like:

```rust
macro_rules! cfg_proc_macros {
    ($($item:item)*) => cfg_feature!("jsonrpc-proc-macros", $($item:item)*)
}
```

There seem like enough copies to justify it :)

jsdw (approved):
LGTM, just one suggestion to reduce the macro repetition

jsonrpsee/src/macros.rs
niklasad1: yes, I agree

PR #742, opened 2022-04-28
refactor: remove SubscriptionSink::inner_send

niklasad1:
No change just to make the easier to follow without checks in two different functions.

### DIFF
--- a/core/src/server/rpc_module.rs
+++ b/core/src/server/rpc_module.rs
@@ -843,7 +843,7 @@ impl SubscriptionSink {
 		}
 
 		let msg = self.build_message(result)?;
-		Ok(self.inner_send(msg))
+		Ok(self.inner.send_raw(msg).is_ok())
 	}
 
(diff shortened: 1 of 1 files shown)


core/src/server/rpc_module.rs
```diff
@@ -843,7 +843,7 @@ impl SubscriptionSink {
 		}
 
 		let msg = self.build_message(result)?;
-		Ok(self.inner_send(msg))
+		Ok(self.inner.send_raw(msg).is_ok())
```
jsdw, line 846: Just to check; we don't care about checking that `self.is_active_subscription()` is true any more?

jsdw: Aah I see; is_closed has that check now!

jsdw (approved):

TarikGul (approved):

niklasad1:
Superseeded by #743

PR #743, opened 2022-04-28
fix(rpc module): close subscription task when a subscription is `unsubscribed` via the `unsubscribe call`

niklasad1:
The reason for this is that the stream passed to `pipe_from_stream` might not produce new items and that can lead to that calls that are `unsubscribed` is not terminated until an item for the underlying stream is produced.

In worst these tasks will not be terminated until the actual connection is closed which this fixes.

After this PR the subscriptions is terminated once the unsubscribe call is received.

### DIFF
--- a/core/src/server/rpc_module.rs
+++ b/core/src/server/rpc_module.rs
@@ -35,7 +35,7 @@ use crate::id_providers::RandomIntegerIdProvider;
 use crate::server::helpers::MethodSink;
 use crate::server::resource_limiting::{ResourceGuard, ResourceTable, ResourceVec, Resources};
 use crate::traits::{IdProvider, ToRpcParams};
-use futures_channel::{mpsc, oneshot};
+use futures_channel::mpsc;
 use futures_util::future::Either;
 use futures_util::pin_mut;
 use futures_util::{future::BoxFuture, FutureExt, Stream, StreamExt, TryStream, TryStreamExt};
@@ -48,7 +48,7 @@ use jsonrpsee_types::{
 use parking_lot::Mutex;
 use rustc_hash::FxHashMap;
 use serde::{de::DeserializeOwned, Serialize};
-use tokio::sync::Notify;
+use tokio::sync::{watch, Notify};
 
 /// A `MethodCallback` is an RPC endpoint, callable with a standard JSON-RPC request,
 /// implemented as a function pointer to a `Fn` function taking four arguments:
@@ -98,7 +98,7 @@ impl<'a> std::fmt::Debug for ConnState<'a> {
 	}
 }
 
-type Subscribers = Arc<Mutex<FxHashMap<SubscriptionKey, (MethodSink, oneshot::Sender<()>)>>>;
+type Subscribers = Arc<Mutex<FxHashMap<SubscriptionKey, (MethodSink, watch::Sender<()>)>>>;
 
 /// Represent a unique subscription entry based on [`RpcSubscriptionId`] and [`ConnectionId`].
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
@@ -794,9 +794,9 @@ impl PendingSubscription {
 		let InnerPendingSubscription { sink, close_notify, method, uniq_sub, subscribers, id } = inner;
 
 		if sink.send_response(id, &uniq_sub.sub_id) {
-			let (tx, rx) = oneshot::channel();
+			let (tx, rx) = watch::channel(());
 			subscribers.lock().insert(uniq_sub.clone(), (sink.clone(), tx));
-			Some(SubscriptionSink { inner: sink, close_notify, method, uniq_sub, subscribers, unsubscribe: Some(rx) })
+			Some(SubscriptionSink { inner: sink, close_notify, method, uniq_sub, subscribers, unsubscribe: rx })
 		} else {
 			None
 		}
@@ -827,7 +827,7 @@ pub struct SubscriptionSink {
 	/// Shared Mutex of subscriptions for this method.
 	subscribers: Subscribers,
 	/// Future that returns when the `unsubscribe method ` has been called.
-	unsubscribe: Option<oneshot::Receiver<()>>,
+	unsubscribe: watch::Receiver<()>,
 }
 
 impl SubscriptionSink {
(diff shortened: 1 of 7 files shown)


core/src/server/rpc_module.rs
```diff
@@ -722,6 +722,8 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 					};
 					let sub_id = sub_id.into_owned();
 
+					tracing::info!("unsubscribe=`{}` id={:?}", unsubscribe_method_name, sub_id);
```
niklasad1, line 725:
```suggestion
```

core/src/server/rpc_module.rs
```diff
@@ -948,21 +968,22 @@ impl SubscriptionSink {
...
-	/// Returns whether this channel is closed without needing a context.
+	/// Returns whether the subscription is closed.
 	pub fn is_closed(&self) -> bool {
-		self.inner.is_closed() || self.close_notify.is_none()
+		self.inner.is_closed() || self.close_notify.is_none() || self.is_active_subscription()
 	}
 
 	fn is_active_subscription(&self) -> bool {
-		Arc::strong_count(&self.active_sub) > 1
+		self.subscribers.lock().contains_key(&self.uniq_sub)
```
niklasad1, line 986: this is a bit annoying with extra mutex accesses but I kept for ease of use.

we need another `AtomicBool` or something because `rx side` of oneshot can't be checked

core/src/server/rpc_module.rs
```diff
@@ -922,11 +938,14 @@ impl SubscriptionSink {
 					break SubscriptionClosed::Failed(err);
 				}
 				Either::Left((Ok(None), _)) => break SubscriptionClosed::Success,
-				Either::Right(((), _)) => {
+				Either::Right((_, _)) => {
+					tracing::info!("subscription stream terminated; by connection reset most likely");
```
niklasad1, line 942:
```suggestion
```

core/src/server/rpc_module.rs
```diff
@@ -881,25 +882,40 @@ impl SubscriptionSink {
...
 		T: Serialize,
 		E: std::fmt::Display,
+		F: FnOnce(SubscriptionClosed, SubscriptionSink),
 	{
-		let close_notify = match self.close_notify.clone() {
+		let conn_closed = match self.close_notify.clone() {
 			Some(close_notify) => close_notify,
-			None => return SubscriptionClosed::RemotePeerAborted,
+			None => {
+				on_close(SubscriptionClosed::RemotePeerAborted, self);
```
jsdw, line 899: I don't really understand the benefit of having this `on_close` closure versus returning, say, `(SubscriptionClosed, Self)`?

And that makes me wonder; would this method take `&mut self` instead then, so it wouldn't need to return `Self` at all, just `SubscriptionClosed`?

And then I wonder, would it be nicer to remove the `SubscriptionClosed::Success` variant and actually return `Result<(), SubscriptionClosed>` rather than `SubscriptionClosed` (and maybe then rename `SubscriptionClosed` to something like `SubscriptionCloseError`?

jsdw: I had to double check, but what I mean is that the closure doesn't protect against it being reused? It might be a little safer if it was handed `&mut self` instead of `self` though, but the safest approach if people need access to some of the functionality from `self`, but not all of it, would be to wrap `self` in some type to make sure it can only be used in a certain way?

jsdw: Dumb question; why can't it be reused?

niklasad1: The refactoring [here](https://github.com/paritytech/jsonrpsee/blob/na-close-tasks-when-unsubscribed/core/src/server/rpc_module.rs#L904) which takes the `unsubscribed oneshot` which is why I wanted the `pipe_from_sink(self)` rather than `pipe_from_sink(&mut self)`

So if I do this:

```rust
  sink.pipe_from(stream1).await;
  // this will compile but not work, unsubscribe oneshot is already consumed via `Option::take`
  sink.pipe_from(stream2).await;
```

But yeah I can do nasty things in the exit closure too you correct,

```rust
  sink.pipe_from(stream1, |_, _| { i can do nasty stuff here as well but async stuff right }).await;
  // this won't compile
  sink.pipe_from(stream2).await;
```

niklasad1: reverted good catch James.

core/src/server/rpc_module.rs
```diff
@@ -891,14 +892,26 @@ impl SubscriptionSink {
...
-		let close_notify = match self.close_notify.clone() {
+		let conn_closed = match self.close_notify.clone() {
 			Some(close_notify) => close_notify,
-			None => return SubscriptionClosed::RemotePeerAborted,
+			None => {
+				return SubscriptionClosed::RemotePeerAborted;
+			}
+		};
+
+		let sub_closed = match self.unsubscribe.take() {
```
jsdw, line 902: Ah I see, so this oneshot unsubscribe channel is removed here, which means we can only use `pipe_from_try_stream` once.

Is it possible to take a mutable ref to it instead to use below (which may need pinning)? Maybe then this method would still be reusable?

niklasad1: it won't work because `send` takes a `&self as well.

I changed it to `tokio::sync::watch` for ease of use.

tests/tests/helpers.rs
```diff
@@ -108,6 +112,54 @@ pub async fn websocket_server_with_subscription() -> (SocketAddr, WsServerHandle
 		})
 		.unwrap();
 
+	module
+		.register_subscription("subscribe_5_ints", "n", "unsubscribe_5_ints", |_, pending, _| {
+			let mut sink = match pending.accept() {
```
dvdplm, line 117: maybe it's better to `unwrap()` here?

tests/tests/helpers.rs
```diff
@@ -108,6 +112,54 @@ pub async fn websocket_server_with_subscription() -> (SocketAddr, WsServerHandle
...
+	module
+		.register_subscription(
+			"subscribe_with_err_on_stream",
+			"n",
+			"unsubscribe_with_err_on_stream",
+			move |_, pending, _| {
+				let mut sink = match pending.accept() {
+					Some(sink) => sink,
+					_ => return,
+				};
```
dvdplm, line 145: same

tests/tests/helpers.rs
```diff
@@ -133,6 +185,34 @@ pub async fn websocket_server() -> SocketAddr {
 	addr
 }
 
+/// Yields at one item then sleeps for an hour.
```
dvdplm, line 188:
```suggestion
/// Yields one item then sleeps for an hour.
```

dvdplm (approved):

tests/tests/helpers.rs
```diff
@@ -108,6 +100,48 @@ pub async fn websocket_server_with_subscription() -> (SocketAddr, WsServerHandle
 		})
 		.unwrap();
 
+	module
+		.register_subscription("subscribe_5_ints", "n", "unsubscribe_5_ints", |_, pending, _| {
```
jsdw, line 104: Now we can call `pipe_from_stream` more than once, should we also have a test to make sure that we can?

jsdw (approved):
Looks great to me!

PR #744, opened 2022-04-28
Make it possible to disable batch requests support

popzxc:
While it doesn't comply to the JSON RPC spec, some servers may explicitly decide to not support batched requests for various reasons. This PR makes it possible.

### DIFF
--- a/http-server/src/server.rs
+++ b/http-server/src/server.rs
@@ -45,7 +45,7 @@ use jsonrpsee_core::server::helpers::{collect_batch_response, prepare_error, Met
 use jsonrpsee_core::server::resource_limiting::Resources;
 use jsonrpsee_core::server::rpc_module::{MethodKind, Methods};
 use jsonrpsee_core::TEN_MB_SIZE_BYTES;
-use jsonrpsee_types::error::ErrorCode;
+use jsonrpsee_types::error::{ErrorCode, ErrorObject, BATCHES_NOT_SUPPORTED_CODE, BATCHES_NOT_SUPPORTED_MSG};
 use jsonrpsee_types::{Id, Notification, Params, Request};
 use serde_json::value::RawValue;
 use tokio::net::{TcpListener, ToSocketAddrs};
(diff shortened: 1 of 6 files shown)


niklasad1 (approved):
LGTM.

I wonder what those reasons are to reject batch requests because I think the `max_response_body_size` could work too for that reason but I realized now that we just use for individual calls it seems.

popzxc:
Well, the server may have computationally expensive requests, for example.
An example of real-world RPC server that explicitly disables batch requests would be [Amazon Managed Blockchain](https://docs.aws.amazon.com/managed-blockchain/latest/ethereum-dev/ethereum-json-rpc.html).

jsdw:
I'm not sure what avoiding computationally expensive code has to do with batches, offhand; what's to stop a user submitting the same as multiple individual messages, which is basically how jsonrpsee treats batches anyway?

Do you have a real use case for this that you might be able to share so that I can better understand the need for it?

popzxc:
> what's to stop a user submitting the same as multiple individual messages, which is basically how jsonrpsee treats batches anyway?

Throttling/rate limiting algorithms, basically. Obviously, it's possible to implement those on the application layer, but it'll be more clunky.

> Do you have a real use case for this that you might be able to share so that I can better understand the need for it?

Sure. We have RPC methods that invoke virtual machine to execute smart contracts. It means that these methods can be arbitrary "heavy", and we need to protect our servers from being overloaded. We're working on the composite solution that involves throttling/rate limiting, and ability to send batch requests kind of stands in the way right now.

niklasad1:
so actually I think the resource limiting stuff could be a good alternative to disabling batch requests completely however I think it's fine to have an option to disable batch requests as well.

popzxc:
I mean, they are not mutually exclusive. We will have the full set: both resource limiting and disabled batches. *Just* with rate-limiting but with enabled batches it's possible to obtain weird result: imagine that you sent the batch with requests that mutate state, and only some of them were executed. It will be confusing to users, so better to prevent this situation.

jsdw:
At the end of the day, the code change is fairly minimal, and it's well tested and looks good to me, so I'm happy to approve :)

jsdw (approved):

niklasad1:
sorry @popzxc can you fix the conflicts so we can merge this?

Issue #748, opened 2022-04-30
Cannot make a request to https endpoint


qezz:
I'm trying to make a request to an HTTPs endpoint, but it fails with a weird error

```rust
use jsonrpsee::core::client::ClientT;
use jsonrpsee::http_client::HttpClientBuilder;

pub async fn http_call() -> anyhow::Result<()> {
    let client = HttpClientBuilder::default()
        .build("https://kusama-rpc.polkadot.io:443")
        .unwrap();
    let response: String = client.request("system_chain", None).await.unwrap();

    tracing::info!("response: {:?}", response);

    Ok(())
}

#[tokio::main]
async fn main() {
    tracing_subscriber::FmtSubscriber::builder()
	.with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
	.try_init()
	.expect("setting default subscriber failed");

    http_call().await.unwrap();
}

```

output
```console
$ RUST_LOG=debug ./target/debug/sample
Apr 30 04:38:06.331 DEBUG hyper_rustls::config: with_native_roots processed 161 valid and 0 invalid certs
Apr 30 04:38:06.334 DEBUG jsonrpsee_http_client::transport: send: {"jsonrpc":"2.0","id":0,"method":"system_chain"}
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Transport(HTTP error: error trying to connect: invalid URL, scheme is not http)', src/main.rs:8:71
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

deps
```toml
tokio = { version = "1", features = ["full"] }
anyhow = "1.0.57"
tracing = "0.1.34"
tracing-subscriber = "0.2.0"
jsonrpsee = { version = "=0.11.0", features = ["full"] }
```

niklasad1:
right, thanks,

the HTTP connector is reused for both http or https I thought it could be reused.

niklasad1:
#750 is the fix... should be in next week I guess.

qezz:
Thank you!

niklasad1:
Fixed by #750

PR #750, opened 2022-04-30
fix(http client): use https connector for https

niklasad1:
Closing https://github.com/paritytech/jsonrpsee/issues/748

Wrapping the `HttpConnector` in the `HttpsConnector` doesn't work, see for futher info https://github.com/rustls/hyper-rustls/issues/169 which this PR fixes.

### DIFF
--- a/client/http-client/src/transport.rs
+++ b/client/http-client/src/transport.rs
@@ -57,13 +57,13 @@ impl HttpTransportClient {
 			return Err(Error::Url("Port number is missing in the URL".into()));
 		}
 
-		let mut connector = HttpConnector::new();
-
-		connector.set_reuse_address(true);
-		connector.set_nodelay(true);
-
 		let client = match target.scheme_str() {
 			Some("http") => {
+				let mut connector = HttpConnector::new();
+
+				connector.set_reuse_address(true);
+				connector.set_nodelay(true);
+
 				let client = Client::builder().build::<_, hyper::Body>(connector);
 				HyperClient::Http(client)
 			}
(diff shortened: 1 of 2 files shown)


client/http-client/src/transport.rs
```diff
@@ -57,13 +57,13 @@ impl HttpTransportClient {
...
-		let mut connector = HttpConnector::new();
-
-		connector.set_reuse_address(true);
-		connector.set_nodelay(true);
-
 		let client = match target.scheme_str() {
 			Some("http") => {
+				let mut connector = HttpConnector::new();
+
+				connector.set_reuse_address(true);
```
niklasad1, line 64: remove these to more "similar" to `HTTPs connector?

tests/tests/integration_tests.rs
```diff
@@ -251,17 +251,15 @@ async fn http_making_more_requests_than_allowed_should_not_deadlock() {
 }
 
 #[tokio::test]
-#[ignore]
 async fn https_works() {
-	let client = HttpClientBuilder::default().build("https://kusama-rpc.polkadot.io").unwrap();
+	let client = HttpClientBuilder::default().build("https://kusama-rpc.polkadot.io:443").unwrap();
```
jsdw, line 255: isn't 443 assumed anyway when "https"?

client/http-client/src/transport.rs
```diff
@@ -74,16 +74,15 @@ impl HttpTransportClient {
...
 						.with_webpki_roots()
 						.https_or_http()
 						.enable_http1()
-						.wrap_connector(connector),
+						.build(),
 					_ => return Err(Error::InvalidCertficateStore),
 				};
-				let client = Client::builder().build::<_, hyper::Body>(connector);
-				HyperClient::Https(client)
+				HyperClient::Https(Client::builder().build::<_, hyper::Body>(connector))
```
jsdw, line 85: I'm not really sure what the impact of this change is; was wrapping the HttpConnector causing an issue?

jsdw: Ah I see; I just read https://github.com/rustls/hyper-rustls/issues/169. Interesting!

My reading/assumption was that the wrapping would work as we had expected!

tests/tests/integration_tests.rs
niklasad1: you are correct but [hyper::Uri](https://github.com/hyperium/http/issues/11) doesn't support default ports

client/http-client/src/transport.rs
jsdw: Does removing them have any obvious impact on anything? I did a bit of reading up on them and they both sound like optimisations anyway (I wasn't sure that nodelay would actually help anything much here; it sounds like it trades a little throughput for lower latency, and reuseaddress sounds like an optimisation around freeing up an address a little quicker mainly)

niklasad1: nah, no real difference might reduce the number of file descriptors for the benchmarks.

let's remove it.

jsdw (approved):
LGTM!

TarikGul (approved):
lgtm

PR #755, opened 2022-05-06
Hide internal macros from public interface

jsdw:
The title says it all :)

### DIFF
--- a/core/src/macros.rs
+++ b/core/src/macros.rs
@@ -1,4 +1,3 @@
-#[macro_export]
 macro_rules! cfg_feature {
     ($feature:literal, $($item:item)*) => {
         $(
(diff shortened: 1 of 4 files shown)


core/src/macros.rs
```diff
@@ -11,19 +10,19 @@ macro_rules! cfg_feature {
 
 macro_rules! cfg_client {
 	($($item:item)*) => {
-		$crate::cfg_feature!("client", $($item)*);
+		cfg_feature!("client", $($item)*);
```
jsdw, line 13: I wasn't sure, but things seem to work fine without properly namespacing the macro!

core/src/macros.rs
```diff
@@ -1,4 +1,3 @@
-#[macro_export]
```
jsdw, line 1: It's this macro_export which makes the macro part of the public interface I believe.

core/src/macros.rs
dvdplm: Hmm, here be dragons. I *think* there are situations when this will not work well, but can't remember what the full story was. `$crate::macros::cfg_feature!` should work though right?

dvdplm (approved):

niklasad1 (approved):

niklasad1:
> Hmm, here be dragons. I think there are situations when this will not work well, but can't remember what the full story was. $crate::macros::cfg_feature! should work though right?

should be alright because they are internal anyway, so if it works in the CI that should be sufficent

niklasad1:
Great James, I looked at the generated docs and looks good :)

PR #756, opened 2022-05-07
feat: Allow user to send custom headers in http client

semtexzv:
As a user, I want to use authorization for some requests, or implement cache-control on top of my RPCs to bitcoin nodes.

This seems like a simplest way to achieve that.

### DIFF
--- a/client/http-client/src/client.rs
+++ b/client/http-client/src/client.rs
@@ -24,12 +24,16 @@
 // IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 // DEALINGS IN THE SOFTWARE.
 
+use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Duration;
 
 use crate::transport::HttpTransportClient;
 use crate::types::{ErrorResponse, Id, NotificationSer, ParamsSer, RequestSer, Response};
 use async_trait::async_trait;
+use hyper::header::{HeaderName, InvalidHeaderName};
+use hyper::http::HeaderValue;
+use hyper::HeaderMap;
 use jsonrpsee_core::client::{CertificateStore, ClientT, IdKind, RequestIdManager, Subscription, SubscriptionClientT};
 use jsonrpsee_core::{Error, TEN_MB_SIZE_BYTES};
 use jsonrpsee_types::error::CallError;
@@ -44,6 +48,7 @@ pub struct HttpClientBuilder {
 	max_concurrent_requests: usize,
 	certificate_store: CertificateStore,
 	id_kind: IdKind,
+	custom_headers: HeaderMap,
 }
 
 impl HttpClientBuilder {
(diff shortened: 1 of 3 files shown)


client/http-client/src/client.rs
```diff
@@ -77,10 +79,17 @@ impl HttpClientBuilder {
 		self
 	}
 
+	///
+	pub fn custom_headers(mut self, headers: HeaderMap) -> Self {
```
niklasad1, line 83: I would rather have a more unified API for the WS client and HTTP client, so if you could add a similar API as https://github.com/paritytech/jsonrpsee/blob/master/client/ws-client/src/lib.rs#L125-#L128 I would be happy with it.

The reason why is to avoid to require users to use `http::HeaderMap`

niklasad1 (commented):
Looks overall good,

Some suggestions on the `custom_headers API` to discuss further though.

client/http-client/src/client.rs
semtexzv: How about now?

niklasad1: sorry, I think it would make more sense to return an error instead of `unwrap` these might fail for example if `HeaderName` is uppercase and similar.

in addition an already existing header will just be overwritten with `append`.

so might be better what you did initially for simplicity and change the WS client API instead.

niklasad1: //cc @jsdw any opinions?

semtexzv: I Added a custom error variant, and made the method fallible. I think this is the nicest we can make this without exposing extraneous APIs.

I also would like to keep the ability to overwrite default headers, what if user wants to send `Content-Type: application/json+jsonrpc` instead of the default `Content-type: application/json`

client/http-client/src/client.rs
```diff
@@ -77,10 +82,20 @@ impl HttpClientBuilder {
 		self
 	}
 
+	/// Adds a custom header, which is sent with every request
```
niklasad1, line 85: can you add some additional docs that explains that "dupllicate" keys will be overwritten by this method?

niklasad1 (approved):
LGTM, please run rustfmt to make the CI happy

client/http-client/src/client.rs
niklasad1: makes sense,

I think the docs should explain that you can overwrite the default headers too perhaps the `doc tests/examples` regarding that would awesome :)

core/src/error.rs
```diff
@@ -27,202 +27,205 @@
...
+    MaxResourcesReached,
+    /// Custom error.
+    #[error("Custom error: {0}")]
+    Custom(String),
+    /// Not implemented for HTTP clients.
+    #[error("Not implemented")]
+    HttpNotImplemented,
+    /// You provided invalid HTTP headers
+    #[error("Invalid http headers provided")]
+    InvalidHeaders,
```
jsdw, line 134:
```suggestion
    #[error("Invalid http header provided")]
    InvalidHeader,
```

(just because this error is emitted when one attempt to set a single header)

client/http-client/src/client.rs
```diff
@@ -77,10 +82,20 @@ impl HttpClientBuilder {
...
+	pub fn add_header(mut self, key: &str, value: &str) -> Result<Self, Error> {
```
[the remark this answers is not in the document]
niklasad1: it's because two different `http libraries` are used, soketto is using to `httparse::Header` which isn't as pedantic as hyper::HeaderMap is. 

that's why I was leaning into exposing `fn add_headers(header: HeaderMap) -> Self` for both instead which makes things more uniform and quite obvious how it works.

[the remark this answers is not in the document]
jsdw: Ah I see! I guess the issue there is that the ws-client doesn't (afaict) depend on hyper, so it would be a shame to add it as a dependency just for that type?

jsdw (approved):
I'm happy with this once fmt and such are fixed, but as @niklasad1 suggested I think it'd be good to move to using something like https://docs.rs/http/latest/http/header/struct.HeaderMap.html consistently across clients and then we can expose a more consistent API for doing this, too.

niklasad1:
@semtexzv are you planning on finishing this or shall we take it over?

It's a useful PR for sure.

niklasad1:
Superseeded by https://github.com/paritytech/jsonrpsee/pull/814

PR #757, opened 2022-05-09
use core::result::Result, avoid duplicate names.

rise0chen:


niklasad1 (commented):
hey, thanks for your PR can you explain the benefit with this change?

rise0chen:
```rust
type Result<T> = core::result::Result<T, Error>;
type ExampleStorageKey = Vec<u8>;

#[rpc(server, client, namespace = "state")]
pub trait Rpc<Hash: Clone, StorageKey>
where
	Hash: std::fmt::Debug,
{
	#[method(name = "getKeys")]
	async fn storage_keys(&self, storage_key: StorageKey, hash: Option<Hash>) -> Result<Vec<StorageKey>>;

	/// Subscription that takes a `StorageKey` as input and produces a `Vec<Hash>`.
	#[subscription(name = "subscribeStorage" => "override", item = Vec<Hash>)]
	fn subscribe_storage(&self, keys: Option<Vec<StorageKey>>);
}
```

Failed to compile this code.
Because `type Result<T> = core::result::Result<T, Error>;`

niklasad1:
ok,

we actually re-export a type alias for what you did here [this](https://github.com/paritytech/jsonrpsee/blob/master/core/src/lib.rs#L61)

hmm, the library doesn't compile for no_std we so have assumed `std::result::Result` is used, any particular reason why you use `core::Result`?

rise0chen:
It is my personal habits.
`std::result::Result` is a re-export of `core::result::Result`.  And, I think `core::result::Result` is more general.

jsdw:
this PR isn't harmful, but I don't think it benefits us either; `Result` should always be available within the scope of the proc macro crate, should it not? So let's keep it simple unless this actually has some benefit somewhere :)

jsdw:
I'll close this, but feel free to reopen or discuss in an issue if there are merits to merging this that I'm not aware of :)

PR #763, opened 2022-05-11
feat: add http health API

niklasad1:


http-server/src/server.rs
```diff
@@ -687,3 +717,42 @@ async fn process_validated_request(
...
+				let res = (callback)(Id::Number(0), Params::new(None), sink.clone(), 0, None).await;
+				middleware.on_result(name, false, request_start);
+				res
+			}
+
+			MethodKind::Subscription(_) | MethodKind::Unsubscription(_) => false,
+		},
+	};
+
+	rx.close();
```
niklasad1, line 750:
```suggestion
```

lexnv (approved):

dvdplm (approved):
❤️

jsdw (approved):

http-server/src/server.rs
```diff
@@ -166,6 +169,14 @@ impl<M> Builder<M> {
 		self
 	}
 
+	/// Enable health endpoint.
+	/// Allows you to expose one of the methods under GET /<path> The method will be invoked with no parameters. Error returned from the method will be converted to status 500 response.
+	/// Expects a tuple with (<path>, <rpc-method-name>).
+	pub fn health_api(mut self, path: impl Into<String>, method: impl Into<String>) -> Self {
+		self.health_api = Some(HealthApi { path: path.into(), method: method.into() });
+		self
+	}
```
bkchr, line 178: The docs are rather confusing. I needed to read the code to fully understand on what this is doing.

And why is the method called `health_api`? Couldn't this theoretically be any kind of endpoint? (which is already implicitly supported, because `path` can be anything)

niklasad1: right, you are correct.

I did the fastest possible way and did the same as the `jsonrpc crate` does.

Sure let's open an issue and do something better.

PR #767, opened 2022-05-12
fix: generate docs for all features on docs.rs

niklasad1:


TarikGul (approved):

client/http-client/Cargo.toml
```diff
@@ -29,3 +29,10 @@ tokio = { version = "1.16", features = ["net", "rt-multi-thread", "macros"] }
 [features]
 default = ["tls"]
 tls = ["hyper-rustls/webpki-tokio"]
+
+[package.metadata.docs.rs]
+all-features = true
+rustdoc-args = ["--cfg", "docsrs"]
```
jsdw, line 35: What does this do?

jsdw (approved):

client/http-client/Cargo.toml
niklasad1: By default `docs.rs` only builds the docs with default-features so we want all features to be documented.

jsdw: Ah thanks, and the `--cfg docsrs` bit is about making it possible to conditionally compile stuff for doc purposes I gather!

niklasad1: should be equivalent to `RUSTDOCFLAGS="--cfg docrs" cargo doc --workspace --no-deps --document-private-items` it seems.

PR #772, opened 2022-05-17
Implement `ping-pong` for WebSocket clients

### LINKED ISSUE #738: [ws server]: add support periodic `WS ping` a.k.a "keep-alive"
WebSocket supports `ping-pong` to check if the connection is still "healthy", it would good to have configuration on the server to send out these periodically at some specified interval should be opt-in by default.

lexnv:
This PR exposes the ping-pong interface of the WebSocket clients.

A new `ping_interval` configuration is exposed for async-clients.
The ping interval represents the amount of time that should be elapsed between
an activity event (frontend event, backend event / `pong` frame received)
and submitting a ping frame to the RPC server.

From the RPC: https://www.rfc-editor.org/rfc/rfc6455#section-5.5.2

> Upon receipt of a Ping frame, an endpoint MUST send a Pong frame in
   response, unless it already received a Close frame.  It SHOULD
   respond with Pong frame as soon as is practical.

It is possible that the `Pong` reply is not submitted in the expected time
(ie if the server received a `state_getKeys` request that is taking minutes to finish).
Therefore the `ping_interval` is set to 5 minutes to capture those cases. 

Closes #738.

### DIFF
--- a/client/transport/src/web.rs
+++ b/client/transport/src/web.rs
@@ -76,7 +76,7 @@ impl TransportReceiverT for Receiver {
 
 				match msg {
 					Message::Bytes(bytes) => Ok(ReceivedMessage::Bytes(bytes)),
-					Message::Text(txt) => Ok(ReceivedMessage::Data(txt)),
+					Message::Text(txt) => Ok(ReceivedMessage::Text(txt)),
 				}
 			}
 			Some(Err(err)) => Err(Error::WebSocket(err)),
--- a/client/transport/src/web.rs
+++ b/client/transport/src/web.rs
@@ -22,6 +22,9 @@ pub enum Error {
 	/// WebSocket error
 	#[error("WebSocket Error: {0:?}")]
 	WebSocket(WebSocketError),
+	/// Operation not supported
+	#[error("Operation not supported")]
+	NotSupported,
 }
 
 /// Sender.
@@ -54,7 +57,7 @@ impl TransportSenderT for Sender {
 
 	async fn send_ping(&mut self) -> Result<(), Self::Error> {
 		tracing::trace!("send ping - not implemented for wasm");
-		Ok(())
+		Err(Error::NotSupported)
 	}
 
 	async fn close(&mut self) -> Result<(), Error> {
--- a/client/transport/src/web.rs
+++ b/client/transport/src/web.rs
@@ -5,7 +5,7 @@ use futures_util::sink::SinkExt;
 use futures_util::stream::{SplitSink, SplitStream, StreamExt};
 use gloo_net::websocket::{futures::WebSocket, Message, WebSocketError};
 use jsonrpsee_core::async_trait;
-use jsonrpsee_core::client::{TransportReceiverT, TransportSenderT};
+use jsonrpsee_core::client::{ReceivedMessage, TransportReceiverT, TransportSenderT};
 
 /// Web-sys transport error that can occur.
 #[derive(Debug, thiserror::Error)]
@@ -52,6 +52,11 @@ impl TransportSenderT for Sender {
 		Ok(())
 	}
 
+	async fn send_ping(&mut self) -> Result<(), Self::Error> {
+		tracing::trace!("send ping - not implemented for wasm");
+		Ok(())
+	}
+
 	async fn close(&mut self) -> Result<(), Error> {
 		Ok(())
 	}
@@ -61,17 +66,15 @@ impl TransportSenderT for Sender {
 impl TransportReceiverT for Receiver {
 	type Error = Error;
 
-	async fn receive(&mut self) -> Result<String, Self::Error> {
+	async fn receive(&mut self) -> Result<ReceivedMessage, Self::Error> {
 		match self.0.next().await {
 			Some(Ok(msg)) => {
 				tracing::trace!("rx: {:?}", msg);
 
-				let txt = match msg {
-					Message::Bytes(bytes) => String::from_utf8(bytes).expect("WebSocket message is valid utf8; qed"),
-					Message::Text(txt) => txt,
-				};
-
-				Ok(txt)
+				match msg {
+					Message::Bytes(bytes) => Ok(ReceivedMessage::Bytes(bytes)),
+					Message::Text(txt) => Ok(ReceivedMessage::Data(txt)),
+				}
 			}
 			Some(Err(err)) => Err(Error::WebSocket(err)),
 			None => Err(Error::SenderDisconnected),
--- a/client/transport/src/ws/mod.rs
+++ b/client/transport/src/ws/mod.rs
@@ -41,6 +41,7 @@ use thiserror::Error;
 use tokio::net::TcpStream;
 
 pub use http::{uri::InvalidUri, Uri};
+use soketto::data::ByteSlice125;
 pub use soketto::handshake::client::Header;
 
 /// Sending end of WebSocket transport.
@@ -191,6 +192,16 @@ impl TransportSenderT for Sender {
 	async fn send(&mut self, body: String) -> Result<(), Self::Error> {
 		tracing::debug!("send: {}", body);
 		self.inner.send_text(body).await?;
+		self.inner.flush().await?;
+		Ok(())
+	}
+
+	/// Sends out a ping request. Returns a `Future` that finishes when the request has been
+	/// successfully sent.
+	async fn send_ping(&mut self, data: &[u8]) -> Result<(), Self::Error> {
+		tracing::debug!("send ping: {:?}", data);
+		let byte_slice = ByteSlice125::try_from(data).expect("Found invalid ping slice");
+		self.inner.send_ping(byte_slice).await?;
 		self.inner.flush().await?;
 		Ok(())
 	}
--- a/client/transport/src/ws/mod.rs
+++ b/client/transport/src/ws/mod.rs
@@ -37,7 +37,7 @@ use jsonrpsee_core::{async_trait, Cow};
 use soketto::connection::Error::Utf8;
 use soketto::data::ByteSlice125;
 use soketto::handshake::client::{Client as WsHandshakeClient, ServerResponse};
-use soketto::{connection, Data};
+use soketto::{connection, Data, Incoming};
 use stream::EitherStream;
 use thiserror::Error;
 use tokio::net::TcpStream;
@@ -223,14 +223,19 @@ impl TransportReceiverT for Receiver {
 
 	/// Returns a `Future` resolving when the server sent us something back.
 	async fn receive(&mut self) -> Result<ReceivedMessage, Self::Error> {
-		let mut message = Vec::new();
+		loop {
+			let mut message = Vec::new();
+			let recv = self.inner.receive(&mut message).await?;
 
-		match self.inner.receive_data(&mut message).await? {
-			Data::Text(_) => {
-				let s = String::from_utf8(message).map_err(|err| WsError::Connection(Utf8(err.utf8_error())))?;
-				Ok(ReceivedMessage::Text(s))
+			match recv {
+				Incoming::Data(Data::Text(_)) => {
+					let s = String::from_utf8(message).map_err(|err| WsError::Connection(Utf8(err.utf8_error())))?;
+					break Ok(ReceivedMessage::Text(s));
+				}
+				Incoming::Data(Data::Binary(_)) => break Ok(ReceivedMessage::Bytes(message)),
+				Incoming::Pong(_) => break Ok(ReceivedMessage::Pong),
+				_ => continue,
 			}
-			Data::Binary(_) => Ok(ReceivedMessage::Bytes(message)),
 		}
 	}
 }
--- a/client/transport/src/ws/mod.rs
+++ b/client/transport/src/ws/mod.rs
@@ -31,11 +31,11 @@ use std::net::{SocketAddr, ToSocketAddrs};
 use std::time::Duration;
 
 use futures_util::io::{BufReader, BufWriter};
-use jsonrpsee_core::client::{CertificateStore, TransportReceiverT, TransportSenderT};
+use jsonrpsee_core::client::{CertificateStore, ReceivedMessage, TransportReceiverT, TransportSenderT};
 use jsonrpsee_core::TEN_MB_SIZE_BYTES;
 use jsonrpsee_core::{async_trait, Cow};
-use soketto::connection;
 use soketto::handshake::client::{Client as WsHandshakeClient, ServerResponse};
+use soketto::{connection, Incoming};
 use stream::EitherStream;
 use thiserror::Error;
 use tokio::net::TcpStream;
@@ -217,11 +217,19 @@ impl TransportReceiverT for Receiver {
 	type Error = WsError;
 
 	/// Returns a `Future` resolving when the server sent us something back.
-	async fn receive(&mut self) -> Result<String, Self::Error> {
+	async fn receive(&mut self) -> Result<ReceivedMessage, Self::Error> {
 		let mut message = Vec::new();
-		self.inner.receive_data(&mut message).await?;
-		let s = String::from_utf8(message).expect("Found invalid UTF-8");
-		Ok(s)
+
+		loop {
+			let recv = self.inner.receive(&mut message).await?;
+
+			if let Incoming::Data(_) = recv {
+				let s = String::from_utf8(message).expect("Found invalid UTF-8");
+				return Ok(ReceivedMessage::Data(s));
+			} else if let Incoming::Pong(pong_data) = recv {
+				return Ok(ReceivedMessage::Pong(Vec::from(pong_data)));
+			}
+		}
 	}
 }
 
--- a/client/transport/src/ws/mod.rs
+++ b/client/transport/src/ws/mod.rs
@@ -30,10 +30,12 @@ use std::io;
 use std::net::{SocketAddr, ToSocketAddrs};
 use std::time::Duration;
 
+use crate::ws::WsError::Custom;
 use futures_util::io::{BufReader, BufWriter};
 use jsonrpsee_core::client::{CertificateStore, ReceivedMessage, TransportReceiverT, TransportSenderT};
 use jsonrpsee_core::TEN_MB_SIZE_BYTES;
 use jsonrpsee_core::{async_trait, Cow};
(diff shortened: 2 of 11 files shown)


niklasad1:
hehe, the idea behind the issue was to implement server-side pings based on what Jaco told us about JS-side of things [here](https://github.com/paritytech/substrate/issues/11239#issuecomment-1109408208)

however, I think this is useful for the client as well.

niklasad1 (commented):
Looks good overall :+1: 

It would be better to make the extra ping/pongs optional and disable it by default.
I think there a might be a use case where one wants avoid this extra overhead to send these extra messages (not exactly sure but currently it's not possible to turn it off except a very long ping interval)

client/transport/src/ws/mod.rs
```diff
@@ -31,16 +31,17 @@ use std::net::{SocketAddr, ToSocketAddrs};
...
 use jsonrpsee_core::{async_trait, Cow};
-use soketto::connection;
 use soketto::handshake::client::{Client as WsHandshakeClient, ServerResponse};
+use soketto::{connection, Incoming};
 use stream::EitherStream;
 use thiserror::Error;
 use tokio::net::TcpStream;
 
 pub use http::{uri::InvalidUri, Uri};
+use soketto::data::ByteSlice125;
```
niklasad1, line 44: move this import up a little bit to the other "ordinary" imports, these ones is only `pub use`

core/src/client/async_client/mod.rs
```diff
@@ -378,19 +393,207 @@ impl SubscriptionClientT for Client {
 	}
 }
 
+/// Handle backend messages.
+///
+/// Returns error if the main background loop should be terminated.
+async fn handle_backend_messages<S: TransportSenderT, R: TransportReceiverT>(
```
niklasad1, line 399: nice :+1:

core/src/client/async_client/mod.rs
```diff
@@ -378,19 +393,207 @@ impl SubscriptionClientT for Client {
...
+
+/// Handle frontend messages.
+///
+/// Returns `true` if the main background loop should be terminated.
+async fn handle_frontend_messages<S: TransportSenderT>(
+	message: Option<FrontToBack>,
+	manager: &mut RequestManager,
+	sender: &mut S,
+	max_notifs_per_subscription: usize,
+) -> bool {
```
niklasad1, line 487: I like `Result<(), Error>` better `bool` for readability.

Is there reason why just to return `bool` here?

core/src/client/async_client/mod.rs
```diff
@@ -378,19 +393,207 @@ impl SubscriptionClientT for Client {
...
+			}
+		}
+		// User dropped the notificationHandler for this method
+		Some(FrontToBack::UnregisterNotification(method)) => {
+			tracing::trace!("[backend] unregistering notification handler: {:?}", method);
+			let _ = manager.remove_notification_handler(method);
+		}
+	}
+
+	return false;
```
niklasad1, line 575:
```suggestion
	false
```

client/transport/src/ws/mod.rs
```diff
@@ -206,11 +217,19 @@ impl TransportReceiverT for Receiver {
...
 		let mut message = Vec::new();
-		self.inner.receive_data(&mut message).await?;
-		let s = String::from_utf8(message).expect("Found invalid UTF-8");
-		Ok(s)
+
+		loop {
+			let recv = self.inner.receive(&mut message).await?;
+
+			if let Incoming::Data(_) = recv {
+				let s = String::from_utf8(message).expect("Found invalid UTF-8");
```
niklasad1, line 227: I think this expect is okay because we send only messages as `text` but it's better that we just return an error here instead as more transports than `WebSocket` has been added.

This is probably my own fault but would be great if that is fixed too.

core/src/client/mod.rs
```diff
@@ -132,12 +132,25 @@ pub trait TransportSenderT: MaybeSend + 'static {
...
 		Ok(())
 	}
 }
 
+/// Message type received from the RPC server.
+/// It can either be plain `String` data, or a `Pong` reply to a previously submitted `Ping`.
+#[derive(Debug)]
+pub enum ReceivedMessage {
+	/// Incoming packet contains plain `String` data.
+   Data(String),
```
niklasad1, line 149: Perhaps add `Binary(Vec<u8>)` as well or just wrap https://docs.rs/soketto/latest/soketto/data/enum.Data.html in `Data`

core/src/client/async_client/mod.rs
```diff
@@ -21,9 +20,11 @@ use async_lock::Mutex;
 use async_trait::async_trait;
 use futures_channel::{mpsc, oneshot};
 use futures_timer::Delay;
-use futures_util::future::{self, Either};
+use futures_util::future::{self, Either, Fuse};
+use futures_util::select;
```
niklasad1, line 24: you have to add the `async-await-macro` feature on `future_util` in the `async-wasm-client` for that to work.

client/transport/src/ws/mod.rs
```diff
@@ -195,6 +196,16 @@ impl TransportSenderT for Sender {
 		Ok(())
 	}
 
+	/// Sends out a ping request. Returns a `Future` that finishes when the request has been
+	/// successfully sent.
+	async fn send_ping(&mut self, data: &[u8]) -> Result<(), Self::Error> {
```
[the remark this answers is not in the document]
lexnv: The WebSocket would define a custom additional data that would be sent along with the ping, and to be received back with the `pong`. (`Socketto` limits this to 125 bytes).
It may be useful for cases where the `pong` data is different from what we provided in the `ping` request, but I believe those cases to be improbable. Another use-case might arise from submitting multiple pings, where we submit `ping-1 ping-2 ping-3` and the server can reply (as per what I understand from the spec) directly to a newer message (ie - `ping-2` or `ping-3`). 
But implementing such logic seems overkill for our needs.

As we don't send any data, I've removed entirely the `data` section 😄

core/src/client/async_client/mod.rs
lexnv: It seems that `select` depends on the `std` feature, which I believe we don't want as default (https://github.com/rust-lang/futures-rs/blob/master/futures-util/src/async_await/select_mod.rs#L306).

Have switched to `select_biased` which would go through the futures in order we have them defined (instead of picking a random path)

niklasad1 (commented):
cool, just fix the `web transport` it needs to have the new changes to the `send trait` with the ping message that you have introduced.

jsonrpsee/Cargo.toml
```diff
@@ -30,7 +30,7 @@ async-client = ["jsonrpsee-core/async-client"]
 http-client = ["jsonrpsee-http-client", "jsonrpsee-types", "jsonrpsee-core"]
 http-server = ["jsonrpsee-http-server", "jsonrpsee-types", "jsonrpsee-core"]
 wasm-client = ["jsonrpsee-wasm-client", "jsonrpsee-types", "jsonrpsee-core"]
-ws-client = ["jsonrpsee-ws-client", "jsonrpsee-types", "jsonrpsee-core"]
+ws-client = ["jsonrpsee-ws-client", "jsonrpsee-types", "jsonrpsee-core", "jsonrpsee-core/async-wasm-client"]
```
niklasad1, line 33: hmm, this shouldn't be needed?

if needed it should be "jsonrpsee-core/async-client"

niklasad1 (approved):
LGTM, nice work

client/transport/src/web.rs
```diff
@@ -52,6 +52,11 @@ impl TransportSenderT for Sender {
 		Ok(())
 	}
 
+	async fn send_ping(&mut self) -> Result<(), Self::Error> {
+		tracing::trace!("send ping - not implemented for wasm");
+		Ok(())
```
jsdw, line 57: It feels like if you call this from wasm, it should result in an error, since calling it from wasm is pointless?

core/src/client/mod.rs
```diff
@@ -132,12 +132,27 @@ pub trait TransportSenderT: MaybeSend + 'static {
...
```
jsdw, line 149: Is there a good reason for this being called `Data` and not something like `Text`? For me, the difference between Data and Bytes is confusing :)

wasm-tests/tests/wasm.rs
```diff
@@ -26,9 +26,12 @@ async fn wasm_ws_transport_works() {
 	let exp = r#"{"jsonrpc":"2.0","result":"Substrate Node","id":1}"#;
 
 	tx.send(req.to_string()).await.unwrap();
-	let rp = rx.receive().await.unwrap();
+	let rp: ReceivedMessage = rx.receive().await.unwrap();
 
-	assert_eq!(exp, &rp);
+	match rp {
+		ReceivedMessage::Data(str) => assert_eq!(exp, &str),
```
jsdw, line 32:
```suggestion
		ReceivedMessage::Data(s) => assert_eq!(exp, &s),
```

(nit: just because str is a type :))

core/src/client/async_client/mod.rs
```diff
@@ -403,163 +620,46 @@ async fn background_task<S, R>(
...
-					Err(e) => {
-						tracing::warn!("[backend]: client request failed: {:?}", e);
-						let _ = request.send_back.map(|s| s.send(Err(Error::Transport(e.into()))));
-					}
-				}
-			}
+		let mut submit_ping = if let Some(duration) = ping_interval {
+			Delay::new(duration).fuse()
+		} else {
+			Fuse::<Delay>::terminated()
```
jsdw, line 626: Perhaps worth a comment like "Polling this will always return pending"? (it's not super obvious from the name to me!)

core/src/client/async_client/mod.rs
```diff
@@ -403,163 +620,46 @@ async fn background_task<S, R>(
...
-
-				if manager.insert_notification_handler(&reg.method, subscribe_tx).is_ok() {
-					let _ = reg.send_back.send(Ok((subscribe_rx, reg.method)));
-				} else {
-					let _ = reg.send_back.send(Err(Error::MethodAlreadyRegistered(reg.method)));
+			},
+			_ = submit_ping => {
+				// Ping was already submitted.
+				// No activity from frontend, backend (replies or pong) for a duration of `ping_interval`.
+				if ping_submitted {
```
jsdw, line 649: I can't work out why `ping_submitted` is needed?

as far as I can see, if a message is sent or received, the select will resolve and we'll run the function again and start a brand new delay etc, so we'll already wait for the ping interval after any other event happens?

jsdw: Ooh, so you set `ping_submitted` to true and then expect it to be set back to false with a pong message before we would want to do another ping; got it!

So if the ping interval is really short, we'll also expect pongs back quicker. It may be worth adding in the docs for setting a ping interval that the client will expect a pong to be returned within that interval else it'll emit an error.

client/transport/src/web.rs
niklasad1: yeah return an error here is better :1st_place_medal:

jsdw:
Nice work; left a couple of minor comments is all!

core/src/client/async_client/mod.rs
```diff
@@ -115,6 +117,23 @@ impl ClientBuilder {
 		self
 	}
 
+	/// Set the interval at which pings frames are submitted (disabled by default).
+	///
+	/// Periodically submitting pings at a defined interval has mainly two benefits:
+	///  - Directly, it acts as a "keep-alive" alternative in the WebSocket world.
+	///  - Indirectly by inspecting trace logs, ensures that the endpoint is still responding to messages.
```
TarikGul, line 124:
```suggestion
	///  - Indirectly by inspecting trace logs, it ensures that the endpoint is still responding to messages.
```

core/src/client/async_client/mod.rs
```diff
@@ -378,13 +398,202 @@ impl SubscriptionClientT for Client {
 	}
 }
 
+/// Handle backend messages.
+///
+/// Returns error if the main background loop should be terminated.
```
TarikGul, line 403:
```suggestion
/// Returns an error if the main background loop should be terminated.
```

core/src/client/async_client/mod.rs
```diff
@@ -378,13 +398,202 @@ impl SubscriptionClientT for Client {
...
+			return Err(Error::Custom("WebSocket receiver dropped".into()));
+		}
+	}
+
+	return Ok(());
+}
+
+/// Handle frontend messages.
+///
+/// Returns error if the main background loop should be terminated.
```
TarikGul, line 493:
```suggestion
/// Returns an error if the main background loop should be terminated.
```

core/src/client/async_client/mod.rs
```diff
@@ -378,13 +398,202 @@ impl SubscriptionClientT for Client {
...
+async fn handle_backend_messages<S: TransportSenderT, R: TransportReceiverT>(
+	message: Option<Result<ReceivedMessage, R::Error>>,
+	manager: &mut RequestManager,
+	sender: &mut S,
+	max_notifs_per_subscription: usize,
+) -> Result<(), Error> {
+
+	// Handle raw messages of form `ReceivedMessage::Bytes` (Vec<u8>) or ReceivedMessage::Data` (String).
+	async fn handle_recv_message<S: TransportSenderT>(
```
TarikGul, line 412: So I am under the impression the reason this is nested is this `fn` is only used locally within the scope of `handle_backend_messages`? I quite like it actually as it distinquishes the difference between handling frontend messages and backend message. 👍  Nice cleanup from it earlier being inside of the `fn background_task`. It's way easier to read/digest.

core/src/client/async_client/mod.rs
```diff
@@ -378,13 +398,202 @@ impl SubscriptionClientT for Client {
...
+	match message {
+		// User dropped the sender side of the channel.
+		// There is nothing to do just terminate.
+		None => {
+			return Err(Error::Custom("[backend]: frontend dropped; terminate client".into()));
+		}
+
+		Some(FrontToBack::Batch(batch)) => {
+			tracing::trace!("[backend]: client prepares to send batch request: {:?}", batch.raw);
+			// NOTE(niklasad1): annoying allocation.
```
TarikGul, line 509: @niklasad1 DQ: Just wondering, I've seen this before and have wanted to ask, what specifically about the clone allocation is annoying here?

core/src/client/async_client/mod.rs
```diff
@@ -395,172 +604,59 @@ async fn background_task<S, R>(
...
-			// User called `notification` on the front-end
-			Either::Left((Some(FrontToBack::Notification(notif)), _)) => {
-				tracing::trace!("[backend]: client prepares to send notification: {:?}", notif);
-				if let Err(e) = sender.send(notif).await {
-					tracing::warn!("[backend]: client notif failed: {:?}", e);
-				}
-			}
+	// Place frontend and backend messages into their own select.
+	// This implies that either messages are received (both front or backend),
+	// or submit ping timer expires (if provided).
```
TarikGul, line 611:
```suggestion
	// or the submitted ping timer expires (if provided).
```

I was confused at first when reading this in my head but this is how it sounded when i read it out loud.

core/src/client/async_client/mod.rs
```diff
@@ -395,172 +604,59 @@ async fn background_task<S, R>(
...
-			Either::Left((Some(FrontToBack::RegisterNotification(reg)), _)) => {
-				tracing::trace!("[backend] registering notification handler: {:?}", reg.method);
-				let (subscribe_tx, subscribe_rx) = mpsc::channel(max_notifs_per_subscription);
-
-				if manager.insert_notification_handler(&reg.method, subscribe_tx).is_ok() {
-					let _ = reg.send_back.send(Ok((subscribe_rx, reg.method)));
-				} else {
-					let _ = reg.send_back.send(Err(Error::MethodAlreadyRegistered(reg.method)));
+			// Message received from the backend.
+			Either::Left((Either::Right((backend_value, frontend )), _))=> {
```
TarikGul, line 638:
```suggestion
			Either::Left((Either::Right((backend_value, frontend)), _))=> {
```

TarikGul (approved):
👍  Great Job!

core/src/client/async_client/mod.rs
niklasad1: we can remove that note but it's just the those are `batch ids` are already owned in this scope and so should be possible to borrow that `in theory` but we use the `HashSet entry which takes ownership of that + `RequestManager` doesn't allow borrows for simplicity.

core/src/client/async_client/mod.rs
niklasad1, line 124: I'm not super excited about having the pongs to be logged as `trace` because the and trace` logs are so damn verbose.

I think the `pongs` should be `info` or `debug`, but I'll tidy up the logs in another PR.

niklasad1 (changes requested):
blocking this because of the `trace logging thingy`

core/Cargo.toml
```diff
@@ -62,6 +63,7 @@ async-wasm-client = [
 	"rustc-hash/std",
 	"futures-timer/wasm-bindgen",
 	"tracing",
+	"futures-util/async-await-macro",
```
niklasad1, line 66:
```suggestion
```

core/Cargo.toml
```diff
@@ -54,6 +54,7 @@ async-client = [
 	"tokio/sync",
 	"tracing",
 	"futures-timer",
+	"futures-util/async-await-macro",
```
niklasad1, line 57:
```suggestion
```

client/transport/src/ws/mod.rs
```diff
@@ -195,6 +197,20 @@ impl TransportSenderT for Sender {
 		Ok(())
 	}
 
+	/// Sends out a ping request. Returns a `Future` that finishes when the request has been
+	/// successfully sent.
```
jsdw, line 201: Since we've been talking about cancel safety a bit, and not really relaetd to this PR, but do you reckon it is worth adding an issue to add docs to functions to note when they aren't cancel safe @niklasad1?

jsdw (approved):

client/transport/src/ws/mod.rs
niklasad1: sounds like a good idea.

niklasad1 (approved):
nice work, forgot about this one

Issue #773, opened 2022-05-18
"Invalid JSON-RPC response" when calling u64 rpc method.


spapinistarkware:
Code example:
```
#[cfg(test)]
mod tests {

#[rpc(server, client)]
pub trait Rpc {
    #[method(name = "foo")]
    async fn foo(&self) -> Result<u64, Error>;
}
#[async_trait]
impl RpcServer for RpcServerImpl {
    async fn foo(&self) -> Result<u64, Error> {
        Ok(0)
    }
}
#[tokio::test]
async fn test_foo() {
    let module = RpcServerImpl.into_rpc();
    let result: String = module.call("foo", EmptyParams::new()).await.unwrap();
}

}
```

I'm trying a minimal example that returns a u64.
This is the error:
```
thread 'gateway::tests::test_foo' panicked at 'internal error: entered unreachable code: Invalid JSON-RPC response is not possible using jsonrpsee; this is bug please file an issue', /home/spapini/.cargo/registry/src/github.com-1ecc6299db9ec823/jsonrpsee-core-0.13.1/src/server/rpc_module.rs:366:9
```

What am I doing wrong?

niklasad1:
You tried to call a method which as not been registered.

`let result: String = module.call("foo", EmptyParams::new()).await.unwrap();`

should work?!

We could fix a better error message though..

spapinistarkware:
Sorry, that was a typo, I fixed it above. The error is not unknown method, so, this ain't it:)

PR #774, opened 2022-05-18
rpc_module: remove expect in `fn call`

niklasad1:
This method is mainly used for testing but the `expect` there was bad and confusing which this fixes.

See #773 for further information.

### DIFF
--- a/core/src/server/rpc_module.rs
+++ b/core/src/server/rpc_module.rs
@@ -355,15 +355,17 @@ impl Methods {
 		tracing::trace!("[Methods::call] Calling method: {:?}, params: {:?}", method, params);
 		let (resp, _, _) = self.inner_call(req).await;
 
-		if let Ok(res) = serde_json::from_str::<Response<T>>(&resp) {
-			return Ok(res.result);
-		}
-
-		if let Ok(err) = serde_json::from_str::<ErrorResponse>(&resp) {
-			return Err(Error::Call(CallError::Custom(err.error_object().clone().into_owned())));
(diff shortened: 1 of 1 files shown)


core/src/server/rpc_module.rs
```diff
@@ -355,15 +355,17 @@ impl Methods {
...
-		if let Ok(res) = serde_json::from_str::<Response<T>>(&resp) {
-			return Ok(res.result);
-		}
-
-		if let Ok(err) = serde_json::from_str::<ErrorResponse>(&resp) {
-			return Err(Error::Call(CallError::Custom(err.error_object().clone().into_owned())));
-		}
-
-		unreachable!("Invalid JSON-RPC response is not possible using jsonrpsee; this is bug please file an issue");
+		let res = match serde_json::from_str::<Response<T>>(&resp) {
```
TarikGul, line 358: Clean change 👍

TarikGul (approved):

core/src/server/rpc_module.rs
niklasad1: just annoying to have to do that `let res binding` but otherwise rustc complains about the lifetime issues FYI

jsdw (approved):

lexnv (approved):

core/src/server/rpc_module.rs
niklasad1: me too but might confuse rustc that `resp` is borrowed twice ^^

PR #781, opened 2022-05-24
fix(ws server): support `*` in host and origin filtering

### LINKED ISSUE #780: [ws server]: host-filtering with `wildcard` as port number doesn't work
Host-filtering with wildcard in the port field doesn't work.

```rust
	let server = WsServerBuilder::default()
		.set_allowed_hosts(["http://localhost:*", "http://127.0.0.1:*"])
		.unwrap()
		.build("127.0.0.1:0")
		.await
		.unwrap();

	let client = Arc::new(client);

	let r = client.request::<String>("say_hello", None).await;
        
        // Fails with 403 because the entire address is compared such as as `http://localhost:31881 !=  http://localhost:*`
	assert!(r.is_ok());
```

I have checked the HTTP server and it actually supports wildcards.

Also see https://github.com/paritytech/substrate/issues/11480 for further information

niklasad1:
Fixes https://github.com/paritytech/jsonrpsee/issues/780 by sharing the access control stuff between the servers.

TLDR: we didn't support `*` when compared incoming `hosts` with some concrete port number such as `localhost:9993` in WS-server.


core/src/error.rs
```diff
@@ -105,9 +105,12 @@ pub enum Error {
 	/// Attempted to stop server that is already stopped.
 	#[error("Attempted to stop server that is already stopped")]
 	AlreadyStopped,
-	/// List passed into `set_allowed_origins` was empty
+	/// List passed into access control based `HTTP` headers.
```
niklasad1, line 108:
```suggestion
	/// List passed into access control based on HTTP header verification.
```

core/src/error.rs
```diff
@@ -105,9 +105,12 @@ pub enum Error {
 	/// Attempted to stop server that is already stopped.
 	#[error("Attempted to stop server that is already stopped")]
 	AlreadyStopped,
-	/// List passed into `set_allowed_origins` was empty
+	/// List passed into access control based on HTTP header verification.
 	#[error("Must set at least one allowed value for the {0} header")]
 	EmptyAllowList(&'static str),
+	/// Access control based on `HTTP` header failed.
```
niklasad1, line 111:
```suggestion
	/// Access control of HTTP header verification failed.
```

http-server/src/lib.rs
```diff
@@ -30,17 +30,13 @@
 //!
 //! `jsonrpsee-http-server` is a [JSON RPC](https://www.jsonrpc.org/specification) HTTPS server library that's is built for `async/await`.
 
-mod access_control;
+//mod access_control;
```
niklasad1, line 33:
```suggestion
```

examples/examples/cors_server.rs
```diff
@@ -65,7 +68,7 @@ async fn main() -> anyhow::Result<()> {
 }
 
 async fn run_server() -> anyhow::Result<(SocketAddr, HttpServerHandle)> {
-	let acl = AccessControlBuilder::new().allow_all_headers().allow_all_origins().allow_all_hosts().build();
+	let acl = AccessControlBuilder::default().allow_all_headers().allow_all_origins().allow_all_hosts().build();
```
niklasad1, line 71:
```suggestion
	let acl = AccessControlBuilder::new().allow_all_headers().allow_all_origins().allow_all_hosts().build();
```

core/src/server/access_control/host.rs
```diff
@@ -139,47 +140,31 @@ impl std::ops::Deref for Host {
...
 }
 
-impl<T> From<Option<Vec<T>>> for DomainsValidation<T> {
-	fn from(other: Option<Vec<T>>) -> Self {
-		match other {
-			Some(list) => DomainsValidation::AllowOnly(list),
-			None => DomainsValidation::Disabled,
+impl AllowHosts {
+	/// Verify a host.
+	pub fn verify(&self, value: &str) -> Result<(), Error> {
```
TarikGul, line 154: Ahh okay nice refactor. So were basically replacing `DomainsValidation`, and `is_host_valid` with this `AllowHosts` impl.

core/src/server/access_control/mod.rs
```diff
@@ -0,0 +1,167 @@
...
+	{
+		let allowed_headers: Vec<String> = list.into_iter().map(Into::into).collect();
+		if allowed_headers.is_empty() {
+			return Err(Error::EmptyAllowList("Header"));
+		}
+		self.allowed_headers = AllowHeaders::Only(allowed_headers);
+		Ok(self)
+	}
+
+	/// Build.
```
TarikGul, line 159: IMO not a necessary comment.

http-server/src/server.rs
```diff
@@ -358,7 +360,7 @@ pub struct Server<M = ()> {
 	max_response_body_size: u32,
 	/// Whether batch requests are supported by this server or not.
 	batch_requests_supported: bool,
-	/// Access control
+	/// Access control.
 	access_control: AccessControl,
 	/// Tracker for currently used resources on the server
```
TarikGul, line 365:
```suggestion
	/// Tracker for currently used resources on the server.
```

TarikGul, line 365:
```suggestion
	/// Tracker for currently used resources on the server.
```

http-server/src/server.rs
```diff
@@ -389,36 +391,60 @@ impl<M: Middleware> Server<M> {
...
 					let health_api = health_api.clone();
 
 					// Run some validation on the http request, then read the body and try to deserialize it into one of
 					// two cases: a single RPC request or a batch of RPC requests.
 					async move {
-						if let Err(e) = access_control_is_valid(&access_control, &request) {
-							return Ok::<_, HyperError>(e);
+						let host = match http_helpers::read_header_value(request.headers(), "host") {
+							Some(origin) => origin,
+							None => return Ok(malformed()),
```
TarikGul, line 412: is `malformed()` returning `Response<Body>`?

http-server/src/server.rs
```diff
@@ -389,36 +391,60 @@ impl<M: Middleware> Server<M> {
...
+						};
+						let maybe_origin = http_helpers::read_header_value(request.headers(), "origin");
+						let headers = request.headers().keys().map(|name| name.as_str());
+						let cors_headers =
+							http_helpers::read_header_values(request.headers(), "access-control-request-headers")
+								.filter_map(|val| val.to_str().ok())
+								.flat_map(|val| val.split(", "))
```
TarikGul, line 419: Can you point me to where I can find the structure of the header you are working with here, I am having a hard time picturing what is happening with the splitting of the strings, then chars.

ws-server/src/server.rs
```diff
@@ -244,8 +245,15 @@ where
...
-				let origin_check = cfg.allowed_origins.verify("Origin", req.headers().origin);
+
+				let host = std::str::from_utf8(req.headers().host)
+					.map_err(|_e| Error::Custom("Host header must be valid UTF-8".into()))?;
+				let origin = req.headers().origin.and_then(|h| std::str::from_utf8(h).ok());
+
+				let host_check = cfg.access_control.verify_host(host);
+				let origin_check = cfg.access_control.verify_origin(origin, host);
+
+				// TODO: "access-control-request-headers" not possible to fetch via soketto.
```
TarikGul, line 256: Is this something worthy of making an issue for in soketto?

TarikGul (commented):
Overall looks great. Just a few questions.

lexnv (approved):

http-server/src/server.rs
niklasad1: Yes, https://github.com/paritytech/jsonrpsee/blob/master/http-server/src/response.rs#L83-#L89

ws-server/src/server.rs
niklasad1: I was wrong, this header only applicable for HTTP Option requests when doing preflight requests.

The websocket  HTTP upgrade request must be `GET` AFAIU.

TarikGul (approved):

core/src/server/access_control/mod.rs
```diff
@@ -0,0 +1,167 @@
...
+	pub fn verify_headers<T, I, II>(&self, header_names: I, cors_headers: II) -> Result<(), Error>
+	where
+		T: AsRef<str>,
+		I: Iterator<Item = T>,
+		II: Iterator<Item = T>,
+	{
+		let header = cors::get_cors_allow_headers(header_names, cors_headers, &self.allowed_headers, |name| name);
+
+		if let cors::AllowCors::Invalid = header {
+			Err(Error::HttpHeaderRejected("access-control-request-headers", "<too long to be logged>".into()))
```
jsdw, line 55: Why "too long to be logged"?

niklasad1: I think I was just lazy because the actual values were moved into the function above.

core/src/server/access_control/host.rs
```diff
@@ -188,43 +173,34 @@ mod tests {
...
-		let valid = is_host_valid(Some("parity.io"), &AllowHosts::Only(vec![]));
-		assert!(!valid);
+		assert!((AllowHosts::Only(vec![].into())).verify("parity.io").is_err());
 	}
 
 	#[test]
 	fn should_accept_if_on_the_list() {
-		let valid = is_host_valid(Some("parity.io"), &AllowHosts::Only(vec!["parity.io".into()]));
-		assert!(valid);
+		assert!((AllowHosts::Only(vec!["parity.io".into()].into())).verify("parity.io").is_ok());
```
jsdw, line 194: if it's on the list without a port, like this, would `.verify("parity.io:1234")` work? In other words, is any port valid if no port is given in the allow list?

core/src/server/access_control/cors.rs
```diff
@@ -221,9 +221,9 @@ impl<T> From<AllowCors<T>> for Option<T> {
 /// Returns correct CORS header (if any) given list of allowed origins and current origin.
 pub(crate) fn get_cors_allow_origin(
```
jsdw, line 222: Nit: could this be a little stricter is `pub(super)`?

core/src/server/access_control/cors.rs
```diff
@@ -239,38 +239,38 @@ pub(crate) fn get_cors_allow_origin(
...
-	cors_allow_headers: &AccessControlAllowHeaders,
+	cors_allow_headers: &AllowHeaders,
 	to_result: F,
 ) -> AllowCors<Vec<O>> {
 	// Check if the header fields which were sent in the request are allowed
-	if let AccessControlAllowHeaders::Only(only) = cors_allow_headers {
+	if let AllowHeaders::Only(only) = cors_allow_headers {
 		let are_all_allowed = headers.all(|header| {
 			let name = &Ascii::new(header.as_ref());
 			only.iter().any(|h| Ascii::new(&*h) == name) || ALWAYS_ALLOWED_HEADERS.contains(name)
```
jsdw, line 276: I'm struggling to understand `ALWAYS_ALLOWED_HEADERS`; I think that we allow any headers specified in AllowHeaders plus any in `ALWAYS_ALLOWED_HEADERS`.

jsdw:
Love that the access control stuff is now consistent across HTTP and WS (as well as the *'s working!). Just a few questions/nits :)

jsdw:
Love that the access control stuff is now consistent across HTTP and WS (as well as the *'s working!). Just a few questions/nits :)

core/src/server/access_control/mod.rs
jsdw: Hrm yeah, I mulled over this a bit and I can see that it'd be a bit of a pain!

core/src/server/access_control/host.rs
niklasad1: no it will be denied unless the port is whitelisted or `*` is used but I will add a test to make it clear.

http-server/src/server.rs
```diff
@@ -389,36 +391,60 @@ impl<M: Middleware> Server<M> {
...
+								.flat_map(|val| val.split(','));
```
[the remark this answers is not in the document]
niklasad1: I was curious myself and it's because you might have `val1 ,val2` or `val1,val2` splitting on "," will yield an empty string then

core/src/http_helpers.rs
```diff
@@ -99,13 +99,25 @@ pub fn read_header_value<'a>(headers: &'a hyper::header::HeaderMap, header_name:
...
+}
+
+/// Get the header values from the `access-control-request-headers` header.
+pub fn get_cors_request_headers<'a>(headers: &'a hyper::header::HeaderMap) -> impl Iterator<Item = &str> {
+	const ACCESS_CONTROL_ALLOW_HEADERS: &str = "access-control-request-headers";
+
+	read_header_values(headers, ACCESS_CONTROL_ALLOW_HEADERS)
+		.iter()
+		.filter_map(|val| val.to_str().ok())
+		// NOTE(niklasad1): we perform split twice to cover both `val1,val2` and `val1, val2`.
```
jsdw, line 113: since presumably any number of spaces could exist, how about something like:

```
.flat_map(|val| val.split(","))
.map(|val| val.trim())
```

core/src/server/access_control/cors.rs
niklasad1: Ok, so your suggestion is to only whitelist `Accept`, `Accept-Language`, `Content-Type` and `Content-Language`?

jsdw (approved):
Happy to see the whitelisted header stuff be mulled over in a separate PR!

PR #782, opened 2022-05-24
Implement `ping-pong` for WebSocket server

lexnv:
This PR exposes the ping-pong interface of the WebSocket server.

This is the counterpart of the client #772, requested by #738 .

### Testing Done
- modified `examples/ws` into two binaries, one for the server, one for the client; started them with trace logs

```bash
2022-05-27T10:24:20.552095Z DEBUG jsonrpsee_ws_server::server: send ping
2022-05-27T10:24:20.552336Z TRACE soketto::connection: bd133ba0: send: (Ping (fin 1) (rsv 000) (mask (0 0)) (len 0))
2022-05-27T10:24:20.552676Z TRACE soketto::connection: bd133ba0: Sender flushing connection
2022-05-27T10:24:20.556269Z TRACE soketto: read 2 bytes
2022-05-27T10:24:20.556340Z TRACE soketto: read 4 bytes
2022-05-27T10:24:20.556374Z TRACE soketto::connection: bd133ba0: recv: (Pong (fin 1) (rsv 000) (mask (1 fba99e45)) (len 0))
2022-05-27T10:24:20.556414Z DEBUG jsonrpsee_ws_server::server: recv pong
```

### DIFF
--- a/ws-server/src/server.rs
+++ b/ws-server/src/server.rs
@@ -342,7 +342,6 @@ async fn background_task(
 						tracing::warn!("WS send ping error: {}; terminate connection", err);
 						break;
 					}
-
 					rx_item = next_rx;
 				}
 			}
@@ -365,7 +364,17 @@ async fn background_task(
 
 		{
 			// Need the extra scope to drop this pinned future and reclaim access to `data`
-			let receive = receiver.receive_data(&mut data);
+			let receive = async {
+				// Identical loop to `soketto::receive_data` with debug logs for `Pong` frames.
+				loop {
+					match receiver.receive(&mut data).await? {
+						soketto::Incoming::Data(d) => break Ok(d),
+						soketto::Incoming::Pong(_) => tracing::debug!("recv pong"),
+						_ => continue,
+					}
+				}
+			};
+
 			tokio::pin!(receive);
 
 			if let Err(err) = method_executors.select_with(Monitored::new(receive, &stop_server)).await {
@@ -976,7 +985,7 @@ async fn send_ws_message(
 }
 
 async fn send_ws_ping(sender: &mut Sender<BufReader<BufWriter<Compat<TcpStream>>>>) -> Result<(), Error> {
-	tracing::debug!("submit ping");
+	tracing::debug!("send ping");
 	// Submit empty slice as "optional" parameter.
 	let slice: &[u8] = &[];
 	// Byte slice fails if the provided slice is larger than 125 bytes.
--- a/ws-server/src/server.rs
+++ b/ws-server/src/server.rs
@@ -35,8 +35,10 @@ use crate::future::{FutureDriver, ServerHandle, StopMonitor};
 use crate::types::error::{ErrorCode, ErrorObject, BATCHES_NOT_SUPPORTED_CODE, BATCHES_NOT_SUPPORTED_MSG};
 use crate::types::{Id, Request};
 use futures_channel::mpsc;
-use futures_util::future::{join_all, FutureExt};
+use futures_timer::Delay;
+use futures_util::future::{join_all, Either, FutureExt};
 use futures_util::io::{BufReader, BufWriter};
+use futures_util::pin_mut;
 use futures_util::stream::StreamExt;
 use jsonrpsee_core::id_providers::RandomIntegerIdProvider;
 use jsonrpsee_core::middleware::Middleware;
@@ -51,7 +53,6 @@ use soketto::data::ByteSlice125;
 use soketto::handshake::{server::Response, Server as SokettoServer};
 use soketto::Sender;
 use tokio::net::{TcpListener, TcpStream, ToSocketAddrs};
-use tokio::select;
 use tokio_util::compat::{Compat, TokioAsyncReadCompatExt};
 
 /// Default maximum connections allowed.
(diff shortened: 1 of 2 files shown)


ws-server/src/server.rs
```diff
@@ -309,19 +315,42 @@ async fn background_task(
...
+							break;
+						}
+					} else {
+						break;
+					}
+				}
+
+				_ = submit_ping.tick() => {
+					// Ping was already submitted.
+					if ping_submitted_tx.load(Ordering::Relaxed) {
```
niklasad1, line 342: looks to me that we should use stricter ordering here,

as `store's` from can re-ordered arbitrary with `Relaxed`, so I think it's possible that the will be set "to the wrong value" i.e, even if the ping was sent it could be overwritten.

Please use `release-acquire or seqcst` for this unless you can prove that I'm wrong :)

niklasad1, line 342: looks to me that we should use stricter ordering here,

as `store's` from can re-ordered arbitrary with `Relaxed`, so I think it's possible that the will be set "to the wrong value" i.e, even if the ping was sent it could be overwritten.

Please use `release-acquire or seqcst` for this unless you can prove that I'm wrong :)

ws-server/src/server.rs
```diff
@@ -309,19 +315,42 @@ async fn background_task(
 	let stop_server2 = stop_server.clone();
 	let sink = MethodSink::new_with_limit(tx, max_response_body_size);
 
+	let mut submit_ping = tokio::time::interval(ping_interval);
+	let ping_submitted_tx = Arc::new(AtomicBool::new(false));
```
niklasad1, line 319: Do we really have to kill the connection if a pong was not received between the provided window?

I'm a little terrified of having more boolean flags that must be set/cleared and those are easy mess up on and so on when the refactoring the code.

lexnv: I don't think we explicitly need to handle `Pong`s if we only care about keeping the connection alive.

That would also simplify the code quite a bit server-side 😄

ws-server/src/server.rs
```diff
@@ -309,19 +315,32 @@ async fn background_task(
...
 	// Send results back to the client.
 	tokio::spawn(async move {
 		while !stop_server2.shutdown_requested() {
-			if let Some(response) = rx.next().await {
-				// If websocket message send fail then terminate the connection.
-				if let Err(err) = send_ws_message(&mut sender, response).await {
-					tracing::warn!("WS send error: {}; terminate connection", err);
-					break;
+			select! {
+				rx_value = rx.next() => {
```
niklasad1, line 326: this is cancel-safe?

niklasad1: also I think you could do 

```
select! {
   Some(rx) = rx.next() => { }
   None = rx.next() => {}
   _ = submit_ping.tick() => {}
}
```

to get rid of that extra `else { break }`  branch.

ws-server/src/server.rs
niklasad1: cool, I guess we could add another issue for that if anyone wants kill their connection once the ping is not ACK:ed by pong with the period but it could also be a footgun that the ping period is too short with a slow connection.

ws-server/src/server.rs
```diff
@@ -309,19 +314,28 @@ async fn background_task(
...
 
 	// Send results back to the client.
 	tokio::spawn(async move {
 		while !stop_server2.shutdown_requested() {
-			if let Some(response) = rx.next().await {
-				// If websocket message send fail then terminate the connection.
-				if let Err(err) = send_ws_message(&mut sender, response).await {
-					tracing::warn!("WS send error: {}; terminate connection", err);
-					break;
+			select! {
```
niklasad1, line 324: check if you could replace this `select!` with something that fetches the dropped future have a look https://github.com/paritytech/jsonrpsee/blob/master/core/src/server/rpc_module.rs#L904-#L955 for inspiration :)

:point_up: should be cancel-safe but would be good not use `select!` because it's a footgun.

niklasad1 (approved):
LGTM, but would be good if you could replace the `select!`

lexnv:
To have symmetry with the `client` implementation, added the following debug logs for `Ping` / `Pong` frames:

```
2022-05-27T10:24:20.552095Z DEBUG jsonrpsee_ws_server::server: send ping
...
2022-05-27T10:24:20.556414Z DEBUG jsonrpsee_ws_server::server: recv pong
```

ws-server/src/server.rs
```diff
@@ -313,15 +318,32 @@ async fn background_task(
...
+
 		while !stop_server2.shutdown_requested() {
-			if let Some(response) = rx.next().await {
-				// If websocket message send fail then terminate the connection.
-				if let Err(err) = send_ws_message(&mut sender, response).await {
-					tracing::warn!("WS send error: {}; terminate connection", err);
-					break;
+			// Ensure select is cancel-safe by fetching and storing the `rx_item` that did not finish yet.
+			// Note: Although, this is cancel-safe already, avoid using `select!` macro for future proofing.
+			match futures_util::future::select(rx_item, Delay::new(ping_interval)).await {
```
jsdw, line 327: Note: this `Delay` thing spawns a new background thread to manage timers.

Since we seem to depend on tokio here anyway, why not use the timer from tokio (which may or may not avoid the extra thread, and would def avoid the extra dependency)?

lexnv: Done! Thanks
!

jsdw (approved):
LGTM :)

PR #783, opened 2022-05-25
fix: skip leading whitespace

polachok:
In json, whitespace is not significant, so we shouldn't rely on the first character to be { or |. serde_json parses such json
 just fine.


niklasad1 (approved):
LGTM, thank you

TarikGul (commented):
👍  Thanks for the contribution

TarikGul (approved):

Issue #784, opened 2022-05-26
[http server]: `health_api` uses the entire `jsonrpc response` in the body


niklasad1:
```bash
curl http://mydomaini:9933/health
{"jsonrpc":"2.0","result":{"peers":28,"isSyncing":false,"shouldHavePeers":true},"id":0}
```

As this is only a pure HTTP get request it would be more accurate to just return the result and exclude the jsonrpc related stuff.
In other words the following would be better:

```
{"peers":28,"isSyncing":false,"shouldHavePeers":true}
```

niklasad1:
Closed by #785

PR #785, opened 2022-05-26
fix(http server): omit jsonrpc details in health API

niklasad1:
Closing https://github.com/paritytech/jsonrpsee/issues/784

Ordinary HTTP GET requests doesn't expect the body the be formatted as `JSON-RPC` responses which this fixes i.e, just regular JSON.

### DIFF
--- a/http-server/src/server.rs
+++ b/http-server/src/server.rs
@@ -170,11 +170,20 @@ impl<M> Builder<M> {
 	}
 
 	/// Enable health endpoint.
-	/// Allows you to expose one of the methods under GET /<path> The method will be invoked with no parameters. Error returned from the method will be converted to status 500 response.
-	/// Expects a tuple with (<path>, <rpc-method-name>).
-	pub fn health_api(mut self, path: impl Into<String>, method: impl Into<String>) -> Self {
-		self.health_api = Some(HealthApi { path: path.into(), method: method.into() });
-		self
+	/// Allows you to expose one of the methods under GET /<path> The method will be invoked with no parameters.
+	/// Error returned from the method will be converted to status 500 response.
+	/// Expects a tuple with (</path>, <rpc-method-name>).
+	///
+	/// Fails if the path is missing `/`.
+	pub fn health_api(mut self, path: impl Into<String>, method: impl Into<String>) -> Result<Self, Error> {
+		let path = path.into();
+
+		if !path.starts_with("/") {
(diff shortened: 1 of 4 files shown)


http-server/src/server.rs
```diff
@@ -754,7 +763,14 @@ async fn process_health_request(
 	middleware.on_response(request_start);
 
 	match data {
-		Some(resp) if success => Ok(response::ok_response(resp)),
+		Some(data) if success => {
+			let json: serde_json::Value =
```
niklasad1, line 767: this is super annoying but no way around that with huge refactorings.

http-server/src/server.rs
```diff
@@ -170,11 +170,20 @@ impl<M> Builder<M> {
...
+	/// Error returned from the method will be converted to status 500 response.
+	/// Expects a tuple with (</path>, <rpc-method-name>).
+	///
+	/// Fails if the path is missing `/`.
+	pub fn health_api(mut self, path: impl Into<String>, method: impl Into<String>) -> Result<Self, Error> {
+		let path = path.into();
+
+		if !path.starts_with("/") {
+			return Err(Error::Custom(format!("Health endpoint path must start with `/` to work, got: {}", path)));
+		}
```
jsdw, line 183: Just wondering; why do we care about the path starting with '/'?

niklasad1: Because we have [this check](https://github.com/paritytech/jsonrpsee/blob/master/http-server/src/server.rs#L458), if `/` is missing that will fail and it's quite tricky to debug.

I have already done that myself but more ideally would be `Path type` instead of a String

jsdw: Ah ok gotcha; that makes sense!

http-server/src/server.rs
jsdw: I wonder whether something like this would be any better (likely faster I'd expect):

```rust
#[derive(Deserialize)]
struct RpcPayload<'a> {
    #[serde(borrow)]
    result: &'a serde_json::value::RawValue
}

let json: RpcPayload = serde_json::from_str(&data).expect("valid JSON-RPC response is valid JSON; qed");
Ok(response::ok_response(json.result.to_string())
```

jsdw (approved):
LGTM, just a suggestion on plucking the result out without the intermediate Value stuff :)

lexnv (approved):

PR #786, opened 2022-05-30
Add resource limiting for `Subscriptions`

### LINKED ISSUE #766: [resource limiting]: add support for subscriptions
Currently only `ordinary calls` are supported by the resource limiting but we should support subscriptions as well.

lexnv:
This PR ensures that resource limiting is available for `Subscriptions`.

The `ResourceGuard` wrapper is kept alive for as long as a `SubscriptionSink` is utilized to properly
increment and decrement the internal resources. Otherwise, the resource limiting would have the
same effect as for `sync` methods, which would acquire the resources and drop them immediately
after the function was called.

This implies that subscription registration behaves like normal methods, returning a resource-builder
that can accept limits. To ensure proper feature exposure, the proc-macro generation is extended
to receive resources while generating subscriptions.

While at it, extend the resource limiting tests with the subscription limit.

Closes #766 .


core/src/server/rpc_module.rs
```diff
@@ -689,39 +689,14 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 			return Err(Error::SubscriptionNameConflict(subscribe_method_name.into()));
 		}
 
-		self.methods.verify_method_name(subscribe_method_name)?;
```
niklasad1, line 692: can you explain this? 

these are still required AFAIU, if the method name is already registered we should still fire an error

niklasad1 (commented):
Looks overall good.

I don't remember exactly but this is tricky because often these subscriptions might be event based or periodic.

So once I register a subscription it's not really "static" as it depends on how often notifications are sent out or whether a given state has been triggered.

So the question is really whether the SubscriptionSink also should be part of the resource guard. I reckon that it might quite hard to reason about once one "register resources" on a subscription. 

For example take `state_subscribeStorage` which depends heavily on the number of storage keys which is a parameter to the subscription, it's impossible to give a accurate `static resource limit` without knowing the number of storage keys. I suppose the amount data sent via the `SubscriptionSink` could be one way to determine that.

core/src/server/rpc_module.rs
niklasad1: The problem is that you do check for the subscription and unsubscription separately.
Thus, if only the unsubscription fails we should still deny both.

For example, if the subscription is successful and the unsubscription fails you can't really unsubscribe :thinking:

lexnv: I see now what you mean! Indeed this was living the door open for having the `unsubscriptions` left initialized, have modified to use the previous way of first verifying the method's name. Good catch! Thanks!

ws-server/src/server.rs
```diff
@@ -595,13 +594,12 @@ async fn background_task(
 															close_notify: cn,
 															id_provider: &*id_provider,
 														};
-														callback(id, params, &sink_batch, conn_state)
+														callback(id, params, &sink_batch, conn_state, Some(guard))
```
niklasad1, line 597: @lexnv shouldn't we need to this on the `HTTP server` as well?

niklasad1: right, forgot about :smile:

niklasad1 (approved):

jsdw:
At the retreat it sounded like you guys decided now to add resource limiting for subscriptions owing to various things? Does that render this PR obsolete?

niklasad1:
> At the retreat it sounded like you guys decided now to add resource limiting for subscriptions owing to various things? Does that render this PR obsolete?

Well, we have discussed this and to implement resource limiting on subscriptions properly  with the abstract metrics is just too hard.

The idea is to implement/investigate to do `backpressure` on the `sink` instead and kill resource limiting completely.

jsdw:
Is this PR still valid and wants to be merged then? Wondering whether to look over it or not :)

niklasad1:
We can merge it anyway but as I pointed it out it won't well with "dynamic keys" but as long as we have resource limiting the PR makes sense.

tests/tests/resource_limiting.rs
```diff
@@ -159,6 +217,28 @@ async fn run_tests_on_ws_server(server_addr: SocketAddr, server_handle: WsServer
 	assert!(pass_mem.is_ok());
 	assert_server_busy(fail_mem);
 
+	// 3 CPU units (manually set for subscriptions) per call, the sink is dropped immediately
+	let (pass1, pass2, pass3) = tokio::join!(
+		client.subscribe::<i32>("subscribe_hello", None, "unsubscribe_hello"),
+		client.subscribe::<i32>("subscribe_hello", None, "unsubscribe_hello"),
+		client.subscribe::<i32>("subscribe_hello", None, "unsubscribe_hello"),
+	);
```
jsdw, line 225: Am I right in thinking that even if there were 10 client.subscribes here, they would all be OK because something is dropped immediately, and there is no race condition where that might not always be true?

jsdw: (if so, could the comment be tweaked to say something like "the sink is dropped immediately, so no resources are claimed and all of the calls succeed"?"

ws-server/src/tests.rs
```diff
@@ -21,7 +21,7 @@
 // SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
 // CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
 // OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR
-// IN background_task WITH THE SOFTWARE OR THE USE OR OTHER
+// IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
```
jsdw, line 24: :D

jsdw (commented):
Looks great to me offhand! I do think that having a consistent resource limiting API across everything is good, and at the very least it allows subscriptions to say "this is the minimum amount of CPU/memory I need available in order to run", even if some subscriptions may use more (it reminds me of resources in kubernetes pods, where you can't know how much they could use, but providing some idea is better than not at all).

Given that you can have a "SUB" resource as in thexe examples, and use it to limit the max subscriptions, I wonder whether this will eventually lead to the explicit "max subscriptions" feature being removed (or at least rendered less useful, since this API is more powerful)?

jsdw (approved):

tests/tests/resource_limiting.rs
lexnv: I believe that to be the case for this particular test.

We submit all the `subscribe` requests from one client to the server. The server handles each client on a dedicated `background_task` task. That means all our submitted `subscribe` requests get handled serially, on the same background task.
The subscription handling would: acquire + update + release resource table, call subscription callback, drop the sink immediately, acquire + update + release resource table.

The resource limiting would reject connections for multiple clients if they attempt to make a subscription call to a method that drops the sink immediately.

PR #789, opened 2022-06-04
feat(clients): add explicit unsubscribe API

niklasad1:
I reckon that sometimes one wants to explicitly wait until the unsubscribe method has been called which this does.

We might want to use tin the benches instead of the drop impl.


core/src/client/mod.rs
```diff
@@ -203,12 +203,25 @@ impl<Notif> Subscription<Notif> {
...
 	}
 
 	/// Return the subscription type and, if applicable, ID.
 	pub fn kind(&self) -> &SubscriptionKind {
-		&self.kind
+		self.kind.as_ref().expect("only None after unsubscribe; qed")
+	}
+
+	/// Unsubscribe and consume the subscription.
+	pub async fn unsubscribe(mut self) -> Result<(), Error> {
```
jsdw, line 215: Is there a sensible way to test this?

jsdw (approved):
LGTM, and using `Option` there is def cleaner :)

core/src/client/mod.rs
niklasad1: I think so, sorry but stuck on my other middleware PR :)

lexnv (approved):
LGTM! 👍

tests/tests/integration_tests.rs
```diff
@@ -62,8 +70,40 @@ async fn ws_subscription_works() {
...
+
+	// It's technically possible to have race-conditions between the notifications and the unsubscribe message.
+	// So let's wait for the first notification and then unsubscribe.
+	let _item = sub.next().await.unwrap().unwrap();
+
+	sub.unsubscribe().await.unwrap();
+
+	let mut success = false;
+
+	// wait until the next slot is opened, as only one concurrent call is allowed.
```
niklasad1, line 91:
```suggestion
	// Wait until the the slot available as only one concurrent call is allowed then we know that subscription call has
	// been finished
```

Issue #758, opened 2022-05-09
[middleware]: expose all details regarding a call.


niklasad1:
Currently the middleware just exposing the method for a given call.

It would be nice to know the kind of call (i,e, method call, subscription and so on) then also the params used in the call.

See https://github.com/paritytech/substrate/pull/8783#discussion_r867388416 for further information.

niklasad1:
Fixed by #793

Issue #747, opened 2022-04-29
[deps]: remove all usage of `env_logger`


niklasad1:
As we use the tracing crate all tests and examples should use `tracing_subscriber`

It's just nice get all logs and reduce needless dependencies in the tree.

niklasad1:
Closed by #793

Issue #699, opened 2022-02-14
Consider extending the parameters of Middleware::on_call


ljedrz:
snarkOS could use the ability to log the source and possibly some of the header of the RPC requests it receives: https://github.com/AleoHQ/snarkOS/issues/1628; I think this could be achieved by extending the parameters of `Middleware::on_call`, but I'm open to other options.

If I didn't miss anything in my partial implementation of a [PR addressing the linked issue](https://github.com/AleoHQ/snarkOS/pull/1635), this is currently not supported.

I know that there is a `DEBUG` log turned on by default, but it was too verbose and I think it only contained information on the response (which was also truncated if it was deemed too long), so we silenced that one.

Possibly related existing issues: https://github.com/paritytech/jsonrpsee/issues/450, https://github.com/paritytech/jsonrpsee/issues/487.

dvdplm:
@ljedrz Thank you for the report. One question: when you say "source", what do you have in mind exactly? Do you mean the IP of the host sending the request?

dvdplm:
@maciejhirsz I think we talked about adding this at some point, right?

maciejhirsz:
Yep, we can expose IP and headers in middleware.

666lcz:
> Yep, we can expose IP and headers in middleware.

Hi @maciejhirsz , do you have any ETA for supporting this? We are trying to implement some sort of rate-limiting by IP address. Ty!

niklasad1:
Closed by #793

PR #793, opened 2022-06-13
middleware refactoring

niklasad1:
Closing #758, #747 and #699

It's a quite significant refactoring where the `sink` is only used for subscriptions as the rest of the calls just returns `MethodResponse` which is a type that has the final JSON-RPC response and status flag whether the calls was successful or not.
In order to intercept the messages to log the full responses in the middleware.

Further, it also modifies the `SubscriptionMethod` to be a future which returns when the subscription call has been answered which in turn forces plenty noise i.e, I had to change lots of code.

~~In addition I noticed that `SubscriptionSink::send(&mut self, msg: M)`  is not needed so I changed it to `SubscriptionSink::send(&self, msg: M)` also then `pipe_from_stream` doesn't need `&mut self` anymore.
rustc may benefit from that to some optimizations....~~

It splits the middleware into two traits for WS and HTTP.

```rust
/// Defines a middleware specifically for HTTP requests with callbacks during the RPC request life-cycle.
/// The primary use case for this is to collect timings for a larger metrics collection solution.
///
/// See [`HttpServerBuilder::set_middleware`](../../jsonrpsee_http_server/struct.HttpServerBuilder.html#method.set_middleware) method
/// for examples.
pub trait HttpMiddleware: Send + Sync + Clone + 'static {
	/// Intended to carry timestamp of a request, for example `std::time::Instant`. How the middleware
	/// measures time, if at all, is entirely up to the implementation.
	type Instant: std::fmt::Debug + Send + Sync + Copy;

	/// Called when a new JSON-RPC request comes to the server.
	fn on_request(&self, remote_addr: SocketAddr, headers: &Headers) -> Self::Instant;

	/// Called on each JSON-RPC method call, batch requests will trigger `on_call` multiple times.
	fn on_call(&self, method_name: &str, params: Params);

	/// Called on each JSON-RPC method completion, batch requests will trigger `on_result` multiple times.
	fn on_result(&self, method_name: &str, success: bool, started_at: Self::Instant);

	/// Called once the JSON-RPC request is finished and response is sent to the output buffer.
	fn on_response(&self, result: &str, _started_at: Self::Instant);
}

/// Defines a middleware specifically for WebSocket connections with callbacks during the RPC request life-cycle.
/// The primary use case for this is to collect timings for a larger metrics collection solution.
///
/// See the [`WsServerBuilder::set_middleware`](../../jsonrpsee_ws_server/struct.WsServerBuilder.html#method.set_middleware)
/// for examples.
pub trait WsMiddleware: Send + Sync + Clone + 'static {
	/// Intended to carry timestamp of a request, for example `std::time::Instant`. How the middleware
	/// measures time, if at all, is entirely up to the implementation.
	type Instant: std::fmt::Debug + Send + Sync + Copy;

	/// Called when a new client connects
	fn on_connect(&self, remote_addr: SocketAddr, headers: &Headers);

	/// Called when a new JSON-RPC request comes to the server.
	fn on_request(&self) -> Self::Instant;

	/// Called on each JSON-RPC method call, batch requests will trigger `on_call` multiple times.
	fn on_call(&self, method_name: &str, params: Params);

	/// Called on each JSON-RPC method completion, batch requests will trigger `on_result` multiple times.
	fn on_result(&self, method_name: &str, success: bool, started_at: Self::Instant);

	/// Called once the JSON-RPC request is finished and response is sent to the output buffer.
	fn on_response(&self, result: &str, started_at: Self::Instant);

	/// Called when a client disconnects
	fn on_disconnect(&self, remote_addr: std::net::SocketAddr);
}
```


core/src/middleware.rs
```diff
@@ -37,22 +39,22 @@
...
-	type Instant: Send + Copy;
+	type Instant: std::fmt::Debug + Send + Sync + Copy;
 
 	/// Called when a new client connects (WebSocket only)
 	fn on_connect(&self) {}
 
-	/// Called when a new JSON-RPC comes to the server.
-	fn on_request(&self) -> Self::Instant;
+	/// Called when a new JSON-RPC request comes to the server.
+	fn on_request(&self, remote_addr: std::net::SocketAddr, headers: &http::HeaderMap) -> Self::Instant;
```
niklasad1, line 48: This is a bit weird as the `WS` will have the same remote addr and headers for the entire connection

Then for `HTTP` it will be useful for each "request".

So what we could do, is the have_
- `on_connect(&self, remote_addr: std::net::SocketAddr, headers: &http::HeaderMap)`
- `on_request(&self)`

Then justify the documentation that `on_connect` for  can be used to inspect headers and so on for stateless protocols, we could also move these to different traits as  I could be useful to separate middleware impls because HTTP/WS is quite different. 

 
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

core/src/server/rpc_module.rs
```diff
@@ -403,7 +426,10 @@ impl Methods {
 
 	/// Execute a callback.
 	async fn inner_call(&self, req: Request<'_>) -> RawRpcResponse {
```
jsdw, line 428: Now that the sink is only needed for subscriptions, I wonder whether inner_call could return an enum, so it would only need to construct the subscription channels etc
in the event that the method being called was a subscription, else it could do a simpler thing with other types of method?

This might also make it a little more obvious when the returned sink etc are actually useful?

ws-server/src/server.rs
```diff
@@ -427,252 +468,79 @@ async fn background_task(
 
 		tracing::debug!("recv {} bytes", data.len());
 
-		let request_start = middleware.on_request();
+		let request_start = middleware.on_request(remote_addr, &headers);
```
jsdw, line 471: So, it looks like those headers will only ever contain Host and Origin from above; what about any custom headers?

core/src/middleware.rs
jsdw: I've been mulling this over; I'm personally leaning towards separate middleware for Http and Ws, since it feels like we have to hack it around to have functions which map nicely to both.

Pros of separate middlware:
- precise traits that better express what's actually happening.
- (maybe this helps perf or something?)

Cons:
- we need to have two similar traits and sets of blanket impls.
- users/we can't write generic middleware that applies to both (though I doubt that's an issue, and one could always write some wrapper `GenericMiddleware` thing that implements both where there is overlap if needed)

ws-server/src/server.rs
niklasad1: https://github.com/paritytech/soketto/pull/60,

it's missing from soketto but I don't want to make this PR any larger.

ws-server/src/server.rs
```diff
@@ -225,10 +229,9 @@ enum HandshakeResponse<'a, M> {
 	},
 }
 
-async fn handshake<M>(socket: tokio::net::TcpStream, mode: HandshakeResponse<'_, M>) -> Result<(), Error>
-where
-	M: Middleware,
-{
+async fn handshake<M: Middleware>(socket: tokio::net::TcpStream, mode: HandshakeResponse<'_, M>) -> Result<(), Error> {
```
jsdw, line 232: Probably out of scope of this PR but it feels like having an `accept_handshake` and `reject_handshake` call (or similar) would make more sense, since this call basically just switches on an enum and has very little shared between them?

core/src/middleware.rs
niklasad1: yeah, I agree it's better to have two different traits.

You most likely want to handle them differently.

ws-server/src/server.rs
```diff
@@ -428,257 +473,83 @@ async fn background_task(
...
+						request_start,
+					};
+
+					let (response, maybe_pending_sub_tx) = process_single_request(data, call).await;
+					middleware.on_response(&response.result, request_start);
+
+					let _ = sink.send_raw(response.result);
+
+					if let Some(pending_sub_tx) = maybe_pending_sub_tx {
+						pending_sub_tx.accept();
```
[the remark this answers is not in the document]
niklasad1: >  1. don't like this pending_sub_tx thing here; it feels leaky (and is an odd one-off for just subscriptions). If we removed it without doing anything else, what would the consequences be? That the middleware is potentially called after another message has been sent down the wire? Do we care enough about that?

that's fair, the issue is that the message would never hit the middleware stuff then. It would just be answered so actually changed it a little bit to get rid of this mess and actually cloned the message to send back to both the `ws task` and `middleware´. As we kind of assume that subscription response will be short just a `subscription ID` that should be okay to clone.

> 2. It looks like we already know what the response for a subscription is going to be; it's always MethodResponse::response(id, &uniq_sub.sub_id, sink.max_response_size() as usize). Thus, do we actually need to try and get that result back from the accept() call to here? We just need to know when accept() is called perhaps, and then can build the identical response to give to middleware when needed? (not sure if that actually is useful to ponder or not!)

Yeah, just I did that but it means that we need to clone every subscription response but that should be quite small anyway so should be ok.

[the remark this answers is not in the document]
jsdw: Just had a quick look over; nice! It def feels cleaner to me now!

niklasad1:
> Do you plan on breaking the middleware into an HttpMiddleware and WsMiddleware in this PR?

Yes, if we are on the same page with the other changes :)

jsdw:
> Yes, if we are on the same page with the other changes :)

I like the simplifying you did; it's def easier for me to follow now, so I'm on board with it! And as you say, the clone should be quite cheap for that subscription response! (I wonder what the benchmarks are like at the mo)

niklasad1:
> I wonder what the benchmarks are like at the mo

I ran the benches in the weekend and those looks still solid :)

ws-server/src/server.rs
```diff
@@ -428,257 +471,84 @@ async fn background_task(
...
-					let (id, code) = prepare_error(&data);
-					sink.send_error(id, code.into());
-					middleware.on_response(request_start);
+						MethodResult::SendAndMiddleware(r) => {
+							middleware.on_response(&r.result, request_start);
+							let _ = sink.send_raw(r.result);
+						}
+					};
 				}
+				.boxed();
```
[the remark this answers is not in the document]
niklasad1: Those are moved to `process_single_request` and `process_batch_request`.

core/src/middleware.rs
```diff
@@ -37,22 +39,22 @@
 pub trait Middleware: Send + Sync + Clone + 'static {
 	/// Intended to carry timestamp of a request, for example `std::time::Instant`. How the middleware
 	/// measures time, if at all, is entirely up to the implementation.
-	type Instant: Send + Copy;
+	type Instant: std::fmt::Debug + Send + Sync + Copy;
```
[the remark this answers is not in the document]
niklasad1: Can you elaborate a bit what you mean with an example?

It's possible to copy it between calls and stuff right now

core/src/server/helpers.rs
```diff
@@ -259,6 +212,106 @@ impl BoundedSubscriptions {
...
+						serde_json::to_string(&ErrorResponse::borrowed(ErrorCode::InternalError.into(), id)).unwrap();
+					Self { result, success: false }
+				}
+			}
+		}
+	}
+
+	/// Create a `MethodResponse` from an error.
+	pub fn error<'a>(id: Id, err: impl Into<ErrorObject<'a>>) -> Self {
+		let result = serde_json::to_string(&ErrorResponse::borrowed(err.into(), id)).unwrap();
```
niklasad1, line 256:
```suggestion
		let result = serde_json::to_string(&ErrorResponse::borrowed(err.into(), id)).expect("valid JSON; qed");

File: ws-server/Cargo.toml [hunk 9]
```diff
@@ -21,6 +21,8 @@ tokio = { version = "1.16", features = ["net", "rt-multi-thread", "macros", "tim
 tokio-util = { version = "0.7", features = ["compat"] }
 tracing-futures = "0.2.5"
 tokio-stream = "0.1.7"
+http = "0.2.7"
+
```
niklasad1, line 25:
```suggestion
```

core/src/middleware.rs
```diff
@@ -26,67 +26,149 @@
...
 	/// Called on each JSON-RPC method completion, batch requests will trigger `on_result` multiple times.
-	fn on_result(&self, _name: &str, _success: bool, _started_at: Self::Instant) {}
+	fn on_result(&self, method_name: &str, success: bool, started_at: Self::Instant);
 
 	/// Called once the JSON-RPC request is finished and response is sent to the output buffer.
-	fn on_response(&self, _started_at: Self::Instant) {}
+	fn on_response(&self, result: &str, started_at: Self::Instant);
+
+	/// Called when a client disconnects
+	fn on_disconnect(&self, remote_addr: std::net::SocketAddr);
```
niklasad1, line 83: this is silly but I reckon that it might be useful to log when peers disconnect

core/src/middleware.rs
```diff
@@ -26,67 +26,149 @@
...
+pub trait HttpMiddleware: Send + Sync + Clone + 'static {
+	/// Intended to carry timestamp of a request, for example `std::time::Instant`. How the middleware
+	/// measures time, if at all, is entirely up to the implementation.
+	type Instant: std::fmt::Debug + Send + Sync + Copy;
+
+	/// Called when a new JSON-RPC request comes to the server.
+	fn on_request(&self, remote_addr: SocketAddr, headers: &Headers) -> Self::Instant;
+
+	/// Called on each JSON-RPC method call, batch requests will trigger `on_call` multiple times.
+	fn on_call(&self, method_name: &str, params: Params);
```
niklasad1, line 48: do we want `method kind` here as well i.e, if the call is a "subscription", "method call" or "notification"?

http-server/src/server.rs
```diff
@@ -591,234 +633,243 @@ async fn process_validated_request(
...
+}
+
+#[derive(Debug, Clone)]
+struct Batch<'a, M: Middleware> {
+	data: Vec<u8>,
+	call: CallData<'a, M>,
+}
+
+#[derive(Debug, Clone)]
+struct CallData<'a, M: Middleware> {
```
niklasad1, line 735: this should be cheap to `copy`

examples/examples/middleware_ws.rs
```diff
@@ -27,30 +27,40 @@
...
+use jsonrpsee::core::client::ClientT;
+use jsonrpsee::core::middleware::{self, Headers, Params};
 use jsonrpsee::ws_client::WsClientBuilder;
 use jsonrpsee::ws_server::{RpcModule, WsServerBuilder};
 
 #[derive(Clone)]
 struct Timings;
 
-impl middleware::Middleware for Timings {
+impl middleware::WsMiddleware for Timings {
```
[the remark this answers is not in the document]
jsdw: That was my take on it yeah; it'd just give the flexibility to keep track of whatever was desired across the lifetime of a single connection :)

core/src/middleware.rs
jsdw: Is it easily accessbile? I guess it wouldn't do any harm to have!

lexnv (approved):
LGTM! 👍 As a follow-up we could extend the middleware with `method_kind` and custom user context/state to extend the `type Instant`

core/src/middleware.rs
niklasad1: I don't do it because it requires to lookup the method name in the HashMap.

should be quite easy to do.

examples/examples/middleware_ws.rs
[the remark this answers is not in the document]
jsdw: As we agreed,  let's think about this in a followup issue/pr!

core/src/middleware.rs
jsdw: Mmm, if it's a performance hit to do so (even when no middleware) then I'm happy not at the mo!

jsdw (approved):
Looks great to me!

Issue #735, opened 2022-04-21
[rpc module]: add easy way to register subscription without accept/reject API


niklasad1:
https://github.com/paritytech/jsonrpsee/pull/728#issuecomment-1099133993

jsdw:
Given the other stuff we're working on, I suspect that we'll just end up closing this and accepting that the accept() API will be useful and not worth trying to avoid (also being able to `pipe_from_stream` on the pending sink makes it more ergonomic already)

lexnv:
Closed by: https://github.com/paritytech/jsonrpsee/pull/799

PR #799, opened 2022-06-20
Return error from subscription callbacks

### LINKED ISSUE #734: [rpc module]: return error from subscriptions to make error handling more ergonomic
Recently the subscription code got refactor where one must reject or accept a subscription manually and because
`SubscriptionClosed` is not treated as an error the closures must not return `Result`.

However, it's quite awkward to so:

Example:

```rust
	module
		.register_subscription("sub_one_param", "sub_one_param", "unsub_one_param", |params, pending, _| {
		       // not possible to return Result here.
                       let (idx, mut sink) = match (params.one::<usize>(), pending.accept()) {
				(Ok(idx), Some(sink)) => (idx, sink),
				_ => return,
			};
		       
                      sink.pipe_from_stream(...);
		})
		.unwrap();
```

So it would be more ergonomic / less awkward if it would be possible to do:

```rust
	module
		.register_subscription("sub_one_param", "sub_one_param", "unsub_one_param", |params, pending, _| {
		       let idx = params.one()?;
                       let sink = pending.accept()?;
		       
                        sink.pipe_from_stream(...);
                        Ok(())
		})
		.unwrap();
```

We could easily support this.

lexnv:
Subscription callbacks are now returning errors for improved ergonimics.

The jsonrpsee proc-macro renders the provided subscription method with
an additional return type. 

```rust
pub trait Rpc {
	#[subscription(name = "subscribe", item = String)]
	fn sub(&self);
}


#[async_trait]
impl RpcServer for RpcServerImpl {
	fn sub(&self, pending: PendingSubscription) -> ReturnTypeSubscription {
		Ok(())
	}
```

While at it, implement `pipe_from_try_stream` on the `PendingSubscription` for better ergonomics,
which wraps the `pending.accept()` and pipes from the provided stream.

### Before

```rust
.register_subscription("sub_one_param", "sub_one_param", "unsub_one_param", |params, pending, _| {
    let idx = match params.one::<usize>() {
        Ok(idx) => idx,
        _ => {
            pending.reject("some special error);
            return;
        }
    };
    tokio::spawn(async move {
        if let Some(sink) = pending.accept() {
             sink.pipe_from_try_stream(stream).await;
             // ...
        }
    });
    Ok(())
}
```

### After

```rust
.register_subscription("sub_one_param", "sub_one_param", "unsub_one_param", |params, pending, _| {
    let idx = match params.one::<usize>()?;

    tokio::spawn(async move {
        pending.pipe_from_try_stream(stream).await;
        // ...
    });
    Ok(())
}
```

Closes #734, #735.


examples/examples/proc_macro.rs
```diff
@@ -28,6 +28,7 @@ use std::net::SocketAddr;
 
 use jsonrpsee::core::{async_trait, client::Subscription, Error};
 use jsonrpsee::proc_macros::rpc;
+use jsonrpsee::types::ReturnTypeSubscription;
```
niklasad1, line 31: you should use `RpcResult<()>` instead no need to introduce another type it will just annoy users of the API.

I think `RpcResult<()`> is more readable as well.

proc-macros/src/render_server.rs
```diff
@@ -400,7 +406,7 @@ impl RpcDescription {
 							#tracing::error!("Failed to parse JSON-RPC params as object: {}", e);
 							let _e: #err = e.into();
 							#pending.reject(_e);
-							return;
+							return Ok(());
```
niklasad1, line 409: we need to document that this is a "hack" i.e, the result of this is not used anyway just make the API more ergonomic such that users can just do `params.parse()?`

niklasad1: I removed it recently just for that reason :sweat_smile:

examples/examples/ws_pubsub_with_params.rs
```diff
@@ -67,9 +67,10 @@ async fn run_server() -> anyhow::Result<SocketAddr> {
 	let mut module = RpcModule::new(());
 	module
 		.register_subscription("sub_one_param", "sub_one_param", "unsub_one_param", |params, pending, _| {
-			let (idx, mut sink) = match (params.one(), pending.accept()) {
-				(Ok(idx), Some(sink)) => (idx, sink),
-				_ => return,
+			let idx = params.one()?;
+			let mut sink = match pending.accept() {
+				Some(sink) => sink,
+				_ => return Ok(()),
```
niklasad1, line 73: better to change `fn accept() - Result<SubscriptionSink, Error` this `return Ok(())` is confusing for most folks.

I get it because at this point jsonrpsee has no use of the error because the connection is closed.

core/src/server/rpc_module.rs
```diff
@@ -750,7 +754,7 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 						claimed,
 					}));
 
-					callback(params, sink, ctx.clone());
+					let _ = callback(params, sink, ctx.clone());
```
jsdw, line 757: Ah I see; the result really isn't used at all!

I feel a bit uneasy about using a result and then throwing away the error; returning an error makes it feel like it'll be made use of in some way. We are just doing it because it plays nicely with the `?` type.

Perhaps we _should_ use a different error type in the response; something like `struct SubscriptionError` which hodls no data and impls From so that `?` can be used transparently on other jsonrpsee errors to convert them into it. This at least signals in the API typings that error information is being discarded when returned?

Other question; don't we want to use an error case here to make sure the connection is closed or something? For example if a subscription does something like:

```
        tokio::spawn(async move {
            pending.pipe_from_stream(params).await;
        });
        Err(SubscriptionError)
```

(or otherwise spawns a thread but still then decides to return an error), I'd expect the error response to lead to the connection being closed?

types/src/error.rs
```diff
@@ -80,6 +80,36 @@ impl<'a> fmt::Display for ErrorResponse<'a> {
 	}
 }
 
+/// The return type of the subscription's method for the rpc server implementation.
+///
+/// **Note**: The error does not contain any data and is discarded on drop.
+pub type ReturnTypeSubscription = Result<(), SubscriptionEmptyError>;
```
niklasad1, line 86:
```suggestion
pub type SubscriptionResult = Result<(), SubscriptionEmptyError>;
```

core/src/server/rpc_module.rs
niklasad1: would be nice with a comment here, I will likely forget about this when looking at the code in a few days :)

examples/examples/proc_macro.rs
niklasad1: I still don't like the name why not `SubscriptionResult`or something shorter.

core/src/server/rpc_module.rs
```diff
@@ -817,19 +816,99 @@ impl PendingSubscription {
...
+			S: TryStream<Ok = T, Error = E> + Unpin,
+			T: Serialize,
+			E: std::fmt::Display,
+	{
+		if let Ok(mut sink) = self.accept() {
+			let result = sink.pipe_from_try_stream(stream).await;
+			(Some(sink), result)
 		} else {
-			None
+			(None, SubscriptionClosed::RemotePeerAborted)
```
niklasad1, line 886: this became quite complicated but fine for now

types/src/error.rs
niklasad1: that one is likely only internal in jsonrpsee, you can rename it to `InnerSubscriptionResult` or something.

and rename :point_up_2: to `SubscriptionResult`

niklasad1 (dismissed):
nice work

core/src/server/rpc_module.rs
```diff
@@ -817,19 +816,99 @@ impl PendingSubscription {
...
+	/// ```no_run
+	///
+	/// use jsonrpsee_core::server::rpc_module::{RpcModule, SubscriptionResult};
+	/// use jsonrpsee_core::error::{Error, SubscriptionClosed};
+	/// use jsonrpsee_types::ErrorObjectOwned;
+	/// use anyhow::anyhow;
+	///
+	/// let mut m = RpcModule::new(());
+	/// m.register_subscription("sub", "_", "unsub", |params, pending, _| {
+	///     // let mut sink = pending.accept().unwrap();
```
jsdw, line 853: Remove this commented out pending.accept() now?

core/src/server/rpc_module.rs
```diff
@@ -817,19 +816,99 @@ impl PendingSubscription {
...
+	///                let err_obj: ErrorObjectOwned = SubscriptionClosed::Success.into();
+	///                sink.close(err_obj);
+	///            }
+	///            (Some(sink), SubscriptionClosed::Failed(e)) => {
+	///                sink.close(e);
+	///            }
+	///            // we don't want to send close reason when the client is unsubscribed or disconnected.
+	///            (Some(_sink), SubscriptionClosed::RemotePeerAborted) => (),
+	///            (None, _) => (),
+	///         }
```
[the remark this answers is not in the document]
niklasad1: awesome idea :100:

core/src/server/rpc_module.rs
```diff
@@ -842,6 +923,55 @@ impl Drop for PendingSubscription {
...
+	/// The connection was accepted and the pipe returned [`SubscriptionClosed::Failed`]
+	/// with the provided error.
+	Failure(SubscriptionSink, ErrorObjectOwned),
+	/// The remote peer closed the connection or called the unsubscribe method.
+	RemotePeerAborted,
+}
+
+impl PipeFromStreamResult {
+	/// Callback that will run the provided function if the result is [`PipeFromStreamResult::Success`].
+	/// After the function runs a new [`PipeFromStreamResult::RemotePeerAborted`] is returned.
```
niklasad1, line 941: Can you explain why `PipeFromStreamResult::RemotePeerAborted` is returned here?

Can't you borrow the `SubscriptionSink` in the closure?

examples/examples/ws_pubsub_broadcast.rs
```diff
@@ -73,10 +73,7 @@ async fn run_server() -> anyhow::Result<SocketAddr> {
 
 	module.register_subscription("subscribe_hello", "s_hello", "unsubscribe_hello", move |_, pending, _| {
 		let rx = BroadcastStream::new(tx.clone().subscribe());
-		let mut sink = match pending.accept() {
-			Some(sink) => sink,
-			_ => return,
-		};
+		let mut sink = pending.accept()?;
```
niklasad1, line 76: can you update the all the examples with the latest changes?

such as:

```rust
    tokio::spawn(async move {
        pending
            .pipe_from_stream(stream)
            .await
            .on_success(|sink| {
                let err_obj: ErrorObjectOwned = SubscriptionClosed::Success.into();
                sink.close(err_obj);
            })
            .on_failure(|sink, err| {
                sink.close(err);
            })
    });
```

proc-macros/src/lib.rs
```diff
@@ -98,8 +99,8 @@ pub(crate) mod visitor;
 ///     async fn async_method(&self, param_a: u8, param_b: String) -> u16;
 ///     fn sync_method(&self) -> String;
 ///
-///     // Note that `subscription_sink` was added automatically.
-///     fn sub(&self, subscription_sink: PendingSubscription);
+///     // Note that `subscription_sink` and `ReturnTypeSubscription` were added automatically.
```
niklasad1, line 102:
```suggestion
///     // Note that `subscription_sink` and `SubscriptionResult` were added automatically.
```

proc-macros/src/lib.rs
```diff
@@ -98,8 +99,8 @@ pub(crate) mod visitor;
...
+///     fn sub(&self, subscription_sink: PendingSubscription) -> ReturnTypeSubscription;
```
niklasad1, line 103:
```suggestion
///     fn sub(&self, subscription_sink: PendingSubscription) -> SubscriptionResult;
```

core/src/server/rpc_module.rs
lexnv: I was thinking of having a sort of `PipeFromStreamResult::ResultConsumed` to keep on chaining `on_success` with `on_failure` for better ergonomics. WDYT?

niklasad1: sure, that works as well

core/src/server/rpc_module.rs
jsdw: Just want to confirm; do we not care about any explicit closing of the subscription here if an error is returned?

core/src/server/rpc_module.rs
jsdw: `ResultConsumed` makes sense but does mean you lose information if you're matching on the enum.

Another option might be something like:

```rust
pub enum PipeFromStreamResult {
	/// The connection was accepted and the pipe returned [`SubscriptionClosed::Success`].
	Success(Option<SubscriptionSink>),
	/// The connection was accepted and the pipe returned [`SubscriptionClosed::Failed`]
	/// with the provided error.
	Failure(Option<(SubscriptionSink, ErrorObjectOwned)>),
	/// The remote peer closed the connection or called the unsubscribe method.
	RemotePeerAborted,
}
```

And then `on_success` and `on_failure` only run if the options are `Some` and `take` the values out of them, but one can still match on the enum and see what the result was. (semantically it's the same as having `SuccessConsumed` and `FailureConsumed` variants, but less matching). I don't love it, but I haven't thought up anything better!

core/src/server/rpc_module.rs
niklasad1: I guess we could log it as `warn`.

Should only possible to fail if `params.parse()` or `SubscriptionEmptyError (connection was closed)`

core/src/server/rpc_module.rs
lexnv: The `Option` approach seems a bit better for the end user 👍

core/src/server/rpc_module.rs
lexnv: Have added a warning here with the `error` as well, just in case we decide to change it later to something else

core/src/server/rpc_module.rs
jsdw: d

core/src/server/rpc_module.rs
```diff
@@ -889,7 +1021,7 @@ impl SubscriptionSink {
 	///
 	/// ```no_run
 	///
-	/// use jsonrpsee_core::server::rpc_module::{RpcModule, SubscriptionResult};
+	/// use jsonrpsee_core::server::rpc_module::RpcModule;
 	/// use jsonrpsee_core::error::{Error, SubscriptionClosed};
 	/// use jsonrpsee_types::ErrorObjectOwned;
 	/// use anyhow::anyhow;
```
jsdw, line 1027: Update this example with new pipe_from_stream .on_success/.on_failure api?

jsdw: Ah; PipeFromStreamResult doesn't appear to be used in these `pipe_from_stream` and `pipe_from_try_stream` yet. Should we be using them here too for consistency?

core/src/server/rpc_module.rs
```diff
@@ -987,6 +1120,7 @@ impl SubscriptionSink {
 	///     let mut sink = pending.accept().unwrap();
 	///     let stream = futures_util::stream::iter(vec![1_usize, 2, 3]);
 	///     tokio::spawn(async move { sink.pipe_from_stream(stream).await; });
+	///     Ok(())
 	/// });
 	/// ```
 	pub async fn pipe_from_stream<S, T>(&mut self, stream: S) -> SubscriptionClosed
```
jsdw, line 1126: Same here; does it make sense to use PipeFromStreamResult? it feels odd having two very similar `pipe_from_stream`/`pipe_from_try_stream` calls that result in different but similar return types, but maybe there's good reason for it?

These calls do take `&mut self` at the mo, and the versions on `PendingSink` consume self. I can imagine this being tricky.

Is there a way to make these APIs consistent? Should these take `mut self` too? @niklasad1 what do you reckon?

types/src/error.rs
```diff
@@ -80,6 +80,36 @@ impl<'a> fmt::Display for ErrorResponse<'a> {
...
+pub type SubscriptionResult = Result<(), SubscriptionEmptyError>;
+
+/// The error returned by the subscription's method for the rpc server implementation.
+///
+/// It contains no data, and neither is the error utilized. It provides an abstraction to make the
+/// API more ergonomic while handling errors that may occur during the subscription callback.
+#[derive(Debug)]
+pub struct SubscriptionEmptyError;
+
+impl From<anyhow::Error> for SubscriptionEmptyError {
```
jsdw, line 95: These From impls could perhaps just be:

```rust
impl <T> From<T> for SubscriptionEmptyError {
    fn from(e: T) -> Self {
        SubscriptionEmptyError
    }
}
```

Since we're not super specific anyway about the types of error that can be converted into this, maybe that's easier?

core/src/server/rpc_module.rs
niklasad1: yes :100:

core/src/server/rpc_module.rs
niklasad1: yeah, that one is tricky for sure.

Recall that `PendingSubscription` must be moved to be converted to a `SubscriptionSink` so it's not possible to unify that API.

So maybe we could make the APIs: `Subscription::accept, Subscription::pipe_from_stream, Subscription::pipe_from_try_stream` private and folks you be forced the use `Pending::pipe_from_stream` at least then they won't be confused.

I don't see a reason why anyone would need those APIs anyway.

lexnv: I like the idea of having `Subscription::pipe_from_stream` private.
One drawback atm is that `Pending::pipe_from_stream` could not be reused, due to it taking `self` and not `&mut self`. Tho, I believe this limitation could be lifted to ensure the API is as permissive as `Subscription::pipe_from_stream`

jsdw: Going down this line, I'm wondering now whether it even makes sense to have a distinction between the `PendingSubscriptionSink` and `SubscriptionSink`.

How about we just return a `SubscriptionSink` straight away again, with the following:

- `pub fn accept(&mut self)` - accepts the connection and sends back the notification to the user without sending anything else. You call this if you want to send the accept message without sending anything else.
- the usual Sink impl would, if `accept` hasn't been explicitly called, call `accept` under the hood prior to sending any messages (if it hasn't been called already), so you don't need to explicitly `accept` at all if you don't care when the accept message is sent back.
- `pub fn pipe_from_stream`/`pipe_from_try_stream`  methods consume `&mut self`, and pretty much work as they do now, internally handling `accept()` too (if it hasn't been called already).
- `PipeFromStreamResult` no longer needs to contain the Sink anywhere. I'm hoping the lifetimes work out so that you can still do something like:

```rust
sink.pipe_from_stream(foo)
    .await
    .on_failure(|e| sink.close(e));
```

Or just trying to pipe a few things and then handling the result eventually (which I'd expect to basically come back the same whenever you try to pipe into a stream that's already been closed or whatever):

```rust
sink.pipe_from_stream(bar).await;
sink.pipe_from_stream(wibble).await;
sink.pipe_from_stream(foo).await
    .on_failure(|e| sink.close(e));
```

But possibly those `on_success`/`on_failure` calls needs a rethink they aren't very useful now.

What do you guys think?

tests/tests/rpc_module.rs
```diff
@@ -348,3 +348,36 @@ async fn subscribe_unsubscribe_without_server() {
...
+		.register_subscription("my_sub", "my_sub", "my_unsub", |_, mut sink, _| {
+			let err = ErrorObject::borrowed(0, &"rejected", None);
+			sink.reject(err.into_owned())?;
+			Ok(())
+		})
+		.unwrap();
+
+	let sub_err = module.subscribe("my_sub", EmptyParams::new()).await.unwrap_err();
+
+	assert!(matches!(sub_err, Error::Call(CallError::Custom(e)) if e.message().contains("rejected") && e.code() == 0));
```
niklasad1, line 382: we shouldn't use `error code == 0` that is not according to the JSON-RPC spec.

types/src/error.rs
lexnv: Unfortunately, that leads to an error of `conflicting implementations of trait std::convert::From`, which is conflicting with std implementation.

core/src/server/rpc_module.rs
```diff
@@ -923,14 +948,29 @@ impl SubscriptionSink {
 		T: Serialize,
 		E: std::fmt::Display,
 	{
+		if self.maybe_accept().is_err() {
```
niklasad1, line 951: :+1:

examples/examples/ws_pubsub_with_params.rs
```diff
@@ -66,36 +66,27 @@ async fn run_server() -> anyhow::Result<SocketAddr> {
...
-			};
+		.register_subscription("sub_one_param", "sub_one_param", "unsub_one_param", |params, mut sink, _| {
+			let idx = params.one()?;
 			let item = LETTERS.chars().nth(idx);
 
 			let interval = interval(Duration::from_millis(200));
 			let stream = IntervalStream::new(interval).map(move |_| item);
 
 			tokio::spawn(async move {
 				match sink.pipe_from_stream(stream).await {
```
niklasad1, line 77: the `on_success/on_failure` stuff is removed?

core/src/server/rpc_module.rs
```diff
@@ -851,25 +832,69 @@ pub struct SubscriptionSink {
...
+		}
+	}
+
+	/// Attempt to accept the subscription and respond the subscription method call.
 	///
-	/// Returns `Ok(true)` if the message could be send
-	/// Returns `Ok(false)` if the sink was closed (either because the subscription was closed or the connection was terminated)
-	/// Return `Err(err)` if the message could not be serialized.
+	/// Fails if the connection was closed, or if called multiple times.
+	pub fn accept(&mut self) -> Result<(), SubscriptionEmptyError> {
```
jsdw, line 861: I'm not really sure what the value of having `accept()` and `maybe_accept()` is, especially given both have the same return type and the error is not informative; I think there is potential for confusion over when to use each.

Is it worth returning a more informative error here (all errors convert into `SubscriptionEmptyError` anyway, which I view as a bit of a hack to help ergonomics in the common case, and don't think we should be using ourselves unless we really do have no information to return)?

I guess I'd go for something like just having `accept(&mut self) -> Result<(), AcceptError>` where `AcceptError` can be something like:

```rust
enum AcceptError {
    AlreadyAccepted,
    RemotePeerAborted,
}
```

(all names and such up for grabs, and should this be nested in a larger error type?)

That way, people can just call it and ignore the error if they want, or pay attention to the error to know why it didn't work out, and no need for a separate `maybe_accept`?

WDYT?

types/src/error.rs
jsdw: Does `impl <T> Into<SubscriptionEmptyError> for T` work or is that the same issue?

jsdw:
Looks good; the unified sink stuff def feels nicer API wise!

My only real blocked is that I don't think we should have `accept()` and `maybe_accept()` (the difference is confusing and one call can express enough info to use in both places). The other comment was an internal API nit so doesn't matter so much.

core/src/server/rpc_module.rs
niklasad1: these errors are still not propagated because of `SubscriptionEmptyError` so the error themselves are just swallowed here.

Thus, I don't get why need to extra verbose stuff `SubscriptionSinkState` but yeah if we want to propagate the error `AcceptError` makes more sense.

core/src/server/rpc_module.rs
```diff
@@ -775,69 +775,50 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
...
-	/// Claimed resources.
-	claimed: Option<ResourceGuard>,
+enum SubscriptionSinkState {
+	/// The subscription is pending and needs to be accepted or rejected
+	/// before utilizing the sink. This is the initial state of the sink.
+	///
+	/// This state contains the request ID of the subscription.
+	Pending(Option<Id<'static>>),
+	/// The subscription was accepted via the [`SubscriptionSink::accept`] call.
+	/// This state contains the future that returns when the unsubscribe method has been called.
```
niklasad1, line 787: I'm still struggling to understand the `Accept which holds` the `rx` which yields once the unsubscribe method has been called.

It just makes me confused as we have `method_sink` and `close_notify` which is sort of something similar which is in the `SubscriptionSink` it just makes everything hard to follow.

Thus, I'm in favor of having everything in the `SubscriptionSink` and just remove `SubscriptionSinkState` because the method sink and close_notify are technically part of that state too.

Alternatively write some nicer abstraction for that.

Thoughts?

lexnv: Sounds like removing it might make our life easier, tho we'd have to be careful internally about how to reason about the states in which we are in.
I could remodel it using `Option<Id<'static>>` for "pending", `Option<watch::Receiver<()>` for "accepted" and treat other cases as "rejected" via the `close_notify` option.

niklasad1: yeah, having different types for the states was one option :D

types/src/error.rs
lexnv: That breaks the orphan rule with https://doc.rust-lang.org/stable/error-index.html#E0210

core/src/server/helpers.rs
```diff
@@ -156,9 +156,10 @@ impl MethodSink {
 
 		if let Err(err) = self.send_raw(json) {
 			tracing::warn!("Error sending response {:?}", err);
+			false
```
Xanewok, line 159: I think a simple two-variant enum could help understand what the `bool` returned from `send_error` specifically means

core/src/server/rpc_module.rs
```diff
@@ -923,14 +899,29 @@ impl SubscriptionSink {
...
 			Some(cn) => cn,
 			None => {
 				return SubscriptionClosed::RemotePeerAborted;
 			}
 		};
 
-		let mut sub_closed = self.unsubscribe.clone();
+		let mut sub_closed = match self.unsubscribe.as_ref() {
+			Some(rx) => rx.clone(),
+			_ => {
```
Xanewok, line 915: style nit: `_ => return ...` would help me understand this is short-circuiting, at a glance

core/src/server/rpc_module.rs
```diff
@@ -740,17 +740,21 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
...
-						id: id.clone().into_owned(),
-						claimed,
-					}));
+						id: Some(id.clone().into_owned()),
+						unsubscribe: None,
+						_claimed: claimed,
+					};
 
-					callback(params, sink, ctx.clone());
+					// The callback returns an empty `SubscriptionError` for improved API ergonomics.
```
Xanewok, line 754: I'm not sure I understand this comment - does it usually do but in this case it's "empty" for the API reasons or is it something else? If it's just an explanation for callback returning `SubscriptionResult`, then I think we're better off leaving that note in the doc comment or near the `F` type near the fn sig above

Xanewok (approved):
That looks like a great ergonomics boost, nice! Left some drive-by notes, don't want to comment on the internals yet.

core/src/server/rpc_module.rs
niklasad1: yeah, we could probably clarify that a bit this but it's just an internal hack to make it's a bit more ergonomic 
for folks to return an error when trying to "accept" a subscription.

if it fails such the params could not be parsed we just send a message to subscription call itself, so this not propagated to the user just something internal hack to circumvent that.

core/src/server/rpc_module.rs
```diff
@@ -775,101 +779,73 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
...
 	method: &'static str,
+	/// Shared Mutex of subscriptions for this method.
+	subscribers: Subscribers,
 	/// Unique subscription.
 	uniq_sub: SubscriptionKey,
-	/// Shared Mutex of subscriptions
-	subscribers: Subscribers,
-	/// Request ID.
-	id: Id<'static>,
+	/// Id of the subscription.
```
niklasad1, line 795:
```suggestion
	/// id of the `subscription call` (i.e. not the same as subscription id) .
```

core/src/server/rpc_module.rs
```diff
@@ -775,101 +779,73 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
...
+	///
+	/// *Note*: Having some value means the subscription was not accepted or rejected yet.
+	id: Option<Id<'static>>,
+	/// Returns when the unsubscribe method has been called.
+	///
+	/// *Note*: Have some values means the subscription was accepted.
+	unsubscribe: Option<watch::Receiver<()>>,
```
niklasad1, line 802: add a type alias for that I reckon that should make the code more readable and the comment should not be necessary

core/src/server/rpc_module.rs
```diff
@@ -1057,7 +1051,10 @@ impl SubscriptionSink {
 
 impl Drop for SubscriptionSink {
 	fn drop(&mut self) {
-		if self.is_active_subscription() {
+		// Subscription was never accepted.
+		if let Some(id) = self.id.take() {
```
niklasad1, line 1055: right, the subscription is only added in `subscribers` if accept has been called successfully?

niklasad1 (approved):
LGTM, some minor comments that are optional to fix.

Would be nice to have some abstraction that decouples the `state` of the `SubscriptionSink` which includes `inner method sink`, `unsubscribe future` and `close_notify` somehow.

I have been thinking that we could just use `tokio::sync::mpsc::unbounded_channel` instead `futures::mpsc::unbounded_channel` and `tokio::sync::Notify` as the sender in tokio has `async fn closed` which returns once the `receiver` has been dropped which would cancel each subscription on that connection.

That would simplify the `SubscriptionSink` slighty :P

core/src/server/rpc_module.rs
lexnv: Yep, `accept` adds the `uniq_sub` to the `self.subscribers` on success. 
However, if the `self.id` was not consumed, neither the `accept` nor `reject` calls were called. Submitting `InvalidParams` in that case keeps backwards compatibility with `PendingSubscription`.
Have moved the comment inside the if statement

jsdw (approved):
LGTM, well done!

Issue #797, opened 2022-06-17
Non-standard JSON-RPC request fields


tgmichel:
Some [http providers](https://github.com/MetaMask/web3-provider-engine/blob/main/index.js#L30) by default include non-standard fields (i.e. `skipCache`) in the request. This results in `Invalid Request` as the `Request` explicitly denies deserialization of unknown fields.

To support this type of requests on `jsonrpsee`, some options:

- Adding a generic parameter to `Request` to allow accepting specific non-standard fields. This generic must implement `Deserialize`, represented as a flatten field in `Request`, and default to an empty type, something like:

```
#[derive(Deserialize, Debug)]
#[serde(deny_unknown_fields)]
pub struct Request<'a, RequestExtension: Deserialize> {
	/// JSON-RPC version.
	pub jsonrpc: TwoPointZero,
	/// Request ID
	#[serde(borrow)]
	pub id: Id<'a>,
	/// Name of the method to be invoked.
	#[serde(borrow)]
	pub method: Cow<'a, str>,
	/// Parameter values of the request.
	#[serde(borrow)]
	pub params: Option<&'a RawValue>,
	#[serde(flatten, default = "EmptyRequestExtension::default")]
        pub extension: RequestExtension,
}

#[derive(Deserialize, Debug, Default)]
struct EmptyRequestExtension;
```

- Other option can be putting `deny_unknown_fields` behind a feature flag, which will optionally accept _any_ non-standard field.

niklasad1:
all right, I see.

Do you know if there's any documentation about these extra fields and what the other servers does with that information?

We could also just remove `deny_unknown_fields`, IIRC we did this to be explicit when testing but question is if we want to handle these "extensions". I guess we don't want to...

tgmichel:
> Do you know if there's any documentation about these extra fields and what the other servers does with that information?

Unforntunately I couldn't find any. I believe in some server implementation there is some cache for response payloads, and this `skipCache` flag indicates wether to use this cache or not.

> question is if we want to handle these "extensions". I guess we don't want to...

I agree, discriminating _valid_ and _invalid_ non-standard fields will likely become a problem (what is a _valid_ non-standard field anyway?) Removing `deny_unknown_fields` (or setting it as default feature) seems reasonable to me.

niklasad1:
Closed by #803

PR #803, opened 2022-06-22
Remove `deny_unknown_fields` from `Request` and `Response`

tgmichel:
Solves #797 

@niklasad1 please let me know if this is prefered over keeping optional `deny_unknown_fields` behind a feature.

### DIFF
--- a/http-server/src/tests.rs
+++ b/http-server/src/tests.rs
@@ -353,6 +353,17 @@ async fn invalid_request_object() {
 }
 
 #[tokio::test]
+async fn unknown_field_is_ok() {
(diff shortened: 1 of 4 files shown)


niklasad1 (approved):
Looks good but we should do the same for the `Response` for consistency.
It's not that likely that other servers impls will add custom fields but better safe than sorry I guess.

I'm happy if we don't have to use more feature flags than necessary :)

lexnv (approved):

jsdw (approved):
A test might be nice but not too important, so LGTM :)

PR #808, opened 2022-06-24
Allow trait bounds to be overridden in macro

### LINKED ISSUE #696: Allow trait bounds to be overridden in macro
First off, apologies if this is wrong; I haven't spent long digging into the macro code here!

So, currently we can generate a client impl using the proc macro like so:

```
#[rpc(client, namespace = "foo")]
pub trait MyRpc<T> {
    #[method(name = "bar")]
    fn method(&self) -> RpcResult<T>;
}
```

I believe this generates something like an `trait MyRpcClient<T>: ClientT where T: Send + Sync + 'static` trait, and an impl for it that means the RPC `Client` impls these methods (very cool btw!).

However, in some cases this general rule of applying `Send + Sync + 'static` to all type params is unnecessarily restrictive. Take the following example:

```
trait Config {
    type Hash: Send + Sync + 'static;
}

#[rpc(client, namespace = "foo")]
pub trait MyRpc<T: Config> {
    #[method(name = "bar")]
    fn method(&self) -> RpcResult<T::Hash>;
}
```

This generates something like `trait MyRpcClient<T>: ClientT where T: Send + Sync + 'static + Config`, but here `T` does not need to be `Send + Sync + 'static` since it's not actually returned or passed in anywhere (we only care about `T::Hash`).

This is quite a common issue in derives too, for instance if we take this struct:

```
#[derive(Clone)]
struct Foo<T> {
    marker: std::marker::PhantomData<T>
}
```

The generated `Clone` impl is something like `impl <T> Clone for Foo<T> where T: Clone {..}`, which is unnecessary since T really doesn't need to be cloneable (it's not instantiated anywhere).

A common solution is to allow people an escape hatch to override the automatically applied trait bounds. The `derivative` library provides this for common traits, so we could impl Clone for all `Foo<T>`'s like so:

```
#[derive(Derivative)]
#[derivative(Clone(bound = ""))]
struct Foo<T> {
    marker: std::marker::PhantomData<T>
}
```

This generates an impl more like `impl <T> Clone for Foo<T> {..}` (we override the Clone bounds to be nothing).

Serde has a similar feature:

```
#[derive(Serialize)]
#[derde(bound = "")]
struct Foo<T> {
    marker: std::marker::PhantomData<T>
}
```

Again, we can manually set the bound to something like "T: MyTrait", or in our case, nothing ("").

Perhaps it would be useful to provide a similar "escape hatch" in jsonrpsee? With such a feature, the earlier example could be made to work with something like:

```
trait Config {
    type Hash: Send + Sync + 'static;
}

#[rpc(client, namespace = "foo", client_bound = "")]
pub trait MyRpc<T: Config> {
    #[method(name = "bar")]
    fn method(&self) -> RpcResult<T::Hash>;
}
```

It would need a little thought;
- does `client_bounds` override _all_ bounds that would be applied, so that they all need to be manual? In this case we might need eg `client_bound = "T::Hash: Serialize + Deserialize"` or something?
- would we need `client_bound` and `server_bound` to allow overriding each?

lexnv:
This PR adds the `server_bounds` and `client_bounds` attributes of the `rpc` macro.

The attributes allow users to specify custom bounds for their traits, as a way to opt-out
of the default `Send + Sync + 'static`  where appropriate.

### Example

Without the custom provided bounds, the RPC macro would add by default 
`Conf : Send + Sync + 'static + jsonrpsee::core::DeserializeOwned` for client implementation and
`Conf : Send + Sync + 'static + jsonrpsee::core::Serialize` for server implementation.
This rule is quite restrictive, as we need just `Conf::Hash` to comply to those rules.
 
```rust
pub trait Config {
	type Hash: Send + Sync + 'static;
}

impl Config for ExampleHash {
	type Hash = Self;
}

#[rpc(server, client, namespace = "foo", client_bounds(Conf::Hash: jsonrpsee::core::DeserializeOwned), server_bounds(Conf::Hash: jsonrpsee::core::Serialize))]
pub trait Rpc<Conf: Config> {
	#[method(name = "bar")]
	fn method(&self) -> Result<Conf::Hash, Error>;
}

```

While at it, add a new example illustrating the custom bounds and a few UI tests to capture
proper error messages.

Closes #696.

### DIFF
--- a/examples/examples/proc_macro_bounds.rs
+++ b/examples/examples/proc_macro_bounds.rs
@@ -1,0 +1,90 @@
+// Copyright 2019-2021 Parity Technologies (UK) Ltd.
+//
+// Permission is hereby granted, free of charge, to any
+// person obtaining a copy of this software and associated
+// documentation files (the "Software"), to deal in the
+// Software without restriction, including without
+// limitation the rights to use, copy, modify, merge,
+// publish, distribute, sublicense, and/or sell copies of
+// the Software, and to permit persons to whom the Software
+// is furnished to do so, subject to the following
+// conditions:
+//
+// The above copyright notice and this permission notice
+// shall be included in all copies or substantial portions
+// of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF
+// ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED
+// TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+// PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT
+// SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+// CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR
+// IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+// DEALINGS IN THE SOFTWARE.
+
+use std::net::SocketAddr;
+
+use jsonrpsee::core::{async_trait, Error};
+use jsonrpsee::proc_macros::rpc;
+use jsonrpsee::ws_client::WsClientBuilder;
+use jsonrpsee::ws_server::{WsServerBuilder, WsServerHandle};
+
+type ExampleHash = [u8; 32];
+
+pub trait Config {
+	type Hash: Send + Sync + 'static;
+}
+
+impl Config for ExampleHash {
+	type Hash = Self;
+}
+
+/// The RPC macro requires `DeserializeOwned` for output types for the client implementation, while the
+/// server implementation requires output types to be bounded by `Serialize`.
+///
+/// In this example, we don't want the `Conf` to be bounded by default to
+/// `Conf : Send + Sync + 'static + jsonrpsee::core::DeserializeOwned` for client implementation and
+/// `Conf : Send + Sync + 'static + jsonrpsee::core::Serialize` for server implementation.
+///
+/// Explicitly, specify client and server bounds to handle the `Serialize` and `DeserializeOwned` cases
+/// just for the `Conf::hash` part.
+#[rpc(server, client, namespace = "foo", client_bounds(Conf::Hash: jsonrpsee::core::DeserializeOwned), server_bounds(Conf::Hash: jsonrpsee::core::Serialize))]
+pub trait Rpc<Conf: Config> {
+	#[method(name = "bar")]
+	fn method(&self) -> Result<Conf::Hash, Error>;
+}
+
+pub struct RpcServerImpl;
+
+#[async_trait]
+impl RpcServer<ExampleHash> for RpcServerImpl {
+	fn method(&self) -> Result<<ExampleHash as Config>::Hash, Error> {
+		Ok([0u8; 32])
+	}
+}
+
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+	tracing_subscriber::FmtSubscriber::builder()
+		.with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
+		.try_init()
+		.expect("setting default subscriber failed");
+
+	let (server_addr, _handle) = run_server().await?;
+	let url = format!("ws://{}", server_addr);
+
+	let client = WsClientBuilder::default().build(&url).await?;
+	assert_eq!(RpcClient::<ExampleHash>::method(&client).await.unwrap(), [0u8; 32]);
+
+	Ok(())
+}
+
+async fn run_server() -> anyhow::Result<(SocketAddr, WsServerHandle)> {
+	let server = WsServerBuilder::default().build("127.0.0.1:0").await?;
+
+	let addr = server.local_addr()?;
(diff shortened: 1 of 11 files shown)


examples/examples/proc_macro_bounds.rs
```diff
@@ -0,0 +1,90 @@
...
+
+use jsonrpsee::core::{async_trait, Error};
+use jsonrpsee::proc_macros::rpc;
+use jsonrpsee::ws_client::WsClientBuilder;
+use jsonrpsee::ws_server::{WsServerBuilder, WsServerHandle};
+
+type ExampleHash = [u8; 32];
+
+pub trait Config {
+	type Hash: Send + Sync + 'static;
```
niklasad1, line 37: IIRC: before/"currently" we add the trait bound the entire trait right?

Can you just verify that we don't add trait bounds on associated types that are used in the RPC trait?

such as:

```rust
pub trait Config {
   type Hash: Send + Sync + 'static;
   type NotUsed;
}

```

niklasad1 (commented):
great.

some UI tests that it actually work as well then this should be ready to go :)

examples/examples/proc_macro_bounds.rs
lexnv: Debugging the proc-macro, it seems that "currently" (without custom `bounds` attribute):

```rust
pub trait Config {
	type Hash: Send + Sync + 'static;
	type NotUsed;
}

#[rpc(server, namespace = "foo")]
pub trait Rpc<Conf: Config> {
	#[method(name = "bar")]
	fn method(&self) -> Result<Conf::Hash, Error>;
}
```

The `rpc` macro implies just: `Conf : Send + Sync + 'static + jsonrpsee :: core :: Serialize`.

Although, this fails to compile due to:

```rust
55  | #[rpc(server, client, namespace = "foo", client_bounds(Conf::Hash: jsonrpsee::core::DeserializeOwned))]
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Serialize` is not implemented for `<Conf as Config>::Hash`
```

niklasad1: Yeah, that's fine I meant with the overrides for the `Hash`

```rust

pub trait Config {
	type Hash: Send + Sync + 'static;
	type NotUsed;
}

#[rpc(server, client, client_bounds(Conf::Hash: jsonrpsee::core::Serialize))]
pub trait Rpc<Conf: Config> {
	#[method(name = "bar")]
	fn method(&self) -> Result<Conf::Hash, Error>;
}
```

Such that it doesn't expand to something like:

```rust
pub trait Rpc<T: Client, C: Config> 
where 
        C::Hash: Serialize,
        C::NotUsed: Send + Sync + Serialize + DeserializeOwned
{
	#[method(name = "bar")]
	fn method(&self) -> Result<Conf::Hash, Error>;
}
```

But as it works in subxt all fine, IIRC we only inspect the parameters and return types anyway add bounds on them so all good.

niklasad1 (approved):
excellent, nice work

some minor suggestions on testing but really good to have this finally

jsdw:
Just wondering; is it possible to define something like:

```rust
#[rpc(server, client, client_bounds(Conf::Hash: jsonrpsee::core::Serialize))]
pub trait Rpc<Conf: Config> {
	fn generic_input<T>(&self, input: T) -> Result<Conf::Hash, Error>;
        fn generic_output<O>(&self, input: T) -> Result<O, Error>;
}
```

And if so, can we control the bounds of `T` and `O`? This may be a dumb question :)

jsdw (approved):
Not sure whether the question above makes sense or not (I can't recall whether it's possible to add generic bounds in that way), but regardless, this looks great; nice job!

lexnv:
Managed to define something similar:

```rust
#[rpc(client, client_bounds(
	Conf::Hash: jsonrpsee::core::DeserializeOwned,
	T: jsonrpsee::core::Serialize + Send + Sync + 'static,
	O: jsonrpsee::core::DeserializeOwned)
)]
pub trait Rpc2<Conf: Config, T, O> {
	#[method(name = "bar")]
	fn generic_input<T>(&self, input: T) -> Result<Conf::Hash, Error>;
	#[method(name = "foo")]
	fn generic_output<T, O>(&self, input: T) -> Result<O, Error>;
}
```
with `T: jsonrpsee::core::Serialize + Send + Sync + 'static` (as it is input for client) and `O: jsonrpsee::core::DeserializeOwned` (output)

niklasad1:
> Just wondering; is it possible to define something like:

```rust
#[rpc(server, client, client_bounds(Conf::Hash: jsonrpsee::core::Serialize))]
pub trait Rpc<Conf: Config> {
	fn generic_input<T>(&self, input: T) -> Result<Conf::Hash, Error>;
        fn generic_output<O>(&self, input: T) -> Result<O, Error>;
}
```

> And if so, can we control the bounds of `T` and `O`? This may be a dumb question :)

The bounds on individual methods doesn't work because the server has  `fn into_rpc(self) -> jsonrpsee::RpcModule<Self>` and then these  "type param" are not `bound` to anything so it won't compile.

It's would need us to find all `individual type params` and generate `fn into_rpc<T1, ... Tn>(self) -> jsonrpsee::RpcModule<Self>
then it would work I guess.

I think the same applies to client trait too i.e, that we don't really parse `type params` on individual methods.

PR #814, opened 2022-07-05
Uniform API for custom headers between clients 

### LINKED ISSUE #809: [clients]: unify APIs to configure custom headers
Currently the WS-client/async client has functionality to configure custom headers to send during the upgrade request/connection.

The HTTP client on the other hand doesn't have this functionality, see https://github.com/paritytech/jsonrpsee/pull/756

So ideally we should have a unified API to `fn set_headers(headers: &[Headers])` or something for that.

lexnv:
Jsonrpsee allows users to set custom headers during the WS handshake (for the ws-clients).

This PR extends the ability to add custom headers for the HTTP clients.

The PR introduces a uniform API for the client builders, relying on `http::HeaderMap`
```rust
pub fn set_headers(mut self, headers: http::HeaderMap) -> Self
```
that behaves as follows:
- ws-client: Custom headers are used only during the handshake
- http-client: Custom headers are submitted with every request

One advantage is that the ws-client and http-client have a uniform API, used for caching or any custom user operations, extending jsonrpsee beyond the realm of the `Substrate`.
One disadvantage is that ws-client uses `soketto` that internally requires an `httparse::Header` object, which leads to conversion from `http::HeaderMap` only if present.

### Next Steps
- [x] Write documentation to illustrate differences between ws/http clients

Closes #809.

### DIFF
--- a/client/http-client/src/client.rs
+++ b/client/http-client/src/client.rs
@@ -39,6 +39,29 @@ use serde::de::DeserializeOwned;
 use tracing_futures::Instrument;
 
 /// Http Client Builder.
+///
+/// # Examples
+///
+/// ```no_run
+///
+/// use jsonrpsee_http_client::HttpClientBuilder;
+///
+/// #[tokio::main]
+/// async fn main() {
+///     // Build custom headers used for every submitted request.
+///     let mut headers = http::HeaderMap::new();
+///     headers.insert("Any-Header-You-Like", http::HeaderValue::from_static("42"));
+///
+///     // Build client
+///     let client = HttpClientBuilder::default()
+///          .set_headers(headers)
+///          .build("wss://localhost:443")
+///          .unwrap();
+///
+///     // use client....
+/// }
+///
+/// ```
 #[derive(Debug)]
 pub struct HttpClientBuilder {
 	max_request_body_size: u32,
--- a/client/http-client/src/client.rs
+++ b/client/http-client/src/client.rs
@@ -46,12 +46,13 @@ use tracing_futures::Instrument;
 /// ```no_run
 ///
 /// use jsonrpsee_http_client::HttpClientBuilder;
+/// use hyper::http::{HeaderMap, HeaderValue};
 ///
 /// #[tokio::main]
 /// async fn main() {
 ///     // Build custom headers used for every submitted request.
-///     let mut headers = http::HeaderMap::new();
-///     headers.insert("Any-Header-You-Like", http::HeaderValue::from_static("42"));
+///     let mut headers = HeaderMap::new();
+///     headers.insert("Any-Header-You-Like", HeaderValue::from_static("42"));
 ///
 ///     // Build client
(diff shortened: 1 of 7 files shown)


client/http-client/src/client.rs
```diff
@@ -47,6 +47,7 @@ pub struct HttpClientBuilder {
 	certificate_store: CertificateStore,
 	id_kind: IdKind,
 	max_log_length: u32,
+	headers: Option<http::HeaderMap>,
```
niklasad1, line 50: should fine with HeaderMap here,

HeaderMap::new() doesn't allocate AFAIK

client/http-client/src/transport.rs
```diff
@@ -100,11 +103,16 @@ impl HttpTransportClient {
 			return Err(Error::RequestTooLarge);
 		}
 
-		let req = hyper::Request::post(&self.target)
+		let mut req = hyper::Request::post(&self.target)
```
niklasad1, line 106: this is not clone right? 

otherwise we could just build the request once and clone it on every call.

however can we pre-compute the "headers" at just set them with `req.headers_mut = self.headers.clone()` on each request?

seems annoying to that on every call.

client/http-client/src/transport.rs
```diff
@@ -90,7 +93,18 @@ impl HttpTransportClient {
...
 		};
-		Ok(Self { target, client, max_request_body_size, max_log_length })
+
+		// Cache request headers: 2 default headers, followed by user custom headers.
+		// Maintain order for headers in case of duplicate keys:
+		// https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.2
+		let mut cached_headers = http::HeaderMap::with_capacity(2 + headers.len());
+		cached_headers.insert(hyper::header::CONTENT_TYPE, hyper::header::HeaderValue::from_static(CONTENT_TYPE_JSON));
+		cached_headers.insert(hyper::header::ACCEPT, hyper::header::HeaderValue::from_static(CONTENT_TYPE_JSON));
+		for (key, value) in headers.iter() {
```
niklasad1, line 103: I think you could take ownership here and skip the clone below

```suggestion
		for (key, value) in headers.into_iter() {
```

niklasad1 (approved):
LGTM,

I like this and I have created this PR https://github.com/paritytech/soketto/pull/60/files which changes soketto to use `HeaderMap` but I need to benchmark it. 

Especially for the client is makes sense as to have this verification as `invalid HTTP headers` can't be created.

`httparse` is really intended for parsing HTTP headers not create them :P

jsdw (approved):
Liking the standardisation! Any idea what the performance impact of this change is?

niklasad1:
For the websocket client is should "only" impact the connection speed and we have no benchmarks for that.... 

On the other hand for the HTTP client the headers are "cloned" for each call which should visible on the benchmarks if we enabled "some custom headers".

PR #815, opened 2022-07-06
ws-server: Handle `soketto::Incoming::Closed` frames

### LINKED ISSUE #813: [ws server]: clean shutdown on `soketto::Incoming::Closed(reason)`
We shouldn't ignore `soketto::Incoming::Closed(reason)`  [here](https://github.com/paritytech/jsonrpsee/blob/master/ws-server/src/server.rs#L397) instead we should terminate the loop early to avoid logging `warn` as `ENOTCONN/ECONNRESET` when the connection in fact was closed cleanly.

lexnv:
Ensure that incoming `Closed` frames are not ignored.

This PR captures the close reason for the connection when detected by the `receiver`.

While at it, the `WS error` log line was polluting internal metrics when clients disconnected suddenly. Have changed it from `warning` to `debug`.

### Logs 
- obtained with a custom client that closes its sender part in the `background_task` 

```rust
2022-07-06T15:21:18.739200Z TRACE soketto::connection: e0000f78: recv, incoming CLOSE: CloseReason { code: 1000, descr: Some("") }
2022-07-06T15:21:18.739225Z DEBUG jsonrpsee_ws_server::server: WS transport: remote peer terminated the connection: 1
```

Closes #813 .

### DIFF
--- a/ws-server/src/server.rs
+++ b/ws-server/src/server.rs
@@ -394,7 +394,12 @@ async fn background_task(
 					match receiver.receive(&mut data).await? {
 						soketto::Incoming::Data(d) => break Ok(d),
 						soketto::Incoming::Pong(_) => tracing::debug!("recv pong"),
-						_ => continue,
+						soketto::Incoming::Closed(reason) => {
+							// Log the close reason for debugging purposes and return the `Closed` error
+							// to avoid logging unnecessary warnings on clean shutdown.
+							tracing::debug!("WS transport: connection: {} closed with reason: {:?} ", conn_id, reason);
+							break Err(SokettoError::Closed);
+						}
 					}
 				}
 			};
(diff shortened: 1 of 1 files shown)


TarikGul (approved):

niklasad1 (approved):
great but I think one of the logs is sufficient.

what about removing https://github.com/paritytech/jsonrpsee/blob/master/ws-server/src/server.rs#L407?

lexnv:
Have removed the `close reason` log instead. I was able to force the `MonitoredError::Selector(SokettoError::Closed)` code-path with multiple client requests and dropped connection, without triggering the `soketto::Incoming::Closed` path. The close reason is also printed by `soketto` for trace-level logs.

PR #823, opened 2022-07-12
Refactor HTTP header validation

### LINKED ISSUE #795: refactor HTTP header verification
I'm struggling to understand `ALWAYS_ALLOWED_HEADERS`; I think that we allow any headers specified in AllowHeaders plus any in `ALWAYS_ALLOWED_HEADERS`.

_Originally posted by @jsdw in https://github.com/paritytech/jsonrpsee/pull/781#discussion_r895569757_

lexnv:
In this PR, the `DEFAULT_ALLOWED_HEADERS` list is trimmed to a minimum number of headers.
The list contains default headers for access control regardless of provided user input.

Furthermore, the `HTTP` server returns the provided list from `verify_headers` for `OPTIONS` methods.

Closes #795.

### DIFF
--- a/core/src/http_helpers.rs
+++ b/core/src/http_helpers.rs
@@ -106,7 +106,7 @@ pub fn read_header_values<'a>(
 }
 
 /// Get the header values from the `access-control-request-headers` header.
-pub fn get_cors_request_headers<'a>(headers: &'a hyper::header::HeaderMap) -> impl Iterator<Item = &str> {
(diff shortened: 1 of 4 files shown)


niklasad1:
@lexnv is there any point of reviewing this PR?

Isn't this going to replaced by the tower middleware/CORS stuff?

lexnv:
Indeed, I will close this as the internal CORS should be replaced by the tower middleware.
Thanks for noticing this!

PR #824, opened 2022-07-13
Benchmarks for different header sizes

lexnv:
This PR adds benchmarks for:
- HTTP round trips with custom headers
- WS handshake with custom headers

### HTTP

|  Benchmark  | 0 KiB header | 1 KiB header | 5 KiB header | 25 KiB header | 100 KiB Header| 
| ---- | ---- | ---- | ---- | ---- |----|
| sync/http_custom_headers_round_trip | **57.518 us** | 51.544 us  | 64.647 us | 64.291 us | 102.75 us
| async/http_custom_headers_round_trip | **65.822 us** | 62.544 us  | 72.057 us | 63.859 us | 131.97 us

### WS

|  Benchmark  | 0 KiB header | 1 KiB header | 2 KiB header | 4 KiB header | 
| ---- | ---- | ---- | ---- | ---- |
| sync/ws_custom_headers_handshake | **86.346 us** | 88.853 us  | 89.603 us | 96.346 us |
| async/ws_custom_headers_handshake | **90.601 us** | 89.546 us  | 89.251 us | 91.919 us |

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -320,7 +320,7 @@ fn http_custom_headers_round_trip(
 	name: &str,
 	request: RequestType,
 ) {
-	for header_size in [1 * KIB, 2 * KIB, 8 * KIB] {
+	for header_size in [KIB, 2 * KIB, 8 * KIB] {
 		let mut headers = HeaderMap::new();
 		headers.insert("key", "A".repeat(header_size).parse().unwrap());
 
@@ -334,7 +334,7 @@ fn http_custom_headers_round_trip(
 /// Bench WS handshake with different header sizes.
 fn ws_custom_headers_handshake(rt: &TokioRuntime, crit: &mut Criterion, url: &str, name: &str, request: RequestType) {
 	let mut group = crit.benchmark_group(request.group_name(name));
-	for header_size in [0, 1 * KIB, 2 * KIB, 4 * KIB] {
+	for header_size in [0, KIB, 2 * KIB, 4 * KIB] {
 		group.bench_function(format!("{}", header_size), |b| {
 			b.to_async(rt).iter(|| async move {
 				let mut headers = HeaderMap::new();
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -1,5 +1,6 @@
 use std::sync::Arc;
 
+use crate::helpers::{ws_handshake, KIB};
 use criterion::*;
 use futures_util::future::{join_all, FutureExt};
 use futures_util::stream::FuturesUnordered;
@@ -101,6 +102,7 @@ trait RequestBencher {
 		ws_concurrent_conn_calls(&rt, crit, &url, "ws_concurrent_conn_calls", Self::REQUEST_TYPE);
 		ws_concurrent_conn_subs(&rt, crit, &url, "ws_concurrent_conn_subs", Self::REQUEST_TYPE);
 		batch_round_trip(&rt, crit, client, "ws_batch_requests", Self::REQUEST_TYPE);
+		ws_custom_headers_handshake(&rt, crit, &url, "ws_custom_headers_handshake", Self::REQUEST_TYPE);
 	}
 
 	fn subscriptions(crit: &mut Criterion) {
@@ -318,9 +320,6 @@ fn http_custom_headers_round_trip(
 	name: &str,
 	request: RequestType,
 ) {
-	// 1 KiB = 1024 bytes
-	const KIB: usize = 1024;
-
 	for header_size in [1 * KIB, 2 * KIB, 8 * KIB] {
 		let mut headers = HeaderMap::new();
 		headers.insert("key", "A".repeat(header_size).parse().unwrap());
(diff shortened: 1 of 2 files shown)


benches/bench.rs
```diff
@@ -307,3 +311,40 @@ fn http_concurrent_conn_calls(rt: &TokioRuntime, crit: &mut Criterion, url: &str
...
+
+/// Bench `round_trip` with different header sizes.
+fn http_custom_headers_round_trip(
+	rt: &TokioRuntime,
+	crit: &mut Criterion,
+	url: &str,
+	name: &str,
+	request: RequestType,
+) {
+	for header_size in [1 * KIB, 2 * KIB, 8 * KIB] {
```
TarikGul, line 323:
```suggestion
	for header_size in [KIB, 2 * KIB, 8 * KIB] {
```
Does it make sense to just have KIB instead of multiplying it by 1?

benches/bench.rs
```diff
@@ -307,3 +311,40 @@ fn http_concurrent_conn_calls(rt: &TokioRuntime, crit: &mut Criterion, url: &str
...
+		let bench_name = format!("{}/{}", name, header_size);
+
+		round_trip(rt, crit, client, &bench_name, request);
+	}
+}
+
+/// Bench WS handshake with different header sizes.
+fn ws_custom_headers_handshake(rt: &TokioRuntime, crit: &mut Criterion, url: &str, name: &str, request: RequestType) {
+	let mut group = crit.benchmark_group(request.group_name(name));
+	for header_size in [0, 1 * KIB, 2 * KIB, 4 * KIB] {
```
TarikGul, line 337:
```suggestion
	for header_size in [0, KIB, 2 * KIB, 4 * KIB] {
```
Same as above

TarikGul (approved):
Small nit, but LGTM

benches/bench.rs
lexnv: That also makes sense for me! Thanks!

benches/bench.rs
lexnv: That also makes sense to me! Thanks!

jsdw:
I'm surprised by the HTTP performance not changing! Are you sure that the benchmarks are actually testinng a handshake/http round trip? I might be reading wrong but offhand it looks like they just build an HttpClient/WsClient, and the `HttpClientBuilder::build()` fn isn't async and so doesn't do much interesting, and the `WsClientBuilder` spawns a background task but I'm not sure whether the bit you're measuring does much?

niklasad1:
> I'm surprised by the HTTP performance not changing! Are you sure that the benchmarks are actually testinng a handshake/http round trip? I might be reading wrong but offhand it looks like they just build an HttpClient/WsClient, and the HttpClientBuilder::build() fn isn't async and so doesn't do much interesting, and the WsClientBuilder spawns a background task but I'm not sure whether the bit you're measuring does much?

The HTTP benchmark is only measuring the actual roundtrip which looks legit to me.
Not sure if the changed `caching` of headers helps here which is created in `HttpClient::new()`

Lemme try benchmarking much bigger headers

jsdw:
I'd be interested in that! And ah yes, the http benchmarking code does more than I initially saw :)

niklasad1 (approved):

benches/bench.rs
```diff
@@ -307,3 +311,48 @@ fn http_concurrent_conn_calls(rt: &TokioRuntime, crit: &mut Criterion, url: &str
...
+				black_box(client.request::<String>(method_name, None).await.unwrap());
+			})
+		});
+	}
+}
+
+/// Bench WS handshake with different header sizes.
+fn ws_custom_headers_handshake(rt: &TokioRuntime, crit: &mut Criterion, url: &str, name: &str, request: RequestType) {
+	let mut group = crit.benchmark_group(request.group_name(name));
+	for header_size in [0, KIB, 2 * KIB, 4 * KIB] {
```
niklasad1, line 345: FWIW, I find it weird that soketto rejects 8 KIB [when it has this](https://github.com/paritytech/soketto/blob/master/src/handshake/server.rs#L24)

benches/bench.rs
```diff
@@ -307,3 +311,48 @@ fn http_concurrent_conn_calls(rt: &TokioRuntime, crit: &mut Criterion, url: &str
...
+
+/// Bench `round_trip` with different header sizes.
+fn http_custom_headers_round_trip(
+	rt: &TokioRuntime,
+	crit: &mut Criterion,
+	url: &str,
+	name: &str,
+	request: RequestType,
+) {
+	let method_name = request.methods()[0];
```
niklasad1, line 323: @lexnv

I changed that we just test this against one specific JSON-RPC call i.e, we don't really care about which call it is.
The only we care about is the overhead of the headers ^^

benches/bench.rs
lexnv: Yep, I think that was the reason why I had to limit the benchmarks to 8 KiB unfortunately.

benches/bench.rs
lexnv: This makes sense to me since we only care about the header benchmarking, it should also reduce the bench time 😄

PR #831, opened 2022-07-25
HTTP server with `tower` interaction

### LINKED ISSUE #841: Support tower_http Service's in front of our Http server
By doing this, we can remove CORS support from the core crate any rely on an external middleware instead. (It also gives users the ability to add their own "proper" http middleware, for example to convert a rest API into jsonrpsee requests or what have you)

lexnv:
This PR allows users to set custom tower `ServiceBuilder`s as middleware
to layer the RPC service.

Exposing such functionality enables users to rely on crates that build upon `tower`,
such as `tower_http` to crate a default layered filtering system (ie, enable CORS from
tower_http, or extensive debugging upon receiving a `response` / `request`).

### Tower HTTP Example

The server is constructed with the tower's `ServiceBuilder` with the following layer on top:
```rust
.layer(
	TraceLayer::new_for_http()
		.on_body_chunk(|chunk: &Bytes, latency: Duration, _: &tracing::Span| {
			tracing::trace!(size_bytes = chunk.len(), latency = ?latency, "sending body chunk")
		})
		.make_span_with(DefaultMakeSpan::new().include_headers(true))
		.on_response(DefaultOnResponse::new().include_headers(true).latency_unit(LatencyUnit::Micros)),
)
```

Output with `RUST_LOG=trace`:

```bash
2022-07-27T12:48:46.665907Z TRACE request{method=POST
uri=/ version=HTTP/1.1 headers={"content-type": "application/json",
"accept": "application/json", "host": "127.0.0.1:9935", "content-length": "45"}}:
tower_http: sending body chunk size_bytes=38 latency=347.166µs
```

### Next Steps
- [x] Documentation
- [x] Enable service with `tower_http`

Closes #841


http-server/src/server.rs
```diff
@@ -344,6 +344,23 @@ impl<M> Builder<M> {
...
+			max_request_body_size: self.max_request_body_size,
+			max_response_body_size: self.max_response_body_size,
+			batch_requests_supported: self.batch_requests_supported,
+			resources: self.resources,
+			tokio_runtime: self.tokio_runtime,
+			middleware: self.middleware,
+			max_log_length: self.max_log_length,
+			health_api: self.health_api,
+		}
+	}
```
jsdw, line 363: Could you just convert the builder straight into the `RPSeeServerSvc` without needing this "pre-build" step (which I assume doesn't really make any sense outside of creating a service?)

(also we'd want to tweak the names ultimately; `RPSeeServerSvc` is quite hideous :D but that's not important here!)

http-server/src/server.rs
```diff
@@ -383,11 +400,208 @@ impl Future for ServerHandle {
...
+	///
+	/// Logs bigger than this limit will be truncated.
+	max_log_length: u32,
+	/// Whether batch requests are supported by this server or not.
+	batch_requests_supported: bool,
+}
+
+impl<M: Middleware> RPSeeSvcData<M> {
+	/// Default behavior for RPSee handling of requests.
+	async fn handle_request(
```
jsdw, line 432: Is all of this copy-pasted from the http impl, and if so, is there a way to avoid the duplication?

examples/examples/tower_http.rs
```diff
@@ -0,0 +1,102 @@
...
+	module.register_method("say_hello", |_, _| Ok("lo"))?;
+
+	println!("[run_server]: Creating RPC service");
+	let rpc_service = HttpServerBuilder::new().set_middleware(Timings).pre_build().to_service(module);
+
+	println!("[run_server]: Tower builder");
+	let service_builder = tower::ServiceBuilder::new()
+		// Mark the `Authorization` request header as sensitive so it doesn't show in logs
+		.layer(SetSensitiveRequestHeadersLayer::new(once(hyper::header::AUTHORIZATION)))
+		.service(rpc_service);
```
jsdw, line 93: Ah ok cool, so basically by taking the HttpServer into something implementing hyper::Servicve, we can layer in arbitrary tower middleware by the looks of it! Offhand that makes sense to me as one way to add middleware. I wonder if we could get ride of the CORS stuff and rely on pre-existing middleware for it (or if not, move our cors stuff into a thing implementing the right trait to use as middleware)

niklasad1: https://docs.rs/tower-http/latest/tower_http/cors/index.html looks quite promising but we need to consolidate the ws stuff anyway.

unless we want to implement some wrapper over soketto or something...

examples/examples/tower_http.rs
```diff
@@ -0,0 +1,130 @@
...
+	let make_service = make_service_fn(move |_| {
+		async move {
+			let mut module = RpcModule::new(());
+			module.register_method("say_hello", |_, _| Ok("lo")).unwrap();
+
+			println!("[run_server]: Creating RPC service");
+			let rpc_svc = HttpServerBuilder::new().set_middleware(Timings).to_service(module).unwrap();
+
+			println!("[run_server]: Tower builder");
+			let tower_svc = tower::ServiceBuilder::new()
```
niklasad1, line 102: Is there a reason why we shouldn't enable this be the default on the HTTP server i.e. `tower middleware`?
I reckon that most folks would find it useful to debug the entire HTTP request/response.

However, the logging/middleware stuff becomes quite complicated because we got:

1. RpcMiddleware
2. Tracing of RPC calls
3. HTTP middleware

It's probably fine but a bit overwhelming to understand the difference :P

http-server/src/server.rs
```diff
@@ -344,6 +344,26 @@ impl<M> Builder<M> {
 			health_api: self.health_api,
 		})
 	}
+
+	/// Returns a service that can be utilised with `tower` compatible crates.
+	pub fn to_service(self, methods: impl Into<Methods>) -> Result<TowerService<M>, Error> {
```
niklasad1, line 349: it might be better to implement this on the `Server struct` and not the builder.

Then you don't need the `Option` as `remote addr`

examples/examples/tower_http.rs
jsdw: I agree with the middleware being confusing bit. Some of it just a naming thing; calling the logging stuff Middleware was probably a mistake, since it isn't really; it's just some basic hooks to log details from and can't effect any change.

There are different sorts of middleware support that are interesting:

- HTTP Service (like this PR; just an obvious thing to support since tower is part of hyper already. This is the lowest hanging fruit I guess. CORS could lilve here, and we keep more basic IP filtering in jsonrpsee itself)
- WS Service (use tower service trait and allow things to be put in front of each request/response pair?)
- Request/Response Service (Like, regardless of HTTP or WS, a general service which would receive requests/responses and be able to act on/modify them? Maybe this is strictly just better than a WS specific thing?)
- The "manual" approach; allow some custom param `T` to be passed into call/subscription handlers so that the user has a way to get custom information in from the outside (worst case by manually hooking up an http service and manually calling things on RpcModule with said custom state; maximum flexibility but hardest to make use of).

http-server/src/server.rs
lexnv: Have managed to tidy up the API, without requiring us to have `Options` on either the service or server, let me know what you think of it 😄

niklasad1: looks really good

examples/examples/tower_http.rs
```diff
@@ -0,0 +1,130 @@
...
+				.layer(SetSensitiveRequestHeadersLayer::new(once(hyper::header::AUTHORIZATION)))
+				.timeout(Duration::from_secs(2))
+				.service(rpc_svc);
+
+			Ok::<_, Infallible>(tower_svc)
+		}
+	});
+
+	let addr = SocketAddr::from(([127, 0, 0, 1], 9935));
+	tokio::spawn(async move { Server::bind(&addr).serve(make_service).await });
```
niklasad1, line 124: this API is very low-level, can't we bind the addr and start the hyper server under the hood? 

I guess you wanted to avoid that with the `Option<SocketAddr>` but I think this is too low-level for must users.
Would be nice to just "plug-in" the tower service in the builder and then just call `server.start` IMO, is that hard or why did you end up with this approach?

lexnv: I think these trait bounds from [hyper 0.14.2](https://github.com/hyperium/hyper/blob/0ff6213be2c3b93a5cb0fa6caf169622131f8466/src/server/server.rs#L536) make it a lot harder than I expected to plug in the `tower::ServiceBuilder<_>` into the HttpBuilder or start method directly.

I agree this is a bit too low-level and takes a lot of tower knowledge to make it work.

I'll keep on trying to find a better way to pass the ServiceBuilder to our server 😄 

A direct passthrough version might look like the following, but still, there are some errors generated by `ConnStreamExec` from the above link

```rust
pub fn start_with_builder<T, U>(mut self, methods: impl Into<Methods>, builder: tower::ServiceBuilder<T>)
where
T: Layer<U>,
<T as Layer<U>>::Service: Send + Service<hyper::Request<Body>>,
<<T as Layer<U>>::Service as Service<hyper::Request<Body>>>::Error: Send + Sync + StdError,
```

niklasad1: much nicer now :muscle:

examples/examples/tower_http.rs
```diff
@@ -0,0 +1,113 @@
...
+		.layer(SetSensitiveRequestHeadersLayer::new(once(hyper::header::AUTHORIZATION)))
+		.timeout(Duration::from_secs(2));
+
+	let server = HttpServerBuilder::new().set_logger(Timings).build("127.0.0.1:0".parse::<SocketAddr>()?).await?;
+	let addr = server.local_addr()?;
+
+	let mut module = RpcModule::new(());
+	module.register_method("say_hello", |_, _| Ok("lo")).unwrap();
+
+	let handler = server.start_with_builder(module, builder)?;
```
niklasad1, line 110: I suggest to rename this method to `Server::start_with_middleware` or something like that because I kind of want not confuse it with the `HttpServerBuilder`

Besides that it looks great now, perhaps remove the `HttpLogger` stuff in the example to keep it easier to follow that it's not really related to the middleware or rename the example then document that the `tower middleware` and `HttpLogger` are not mutually exclusive.

http-server/src/server.rs
```diff
@@ -539,7 +591,14 @@ pub struct TowerService<L> {
...
 impl<L: Logger> hyper::service::Service<hyper::Request<hyper::Body>> for TowerService<L> {
 	type Response = hyper::Response<hyper::Body>;
-	type Error = hyper::Error;
+
+	// NOTE(lexnv): The `handle_request` method returns `Result<_, Infallible>`.
+	// This is because the RPC service will always return a valid HTTP response (ie return `Ok(_)`).
+	//
+	// The following associated type is required by the `impl<B, U, L: Logger> Server<B, L>` bounds.
+	// It satisfies the server's bounds when the `tower::ServiceBuilder<B>` is not set (ie `B: Identity`).
+	type Error = Box<(dyn StdError + Send + Sync + 'static)>;
```
niklasad1, line 600:
```suggestion
	type Error = Box<dyn StdError + Send + Sync + 'static>;
```

http-server/src/server.rs
```diff
@@ -578,80 +640,28 @@ pub struct Server<L = ()> {
 	tokio_runtime: Option<tokio::runtime::Handle>,
 	logger: L,
 	health_api: Option<HealthApi>,
+	service_builder: tower::ServiceBuilder<B>,
 }
 
-impl<L: Logger> Server<L> {
+impl<B, U, L: Logger> Server<B, L>
+where
+	B: Layer<TowerService<L>> + Send + 'static,
```
niklasad1, line 648: :smile: 

hopefully, we don't have to bother with these trait bounds anymore

niklasad1 (approved):
great, we should probably check how much overhead tower introduces.

follow-up PR remove the CORS stuff and rely on tower?!

http-server/src/server.rs
```diff
@@ -51,15 +54,18 @@ use jsonrpsee_core::tracing::{rx_log_from_json, rx_log_from_str, tx_log_from_str
 use jsonrpsee_core::TEN_MB_SIZE_BYTES;
 use jsonrpsee_types::error::{ErrorCode, ErrorObject, BATCHES_NOT_SUPPORTED_CODE, BATCHES_NOT_SUPPORTED_MSG};
 use jsonrpsee_types::{Id, Notification, Params, Request};
+use serde::de::StdError;
```
niklasad1, line 57: you could most likely import `std::error::Error as StdError` instead...

http-server/src/server.rs
```diff
@@ -609,7 +761,7 @@ struct ProcessValidatedRequest<L: Logger> {
 /// Process a verified request, it implies a POST request with content type JSON.
 async fn process_validated_request<L: Logger>(
 	input: ProcessValidatedRequest<L>,
-) -> Result<hyper::Response<hyper::Body>, HyperError> {
+) -> Result<hyper::Response<hyper::Body>, Infallible> {
```
jsdw, line 764: Is there a reason to return a `Result<_, Infallible>` at all, and not just scrap the `Result` entirely?

http-server/src/server.rs
```diff
@@ -384,9 +439,185 @@ impl Future for ServerHandle {
...
+	}
+}
+
+/// JsonRPSee service compatible with `tower`.
+///
+/// # Note
+/// This is similar to [`hyper::service::service_fn`].
+#[derive(Debug)]
+pub struct TowerService<L> {
+	inner: ServiceData<L>,
```
jsdw, line 589: Is there a benefit to having this inner `ServiceData<L>` as well as the outer `TowerService<L>` instead of just having `TowerService<L>` have all of the props etc from `ServiceData` in it?

http-server/src/server.rs
```diff
@@ -384,9 +439,185 @@ impl Future for ServerHandle {
...
+
+	type Future = Pin<Box<dyn Future<Output = Result<Self::Response, Self::Error>> + Send>>;
+
+	/// Opens door for back pressure implementation.
+	fn poll_ready(&mut self, _: &mut Context) -> Poll<Result<(), Self::Error>> {
+		Poll::Ready(Ok(()))
+	}
+
+	fn call(&mut self, request: hyper::Request<hyper::Body>) -> Self::Future {
+		let data = self.inner.clone();
```
jsdw, line 610: Do we have to worry about performance at all cloning all of these?

(It looked like a bunch of them were cloned previously so I'm guessing not).

And ah ok, is the inner struct so that you can clone it here? I guess you can just clone `&mut self` though and have just one struct?

http-server/src/server.rs
```diff
@@ -384,9 +439,185 @@ impl Future for ServerHandle {
...
+		// Note that `handle_request` will never return error.
+		// The dummy error is set in place to satisfy the server's trait bounds regarding the
+		// `tower::ServiceBuilder` and the error will never be mapped.
+		Box::pin(data.handle_request(request).map_err(|_| "".into()))
```
jsdw, line 614: If `handle_request` didn't return a result I guess you could just do `Ok(data.handle_request(..))` instead?

http-server/src/server.rs
jsdw: Do you mean there is some way to ismplify them?

Nit: If not, perhaps they should just be on the `start` method and not on the whole impl block (not a big deal but just that `local_addr()` doesn't need them)

jsdw: I'm almost wondering whether it's worth doing something like this (assuming it's possible):

```rust
trait HttpServiceLayer {} // name up for debate; just a placeholder!

impl <B, U> HttpServiceLayer for B where
   // ... complicated trait bounds copied here
```

And then we'd just have to do `impl <B: HttpServiceLayer, L: Logger> Server<B, L> { .... }` here.

Also then we might use the same bounds on `set_middleware` just to catch any issues with bad user submitted middleware sooner (optional though; better or not?).

niklasad1: I mean that the these trait bounds are super hard to get right and takes a lot of time to figure out what the compilation comes from :P

http-server/src/server.rs
niklasad1: I think the hyper service fn requires `Result` but I could be wrong.

http-server/src/server.rs
niklasad1: I think it's similar as it was before because of the `FnMut` closure in the hyper service fn

http-server/src/server.rs
niklasad1: orphan rule?

http-server/src/server.rs
jsdw: If that's true it may still be cleaner to pass `Ok(result)` or something than have the "wrong" type signature on the whole fn? But shrug :)

http-server/src/server.rs
niklasad1: good idea, fixed now.

examples/examples/tower_http.rs
```diff
@@ -0,0 +1,87 @@
+// Copyright 2019-2021 Parity Technologies (UK) Ltd.
```
niklasad1, line 1: I think we should rename this to `http_middleware`, thoughts?

http-server/src/server.rs
niklasad1: Fixed

http-server/src/server.rs
jsdw: Oooh gotcha!

jsdw: (I had a go at splitting the trait bounds into a separate trait, but it was rather messy so I'm happy with it as is!)

jsdw (approved):
LGTM, nice work Alex!

My remaining comments are nits and can be ignored!

http-server/src/server.rs
niklasad1: Look below it implements `hyper::service::Service<hyper::Request<hyper::Body>>` which can't be done because both the trait and type are not defined by "us".

So one need the `newtype` to implement that trait AFAIU.

http-server/src/server.rs
```diff
@@ -384,9 +437,177 @@ impl Future for ServerHandle {
...
+	type Future = Pin<Box<dyn Future<Output = Result<Self::Response, Self::Error>> + Send>>;
+
+	/// Opens door for back pressure implementation.
+	fn poll_ready(&mut self, _: &mut Context) -> Poll<Result<(), Self::Error>> {
+		Poll::Ready(Ok(()))
+	}
+
+	fn call(&mut self, request: hyper::Request<hyper::Body>) -> Self::Future {
+		let data = self.inner.clone();
+		// Note that `handle_request` will never return error.
```
niklasad1, line 601:
```suggestion
```

http-server/src/server.rs
```diff
@@ -384,9 +437,176 @@ impl Future for ServerHandle {
...
+	type Future = Pin<Box<dyn Future<Output = Result<Self::Response, Self::Error>> + Send>>;
+
+	/// Opens door for back pressure implementation.
+	fn poll_ready(&mut self, _: &mut Context) -> Poll<Result<(), Self::Error>> {
+		Poll::Ready(Ok(()))
+	}
+
+	fn call(&mut self, request: hyper::Request<hyper::Body>) -> Self::Future {
+		let data = self.inner.clone();
+		// The dummy error is set in place to satisfy the server's trait bounds regarding the
```
niklasad1, line 601:
```suggestion
```

http-server/src/server.rs
```diff
@@ -384,9 +437,175 @@ impl Future for ServerHandle {
...
+	type Future = Pin<Box<dyn Future<Output = Result<Self::Response, Self::Error>> + Send>>;
+
+	/// Opens door for back pressure implementation.
+	fn poll_ready(&mut self, _: &mut Context) -> Poll<Result<(), Self::Error>> {
+		Poll::Ready(Ok(()))
+	}
+
+	fn call(&mut self, request: hyper::Request<hyper::Body>) -> Self::Future {
+		let data = self.inner.clone();
+		// `tower::ServiceBuilder` and the error will never be mapped.
```
niklasad1, line 601:
```suggestion
```

http-server/src/server.rs
```diff
@@ -409,14 +627,33 @@ pub struct Server<L = ()> {
...
-impl<L: Logger> Server<L> {
+impl<B, L> Server<B, L> {
 	/// Returns socket address to which the server is bound.
 	pub fn local_addr(&self) -> Result<SocketAddr, Error> {
 		self.local_addr.ok_or_else(|| Error::Custom("Local address not found".into()))
 	}
+}
 
+// Required trait bounds for the middleware service.
+// TODO: consider introducing a marker trait for this.
```
niklasad1, line 641:
```suggestion
```

PR #837, opened 2022-08-02
client: Implement `notify_on_disconnect`

### LINKED ISSUE #770: [async client]: add `notify_on_disconnect` hook when the connection is closed
We already have an API for `is_connected` but I imagine it would be easier for folks to have future/callback that fires when the connection has been closed instead of polling this function in an async context.

So would nice neat to have an API:

```
/// Returns only if the connection has been disconnected.
///
/// Should be cancel-safe if possible
async fn notify_on_disconnect() { } 
```

lexnv:
This PR introduces the client's `notify_on_disconnect` functionality.

When called, the method completes when the client is disconnected 
or the client's background task encountered an error.
The method is cancel-safe, as the client disconnects or background task
errors are unrecoverable.

### Testing Done

- Server: Started a process with WS server that terminates in 60 seconds
- Client: Submitted a request (to verify connectivity) and called `client.notify_on_disconnect().await;`

### Logs

Server
```
2022-08-02T10:00:37.479952Z DEBUG jsonrpsee_ws_server::server: recv pong
2022-08-02T10:00:37.601751Z  INFO ws_server: Server Shutdown
```

Client
```
2022-08-02T10:00:37.604050Z ERROR jsonrpsee_core::client::async_client: Error: Connection(Closed) terminating client
2022-08-02T10:00:37.604176Z  WARN jsonrpsee_core::client::async_client: Transport(WebSocket connection error: connection closed

Caused by:
    connection closed)
2022-08-02T10:00:37.604617Z TRACE soketto::connection: ebce45eb: closing connection
2022-08-02T10:00:37.604655Z TRACE soketto::connection: ebce45eb: send: (Close (fin 1) (rsv 000) (mask (1 33031046)) (len 2))
2022-08-02T10:00:37.605058Z TRACE soketto::connection: ebce45eb: Sender flushing connection
2022-08-02T10:00:37.605205Z TRACE mio::poll: deregistering event source from poller
2022-08-02T10:00:37.605375Z  INFO ws_client: Client stop...
```

### Next
- [x] Handle tokio::SendError and futures_utils::SendError conversion
- [x] Handle WS part

Closes #770.

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -16,7 +16,6 @@ thiserror = "1"
 serde = { version = "1.0", default-features = false, features = ["derive"] }
 serde_json = { version = "1", features = ["raw_value"] }
 tracing = "0.1.34"
-tokio = { version = "1.16", features = ["sync"] }
 
 # optional deps
 arrayvec = { version = "0.7.1", optional = true }
@@ -29,6 +28,7 @@ rustc-hash = { version = "1", optional = true }
 rand = { version = "0.8", optional = true }
 soketto = { version = "0.7.1", optional = true }
 parking_lot = { version = "0.12", optional = true }
+tokio = { version = "1.16", optional = true }
 wasm-bindgen-futures = { version = "0.4.19", optional = true }
 futures-timer = { version = "3", optional = true }
 globset = { version = "0.4", optional = true }
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -52,7 +52,7 @@ server = [
 	"unicase",
 	"http",
 ]
-client = ["futures-util/sink", "futures-channel/sink", "futures-channel/std"]
+client = ["futures-util/sink", "futures-channel/sink", "futures-channel/std", "tokio/sync"]
 async-client = [
 	"async-lock",
 	"client",
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -29,7 +29,6 @@ rand = { version = "0.8", optional = true }
 soketto = { version = "0.7.1", optional = true }
 parking_lot = { version = "0.12", optional = true }
 tokio = { version = "1.16", optional = true }
-tokio-stream = { version = "0.1.7", optional = true }
 wasm-bindgen-futures = { version = "0.4.19", optional = true }
 futures-timer = { version = "3", optional = true }
 globset = { version = "0.4", optional = true }
@@ -61,7 +60,6 @@ async-client = [
 	"tokio/macros",
 	"tokio/rt",
 	"tokio/sync",
-	"tokio-stream",
 	"tracing-futures",
 	"futures-timer",
 ]
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -29,6 +29,7 @@ rand = { version = "0.8", optional = true }
 soketto = { version = "0.7.1", optional = true }
 parking_lot = { version = "0.12", optional = true }
 tokio = { version = "1.16", optional = true }
+tokio-stream = { version = "0.1.7", optional = true }
 wasm-bindgen-futures = { version = "0.4.19", optional = true }
 futures-timer = { version = "3", optional = true }
 globset = { version = "0.4", optional = true }
@@ -60,6 +61,7 @@ async-client = [
 	"tokio/macros",
 	"tokio/rt",
 	"tokio/sync",
+	"tokio-stream",
 	"tracing-futures",
 	"futures-timer",
 ]
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -16,6 +16,7 @@ thiserror = "1"
 serde = { version = "1.0", default-features = false, features = ["derive"] }
 serde_json = { version = "1", features = ["raw_value"] }
 tracing = "0.1.34"
+tokio = { version = "1.16", features = ["sync"] }
 
 # optional deps
 arrayvec = { version = "0.7.1", optional = true }
@@ -28,7 +29,6 @@ rustc-hash = { version = "1", optional = true }
 rand = { version = "0.8", optional = true }
 soketto = { version = "0.7.1", optional = true }
 parking_lot = { version = "0.12", optional = true }
-tokio = { version = "1.16", optional = true }
 wasm-bindgen-futures = { version = "0.4.19", optional = true }
 futures-timer = { version = "3", optional = true }
 globset = { version = "0.4", optional = true }
@@ -52,7 +52,7 @@ server = [
 	"unicase",
 	"http",
 ]
-client = ["futures-util/sink", "futures-channel/sink", "futures-channel/std", "tokio/sync"]
+client = ["futures-util/sink", "futures-channel/sink", "futures-channel/std"]
 async-client = [
 	"async-lock",
 	"client",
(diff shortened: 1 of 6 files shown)


core/src/client/async_client/mod.rs
```diff
@@ -232,6 +248,16 @@ impl Client {
 		*err_lock = next_state;
 		err
 	}
+
+	/// Completes when the client is disconnected or the client's background task encountered an error.
+	///
+	/// # Cancel safety
+	///
+	/// This method is cancel safe. Once the client is disconnected, it stays disconnected forever and all
+	/// future calls to this method will return immediately.
```
jsdw, line 257: I love the thought to cancel safety, though most of the text isn't really related to cancel safety so I'd move it up; something like:

```suggestion
	/// Completes when the client is disconnected or the client's background task encountered an error. If the client is already disconnected, the future produced by this method will complete immediately.
	///
	/// # Cancel safety
	///
	/// This method is cancel safe.
```

core/src/client/async_client/mod.rs
```diff
@@ -633,10 +637,12 @@ async fn background_task<S, R>(
 	});
 	futures_util::pin_mut!(backend_event);
 
+	let mut frontend_stream = tokio_stream::wrappers::ReceiverStream::new(frontend);
+
 	// Place frontend and backend messages into their own select.
 	// This implies that either messages are received (both front or backend),
 	// or the submitted ping timer expires (if provided).
-	let next_frontend = frontend.next();
```
niklasad1, line 639: can't we use `frontend.recv()` to avoid having the `tokio_stream` dependency?

niklasad1 (approved):

core/src/client/async_client/mod.rs
```diff
@@ -232,11 +231,15 @@ impl Client {
...
-impl Drop for Client {
-	fn drop(&mut self) {
-		self.to_back.close_channel();
+	/// Completes when the client is disconnected or the client's background task encountered an error.
+	/// If the client is already disconnected, the future produced by this method will complete immediately.
+	///
+	/// # Cancel safety
+	///
+	/// This method is cancel safe.
+	pub async fn notify_on_disconnect(&self) {
```
niklasad1, line 241: I think I suggested this name but I find it a bit long.

Maybe `on_disconnect` or `disconnect` are better.

core/src/client/async_client/mod.rs
```diff
@@ -232,11 +231,15 @@ impl Client {
...
 	}
-}
 
-impl Drop for Client {
-	fn drop(&mut self) {
-		self.to_back.close_channel();
+	/// Completes when the client is disconnected or the client's background task encountered an error.
+	/// If the client is already disconnected, the future produced by this method will complete immediately.
+	///
+	/// # Cancel safety
```
niklasad1, line 238: this is really great, we should adopt this idiom IMO to document cancel safety.

most things here are cancel-safe but really great to be explicit.

jsdw: Agreed! It's so useful to know :)

core/Cargo.toml
```diff
@@ -28,7 +29,6 @@ rustc-hash = { version = "1", optional = true }
 rand = { version = "0.8", optional = true }
 soketto = { version = "0.7.1", optional = true }
 parking_lot = { version = "0.12", optional = true }
-tokio = { version = "1.16", optional = true }
```
jsdw, line 31: Is there a downside to tokio becoming non-optional?

niklasad1: bigger dependency tree :(

hmm, I think it's because of `TokioSendError` impl

core/src/client/async_client/mod.rs
```diff
@@ -633,10 +637,21 @@ async fn background_task<S, R>(
...
 
+	let frontend_event = futures_util::stream::unfold(frontend, |mut frontend| async {
+		let res = frontend.recv().await;
+		if let Some(res) = res {
+			Some((res, frontend))
+		} else {
+			None
+		}
+	});
+	futures_util::pin_mut!(frontend_event);
```
jsdw, line 648: Just curious; so this crates a `Stream` from `frontend`. Why do we need it though? Is it not possible to use `frontend.recv()` instead of `frontend.next()` below?

core/src/error.rs
```diff
@@ -134,6 +136,31 @@ pub enum Error {
...
+		// The error wraps the value passed to `send`.
+		// Tokio's `send` can only fail if the receive half of the channel is closed
+		// (either `close` was called, or the receiver was dropped).
+		Error::Internal(InternalError::Disconnected)
+	}
+}
+
+impl From<futures_channel::mpsc::SendError> for Error {
+	fn from(err: futures_channel::mpsc::SendError) -> Self {
+		if err.is_full() {
```
jsdw, line 156: I think tokio hid the enum because they could add other states and require other checks in a non breaking change.

I wonder whether it's worth documenting/future-proofing this by being more explicit, like:

```rust
if err.is_full() { Full }
else if err.is_disconnected() { Disconnected }
else { Unknown }
```

what do you think?

core/src/error.rs
```diff
@@ -229,3 +256,27 @@ impl From<hyper::Error> for Error {
 		Error::Transport(hyper_err.into())
 	}
 }
+
+/// Internal error for the frontend/backend channel.
+#[derive(Debug)]
+pub enum InternalError {
```
jsdw, line 262: I like having our own error instead of using some SendError actually, even if it adds some code.

core/Cargo.toml
jsdw: I guess I'm wondering; is it worth the extra dependencies etc to take this approach rather than the original one of an extra Notify thing?

I think I like this approach overall but would be good to get your opinion too on which you prefer, now you've been both!

jsdw:
One other thing; can we add a couple of tests for this new function?

core/Cargo.toml
lexnv: Have picked the original approach and utilized `futures_channel` instead of `tokio` to avoid having an extra dependency. Mainly because the tokio's sender channel cannot be closed directly. It is only closed when all senders are dropped, and we have it cloned in the subscriptions.

For now, utilizing this approach seems like a good tradeoff between dependencies/code to maintain. We can in the future change it under the hood, as it would likely not imply any breaking changes.

core/src/client/async_client/mod.rs
```diff
@@ -186,16 +197,18 @@ impl ClientBuilder {
...
-			background_task(sender, receiver, from_front, err_tx, max_notifs_per_subscription, None).await;
+			background_task(sender, receiver, from_front, err_tx, max_notifs_per_subscription, None, on_close_tx).await;
 		});
 		Client {
 			to_back,
 			request_timeout: self.request_timeout,
 			error: Mutex::new(ErrorFromBack::Unread(err_rx)),
 			id_manager: RequestIdManager::new(self.max_concurrent_requests, self.id_kind),
 			max_log_length: self.max_log_length,
+			notify: Mutex::new(Some(on_close_rx)),
```
niklasad1, line 211: no more `tokio::sync::Notify`?
doesn't it compile for WASM or what's the reason?

jsdw: This perhaps?: https://github.com/paritytech/jsonrpsee/pull/837#discussion_r938031389

lexnv: Yep, that's the explanation, mainly to reduce dependencies on tokio 🙏

jsdw (approved):

PR #845, opened 2022-08-03
Refactor `Middleware` to `Logger`

### LINKED ISSUE #843: Rename HttpMiddleware/WsMiddleware to avoid confusion
"Middleware" is a term that has some certain meaning/connotations around being able to intercept and modify requests/responses. The things that we call middleware have no ability to effect the requests/responses and really just exist as hooks for metrics and logging purposes.

Let's call it something like `RpcMetrics` or what have you. This leaves the door open for "real" middleware, too, at some point in the future.

lexnv:
Change the name of the Middleware traits (HTTP / WS) to better reflect
its purposes.

Closes #843.


http-server/src/server.rs
```diff
@@ -507,7 +507,7 @@ impl<M: Middleware> Server<M> {
 								let origin = return_origin_if_different_from_host(request.headers()).cloned();
 								let mut res = process_validated_request(ProcessValidatedRequest {
 									request,
-									middleware,
+									metrics: metrics,
```
jsdw, line 510:
```suggestion
									metrics,
```

jsdw (approved):
I'm not certain on the word `Metrics` on it's own but I could live with it! `Logger` was another word that came to mind :)

niklasad1:
yeah, I'm not a fan of using `HttpMetrics` either we should really have like `HttpRpcLogging/Metrics` because we need to make it clear that you can't really do any logging/metrics on the underlying HTTP/WS instead it gives timings on the specific JSON-RPC calls except some edge-cases just as `on_connect` but perhaps that can be removed with `tower middleware`...

jsdw:
I mean, since `on_connect` and `on_disconnect` exist, we do allow some logging based on the whole http/ws connection too?

My fave really would be `HttpLogger` or `WsLogger`; the user would then impl eg `HttpLogger` for some struct and it would do some sort of logging of http things (mainly the rpc request bit but also the whole http cycle via the on_connect/on_disconnect too right?).

`impl HttpMetrics for SomeStruct` feels a bit wonky/less clear to me (and, like, we might log metrics or we might log some other stuff, but _logging_ something is really what we're enabling, basically.

niklasad1:
with that said, shouldn't we expose the entire HTTP request then instead of just the headers?

then I'm on board with `HttpLogger` or whatever :P

jsdw:
If it's easy to do then I don't see why not!

niklasad1 (approved):
LGTM, nice work

PR #846, opened 2022-08-03
fix(tracing): use tracing instrument macro

### LINKED ISSUE #844: [tracing]: refactor to use `tracing attribute macros` for ease of use
[See comment](https://github.com/paritytech/jsonrpsee/pull/835#issuecomment-1199117485)

Ideally we should move to exclusively to just use https://docs.rs/tracing/latest/tracing/attr.instrument.html if possible
but we have to add some additional to add the correct span id

niklasad1:
Closes https://github.com/paritytech/jsonrpsee/issues/844

It relies on `instrument` macro so we don't have to bother thinking about whether it's async or sync context :)
I introduced a few bugs related that in the past and hopefully never again now


http-server/src/server.rs
```diff
@@ -805,15 +796,12 @@ where
...
-			execute_call(Call { name, params, id, call }).await
-		}
-		.instrument(trace.into_span())
-		.await
+		rx_log_from_json(&req, call.max_log_length);
+		let params = Params::new(req.params.map(|params| params.get()));
+		let name = &req.method;
+		let id = req.id;
+
+		execute_call(Call { name, params, id, call }).instrument(trace.into_span()).await
```
niklasad1, line 804: ideally we should get rid of this one too.

ws-server/src/server.rs
```diff
@@ -895,18 +891,13 @@ where
...
 
-			let params = Params::new(req.params.map(|params| params.get()));
-			let name = &req.method;
-			let id = req.id;
-
-			execute_call(Call { name, params, id, call }).await
-		}
-		.instrument(trace.into_span())
-		.await
+		execute_call(Call { name, params, id, call }).instrument(trace.into_span()).await
```
niklasad1, line 900: ideally we should get rid of this one too.

lexnv (approved):

core/src/client/async_client/mod.rs
```diff
@@ -394,6 +381,7 @@ impl SubscriptionClientT for Client {
 	///
 	/// The `subscribe_method` and `params` are used to ask for the subscription towards the
 	/// server. The `unsubscribe_method` is used to close the subscription.
+	#[instrument(name = "subscription", skip(self, params), level = "trace")]
```
niklasad1, line 384: note this differs a bit from the servers where don't distinguish between subscriptions and calls because it's so messy to support.

do we want both `subscribe and unsubscribe method`? here or only subscribe?

server/src/transport/ws.rs
```diff
@@ -267,7 +251,7 @@ pub(crate) async fn execute_call<L: Logger>(c: Call<'_, L>) -> MethodResult {
 
 	let r = response.as_inner();
 
-	rx_log_from_str(&r.result, max_log_length);
+	tx_log_from_str(&r.result, max_log_length);
```
niklasad1, line 254: ooops, bug fix :P

server/src/transport/http.rs
```diff
@@ -289,6 +275,14 @@ pub(crate) async fn execute_call<L: Logger>(c: Call<'_, L>) -> MethodResponse {
 	response
 }
 
+#[instrument(name = "notification", fields(method = notif.method.as_ref()), skip(notif, max_log_length), level = "TRACE")]
+fn execute_notification(notif: Notif, max_log_length: u32) -> MethodResponse {
+	rx_log_from_json(&notif, max_log_length);
+	let response = MethodResponse { result: String::new(), success: true };
+	tx_log_from_str(&response.result, max_log_length);
```
niklasad1, line 282: this is silly it would just be an empty string but yeye for consistency

jsdw (approved):
Nice!

PR #848, opened 2022-08-10
Add `bytes_len()` to Params

mwtian:
This is useful for middleware, e.g. monitoring request throughput and tracing.

### DIFF
--- a/types/src/params.rs
+++ b/types/src/params.rs
@@ -142,7 +142,7 @@ impl<'a> Params<'a> {
 		Params(self.0.map(|s| Cow::owned(s.into_owned())))
 	}
 
-	/// Return the byte length of the underlying data.
+	/// Return the length of underlying JSON string in number of bytes.
 	pub fn len_bytes(&self) -> usize {
 		match self.0 {
 			Some(ref cow) => cow.len(),
--- a/types/src/params.rs
+++ b/types/src/params.rs
@@ -474,7 +474,7 @@ mod test {
 		assert_eq!(none.len(), 0);
 
 		let array_params = Params::new(Some("[1, 2, 3]"));
-		assert_eq!(array_params.len(), 9);
+		assert_eq!(array_params.len_bytes(), 9);
 		let arr: Result<[u64; 3], _> = array_params.parse();
 		assert!(arr.is_ok());
 
--- a/types/src/params.rs
+++ b/types/src/params.rs
@@ -491,7 +491,7 @@ mod test {
 		assert!(one.is_ok());
 
 		let object_params = Params::new(Some(r#"{"beef":99,"dinner":0}"#));
-		assert_eq!(object_params.len(), 22);
+		assert_eq!(object_params.len_bytes(), 22);
 		let obj: Result<JsonValue, _> = object_params.parse();
 		assert!(obj.is_ok());
 	}
--- a/types/src/params.rs
+++ b/types/src/params.rs
@@ -141,6 +141,14 @@ impl<'a> Params<'a> {
 	pub fn into_owned(self) -> Params<'static> {
 		Params(self.0.map(|s| Cow::owned(s.into_owned())))
 	}
+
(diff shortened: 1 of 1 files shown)


niklasad1 (approved):
LGTM

types/src/params.rs
```diff
@@ -141,6 +141,14 @@ impl<'a> Params<'a> {
 	pub fn into_owned(self) -> Params<'static> {
 		Params(self.0.map(|s| Cow::owned(s.into_owned())))
 	}
+
+	/// Return the byte length of the underlying data.
+	pub fn len(&self) -> usize {
```
niklasad1, line 146: So, the docs on the type itself clear states that it's a wrapper of a `JSON string` thus it should okay to use
`len` here i.e, not to be confused with the length of params themselves.

/cc @lexnv @jsdw

lexnv (approved):

types/src/params.rs
mwtian: Makes sense. Also I can definitely rename it to something like `data_len()` or `byte_len()`, if it is clearer.

niklasad1, line 146:
```suggestion
	pub fn len_bytes(&self) -> usize {
```

niklasad1: yeah, let's do `len_bytes()` to be on the safe side

types/src/params.rs
```diff
@@ -141,6 +141,14 @@ impl<'a> Params<'a> {
 	pub fn into_owned(self) -> Params<'static> {
 		Params(self.0.map(|s| Cow::owned(s.into_owned())))
 	}
+
+	/// Return the byte length of the underlying data.
```
niklasad1, line 145:
```suggestion
	/// Return the length of underlying JSON string in number of bytes.
```

types/src/params.rs
jsdw: I don't mind whether it's `len()` or `byte_len()` given the comment. If it is `len()` though than iirc clippy will normally want you to also add an `is_empty()` function to be happy.

mwtian: Sounds good. Renamed to `len_bytes()` as suggested.

types/src/params.rs
```diff
@@ -463,8 +471,10 @@ mod test {
 		let none = Params::new(None);
 		assert!(none.sequence().next::<u64>().is_err());
 		assert!(none.parse::<Option<u64>>().is_ok());
+		assert_eq!(none.len(), 0);
```
niklasad1, line 474:
```suggestion
		assert_eq!(none.len_bytes(), 0);
```

types/src/params.rs
```diff
@@ -463,8 +471,10 @@ mod test {
...
 
 		let array_params = Params::new(Some("[1, 2, 3]"));
+		assert_eq!(array_params.len(), 9);
```
niklasad1, line 477:
```suggestion
		assert_eq!(array_params.len_bytes(), 9);
```

types/src/params.rs
```diff
@@ -476,10 +486,12 @@ mod test {
 		assert!(seq.next::<u64>().is_err());
 
 		let array_one = Params::new(Some("[1]"));
+		assert_eq!(array_one.len(), 3);
```
niklasad1, line 489:
```suggestion
		assert_eq!(array_one.len_bytes(), 3);
```

types/src/params.rs
```diff
@@ -476,10 +486,12 @@ mod test {
...
 		let one: Result<u64, _> = array_one.one();
 		assert!(one.is_ok());
 
 		let object_params = Params::new(Some(r#"{"beef":99,"dinner":0}"#));
+		assert_eq!(object_params.len(), 22);
```
niklasad1, line 494:
```suggestion
		assert_eq!(object_params.len_bytes(), 22);
```

mwtian:
Thanks!

Issue #847, opened 2022-08-08
[proc macros/jsonrpsee]: less deps when defining client API only


niklasad1:
Currently in the `wrapper/facade crate jsonrpsee` it's not possible to just import some minor deps for the client traits.
If I want to something like:

```rust
use jsonrpsee::{core::RpcResult, proc_macros::rpc};
#[rpc(client)]
pub trait Api {
	#[method(name = "call")]
	fn call(&self, x: usize) -> RpcResult<()>;
}
```

When trying to do that one must either enable `async-client` feature which brings a few needless deps including tokio

It's also possible though to import `jsonrpsee-core/client` manually to workaround that but we should probably expose a feature for that.

niklasad1:
Closed by #849

PR #849, opened 2022-08-10
fix(jsonrpsee): less deps when defining RPC API.

niklasad1:
Closing https://github.com/paritytech/jsonrpsee/issues/847 by introducing two new features `server-core` and `client-core` to only import relevant types for defining the RPC API i.e, not import transport related deps.

It makes things better such that if one doesn't want the transport stuff just the `client` or `server` types from core it reduces the dependency tree a lot. The typical use-case if some third-party crate or library just defines the RPC API and some other crate such a client application just want to re-use the API definition.


jsonrpsee/Cargo.toml
```diff
@@ -34,9 +35,11 @@ ws-client = ["jsonrpsee-ws-client", "jsonrpsee-types", "jsonrpsee-core"]
 ws-server = ["jsonrpsee-ws-server", "jsonrpsee-types", "jsonrpsee-core"]
 macros = ["jsonrpsee-proc-macros", "jsonrpsee-types", "jsonrpsee-core/client", "tracing"]
 
-client = ["http-client", "ws-client", "wasm-client"]
-server = ["http-server", "ws-server"]
-full = ["client", "server", "macros", "async-client", "client-ws-transport"]
+client-full = ["client", "http-client", "ws-client", "wasm-client", "client-ws-transport", "client-web-transport", "async-client"]
+client = ["macros"]
```
niklasad1, line 39: I guess we could rename this to client-core as well and keep the client feature, not sure what's the best

lexnv (approved):
LGTM!

jsonrpsee/Cargo.toml
jsdw: I think I prefer "client" being kept as-is and "client-core" for those "optimisation" cases where you want less. Also it wouldn't be a breaking change I guess that way so it's a win win really :)

niklasad1: fair enough, I'm convinced

TarikGul (approved):

PR #850, opened 2022-08-10
Fix clippy warnings

lexnv:
Fix clippy warnings.

Use `actions-rs/clippy-check@v1` which posts all lints as annotations for the pushed commit.

### Clippy Check

github_lint


niklasad1 (commented):
The changes looks good but I'm still against failing the CI just because some opinionated style was denied by clippy.

It's enough for the GHA to show to errors/warnings but it seems to be gone these days

TarikGul:
Is there a specific PR these warnings are coming from that I can check out?

niklasad1:
> Is there a specific PR these warnings are coming from that I can check out?

https://gitlab.parity.io/parity/mirrors/jsonrpsee/-/jobs/1732205

So before the GitHub UI show this warnings inline in the files when reviewing the code, not sure why it doesn't show up anymore

niklasad1 (approved):
:ok_hand:

TarikGul (approved):
Nice job, LGTM!

PR #851, opened 2022-08-11
Remove CORS logic

### LINKED ISSUE #795: refactor HTTP header verification

### LINKED ISSUE #842: Remove CORS support from Jsonrpsee and rely on tower middleware instead
Depends on us having #841.

Let's simplify Jsonrpsee by getting the CORS stuff out and relying on a pre-existing middleware instead.

When we do this we should have at least one example showing how to use the third party CORS middleware (eg this one; https://docs.rs/tower-http/latest/tower_http/cors/index.html)

lexnv:
This PR removes the CORS logic from the HTTP server.

JSONRPSee exposes the ability to add custom tower service builders,
which are capable of handling many use-cases, including CORS layers.

The CORS example is adjusted to work with both the internal access control
and the upstream CORS layers from the `tower_http` crate.

Internal code is modified to support filtering for HOST and ORIGIN via
the access control, to keep parity with substrate.

### Differences

JSONRPSee previously filtered the HOST, ORIGIN, and all the headers of a request.
This behavior causes the RPC server to respond with an error, even for OPTIONS
requests, if the request's headers are not whitelisted.

The header filtering is removed from the codebase, relying upon browsers to submit
a preflight request (OPTIONS).

Users that need to implement CORS can rely upon upstream code to achieve the same
behavior.

The RPC cors example was modified:
- for internal CORS logic (removed)

```rust
let acl = AccessControlBuilder::new()
		.set_allowed_headers(vec!["Content-Type"])?
		.allow_all_origins()
		.allow_all_hosts()
		.build();

```

- for upstream CORS logic
```rust 
let acl = AccessControlBuilder::new().allow_all_hosts().build();
	let cors = CorsLayer::new()
		// Allow `POST` when accessing the resource
		.allow_methods([Method::POST])
		// Allow requests from any origin
		.allow_origin(Any)
		.allow_headers([hyper::header::CONTENT_TYPE]);
	let middleware = tower::ServiceBuilder::new().layer(cors);
```

Requests were submitted via CURL, with / without a custom `Dummy_header`:

```bash
curl -ik  127.0.0.1:49308  -X OPTIONS -H "Content-Type: application/json" -H "Dummy_header: tmp" -H "Origin: http://not.com" -d '{"jsonrpc": "2.0", "method":  "say_hello", "id": 1}'

curl -ik  127.0.0.1:49308  -X POST  -H "Content-Type: application/json" -H "Dummy_header: tmp" -H "Origin: http://not.com" -d '{"jsonrpc": "2.0", "method":  "say_hello", "id": 1}'
```

Method | Old Behavior | New Behavior
---- | ---- | ----
POST | `HTTP/1.1 403 Forbidden; content-type: text/plain; content-length: 120 ;date: Wed, 10 Aug 2022 12:49:03 GMT; Requested headers are not allowed for CORS. CORS headers would not be sent and any side-effects were cancelled as well.` | `HTTP/1.1 200 OK; Content-type: application/json; charset=utf-8; access-control-allow-origin: *; vary: origin; vary: access-control-request-method; vary: access-control-request-headers; content-length: 49; date: Wed, 10 Aug 2022 12:40:31 GMT`
OPTIONS | `HTTP/1.1 403 Forbidden; content-type: text/plain; content-length: 120; date: Wed, 10 Aug 2022 12:49:25 GMT; Requested headers are not allowed for CORS. CORS headers would not be sent and any side-effects were cancelled as well.` | `HTTP/1.1 200 OK; access-control-allow-origin: *; vary: origin; vary: access-control-request-method; vary: access-control-request-headers; access-control-allow-methods: POST; access-control-allow-headers: content-type; content-length: 0; date: Wed, 10 Aug 2022 12:38:42 GMT`

Closes #795, closes #842.


niklasad1:
For simplicity and to follow along this PR I suggest that you keep:

```
let acl = AccessControlBuilder::new()
		.allow_all_origins()
		.allow_all_hosts()
		.build();
```

we still have `origin and host` filter left in the HTTP server AFAIU so I think we should keep it to have some sort of uniform API between the ws and http server but as you showed in the example users can use the origin filtering from the middleware as well.

http-server/src/server.rs
```diff
@@ -521,13 +513,10 @@ impl<L: Logger> ServiceData<L> {
...
 				};
 
-				let allowed_headers = acl.allowed_headers().to_cors_header_value();
-				let allowed_header_bytes = allowed_headers.as_bytes();
-
 				let res = hyper::Response::builder()
 					.header("access-control-allow-origin", origin)
 					.header("access-control-allow-methods", "POST")
-					.header("access-control-allow-headers", allowed_header_bytes)
+					.header("access-control-allow-headers", "*".as_bytes())
```
niklasad1, line 519: hmm, this is now handled by the tower middleware and we don't know which ones are whitelisted.

so `*` might be wrong?

jsdw: Hm yeah, can't this entire OPTIONs handling go away? We don't care about it at all here any more if the CORS middleware takes care of it.

jsdw: (and remember to update the comment on the Method::POST match too, and presumable we don't need to return an "access-control-allow-origin" header either on that?)

jsdw:
I'd recommend running a search for "cors" to see what's left; there appear to be some other bits and pieces around and I'm wondering whether we can remove those too?

http-server/src/server.rs
niklasad1: right I guess we could rid of https://github.com/paritytech/jsonrpsee/pull/851/files#diff-f82cd7c573676fe64905b975da3d3c8cdb9a251297b53da66aabc2d027cdeaccR532 too then

lexnv: Yep, have entirely removed those. The functionality works as before. To satisfy CORS, we still need to rely on upstream layers, but that's ok

niklasad1: that much better, removed code is a win :)

core/src/server/access_control/host.rs
```diff
@@ -127,7 +127,7 @@ impl Host {
 }
 
 impl Pattern for Host {
-	fn matches<T: AsRef<str>>(&self, other: T) -> bool {
+	fn matches<'a, T: AsRef<str> + PartialEq<&'a str>>(&self, other: T) -> bool {
```
niklasad1, line 130: I don't follow why this additional lifetime and `PartialEq` trait bound was added?

core/src/server/access_control/matcher.rs
```diff
@@ -32,7 +32,7 @@ use tracing::warn;
 /// Pattern that can be matched to string.
 pub(crate) trait Pattern {
 	/// Returns true if given string matches the pattern.
-	fn matches<T: AsRef<str>>(&self, other: T) -> bool;
+	fn matches<'a, T: AsRef<str> + PartialEq<&'a str>>(&self, other: T) -> bool;
```
niklasad1, line 35: I don't follow why this additional lifetime and PartialEq trait bound was added?

http-server/src/response.rs
```diff
@@ -56,22 +56,9 @@ pub fn method_not_allowed() -> hyper::Response<hyper::Body> {
 	)
 }
 
-/// Create a text/plain response for invalid CORS "Origin" headers.
+/// Create a text/plain response for invalid "Origin" headers.
 pub fn invalid_allow_origin() -> hyper::Response<hyper::Body> {
```
niklasad1, line 60:
```suggestion
pub fn origin_rejected(rejected_origin: Option<String>) -> hyper::Response<hyper::Body> {
```

http-server/src/response.rs
```diff
@@ -56,22 +56,9 @@ pub fn method_not_allowed() -> hyper::Response<hyper::Body> {
...
-}
-
-/// Create a text/plain response for invalid CORS "Allow-*" headers.
-pub fn invalid_allow_headers() -> hyper::Response<hyper::Body> {
-	from_template(
-        hyper::StatusCode::FORBIDDEN,
-        "Requested headers are not allowed for CORS. CORS headers would not be sent and any side-effects were cancelled as well.\n".to_owned(),
-		TEXT,
-    )
+	from_template(hyper::StatusCode::FORBIDDEN, "Origin of the request is not whitelisted.\n".to_owned(), TEXT)
```
niklasad1, line 61:
```suggestion
	from_template(hyper::StatusCode::FORBIDDEN, format!("Origin: `{}` of the request is not whitelisted.\n, maybe_origin.unwrap_or("")), TEXT)
```

examples/examples/cors_server.rs
```diff
@@ -68,10 +73,34 @@ async fn main() -> anyhow::Result<()> {
 }
 
 async fn run_server() -> anyhow::Result<(SocketAddr, HttpServerHandle)> {
-	let acl = AccessControlBuilder::new().allow_all_headers().allow_all_origins().allow_all_hosts().build();
+	// RPC access control that allows all hosts and all origins.
+	// Note: the access control does not modify the response headers,
+	// it only acts as a filter.
+	// If you need the ORIGIN header to be mirrored back in the response,
```
niklasad1, line 79: this is a bit awkward i.e, if I enable origin filtering but forget to enable it in the `CORS middleware` the correct value is not sent back if one performs a preflight request....

but nothing really we can do about it... as we don't control the CORS middleware

core/src/server/access_control/host.rs
lexnv: This was added for checking the `OriginType` against "null" str, but could be removed

core/src/server/access_control/origin.rs
```diff
@@ -0,0 +1,359 @@
...
+}
+
+/// Origin type allowed to access.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum OriginType {
+	/// Specific origin.
+	Origin(Origin),
+	/// Null origin (file:///, sandboxed iframe)
+	Null,
+	/// Any non-null origin.
```
niklasad1, line 133: I don't really understand why this was re-named to `AnyNonNull` this naming is not clear to me. I still like `Any` or `Wildcard` better.

I guess the comment is from the old code but I suggest that we fix it because it's doesn't say anything useful at all.

> For requests without credentials, the literal value "*" can be specified as a wildcard; the value tells browsers to allow requesting code from any origin to access the resource

So I suggest stating `Allow all origins` or something like that

niklasad1, line 133:
```suggestion
	/// Allow all origins i.e, the literal value "*" which is regarded as a wildcard
```

core/src/server/access_control/origin.rs
```diff
@@ -0,0 +1,359 @@
...
+impl ops::Deref for Origin {
+	type Target = str;
+	fn deref(&self) -> &Self::Target {
+		&self.host_with_proto
+	}
+}
+
+/// Origin type allowed to access.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum OriginType {
```
niklasad1, line 128: I suggested changing this to `Origin` and to rename the `inner origin type` to `InnerOrigin`.

we shouldn't use `type` in `type names` that's implied anyway

core/src/server/access_control/origin.rs
lexnv: Done! Picked `Wildcard`, as it seems a bit more expressive, thanks!

core/src/server/access_control/origin.rs
lexnv: Done! Thanks! 👍

niklasad1 (approved):
nice looks clean :+1:

examples/examples/cors_server.rs
jsdw: I think that's OK isn't it? I mean, if we assume that enabling origin filtering in jsonrpsee has nothing to do with CORS any more and will just block requests from origins people don't want accessing it.

(which I know sortof overlaps the CORS functionality, but I guess we're only keeping it here because we want to maintain consistency with WS connections and they don't have any CORS stuff)

niklasad1: yupp, you are correct those are redundant.

jsdw (approved):
Looks great! Much clearer IMO now that we offer basic filtering and there's a separate CORS thing if you want proper CORS :)

PR #852, opened 2022-08-11
fix(client): adjust TransportSenderT

niklasad1:
This is trait contains `WebSocket` specific details and it's difficult to fix it properly with an extension trait
in the current design.

So this PR documents and marks it clearly that these methods are optional to implement, kind of ugly but better.


core/src/client/mod.rs
```diff
@@ -132,11 +132,19 @@ pub trait TransportSenderT: MaybeSend + 'static {
 	/// Send.
 	async fn send(&mut self, msg: String) -> Result<(), Self::Error>;
 
+	/// This is optional because it's most likely on relevant for WebSocket transports only.
```
lexnv, line 135:
```suggestion
	/// This is optional because it's most likely relevant for WebSocket transports only.
```

core/src/client/mod.rs
```diff
@@ -132,11 +132,19 @@ pub trait TransportSenderT: MaybeSend + 'static {
...
+	///
 	/// Send ping frame (opcode of 0x9).
-	async fn send_ping(&mut self) -> Result<(), Self::Error>;
+	async fn optional_send_ping(&mut self) -> Result<(), Self::Error> {
+		Ok(())
+	}
 
-	/// If the transport supports sending customized close messages.
-	async fn close(&mut self) -> Result<(), Self::Error> {
+	/// This is optional because it's most likely on relevant for WebSocket transports only.
```
lexnv, line 143:
```suggestion
	/// This is optional because it's most likely relevant for WebSocket transports only.
```

lexnv (approved):

lexnv:
The `multiple_blocking_calls_overlap` test causes the CI to fail, which is highly sensitive
to scheduling noise.

`thread 'multiple_blocking_calls_overlap' panicked at 'Expected less than 60ms, got 63.3412ms'`

I suggest increasing the threshold to 80ms for this case 😄

core/src/client/mod.rs
```diff
@@ -132,11 +132,19 @@ pub trait TransportSenderT: MaybeSend + 'static {
...
 	/// Send ping frame (opcode of 0x9).
-	async fn send_ping(&mut self) -> Result<(), Self::Error>;
+	async fn optional_send_ping(&mut self) -> Result<(), Self::Error> {
+		Ok(())
+	}
 
-	/// If the transport supports sending customized close messages.
-	async fn close(&mut self) -> Result<(), Self::Error> {
+	/// This is optional because it's most likely relevant for WebSocket transports only.
+	/// You should only implement this is your transport supports sending periodic pings.
```
jsdw, line 144:
```suggestion
	/// You should only implement this is your transport supports being closed.
```

jsdw (approved):
I like that the default impls don't fail now; it makes much more sense to have default impls when they can safely be ignored if the implementor doesn't care about that functionality :)

For that reason, I'm not sure they need the `optional_` prefix; they are optional now in the same way that any trait fn with a default impl is, so I don't think the prefix adds anything.

niklasad1:
> For that reason, I'm not sure they need the optional_ prefix; they are optional now in the same way that any trait fn with a default impl is, so I don't think the prefix adds anything.

Removed it now, I guess the documentation is clear enough now so we don't need that.
I was almost considering doing `doc(hidden)` to hide this from other users but I think this should be ok.

Issue #298, opened 2021-05-03
Confusing error when providing the wrong type to be decoded


chevdor:
The following fails:
```
use jsonrpsee::http_client::{traits::Client, HttpClientBuilder, JsonValue, Error};
use std::env;
use tokio::runtime::Runtime;

fn get_http_node() -> String {
	env::var("POLKADOT_HTTP").unwrap_or("http://localhost:9933".to_string())
}

fn main() {
	let url = String::from(get_http_node());
	println!("Connecting to {:?}", &url);
	let code = JsonValue::from("0x3a636f6465"); // :code in hex
	let params = vec![code.into()];

	let rt = Runtime::new().unwrap();
	let client = HttpClientBuilder::default().build(url).expect("Building client");
	let response: Result<Vec<u8>, _> = rt.block_on(client.request("state_getStorage", params.into()));
	println!("response: {:?}", response);

	assert!(response.unwrap().len() > 0);
}
```

with:
```
response: Err(ParseError(Error("missing field `error`", line: 1, column: 3752798)))
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ParseError(Error("missing field `error`", line: 1, column: 3752798))', examples/http-polkadot.rs:51:22
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Debugging the request, I see that we get a big answer from the node.
So I assume the request is sent properly and we get the answer. Decoding however fails at https://github.com/paritytech/jsonrpsee/blob/master/http-client/src/client.rs#L76

chevdor:
@niklasad1 thanks for the hints.

Using the following does not solve the issue but does the trick:
```
let response: Result<String, _> = rt.block_on(client.request("state_getStorage", params.into()));
```

jsdw:
This will hopefully be closed by https://github.com/paritytech/jsonrpsee/pull/853

niklasad1:
Closed by #853

PR #853, opened 2022-08-11
fix(http client): more user-friendly error messages when decoding fails

niklasad1:
This PR changes the http client to decode the reponse as `Response<serde_json::Value> -> Response<R>`
In order to get a more user-friendly error message

Before:

```
response: Err(ParseError(Error("missing field `error`", line: 1, column: 3752798)))
```

After

```
2022-08-11T13:57:12.033043Z  INFO http: r: Parse error: invalid type: string "lo", expected u8
```

Not sure about the overhead of doing this but the `async-client` already does this and it's more user-friendly.

Closing #298


lexnv (approved):

niklasad1:
I actually benches this I couldn't find any significant regression when doing:

```diff
diff --git a/benches/bench.rs b/benches/bench.rs
index 0fd6c8e..6a28838 100644
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -4,10 +4,10 @@ use crate::helpers::{ws_handshake, KIB};
 use criterion::*;
 use futures_util::future::{join_all, FutureExt};
 use futures_util::stream::FuturesUnordered;
-use helpers::{http_client, ws_client, SUB_METHOD_NAME, UNSUB_METHOD_NAME};
+use helpers::{http_client, to_human_friend_size, ws_client, SUB_METHOD_NAME, UNSUB_METHOD_NAME};
 use jsonrpsee::core::client::{ClientT, SubscriptionClientT};
 use jsonrpsee::http_client::HeaderMap;
-use jsonrpsee::types::{Id, ParamsSer, RequestSer};
+use jsonrpsee::types::{Id, ParamsSer, RequestSer, Response};
 use pprof::criterion::{Output, PProfProfiler};
 use tokio::runtime::Runtime as TokioRuntime;
 
@@ -16,7 +16,7 @@ mod helpers;
 criterion_group!(
 	name = types_benches;
 	config = Criterion::default().with_profiler(PProfProfiler::new(100, Output::Flamegraph(None)));
-	targets = jsonrpsee_types_v2
+	targets = types_serialize_request, types_decode_response
 );
 criterion_group!(
 	name = sync_benches;
@@ -62,8 +62,8 @@ fn v2_serialize(req: RequestSer<'_>) -> String {
 	serde_json::to_string(&req).unwrap()
 }
 
-pub fn jsonrpsee_types_v2(crit: &mut Criterion) {
-	crit.bench_function("jsonrpsee_types_v2_array_ref", |b| {
+pub fn types_serialize_request(crit: &mut Criterion) {
+	crit.bench_function("create_and_serialize_request_ref", |b| {
 		b.iter(|| {
 			let params = &[1_u64.into(), 2_u32.into()];
 			let params = ParamsSer::ArrayRef(params);
@@ -72,7 +72,7 @@ pub fn jsonrpsee_types_v2(crit: &mut Criterion) {
 		})
 	});
 
-	crit.bench_function("jsonrpsee_types_v2_vec", |b| {
+	crit.bench_function("create_and_serialize_request_alloc", |b| {
 		b.iter(|| {
 			let params = ParamsSer::Array(vec![1_u64.into(), 2_u32.into()]);
 			let request = RequestSer::new(&Id::Number(0), "say_hello", Some(params));
@@ -81,6 +81,46 @@ pub fn jsonrpsee_types_v2(crit: &mut Criterion) {
 	});
 }
 
+pub fn types_decode_response(crit: &mut Criterion) {
+	let serialized_vals: Vec<String> = [1 * KIB, 10 * KIB, 100 * KIB, 1_024 * KIB, 10_000 * KIB]
+		.into_iter()
+		.map(|size| serde_json::to_string(&Response::new("A".repeat(size), Id::Null)).unwrap())
+		.collect();
+
+	{
+		let mut group = crit.benchmark_group("decode_response");
+		for ser_json in &serialized_vals {
+			group.bench_with_input(
+				BenchmarkId::from_parameter(&to_human_friend_size(ser_json.len())),
+				&ser_json.len(),
+				|b, _| b.iter(|| black_box(serde_json::from_str::<Response<String>>(&ser_json).unwrap())),
+			);
+		}
+
+		group.finish();
+	}
+
+	{
+		let mut group = crit.benchmark_group("decode_response_twice");
+		for ser_json in &serialized_vals {
+			group.bench_with_input(
+				BenchmarkId::from_parameter(&to_human_friend_size(ser_json.len())),
+				&ser_json.len(),
+				|b, _| {
+					b.iter(|| {
+						black_box({
+							let raw_response = serde_json::from_str::<Response<serde_json::Value>>(&ser_json).unwrap();
+							serde_json::from_value::<String>(raw_response.result).unwrap()
+						})
+					})
+				},
+			);
+		}
+
+		group.finish();
+	}
+}
+
 trait RequestBencher {
 	const REQUEST_TYPE: RequestType;
 
diff --git a/benches/helpers.rs b/benches/helpers.rs
index 93d6b99..d8593d8 100644
--- a/benches/helpers.rs
+++ b/benches/helpers.rs
@@ -16,6 +16,8 @@ pub(crate) const ASYNC_METHODS: [&str; 3] = [SYNC_FAST_CALL, SYNC_MEM_CALL, SYNC
 
 // 1 KiB = 1024 bytes
 pub(crate) const KIB: usize = 1024;
+// 1 MB = 1024*1024 bytes
+pub(crate) const MB: usize = 1024 * 1024;
 
 /// Run jsonrpc HTTP server for benchmarks.
 #[cfg(feature = "jsonrpc-crate")]
@@ -207,3 +209,14 @@ pub(crate) async fn ws_handshake(url: &str, headers: HeaderMap) {
 	let uri: Uri = url.parse().unwrap();
 	WsTransportClientBuilder::default().max_request_body_size(u32::MAX).set_headers(headers).build(uri).await.unwrap();
 }
+
+pub(crate) fn to_human_friend_size(len: usize) -> String {
+	const EXCLUSIVE_KIB: usize = KIB - 1;
+	const EXCLUSIVE_MB: usize = MB - 1;
+
+	match len {
+		b @ 0..=EXCLUSIVE_KIB => format!("{}b", b),
+		b @ KIB..=EXCLUSIVE_MB => format!("{}kb", b / KIB),
+		b @ _ => format!("{}mb", b / MB),
+	}
+}
```

but it might be worse on some more complicated type than String

client/http-client/src/client.rs
```diff
@@ -207,16 +207,19 @@ impl ClientT for HttpClient {
 				}
 			};
```
jsdw, line 208: Dumb (unrelated to this PR) question: what's the purpose of this async block?

client/http-client/src/client.rs
```diff
@@ -207,16 +207,19 @@ impl ClientT for HttpClient {
...
+			let response: Response<serde_json::Value> =
+				match serde_json::from_slice::<Response<serde_json::Value>>(&body) {
+					Ok(response) => response,
+					Err(_) => {
+						let err: ErrorResponse = serde_json::from_slice(&body).map_err(Error::ParseError)?;
+						return Err(Error::Call(CallError::Custom(err.error_object().clone().into_owned())));
+					}
+				};
+
+			let result = serde_json::from_value(response.result).map_err(Error::ParseError)?;
```
jsdw, line 219: So, we deseriazlize the bytes into `Response<Value>` and then that `Value` into R?

It might be worth a comment explaining this split (where previously it deserialized all at once); somebody like me would look at this and think "ooh I can skip that Value step" and change it back!

Also, going via `Value` feels like quite a performance hit over going straight to `R`. If we need to do this in two steps, perhaps instead of going to `Value` we could go to `RawValue` to avoid the allocations, and then turn that into an `R`?

client/http-client/src/client.rs
```diff
@@ -267,7 +270,8 @@ impl ClientT for HttpClient {
 					Some(pos) => *pos,
 					None => return Err(Error::InvalidRequestId),
 				};
-				responses[pos] = rp.result
+				let result = serde_json::from_value(rp.result).map_err(Error::ParseError)?;
```
jsdw, line 273: Same here; a comment for the two-part decoding and use RawValue if possible to avoid allocating?

jsdw:
Ooh I see; you go via Value to benefit form the better error messages when Value fails to deserialize into the type. I'm not sure whether or not RawValue woule give you that (still not sure whether it would actually work; I'd need to try it).

I am a bit worries about the perf things; making `Value`s is quite allocation heavy, much more so for complicated types than simpler ones. I wonder what benchmarks owuld best highlight that sort of difference.

client/http-client/src/client.rs
niklasad1: it's part of the `tracing stuff` bug fix, it should be fixed by https://github.com/paritytech/jsonrpsee/pull/846

niklasad1:
> Ooh I see; you go via Value to benefit form the better error messages when Value fails to deserialize into the type. I'm not sure whether or not RawValue woule give you that (still not sure whether it would actually work; I'd need to try it).

> I am a bit worries about the perf things; making Values is quite allocation heavy, much more so for complicated types than simpler ones. I wonder what benchmarks owuld best highlight that sort of difference.

Yeah me too that's why benched it, we could take another route which I implemented: https://github.com/paritytech/jsonrpsee/compare/na-more-user-friendly-decoding-errors?expand=1 but the error message is not as good but it should be cheaper.

client/http-client/src/client.rs
niklasad1: yeah, it works ty for the pro-tip :)

client/http-client/src/client.rs
```diff
@@ -207,16 +207,20 @@ impl ClientT for HttpClient {
 				}
 			};
 
-			let response: Response<_> = match serde_json::from_slice(&body) {
+			// NOTE: it's decoded first `JsonRawValue` and then to `R` below to get
```
niklasad1, line 210:
```suggestion
			// NOTE: it's decoded first to `JsonRawValue` and then to `R` below to get
```

client/http-client/src/client.rs
```diff
@@ -254,7 +258,9 @@ impl ClientT for HttpClient {
 				Ok(Err(e)) => return Err(Error::Transport(e.into())),
 			};
 
-			let rps: Vec<Response<_>> =
+			// NOTE: it's decoded first `JsonRawValue` and then to `R` below to get
```
niklasad1, line 261:
```suggestion
			// NOTE: it's decoded first to `JsonRawValue` and then to `R` below to get
```

TarikGul (approved):

jsdw (approved):

PR #859, opened 2022-08-22
middleware: Implement proxy URI paths to RPC methods

### LINKED ISSUE #765: [http server]: health_api is rather confusing because it's not exclusive for `/health`
https://github.com/paritytech/jsonrpsee/pull/763#discussion_r870142939

Additionally the API is quite easy to misuse if one passes in something like a path with `/` and similar.

lexnv:
This PR moves the internal logic of system health API to a
custom tower layer implementation.

The core of this implementation is the `ProxyRequst`, which proxies 
`GET/ path` requests to the specified RPC method calls.

## Request

The `GET /path` requests are modified into valid `POST` requests for
calling the RPC method. This middleware adds appropriate headers to the
request, and completely alters the request `BODY`.

## Response

The response of the RPC method is stripped down to contain only the method's
response, removing any RPC 2.0 spec logic regarding the response's body.

Closes #765.


http-server/src/middlewares/proxy_request.rs
```diff
@@ -0,0 +1,174 @@
...
+
+	fn call(&mut self, mut req: Request<Body>) -> Self::Future {
+		let modify = self.path.as_str() == req.uri() && req.method() == Method::GET;
+
+		// Proxy the request to the appropriate method call.
+		if modify {
+			// RPC methods are accessed with `POST`.
+			*req.method_mut() = Method::POST;
+			// Precautionary remove the URI.
+			*req.uri_mut() = "/".parse().unwrap();
```
niklasad1, line 92: it's safe to `Uri::from_static` here because it's infallible.

http-server/src/middlewares/proxy_request.rs
```diff
@@ -0,0 +1,174 @@
...
+
+			// Requests must have the following headers:
+			req.headers_mut().insert(CONTENT_TYPE, HeaderValue::from_static("application/json"));
+			req.headers_mut().insert(ACCEPT, HeaderValue::from_static("application/json"));
+
+			// Adjust the body to reflect the method call.
+			let body =
+				Body::from(format!("{{\"jsonrpc\":\"2.0\",\"method\":\"{}\",\"params\":null,\"id\":1}}", self.method));
```
niklasad1, line 100: you could also use https://docs.rs/jsonrpsee-types/0.15.1/jsonrpsee_types/request/struct.RequestSer.html instead of doing this manually then to serialize it to a JSON String

http-server/src/middlewares/proxy_request.rs
```diff
@@ -0,0 +1,174 @@
...
+///
+/// # Response
+///
+/// The response of the RPC method is stripped down to contain only the method's
+/// response, removing any RPC 2.0 spec logic regarding the response' body.
+#[derive(Debug, Clone)]
+pub struct ProxyRequest<S> {
+	inner: S,
+	path: String,
+	method: String,
```
jsdw, line 58: Perhaps `Arc<str>` would be better since you clone them per request and don't need to modify them :)

http-server/src/middlewares/proxy_request.rs
```diff
@@ -0,0 +1,174 @@
...
+			req.headers_mut().insert(ACCEPT, HeaderValue::from_static("application/json"));
+
+			// Adjust the body to reflect the method call.
+			let body =
+				Body::from(format!("{{\"jsonrpc\":\"2.0\",\"method\":\"{}\",\"params\":null,\"id\":1}}", self.method));
+			req = req.map(|_| body);
+		}
+
+		// Depending on `modify` adjust the response.
+		ResponseFuture::PollFuture { future: self.inner.call(req), modify }
```
jsdw, line 105: The alternative to this would be to have `type Future = Pin<Box<dyn Future + Send + Sync + 'static>>` (or something along those lines), and then you can wrap this stuff into something like an `let fut = async move {}; Box::pin(fut)` type thing; then you can just use async/await syntax here (at the cost of a sigle allocation per call, which I think is probably worth the extra simplicity).

jsdw: (and if you do this, check out https://docs.rs/hyper/latest/hyper/body/fn.to_bytes.html to aggregate the Body bytes easily :))

http-server/src/lib.rs
```diff
@@ -35,6 +35,9 @@ mod server;
 /// Common builders for RPC responses.
 pub mod response;
 
+/// Common tower middlewares exposed for RPC interaction.
+pub mod middlewares;
```
niklasad1, line 39:
```suggestion
pub mod middleware;
```
? :)

http-server/src/middlewares/proxy_request.rs
```diff
@@ -0,0 +1,174 @@
...
+impl<S> ProxyRequest<S> {
+	/// Creates a new [`ProxyRequest`].
+	///
+	/// The request `GET /path` is redirected to the provided method.
+	pub fn new(inner: S, path: impl Into<String>, method: impl Into<String>) -> Self {
+		Self { inner, path: path.into(), method: method.into() }
+	}
+}
+
+impl<S> Service<Request<Body>> for ProxyRequest<S>
```
niklasad1, line 70: I suppose it's also possible for third parties to construct their own middleware service with specific logic such as https://github.com/paritytech/jsonrpsee/issues/81?

i.e, to parse params from the URI and pass the parsed params call the method with the parsed params instead of params in the JSON-RPC call?

http-server/src/middlewares/proxy_request.rs
```diff
@@ -0,0 +1,174 @@
+//! Middleware that proxies requests at a specified URI to internal
```
jsdw, line 1: Name bikeshedding; `ProxyRequest` feels a bit vague to me.

How about something like `GetRequestToRpc` or `ProxyGetRequest` just for a touch more detail?

tests/tests/helpers.rs
```diff
@@ -223,13 +224,15 @@ pub async fn http_server() -> (SocketAddr, HttpServerHandle) {
 }
 
 pub async fn http_server_with_access_control(acl: AccessControl, cors: CorsLayer) -> (SocketAddr, HttpServerHandle) {
-	let middleware = tower::ServiceBuilder::new().layer(cors);
+	let middleware = tower::ServiceBuilder::new()
+		// Proxy `GET /health` requests to internal `system_health` method.
+		.layer(ProxyRequestLayer::new("/health", "system_health"))
```
jsdw, line 229: Love this; nice and simple to use :)

http-server/src/middlewares/proxy_request.rs
jsdw: Yeah I don't see why not!

A `ProxyPostRequest` middleware could exist for instance which takes args in the body, or this one could decode named params from querystring or whatever else :)

http-server/src/middlewares/proxy_request.rs
niklasad1: ProxyGetRequest is a better name for it

http-server/src/middlewares/proxy_request.rs
lexnv: Thanks! Done!

http-server/src/middlewares/proxy_request.rs
lexnv: Thanks! It clarifies the code a lot!

http-server/src/middlewares/proxy_request.rs
lexnv: I like it a lot better now, that we can use basic async syntax, thanks!

http-server/src/middlewares/proxy_request.rs
lexnv: It makes sense, thanks for the optimization feedback 😄 !

http-server/src/lib.rs
lexnv: It sounds better, have modified it! Thanks!

http-server/src/middlewares/proxy_request.rs
lexnv: Yep, users can create a new layer extract the params from the request's URI for sure, and pass it down similar to how we do here! 😄

http-server/src/middlewares/proxy_request.rs
lexnv: I'm sold for `ProxyGetRequest` does eliminate any confusion! Thanks!

http-server/Cargo.toml
```diff
@@ -21,6 +21,7 @@ serde_json = { version = "1.0", features = ["raw_value"] }
 serde = "1"
 tokio = { version = "1.16", features = ["rt-multi-thread", "macros"] }
 tower = "0.4.13"
+pin-project-lite = "0.2.9"
```
niklasad1, line 24: we don't need this anymore?

lexnv: Nope! Thanks for noticing this!

niklasad1 (approved):
nice, LGTM

http-server/src/middleware/proxy_request.rs
```diff
@@ -0,0 +1,152 @@
+//! Middleware that proxies requests at a specified URI to internal
```
jsdw, line 1: let's rename this to `proxy_get_request` just to be consistent with our struct rename?

lexnv: It makes sense, yep

jsdw (approved):
Looks great; the boxing is much tidier IMO, and the code in general a solid improvement on havign the health stuff integrated :)

(it did occur to me that it adds a box for every single request, not just health ones, but I guess I still wouldn't worry unless benchmarks suffer; I have a feeling 1 box won't be noticable :))

PR #860, opened 2022-08-22
Enable `cargo-nextest` for CI

lexnv:
To avoid issues with the CI, the nextest was replaced with `cargo test`.

This PR ensures we utilize it again the `cargo nextest` functionality, via
installing it locked ([commit](https://github.com/nextest-rs/nextest/commit/f8fc8ae820524f6171868c384654f94e5655d79a))

Issue reference: https://github.com/nextest-rs/nextest/issues/306

## Time Delta

Location | Sample 1 | Sample 2 | After  1 | After 2 | 
|----|----|----|----|----|
CI (ubuntu) | 3m 10s | 3m 1s | 4m 47s | 3m 23s |
Windows | 9m 33s |  10m 15s  |  11m 18s | 11m 34s |
MacOS | 8m 18s | 5m 43s   | 8m 36s  | 9m 48s |

_Sample 1: https://github.com/paritytech/jsonrpsee/commit/6fdb67f99324ff50b8f629a3eb563fd314af9aae
Sample 2: https://github.com/paritytech/jsonrpsee/commit/bd31557da8246dcc45555bf2dcbb65585a834bbb._

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,6 +109,7 @@ jobs:
         uses: baptiste0928/cargo-install@v1
         with:
           crate: cargo-nextest
+          locked: true
           version: 0.9
 
       - name: Rust Cache
@@ -135,6 +136,7 @@ jobs:
         uses: baptiste0928/cargo-install@v1
         with:
           crate: cargo-nextest
+          locked: true
           version: 0.9
 
       - name: Rust Cache
@@ -161,6 +163,7 @@ jobs:
         uses: baptiste0928/cargo-install@v1
         with:
           crate: cargo-nextest
+          locked: true
           version: 0.9
 
       - name: Rust Cache
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,29 +91,6 @@ jobs:
       - name: Cargo check all targets and features
         run: cargo hack check --workspace --each-feature --all-targets
 
-  tests_ubuntu:
-    name: Run tests Ubuntu
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout sources
-        uses: actions/checkout@v3.0.2
-
-      - name: Install Rust stable toolchain
-        uses: actions-rs/toolchain@v1.0.7
-        with:
-          profile: minimal
-          toolchain: stable
-          override: true
-
-      - name: Install latest nextest release
-        uses: taiki-e/install-action@nextest
-
-      - name: Rust Cache
-        uses: Swatinem/rust-cache@v2.0.0
-
-      - name: Cargo nextest
-        run: cargo nextest run --workspace
-
   tests_macos:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -105,14 +105,18 @@ jobs:
           toolchain: stable
           override: true
 
+      - name: Install cargo-nextest
+        uses: baptiste0928/cargo-install@v1
+        with:
+          crate: cargo-nextest
```
niklasad1, line 111: we could also use the pre-built binaries https://nexte.st/book/release-urls.html

lexnv: Yep, it makes sense! Have changed to reflect: https://nexte.st/book/pre-built-binaries.html#using-pre-built-binaries-in-ci

niklasad1 (approved):
nice, 

it still doesn't support doc tests right?
might be good to add another job for that..

alvicsam:
> it still doesn't support doc tests right?

No, still not https://github.com/nextest-rs/nextest/issues/16

I can suggest editing the `test` job in [.gitlab-ci.yml](https://github.com/paritytech/jsonrpsee/blob/master/.gitlab-ci.yml#L101-L106). It runs on the parity CI runners which are way faster than GitHub runners. In this case the `Continuous integration / Run tests Ubuntu (pull_request) ` GHA job can be removed. We also have a mac-runner but it's needed to test nextest on it. Unfortunately, the we don't have windows runners so for them the only option is to use GHA.

.gitlab-ci.yml
```diff
@@ -103,7 +103,9 @@ test:
   <<:                              *docker-env
   <<:                              *common-refs
   script:
-    - cargo test
+    - cargo install cargo-nextest --locked
```
alvicsam, line 106: This is not necessary, the image already contains `cargo-nextest` (not the latest version though)

lexnv: Thanks! I'll remove it, does `cargo-hack` also exist on the image?

alvicsam: No, we can add it if needed. But it might take some time when it reaches the production image

alvicsam (approved):

jsdw:
So to confirm, it looks like `nextest` is actually taking longer than the old `cargo test` as it stands?

(Does that include the additional `cargo test` invocation that we need to run separate doc tests, if not supported by `nextest`?)

That being the case, I'd prefer to keep it simple and stick with `cargo test` until there is a clear advantage, or am I missing something?

niklasad1:
> So to confirm, it looks like nextest is actually taking longer than the old cargo test as it stands?

wow, I thought nextest was suppose to faster... but I think the UI/proc macro tests are the bottleneck

jsdw:
Is the verdict to close this PR since nexttest is actually slower anyway (plus doesn't run doctests)?

PR #864, opened 2022-08-26
Optimize serialization for client parameters

### LINKED ISSUE #862: Avoid mandatory encoding into serde_json::Value for parameters
Currently `client.request()`, `client.subscribe()` etc take in params that are `Option<ParamSer>`, which looks like:

```rust
#[derive(Serialize, Debug, Clone)]
#[serde(untagged)]
pub enum ParamsSer<'a> {
	/// Positional params (heap allocated).
	Array(Vec<JsonValue>),
	/// Positional params (slice).
	ArrayRef(&'a [JsonValue]),
	/// Params by name.
	Map(BTreeMap<&'a str, JsonValue>),
}
```

This makes encoding params to `serde_json::Value`'s mandatory, and said values are fairly allocation heavy.

How about we: 
1. Accept something like `Option<impl ToRpcParams>` instead of `Option<ParamSer>`on these methods.
2. Consider modifying `ToRpcParams` so that implementors can maximise their efficiency, so:
    - Have `to_rpc_params` take ownership of `self`; this can avoid unnecessary clones (eg if I already have the `Box<RawValue>` that it wants returned), and we only need to do this once, right? Also, you can impl it on a reference type if need be to take a reference.
    - Remove the supertrait of `Serialize`; let the implementor decide how it wants to provide the `Box<RawValue>`; it shouldn't need to be able to be serialized itself (again, if we already have a `RawValue` for instance). 
3. Impl `ToRpcParams` for `ParamSer` if we want to keep using it, orrrr (I think this is my preference, but could possibly be a separate issue:) build more efficient named and unnamed param encoding structs like follows (and update the `rpc_params!` macro to build unnamed params):

```rust
// Build named params:
pub struct NamedParamsBuilder(String);
// The finished, ready-to-go JSON:
pub struct NamedParams(String);

impl NamedParamsBuilder {
    pub fn new() -> Self {
        NamedParams("{".to_string())
    }
    pub fn insert<P: Serialize>(&mut self, name: &str, value: P) -> Result<(), Error> {
        self.0.extend(serde_json::to_string(name)?);
        self.0.push(':');
        self.0.push(value.serialize()?);
        Ok(())
    }
    pub fn build(self) -> NamedParams {
        self.0.push('}');
        NamedParams(self.0)
    }
}

impl ToRpcParams for NamedParams {
    fn to_rpc_params(self) {
        self.0.to_raw_value() // whatever the method to go from String to Box<RawValue> is.
    }
}

pub struct UnnamedParams(String);

impl UnnamedParams {
    // ... Similar sort of thing
}

// And now users can do something more efficient or whatever they prefer..

```

lexnv:
This PR adds a generic way of providing RPC parameters to the clients.

The previous method required the users to serialize their parameters
to a hardcoded `serde_json::Value`, which is allocation heavy, especially
when the valid parameters are already constructed.

To constrain users to correct parameter types, two alternative methods
are built upon the internal `ParameterBuilder`: the `UnnamedParams` and
`NamedParams`.

The former is the equivalent of the json array, while the latter represents a
json map key-value.

The `ClientT` interface has been updated to receive the `ToRpcParams` trait
as parameters, giving users an escape route that provides a raw constructed
value. This has the benefit of avoiding a `Raw` -> `json::Value` -> `Raw`
conversion.  

To accommodate this change, the batch parameters need to receive `Raw` inputs.
This is because users wish to provide some methods without parameters
(deduced to the `()` empty tuple type), while some methods need to have named or
unnamed inputs (`UnnamedParams` and `NamedParams`).

For improving the user's experience, a batch parameter builder is introduced to
abstract the low-level interface required by submitting batch requests.

Closes #862.


core/src/client/mod.rs
```diff
@@ -173,20 +182,21 @@ pub trait TransportReceiverT: 'static {
...
+		// ToRpcParams is implemented for the empty tuple.
+		()
+	};
 	($($param:expr),*) => {
 		{
-			let mut __params = vec![];
+			let mut __params = $crate::client::__reexports::UnnamedParamsBuilder::new();
 			$(
-				__params.push($crate::client::__reexports::to_json_value($param).expect("json serialization is infallible; qed."));
+				__params.insert($param).expect("json serialization is infallible; qed.");
```
jsdw, line 195: Just FYI, Json serialisation isn't infallible?

For example object keys just be strings, so trying to serialize eg a Map<bool, Foo> would fail.

Happy to sort this out in a future PR though to avoid bloating this one; I'd lean towards doing something like I did in subxt and having the macro return a Result, so that the caller can unwrap or "?" It.

Whatcha think @niklasad1 ?

niklasad1: yepp, you are correct @jsdw 

as long as we document that it will panic if serialization fails I'm fine with it and remove the incorrect expect :)

it's intended to work as the `serde_json::json! macro` but we can do what you did in subxt here as well.

client/http-client/src/client.rs
```diff
@@ -225,7 +234,7 @@ impl ClientT for HttpClient {
 		.await
 	}
 
-	async fn batch_request<'a, R>(&self, batch: Vec<(&'a str, Option<ParamsSer<'a>>)>) -> Result<Vec<R>, Error>
+	async fn batch_request<R>(&self, batch: Vec<(&str, Option<Box<RawValue>>)>) -> Result<Vec<R>, Error>
```
niklasad1, line 237: this looks like a more low-level API than before, can you explain why?

it's almost like we should provide two interfaces for this `one high-level` i.e. with generics and `one ow level` with RawValue as James implemented in subxt.

types/src/params.rs
```diff
@@ -408,6 +416,248 @@ impl<'a> Id<'a> {
...
+	fn to_rpc_params(self) -> Result<Option<Box<RawValue>>, serde_json::Error>;
+}
+
+/// Initial number of parameters in a batch request.
+const BATCH_PARAMS_NUM_CAPACITY: usize = 4;
+
+/// Parameter builder that serializes RPC parameters to construct a valid batch parameter.
+/// This is the equivalent of chaining multiple RPC requests.
+#[derive(Debug)]
+pub struct BatchParamsBuilder<'a>(Vec<(&'a str, Option<Box<RawValue>>)>);
```
niklasad1, line 634: hmm, can we call this `BatchRequestBuilder` instead as it also contains the `method name` which isn't a `param` according to JSON-RPC spec :P

sorry, I'm a bit damaged after reading the JSON-RPC spec a few times :)

lexnv: `BatchRequestBuilder` sounds good to me, have updated it!

core/src/client/mod.rs
lexnv: Yep, it makes sense. For this PR have changed the `except` message and added documentation. Will come up with a followup PR to return a `Result` here

client/http-client/src/client.rs
```diff
@@ -225,7 +234,7 @@ impl ClientT for HttpClient {
 		.await
 	}
 
-	async fn batch_request<'a, R>(&self, batch: Vec<(&'a str, Option<ParamsSer<'a>>)>) -> Result<Vec<R>, Error>
```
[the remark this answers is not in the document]
niklasad1: Alex's already introduced the type `BatchRequestBuilder` which does what you suggested so I don't understand what you mean by your comment.

I suppose having `BatchRequestBuilder::build() -> BatchParams<'a>` could help make the RPC trait a it more readable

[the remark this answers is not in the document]
lexnv: I think James suggested we use `BatchReuqstBuilder` directly as a parameter here, similar to:

```rust
async fn batch_request<R>(&self, batch: BatchReuqstBuilder) -> Result<Vec<R>, Error>
```

client/http-client/src/client.rs
lexnv: I wanted to provide a way for users to construct the batch request as they wanted without any overhead, but the `BatchRequestBuilder` provides just that 😄

core/src/client/mod.rs
jsdw: That makes sense @niklasad1! I pondered this and, indeed, it's only programmer error that can led to the panic (ie providing fallible things to the macro), so I'm ok with keeping the panic behaviour and not returning a result!

Maybe subxt's version should be the same..

client/http-client/src/client.rs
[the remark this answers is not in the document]
jsdw: Ah yup I hadn't noticed that the type already existed; I was just suggesting that such a type be used in the method indeed :)

client/http-client/src/client.rs
```diff
@@ -225,10 +234,11 @@ impl ClientT for HttpClient {
...
+	async fn batch_request<'a, R>(&self, batch: BatchRequestBuilder<'a>) -> Result<Vec<R>, Error>
```
niklasad1, line 237: yeah, this API is a bit different than the others i.e, takes the concrete type `BatchRequestBuilder` instead of `ToRpcParams` I think that makes sense as it supports both `positional params (array)` and `named params (map)` but not possible for folks to use there own impl of this.

Indeed it makes the API easier to understand and cleaner, cool

core/src/server/rpc_module.rs
```diff
@@ -337,9 +337,9 @@ impl Methods {
 
 	/// Helper to call a method on the `RPC module` without having to spin up a server.
 	///
-	/// The params must be serializable as JSON array, see [`ToRpcParams`] for further documentation.
+	/// The params must be serializable as JSON array, see [`jsonrpsee_types::ToRpcParams`] for further documentation.
```
niklasad1, line 340: the updated doc is wrong

```suggestion
	/// The params must be serializable as JSON array, see [`ToRpcServerParams`] for further documentation.
```

core/src/server/rpc_module.rs
```diff
@@ -34,7 +34,7 @@ use crate::error::{Error, SubscriptionClosed};
 use crate::id_providers::RandomIntegerIdProvider;
 use crate::server::helpers::{BoundedSubscriptions, MethodSink, SubscriptionPermit};
 use crate::server::resource_limiting::{ResourceGuard, ResourceTable, ResourceVec, Resources};
-use crate::traits::{IdProvider, ToRpcParams};
+use crate::traits::{IdProvider, ToRpcServerParams};
```
niklasad1, line 37: it might be possible to remove `ToRpcServerParams` completely but let's tackle it another PR

core/src/server/rpc_module.rs
```diff
@@ -450,7 +450,7 @@ impl Methods {
 
 	/// Helper to create a subscription on the `RPC module` without having to spin up a server.
 	///
-	/// The params must be serializable as JSON array, see [`ToRpcParams`] for further documentation.
+	/// The params must be serializable as JSON array, see [`jsonrpsee_types::ToRpcParams`] for further documentation.
```
niklasad1, line 453:
```suggestion
	/// The params must be serializable as JSON array, see [`ToRpcServerParams`] for further documentation.
```

examples/examples/core_client.rs
```diff
@@ -42,7 +43,7 @@ async fn main() -> anyhow::Result<()> {
 
 	let (tx, rx) = WsTransportClientBuilder::default().build(uri).await?;
 	let client: Client = ClientBuilder::default().build_with_tokio(tx, rx);
-	let response: String = client.request("say_hello", None).await?;
+	let response: String = client.request("say_hello", rpc_params![]).await?;
```
niklasad1, line 46: maybe we could have a type definition for empty params as well.... I reckon that some folks may not want to use the macro :P

proc-macros/src/render_client.rs
```diff
@@ -149,39 +150,38 @@ impl RpcDescription {
...
+				// Extract parameter names.
+				let param_names = extract_param_names(&signature.sig);
+				// Combine parameter names and values to pass them as parameters.
+				let params_insert = param_names.iter().zip(params).map(|pair| {
+					let name = pair.0;
+					// Throw away the type.
+					let (value, _value_type) = pair.1;
+					quote!(#name, #value)
+				});
+				quote!({
```
niklasad1, line 170: :+1:

types/src/params.rs
```diff
@@ -408,6 +416,256 @@ impl<'a> Id<'a> {
...
+	/// Finish the building process and return a JSON compatible string.
+	pub fn build(mut self) -> String {
+		let idx = self.bytes.len() - 1;
+		if self.bytes[idx] == b',' {
+			self.bytes[idx] = self.end as u8;
+		} else {
+			self.bytes.push(self.end as u8);
+		}
+
+		// Safety: This is safe because we do not emit invalid UTF-8.
```
niklasad1, line 473:
```suggestion
		// Safety: This is safe because JSON does not emit invalid UTF-8.
```

types/src/params.rs
```diff
@@ -408,6 +416,256 @@ impl<'a> Id<'a> {
...
+#[derive(Debug)]
+struct ParamsBuilder {
+	bytes: Vec<u8>,
+	end: char,
+}
+
+impl ParamsBuilder {
+	/// Construct a new [`ParamsBuilder`] with custom start and end tokens.
+	/// The inserted values are wrapped by the _start_ and _end_ tokens.
+	fn new(start: char, end: char) -> Self {
```
niklasad1, line 439: we could also get rid of the `fn new(start: char, end: char) -> Self`  here and provide two different constructs such as `fn positional() -> Self` and `fn named() -> Self `

For example

```rust
fn positional() -> Self {
    let mut bytes = Vec::with_capacity(PARAM_BYTES_CAPACITY);
    bytes.push(b'[');
    ParamsBuilder { bytes, end: b']' }
}
```


anything other than that isn't valid `JSON-RPC params` i.e, either `[ .... ]` or `{ .... }`

types/src/params.rs
```diff
@@ -408,6 +416,256 @@ impl<'a> Id<'a> {
...
+	}
+
+	/// Finish the building process and return a valid batch parameter.
+	pub fn build(self) -> Vec<(&'a str, Option<Box<RawValue>>)> {
+		self.0
+	}
+}
+
+/// Custom implementation to provide a type which contains to RPC parameters.
+impl ToRpcParams for () {
```
niklasad1, line 663: maybe `pub type EmptyParams = ()`?

niklasad1 (approved):
Looks great,

Minor grumble on the naming `UnnamedParams`, I would want something more aligned with the JSON-RPC spec.

I think `PositionalParams` or `ArrayParams` would be better but I'm not completely convinced what do you guys think @jsdw @lexnv? :)

core/src/server/rpc_module.rs
lexnv: Ops, was following `cargo test` suggestions and completely forgot about the other `server` trait. Thanks!

types/src/params.rs
lexnv: It makes things explicit for us to provide the `positional` and `named` constructors.
I would still like to keep the `new` constructor, as we have just one part of the code to modify when things need changing. I made sure to make it private tho, and didn't think another file was worth it, so I also added it into a new module

benches/bench.rs
```diff
@@ -67,15 +68,21 @@ pub fn jsonrpsee_types_v2(crit: &mut Criterion) {
 		b.iter(|| {
```
jsdw, line 68: Perhaps rename/change this and the bench below to bench building named and unnamed params, and also have a bench to set a "baseline" where we just contruct a RawValue?

types/src/params.rs
```diff
@@ -408,6 +416,275 @@ impl<'a> Id<'a> {
...
+		Self(params_builder::ParamsBuilder::positional())
+	}
+}
+
+/// Unnamed RPC parameters stored as a JSON Array object `[ value0, value1, .., valueN ]`.
+#[derive(Clone, Debug)]
+pub struct UnnamedParams(String);
+
+impl ToRpcParams for UnnamedParams {
+	fn to_rpc_params(self) -> Result<Option<Box<RawValue>>, serde_json::Error> {
```
jsdw, line 593: Nit:

Just thinking about other people implementing this; who knows what errors they might want to express in their implementations.

Is it worth doing something like:

```rust
trait ToRpcParams {
    fn to_rpc_params(self) -> Result<Option<Box<RawValue>>, jsonrpsee::Error>;
}
```

So that people have more control over what sorts of errors they spit out?

types/src/params.rs
```diff
@@ -408,6 +416,275 @@ impl<'a> Id<'a> {
...
+		/// Insert a plain value into the builder.
+		pub fn insert<P: Serialize>(&mut self, value: P) -> Result<(), serde_json::Error> {
+			serde_json::to_writer(&mut self.bytes, &value)?;
+			self.bytes.push(b',');
+
+			Ok(())
+		}
+
+		/// Finish the building process and return a JSON compatible string.
+		pub fn build(mut self) -> String {
```
jsdw, line 479: Nit: we allocate even if no params are provided (but the ToRpcParams trait allows us to avoid that by returning an Option).

Perhaps this should return `Option<String>`, and avoid allocating itself if no params are given, to take advantage of that?

jsdw:
Looks awesome; nice work @lexnv!

I left a couple of minor nits is all :)

On the naming, the spec isn't super picky itself, but I'd be fine with any of:
- NamedParams / UnnamedParams (not from the spec but imo easy to understand)
- ObjectParams / ArrayParams
- NamedParams / PositionalParams

Staring at them now, I think I'd lean towards `ObjectParams / ArrayParams` for using the spec wording and also being clear (PositionalParams feels weird to me) :)

types/src/params.rs
lexnv: I do like the idea of exposing a more generic error here. The only inconvenient I can think of is that the general [jsonrpsee::Error](https://github.com/paritytech/jsonrpsee/blob/49f260eb3ce41ad90704f783039add98661cde5f/core/src/error.rs#L59) is defined in the `core` create. 

Another good point for returning the `jsonrpsee::Error` is that's the same error we expect to return from the [client's trait](https://github.com/paritytech/jsonrpsee/blob/49f260eb3ce41ad90704f783039add98661cde5f/core/src/client/mod.rs#L58).

To achieve that, I believe we should move the `core/src/error` to `types/src/error`. Currently, we are limited by the fact the `core` crate depends on the `types` crate, and we'd want the `types` crate to also depend on `core::Error`.

I suggest we should move the files if possible. @jsdw @niklasad1 What do you think about it?

types/src/params.rs
lexnv: I was thinking that general users will want to use the `rpc_params` macro for building the parameters, as opposed to opting for the builders themselves.
While at the same time, experience users might want to implement `ToRpcParams` for their needs.

A nice thing about it is that it can generate either: `EmptyParams` (no allocation if no arguments are provided), `Unnamed/Named Params` (which does allocate). This way the `Unnamed/Named` are committed to allocation, since we expect at least one parameter and hence the `128` bytes allocation to avoid resizing. 

Do you mean to modify it in the `ParamsBuilder` or everywhere using it (Named and Unnamed)? 🤔

jsdw:
It looks like `ParamsSer` is still around; can we remove that type entirely now?

benches/bench.rs
```diff
@@ -63,19 +65,27 @@ fn v2_serialize(req: RequestSer<'_>) -> String {
...
 
 pub fn jsonrpsee_types_v2(crit: &mut Criterion) {
-	crit.bench_function("jsonrpsee_types_v2_array_ref", |b| {
+	// Construct the serialized request using the `ParamsSer` directly.
+	crit.bench_function("jsonrpsee_types_baseline_params", |b| {
 		b.iter(|| {
 			let params = &[1_u64.into(), 2_u32.into()];
 			let params = ParamsSer::ArrayRef(params);
+			let params = serde_json::to_string(&params).unwrap();
+			let params = serde_json::value::RawValue::from_string(params).unwrap();
```
jsdw, line 74: so here I thought we could just manually write eg "[1,2]" and convert that to a RawValue and then we'd have a baseline doing the "fastest thing" to benchmark our ArrayParamsBuilder against :)

we could do the same for ObjectParams too but we can add that another time!

core/src/client/mod.rs
```diff
@@ -172,21 +183,26 @@ pub trait TransportReceiverT: 'static {
...
+	() => {
+		$crate::client::__reexports::EmptyParams
+	};
 	($($param:expr),*) => {
 		{
-			let mut __params = vec![];
+			let mut __params = $crate::client::__reexports::ArrayParamsBuilder::new();
 			$(
-				__params.push($crate::client::__reexports::to_json_value($param).expect("json serialization is infallible; qed."));
+				__params.insert($param).expect(format!("Parameter `{}` cannot be serialized", stringify!($param)).as_str());
```
jsdw, line 201: Nit: we need to allocate a string to pass to `expect` even if we hope it's never called.

I wonder whether the compiler is clever here. But maybe safer to avoid the alloc unless we actually hit an error with something like:

```rust
if let Err(e) = __params.insert($param) {
    panic!("Parameter `{}` cannot be serialized: {}", stringify!($param), e)
}
```

Also, no need to use `__params` really (no chance of name conflicts); `params` would be fine :)

jsdw: (also  also; you know this already but we can remove the "empty" case and return an empty ArrayParamsBuilder.build() too so that the macro is consistently the same type and such (it is afterall very much an "array params" builder, in the same way `vec![]` always builds `Vec`s)

core/src/client/mod.rs
```diff
@@ -38,14 +38,21 @@ use futures_channel::{mpsc, oneshot};
...
 #[doc(hidden)]
 pub mod __reexports {
-	pub use crate::to_json_value;
-	pub use jsonrpsee_types::ParamsSer;
+	// Needs to be in scope for `ArrayParams` to implement it.
+	pub use crate::traits::ToRpcParams;
+	// Main builder object for constructing the rpc parameters.
+	pub use crate::params::ArrayParamsBuilder;
+	// Empty rpc parameters for empty macro.
+	pub use crate::params::EmptyParams;
```
jsdw, line 55: Ah ok, these are reexported for the macro. If this is still needed, I guess we'll be able to remove EmptyParams from this list when we update the macro :)

core/src/params.rs
```diff
@@ -0,0 +1,248 @@
...
+		/// Insert a plain value into the builder.
+		pub(crate) fn insert<P: Serialize>(&mut self, value: P) -> Result<(), serde_json::Error> {
+			serde_json::to_writer(&mut self.bytes, &value)?;
+			self.bytes.push(b',');
+
+			Ok(())
+		}
+
+		/// Finish the building process and return a JSON compatible string.
+		pub(crate) fn build(mut self) -> String {
```
jsdw, line 94: You know this already but just to write it down:

Return `Option<String>` from this, so that it can be zero alloc if no params are inserted.
Then `ObjectParams::new()` and `ArrayParams::new()` also become zero alloc and the `to_rpc_params` macro can just spit out `ArrayParams` always and be zero alloc if empty :)

core/src/params.rs
```diff
@@ -0,0 +1,248 @@
...
+
+impl Default for ArrayParamsBuilder {
+	fn default() -> Self {
+		Self(params_builder::ParamsBuilder::positional())
+	}
+}
+
+/// Array RPC parameters stored as a JSON Array object `[ value0, value1, .., valueN ]`.
+#[derive(Clone, Debug)]
+pub struct ArrayParams(String);
```
jsdw, line 205: Just a thought; do we need `ArrayParamsBuilder` + `ArrayParams` and same for objectparams?

We could just have `ArrayParams` have the functions that the current builder does, and also impl `ToRpcParams` (since that takes ownership of self anyway); would be simpler that way?

core/src/params.rs
```diff
@@ -0,0 +1,248 @@
...
+
+/// Initial number of parameters in a batch request.
+const BATCH_PARAMS_NUM_CAPACITY: usize = 4;
+
+/// Request builder that serializes RPC parameters to construct a valid batch parameter.
+/// This is the equivalent of chaining multiple RPC requests.
+#[derive(Clone, Debug, Default)]
+pub struct BatchRequestBuilder<'a>(Vec<(&'a str, Option<Box<RawValue>>)>);
+
+impl<'a> BatchRequestBuilder<'a> {
```
jsdw, line 221: Lovely :)

core/src/traits.rs
```diff
@@ -24,29 +24,97 @@
...
+///
+/// Please consider using the [`crate::params::ArrayParamsBuilder`] and [`crate::params::ObjectParamsBuilder`] than
+/// implementing this trait.
+///
+/// # Examples
+///
+/// - Implementation for hard-coded strings
+///
+/// ```rust
+///
```
jsdw, line 46: Nit: remove this empty line?

core/src/traits.rs
```diff
@@ -24,29 +24,97 @@
 // IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 // DEALINGS IN THE SOFTWARE.
 
+use crate::Error;
 use jsonrpsee_types::SubscriptionId;
 use serde::Serialize;
 use serde_json::value::RawValue;
 
-/// Marker trait for types that can be serialized as JSON array/sequence.
+/// Marker trait for types that can be serialized as JSON compatible strings.
```
jsdw, line 32:
```suggestion
/// Marker trait for types that can be serialized to compatible RPC parameters.
```

core/src/traits.rs
```diff
@@ -24,29 +24,97 @@
...
+/// This trait ensures the correctness of the RPC parameters.
+///
+/// # Note
+///
+/// Please consider using the [`crate::params::ArrayParamsBuilder`] and [`crate::params::ObjectParamsBuilder`] than
+/// implementing this trait.
+///
+/// # Examples
+///
+/// - Implementation for hard-coded strings
```
jsdw, line 43:
```suggestion
/// ## Implementation for hard-coded strings
```

core/src/traits.rs
```diff
@@ -24,29 +24,97 @@
...
+///
+/// impl ToRpcParams for ManualParam {
+///     fn to_rpc_params(self) -> Result<Option<Box<RawValue>>, Error> {
+///         // Manually define a valid JSONRPC parameter.
+///         RawValue::from_string("[1, \"2\", 3]".to_string()).map(Some).map_err(Error::ParseError)
+///     }
+/// }
+/// ```
+///
+/// - Implementation for JSON serializable structures
```
jsdw, line 61:
```suggestion
/// ## Implementation for JSON serializable structures
```

core/src/traits.rs
```diff
@@ -24,29 +24,97 @@
...
+/// };
+///
+/// impl ToRpcParams for SerParam {
+///     fn to_rpc_params(self) -> Result<Option<Box<RawValue>>, Error> {
+///         let s = String::from_utf8(serde_json::to_vec(&self)?).expect("Valid UTF8 format");
+///         RawValue::from_string(s).map(Some).map_err(Error::ParseError)
+///     }
+/// }
+/// ```
+pub trait ToRpcParams {
```
jsdw, line 82: Love the examples :)

examples/examples/proc_macro.rs
```diff
@@ -1,101 +1,17 @@
...
-//
-// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF
-// ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED
-// TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
-// PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT
-// SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
-// CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
-// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR
-// IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
-// DEALINGS IN THE SOFTWARE.
```
jsdw, line 25: The other exampels have a copyright notice; did you intend to remove this one?

jsdw:
Left a few comments for small tweaks but overall love it!

core/src/client/mod.rs
lexnv: Yep that is the case: https://rust.godbolt.org/z/E33fsWW13, have added the `if let Err` pattern to avoid an allocation. Thanks!

examples/examples/proc_macro.rs
```diff
@@ -24,78 +24,20 @@
 // IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 // DEALINGS IN THE SOFTWARE.
 
-use std::net::SocketAddr;
+//! Example of using proc macro to generate working client and server.
```
jsdw, line 27: It only generates a working client now by the looks of it :)

examples/examples/proc_macro.rs
```diff
@@ -24,78 +24,20 @@
...
-use jsonrpsee::proc_macros::rpc;
-use jsonrpsee::types::SubscriptionResult;
-use jsonrpsee::ws_client::WsClientBuilder;
-use jsonrpsee::ws_server::{SubscriptionSink, WsServerBuilder, WsServerHandle};
+use jsonrpsee::{core::RpcResult, proc_macros::rpc};
 
-type ExampleHash = [u8; 32];
-type ExampleStorageKey = Vec<u8>;
+#[rpc(client)]
+pub trait Rpc {
```
jsdw, line 32: No rush for this PR but would be good to show how the output of this macro is then used?

core/src/params.rs
```diff
@@ -0,0 +1,246 @@
...
+
+impl ToRpcParams for ArrayParams {
+	fn to_rpc_params(self) -> Result<Option<Box<RawValue>>, Error> {
+		if let Some(json) = self.0.build() {
+			RawValue::from_string(json).map(Some).map_err(Error::ParseError)
+		} else {
+			Ok(None)
+		}
+	}
+}
```
jsdw, line 220: Lovely!

core/src/client/mod.rs
```diff
@@ -172,21 +181,28 @@ pub trait TransportReceiverT: 'static {
...
+///
+/// # Panics
+///
+/// Panics if the serialization of parameters fails.
 #[macro_export]
-/// Convert the given values to a [`jsonrpsee_types::ParamsSer`] as expected by a jsonrpsee Client (http or websocket).
 macro_rules! rpc_params {
+	() => {
+		$crate::client::__reexports::ArrayParams::new()
+	};
```
jsdw, line 194: Just delete this case entirely; the one below will catch it no? :)

jsdw (approved):
Amazing!

Left a couple of tiny nits but looks great :)

examples/examples/proc_macro.rs
lexnv: Ops, seems that I was using this example for internal testing of the proc_macro. Have reverted it to origin/master, thanks!

lexnv:
Nope, we can go ahead and merge this for now 😄

PR #866, opened 2022-08-29
fix(http_server): in http2 host is not passed in headers

polachok:
In http2 host is not passed in headers which leads to this function always returning None and all http2 requests erroring with `Parse error`.

### DIFF
--- a/http-server/src/server.rs
+++ b/http-server/src/server.rs
@@ -449,6 +449,10 @@ impl<L: Logger> ServiceData<L> {
 
 		let host = match http_helpers::read_header_value(request.headers(), "host") {
 			Some(origin) => origin,
+			None if request.version() == hyper::Version::HTTP_2 => match request.uri().host() {
+				Some(origin) => origin,
+				None => return response::malformed(),
+			},
 			None => return response::malformed(),
 		};
 		let maybe_origin = http_helpers::read_header_value(request.headers(), "origin");


niklasad1 (changes requested):
Hey, we haven't really tested and targeted `HTTP2` yet.

However,  I'm not sure if hyper rejects HTTP requests without the `host header` for HTTP1.1 so 
I would prefer to check that `request.version() == Version::HTTP_2` and then disable the host filtering.

/cc @lexnv another reason to get rid of the access control stuff in favor of `tower http` :P

polachok:
Wouldn't that be a security flaw? http server supports both protocols over the same port, so skipping the check would create a way to get around access control restrictions.

niklasad1:
I meant 

```rust
let host = match http_helpers::read_header_value(request.headers(), "host") {
  Some(host) => host,
  None if request == Version::HTTP_2 => request.uri().host(),
  None => return response::malformed(),
};
```

so I think it was possible to circumvent the access control if your example if hyper doesn't reject a `HTTP 1.1 request` with a missing host header....

niklasad1 (approved):
thanks, LGTM

lexnv (approved):
LGTM! Thanks for contributing!

niklasad1:
@polachok 

dumb question: I can't find anything regarding that `http2 host` is not sent via the `host header` and I tested with curl http2 which indeed is using the host header, is this when using the streaming mode or something?

I tried a simple request with cURL and it indeed works:

http2 request
```
 $ curl --http2 http://127.0.0.1:46399 -X POST -H 'Content-Type: application/json' -d '{"id":"1","method":"add","jsonrpc":"2.0" }'
```

server
```
2022-09-01T09:23:54.440026Z TRACE jsonrpsee_server::server: Request { method: POST, uri: /, version: HTTP/1.1, headers: {"host": "127.0.0.1:46399", "user-agent": "curl/7.85.0", "accept": "*/*", "connection": "Upgrade, HTTP2-Settings", "upgrade": "h2c", "http2-settings": "AAMAAABkAAQCAAAAAAIAAAAA", "content-type": "application/json", "content-length": "42"}, body: Body(Streaming) }
2022-09-01T09:23:54.440283Z TRACE method_call{method=add}: jsonrpsee_core::tracing: recv="{\"jsonrpc\":\"2.0\",\"id\":\"1\",\"method\":\"add\",\"params\":null}"
```

To conclude we need some tests for http2 as well which can be done in another PR.

polachok:
>http2 request

Not a http2 request. In your own output it shows version HTTP/1.1.

To make a http2 request with curl you have to add `--http2-prior-knowledge`

PR #873, opened 2022-09-08
server: Optimize polling the `ServerHandle`

lexnv:
## Why

Polling by the following method is basically busy polling that causes strain on the
Tokyo async task scheduler, which may negatively impact other use-cases. This
method is invoked when users want to wait until the server exits via the `ServerHandle`. 

```rust
fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
  if self.is_stopped() {
	  Poll::Ready(())
  } else {
	  cx.waker().wake_by_ref();
	  Poll::Pending
  }
}
```

## Impact

This patch effectively reduces the number of polls by a factor of __150_000 x__.

Number of Polls | Before | After
|----|----|----|
ws_server_notify_client_on_disconnect_with_closed_server | 309281 | 2 |
ws_server_notify_client_on_disconnect |  618285 | 4 |

## How

Initial attempts to resolve this issue were using a `let future = Box::pin(self.inner.close())` 
to provide a pollable future for the `fn poll`. This however was not resuming code execution
after the `.await` point because the `fn poll` of the `Future` trait needs to poll on the
same future each time it is invoked by the scheduler, as opposed to creating new one with `Box::pin`.

This is a case of self-referencing structs, which is completely sane for this scenario.
We need to have access to a `watch::Sender<()>` while also having access to its
internal `watch::Sender<()>::closed()` method for polling purposes.
However, we need to circumvent the compiler's restrictions if we wish to optimize polling.

The boxed future needs to live for long enough to ensure the compiler's compliance.
This means that the sender must be 'static, as the future relies upon `self.inner.closed()` to live
long enough.

Obtain a static reference by leaking the box on the heap, then reconstruct and clean-up the memory manually.

Using `Box::leak` as opposed to `Box::into_raw` has two main advantages:
 - It provides `&'a mut T` that can be shared `Send` bounded for the Tokyo's task
 - Allows us to safely use the reference, while still being able to convert it to a pointer.

This wouldn't be necessary if the future we are waiting for was independent of `self`.

## Memory Safety

We leaked the `watch::Sender<()>` in the `new` constructor and we wish to free up
the heap memory in the `Drop`'s implementation.

To go from `&'a [mut] T` to `*[mut] T` we normally need to go through `Box::leak` first.

This is because `Box` is recognized as a "unique pointer" by the compiler's borrow checker,
but internally its nothing more than a raw pointer.
Directly turning the reference into a raw pointer would not permit aliased raw accesses,
and is rejected by the compiler.

However, we used exactly `Box::leak` to get the `&'static watch::Sender<()>` reference in the
`new` constructor.
Therefore, it is safe to turn that reference into a pointer, but not by casting.

The final step is to reconstruct the `Box` created by `Box::new(tx)` in the `new` constructor
to automatically clean up the memory.

## Testing Done

Jsonrpsee has two tests set in place for this exact scenario, namely `ws_server_notify_client_on_disconnect_with_closed_server` and 
`ws_server_notify_client_on_disconnect`.

To measure the number of polls the following patch was utilized (adjusted for before/after):

```rust
static COUNTER: AtomicUsize = AtomicUsize::new(0);

impl Future for ServerHandle {
	type Output = ();

	fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
		COUNTER.fetch_add(1, Ordering::SeqCst);

		match self.future.as_mut().poll(cx) {
			Poll::Ready(()) => {
				println!("Polled: {:?}", COUNTER);
				Poll::Ready(())
			}
			_ => Poll::Pending
		}
	}
}
```


server/src/future.rs
```diff
@@ -170,40 +172,89 @@ impl StopHandle {
...
+		//
+		// Obtain a static reference by leaking the box on the heap,
+		// then reconstruct and clean-up the memory manually.
+		//
+		// Using `Box::leak` as opposed to `Box::into_raw` has two main advantages:
+		//  - It provides `&'a mut T` that can be shared `Send` bounded for the tokio's task
+		//  - Allows us to safely use the reference, while still being able to convert it to a pointer.
+		//
+		// This wouldn't be necessary if the future we are waiting for was independent of `self`.
+		let tx = Box::new(tx);
```
niklasad1, line 206: dq: why are you boxing this twice?

server/src/future.rs
```diff
@@ -170,40 +172,89 @@ impl StopHandle {
...
-#[derive(Debug, Clone)]
-pub struct ServerHandle(Arc<watch::Sender<()>>);
+pub struct ServerHandle {
+	inner: &'static watch::Sender<()>,
+	future: BoxFuture<'static, ()>,
+}
+
+impl std::fmt::Debug for ServerHandle {
+	fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+		write!(f, "{:?}", self.inner)
```
niklasad1, line 182: maybe we could use `debug struct` instead `write!` here

niklasad1 (approved):
I ran `valgrind` against this PR and looks good to me.

Very nice documentation regarding the unsafe stuff.

niklasad1:
@lexnv 

nicely done but we decided to skip implementing `Future` for the `ServerHandle` instead `ServerHandle::stopped` can be used.

It's nice to impl Future as `API` but not worth the effort having these amount of unsafe code it's hard enough to write correct rust code :)

PR #883, opened 2022-09-22
more stable benches

niklasad1:
I have noticed that some benchmarks generates warnings 

```
Benchmarking async/http_round_trip/memory_intense: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 7.4s, enable flat sampling, or reduce sample count to 50.

```

This PR changes the "measurement sampling time" to be configurable via environment variables and changes the default values. 

The default values for the benchmarks may be a bit extreme but I want ensure that the benches doesn't vary that much just because we are not sampling enough i.e, run the benchmarks not enough times to get reliable results.

It will make the benches super slow to run but because this runs in cronjob it should ok.

### DIFF
--- a/benches/README.md
+++ b/benches/README.md
@@ -44,5 +44,5 @@ It's also possible to run profiling on individual benchmarks by:
 Some of the benchmarks are quite expensive to run and doesn't run with enough samples with the default values
 provided by criterion. Currently the default values are very conversative which can be modified by the following environment variables:
 
-    - "SLOW_SAMPLE_TIME" - sets the measurement time for slow benchmarks (default is 250 seconds)
-    - "SAMPLE_TIME" - sets the measurement time for fast benchmarks (default is 10 seconds)
+    - "SLOW_MEASUREMENT_TIME" - sets the measurement time for slow benchmarks (default is 250 seconds)
+    - "MEASUREMENT_TIME" - sets the measurement time for fast benchmarks (default is 10 seconds)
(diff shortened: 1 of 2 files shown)


benches/bench.rs
```diff
@@ -23,7 +23,7 @@ fn slow_sample_time() -> Duration {
 }
 
 fn sample_time() -> Duration {
-	std::env::var("SAMPLE_TIME").map_or(Duration::from_secs(10), |val| {
```
jsdw: Dumb question; what does this effect/do?

jsdw: Ooh I see; it's how long it'll repeat the benchmark for to gather data.

jsdw (approved):

benches/bench.rs
niklasad1: is measurement time better?

benches/README.md
```diff
@@ -38,3 +38,11 @@ This will generate a flamegraph for the specific benchmark in `./target/criterio
...
 
 `$ cargo bench --bench bench -- --profile-time=60 sync/http_concurrent_conn_calls/1024`
+
+## Measurement time of benchmarks
+
+Some of the benchmarks are quite expensive to run and doesn't run with enough samples with the default values
+provided by criterion. Currently the default values are very conversative which can be modified by the following environment variables:
+
+    - "SLOW_SAMPLE_TIME" - sets the measurement time for slow benchmarks (default is 250 seconds)
+    - "SAMPLE_TIME" - sets the measurement time for fast benchmarks (default is 10 seconds)
```
niklasad1, line 48:
```suggestion
    - "SAMPLE_TIME" - sets the measurement time for fast benchmarks (default is 50 seconds)
```

lexnv (approved):

benches/bench.rs
jsdw: that'd be my choice personally :)

PR #884, opened 2022-09-26
[ci] Additional variables for benchmarks execution

alvicsam:
Added `MEASUREMENT_TIME: 3` and `SLOW_MEASUREMENT_TIME: 40` variables to benchmark job. With these variables it takes ~2h to run benchmarks: https://gitlab.parity.io/parity/mirrors/jsonrpsee/-/jobs/1885548

cc https://github.com/paritytech/ci_cd/issues/302

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -110,7 +110,7 @@ benchmarks:
   <<:                              *docker-env
   <<:                              *collect-artifacts
   <<:                              *schedule-refs
-  timeout:                         12h
+  timeout:                         3h
   variables:
     MEASUREMENT_TIME:              3
     SLOW_MEASUREMENT_TIME:         40
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -113,7 +113,7 @@ benchmarks:
   timeout:                         3h
   variables:
     MEASUREMENT_TIME:              3
-    SLOW_MEASUREMENT_TIME:         40
+    SLOW_MEASUREMENT_TIME:         60
   script:
     - cargo bench -p jsonrpsee-benchmarks -- --output-format bencher | tee output.txt
     - mkdir artifacts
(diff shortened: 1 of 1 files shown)


niklasad1:
I suggest to make slow stuff `60s` but I have done some investigation and with some improvement it still took ~3h to complete.

I need to refactor the benches to be faster but the bad the thing is that we can't compare with the old data then :( 

See https://github.com/paritytech/jsonrpsee/pull/885

alvicsam:
> I need to refactor the benches to be faster but the bad the thing is that we can't compare with the old data then :(

If we consider some data not accurate maybe it's not that bad?

> I still think 3h is way too long.

But why? Because for local run it's too long?

niklasad1:
> If we consider some data not accurate maybe it's not that bad?

Probably yes, but I can try backporting the benches to the "most recent" releases to have some data compare against at least.

> But why? Because for local run it's too long?

Yeah or unnecessary load on the benchmarks itself, we are testing with some extreme limits such a `calls with 1 MB payload`, `slow call with a sleep for 1 ms` and up to 1024 concurrent requests.

I think we can relax the `payload size` and `sleep time` to make it run "reasonable time (less than 1 hour)"  without any impact. Trying it right now locally...

alvicsam:
Also it's possible to benchmarks with cargo-nextest after criterion support is added to nextest: https://github.com/nextest-rs/nextest/issues/96. It can significantly reduce time

niklasad1 (approved):

sergejparity (approved):

.gitlab-ci.yml
```diff
@@ -110,6 +110,10 @@ benchmarks:
   <<:                              *docker-env
   <<:                              *collect-artifacts
   <<:                              *schedule-refs
+  timeout:                         3h
+  variables:
+    MEASUREMENT_TIME:              3
```
niklasad1, line 115:
```suggestion
    MEASUREMENT_TIME:              10
```

alvicsam:
Okay, I'd like to wait https://gitlab.parity.io/parity/mirrors/jsonrpsee/-/pipelines/217578 to check how long the benches will take with this settings. If the job succeeds I'll merge this PR

PR #885, opened 2022-09-26
fix(benches): relax `default measurement time`, split `benches` and run `fast call` only on batches

niklasad1:
The default `measurement_time` was too extreme which caused the GHA cronjob to not complete in 4 hours...

This PR changes the batch benches to run only on `fast call`, split benches and relax default measurement time

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -22,7 +22,7 @@ fn measurement_time_slow() -> Duration {
 	})
 }
 
-fn measurement_time() -> Duration {
+fn measurement_mid() -> Duration {
 	std::env::var("MEASUREMENT_TIME").map_or(Duration::from_secs(10), |val| {
 		Duration::from_secs(val.parse().expect("SAMPLE_TIME must be an integer"))
 	})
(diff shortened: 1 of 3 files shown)


lexnv (approved):

benches/helpers.rs
```diff
@@ -16,6 +18,7 @@ pub(crate) const ASYNC_METHODS: [&str; 3] = [SYNC_FAST_CALL, SYNC_MEM_CALL, SYNC
 
 // 1 KiB = 1024 bytes
 pub(crate) const KIB: usize = 1024;
+pub(crate) const SLOW_CALL: Duration = Duration::from_micros(50);
```
niklasad1, line 21: changes the `slow_call` from 1ms to 50us

the bad thing is that we can't really compare old results when introducing this

benches/helpers.rs
```diff
@@ -26,14 +29,14 @@ pub async fn http_server(handle: tokio::runtime::Handle) -> (String, jsonrpc_htt
 	let mut io = IoHandler::new();
 	io.add_sync_method(SYNC_FAST_CALL, |_| Ok(Value::String("lo".to_string())));
 	io.add_method(ASYNC_FAST_CALL, |_| async { Ok(Value::String("lo".to_string())) });
-	io.add_sync_method(SYNC_MEM_CALL, |_| Ok(Value::String("A".repeat(1 * 1024 * 1024))));
-	io.add_method(ASYNC_MEM_CALL, |_| async { Ok(Value::String("A".repeat(1 * 1024 * 1024))) });
+	io.add_sync_method(SYNC_MEM_CALL, |_| Ok(Value::String("A".repeat(KIB))));
```
niklasad1, line 32: changes the `memory intense call` from 1MB to 1KB

the bad thing is that we can't really compare old results when introducing this

benches/helpers.rs
```diff
@@ -16,6 +18,8 @@ pub(crate) const ASYNC_METHODS: [&str; 3] = [SYNC_FAST_CALL, SYNC_MEM_CALL, SYNC
 
 // 1 KiB = 1024 bytes
 pub(crate) const KIB: usize = 1024;
+pub(crate) const MIB: usize = KIB * KIB;
```
niklasad1, line 21:
```suggestion
pub(crate) const MIB: usize = 1024 * KIB;
```

benches/bench.rs
```diff
@@ -254,35 +300,43 @@ fn sub_round_trip(rt: &TokioRuntime, crit: &mut Criterion, client: Arc<impl Subs
...
+	rt: &TokioRuntime,
+	crit: &mut Criterion,
+	url: &str,
+	name: &str,
+	request: RequestType,
+	concurrent_conns: &[usize],
+) {
+	let method = request.methods()[0];
+	let bench_name = format!("{}/{}", name, method);
+	let mut group = crit.benchmark_group(request.group_name(&bench_name));
```
niklasad1, line 338: tweak the name, it will `ws_concurrent_conn_calls/fast_call/<number of concurrent calls>`

benches/bench.rs
```diff
@@ -254,35 +300,43 @@ fn sub_round_trip(rt: &TokioRuntime, crit: &mut Criterion, client: Arc<impl Subs
...
-		for batch_size in [2, 5, 10, 50, 100usize].iter() {
-			let mut batch = BatchRequestBuilder::new();
-			for _ in 0..*batch_size {
-				batch.insert(method, ArrayParams::new()).unwrap();
-			}
-			group.throughput(Throughput::Elements(*batch_size as u64));
-			group.bench_with_input(BenchmarkId::from_parameter(batch_size), batch_size, |b, _| {
-				b.to_async(rt).iter(|| async { client.batch_request::<String>(batch.clone()).await.unwrap() })
-			});
+	let method = request.methods()[0];
```
niklasad1, line 311: run only `fast_call` on the batch bench

jsdw: Worth a comment in the code for this?

jsdw (approved):

benches/bench.rs
niklasad1: added asserts for these which should be verbose enough I hope? :)

PR #886, opened 2022-09-28
server: add `transport protocol details` to the logger trait

### LINKED ISSUE #882: [server]: `metric/logger` not able to distinguish between `WS` and `HTTP` calls
After the refactoring of the `server` to support both `HTTP` and `WS` it is not possible to know whether a call was a `HTTP` or `WS`.

Currently we have [this trait](https://github.com/paritytech/jsonrpsee/blob/master/server/src/logger.rs#L68-#L90) so I can think of having two different trait or having another parameter such as:

```rust
pub trait Logger: Send + Sync + Clone + 'static {
	/// Intended to carry timestamp of a request, for example `std::time::Instant`. How the trait
	/// measures time, if at all, is entirely up to the implementation.
	type Instant: std::fmt::Debug + Send + Sync + Copy;

	/// Called when a new client connects
	fn on_connect(&self, _remote_addr: SocketAddr, _request: &HttpRequest, protocol: &str);

	/// Called when a new JSON-RPC request comes to the server.
	fn on_request(&self) -> Self::Instant;

	/// Called on each JSON-RPC method call, batch requests will trigger `on_call` multiple times.
	fn on_call(&self, method_name: &str, params: Params, kind: MethodKind, protocol: &str);

	/// Called on each JSON-RPC method completion, batch requests will trigger `on_result` multiple times.
	fn on_result(&self, method_name: &str, success: bool, started_at: Self::Instant, protocol: &str);

	/// Called once the JSON-RPC request is finished and response is sent to the output buffer.
	fn on_response(&self, result: &str, started_at: Self::Instant, protocol: &str);

	/// Called when a client disconnects
	fn on_disconnect(&self, _remote_addr: SocketAddr);
}
```

niklasad1:
Close https://github.com/paritytech/jsonrpsee/issues/882

In addition I added that `http` also calls `Logger::on_connect` and `Logger::on_disconnect`, these could also be fetched via middleware but someone wants to have a metrics for that let's provide it for a consistent API.

companion: https://github.com/paritytech/substrate/compare/na-jsonrpsee-v0.16-comp?expand=1

### DIFF
--- a/examples/examples/logger.rs
+++ b/examples/examples/logger.rs
@@ -28,7 +28,7 @@ use std::net::SocketAddr;
 use std::time::Instant;
 
 use jsonrpsee::core::client::ClientT;
-use jsonrpsee::server::logger::{self, HttpRequest, MethodKind, Params};
+use jsonrpsee::server::logger::{self, HttpRequest, MethodKind, Params, TransportProtocol};
 use jsonrpsee::server::ServerBuilder;
 use jsonrpsee::ws_client::WsClientBuilder;
 use jsonrpsee::{rpc_params, RpcModule};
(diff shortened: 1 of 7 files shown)


TarikGul (approved):
Looks solid!

lexnv (approved):

jsdw (approved):

PR #889, opened 2022-10-03
gitlab benches: show results on github pages

niklasad1:
Currently, the benches are also executed in gitlab and the results are pushed every day to the branch `gh-pages` at `<dd-mm-yy>/output.txt` which this PR reads and construct displayable results using the `rhysd/github-action-benchmark@v1` GHA.

Similar to what already has but will display the results on https://paritytech.github.io/jsonrpsee/bench/dev2/ if this works

### DIFF
--- a/.github/workflows/benchmarks_gitlab.yml
+++ b/.github/workflows/benchmarks_gitlab.yml
@@ -23,7 +23,7 @@ jobs:
       - name: Set bench file
         id: step_one
         run: |
-          echo "bench_file=$(date "+%d-%m-%Y")/output.txt" >> $GITHUB_ENV
+          echo "bench_file=bench/$(date "+%d-%m-%Y")/output.txt" >> $GITHUB_ENV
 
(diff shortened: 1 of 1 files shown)


jsdw (approved):

alvicsam (approved):

.github/workflows/benchmarks_gitlab.yml
```diff
@@ -0,0 +1,39 @@
...
+        uses: actions-rs/toolchain@v1.0.7
+        with:
+          profile: minimal
+          toolchain: nightly
+          override: true
+
+      - name: Set bench file
+        id: step_one
+        run: |
+          echo "bench_file=$(date "+%d-%m-%Y")/output.txt" >> $GITHUB_ENV
```
niklasad1, line 26:
```suggestion
          echo "bench_file=bench/$(date "+%d-%m-%Y")/output.txt" >> $GITHUB_ENV
```

PR #894, opened 2022-10-04
fix(gitlab bench): fix path

niklasad1:
See https://github.com/paritytech/jsonrpsee/actions/runs/3184395309 it seems add an additional `/jsonrpsee path` for some reason :P

PR #895, opened 2022-10-05
[ci] Fix checkout ref for GHA benchmarks-gitlab

alvicsam:
Action [fails](https://github.com/paritytech/jsonrpsee/actions/runs/3184395309/jobs/5192746100) because it checkout master and can't find `output.txt`

### DIFF
--- a/.github/workflows/benchmarks_gitlab.yml
+++ b/.github/workflows/benchmarks_gitlab.yml
@@ -9,9 +9,10 @@ jobs:
     name: Benchmarks gitlab
     runs-on: ubuntu-latest
     steps:
-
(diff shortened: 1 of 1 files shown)


niklasad1:
Have you tested it?

I created #894 but maybe this is correct fix not sure

alvicsam:
I haven't noticed https://github.com/paritytech/jsonrpsee/pull/894, this pr has the same changes. Closing it

niklasad1:
sorry for hijacking your PR and thank you, my initial fix was wrong so I used your fix but as it was already approved and could be merged faster.

also tried it with `act` locally hopefully it will work now :pray:

PR #896, opened 2022-10-05
fix: proper error on invalid batches

polachok:
Currently, if batch requests fails to parse inner requests, it returns a single object:

```
{
  "jsonrpc": "2.0",
  "error": {
    "code": -32700,
    "message": "Parse error"
  },
  "id": null
}
```

But according to spec (see below) it should return an array:
```
[
  {
    "jsonrpc": "2.0",
    "error": {
      "code": -32600,
      "message": "Invalid request"
    },
    "id": "kek"
  },
  {
    "jsonrpc": "2.0",
    "error": {
      "code": -32600,
      "message": "Invalid request"
    },
    "id": null
  }
]
```

From spec:

> The Server should respond with an Array containing the corresponding Response objects, after all of the batch Request objects have been processed. A Response object SHOULD exist for each Request object, except that there SHOULD NOT be any Response objects for notifications. The Server MAY process a batch rpc call as a set of concurrent tasks, processing them in any order and with any width of parallelism.
> 
> 
> If the batch rpc call itself fails to be recognized as an valid JSON or as an Array with at least one value, the response from the Server MUST be a single Response object.

### DIFF
--- a/core/src/server/helpers.rs
+++ b/core/src/server/helpers.rs
@@ -160,6 +160,24 @@ pub fn prepare_error(data: &[u8]) -> (Id<'_>, ErrorCode) {
 	}
 }
 
+/// Figure out if this is a sufficiently complete request that we can extract [`Id`] out of, or just plain
+/// unparseable garbage.
+pub fn prepare_batch_error(data: &[u8]) -> Result<(Vec<Id<'_>>, ErrorCode), ErrorCode> {
+	match serde_json::from_slice::<Vec<&serde_json::value::RawValue>>(data) {
+		Ok(values) => {
+			let ids = values
+				.into_iter()
+				.map(|value| match serde_json::from_str::<InvalidRequest>(value.get()) {
+					Ok(req) => req.id,
+					Err(_) => Id::Null,
+				})
+				.collect();
+			Ok((ids, ErrorCode::InvalidRequest))
+		}
+		Err(_) => Err(ErrorCode::ParseError),
+	}
+}
+
 /// A permitted subscription.
 #[derive(Debug)]
 pub struct SubscriptionPermit {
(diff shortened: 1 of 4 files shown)


core/src/server/helpers.rs
```diff
@@ -160,6 +160,24 @@ pub fn prepare_error(data: &[u8]) -> (Id<'_>, ErrorCode) {
...
 }
 
+/// Figure out if this is a sufficiently complete request that we can extract [`Id`] out of, or just plain
+/// unparseable garbage.
+pub fn prepare_batch_error(data: &[u8]) -> Result<(Vec<Id<'_>>, ErrorCode), ErrorCode> {
+	match serde_json::from_slice::<Vec<&serde_json::value::RawValue>>(data) {
+		Ok(values) => {
+			let ids = values
+				.into_iter()
+				.map(|value| match serde_json::from_str::<InvalidRequest>(value.get()) {
```
niklasad1, line 170: FYI, because we are using `serde::RawValue` and it doesn't work for untagged enums.

For example

```json
[
        {"jsonrpc": "2.0", "method": "foo", "id": "1"},
        // notification (this shouldn't be answered)
        {"jsonrpc": "2.0", "method": "foo"},
        // invalid call
        {"foo": "boo"},
]

// according the spec this should be returned
[
        {"jsonrpc": "2.0", "result": 7, "id": "1"},
        {"jsonrpc": "2.0", "error": {"code": -32600, "message": "Invalid Request"}, "id": null},
]
```

Strictly speaking we are not following the spec when for example a mixed batch with calls and notifications doesn't work because of the serde::RawValue thingy...

So if we going to change this according to the spec would seems useful to i.e, if a one call fails in batch you would still want the result of the results

With that said, you need check if the `value` is `Request, Notification or InvalidRequest` to do this correctly. If all those fails then regard it as ParseError.

Thus, we would need some test for that as well


Can you add a test for a batch such as:

niklasad1: Your code would just return:

```json
[invalid null, invalid null, invalid null]
```

which isn't correct but because of `RawValue` stuff which is tricky for sure ^^

niklasad1 (commented):
Overall this seems like a nice thing to have but I have some comments :)

Basically, we did this simple because of the limitations of `serde::RawValue` but I think with some extra fixes to your PR that could be fixed as well.

niklasad1:
superseeded by #917

PR #897, opened 2022-10-05
fix gha gitlab bench again

niklasad1:
Finally works take a look at: https://paritytech.github.io/jsonrpsee/bench/dev2/

I enabled it in the CI pipeline but removed have a look a the `test test commit`


.github/workflows/benchmarks_gitlab.yml
```diff
@@ -22,20 +23,27 @@ jobs:
...
+      - name: Copy bench results
         id: step_one
         run: |
-          echo "bench_file=bench/$(date "+%d-%m-%Y")/output.txt" >> $GITHUB_ENV
+          cp bench/$(date "+%d-%m-%Y")/output.txt output.txt
+
+      - name: Switch branch
+        id: step_two
+        run: |
+          git checkout master
```
niklasad1, line 37: apparently the github action does some git magic under the hood and doesn't work if we are already on gh-pages branch

.github/workflows/benchmarks_gitlab.yml
```diff
@@ -22,20 +23,27 @@ jobs:
...
 
-      - name: Set bench file
+      - name: Rust Cache
+        uses: Swatinem/rust-cache@v2.0.0
+
+      - name: Copy bench results
         id: step_one
         run: |
-          echo "bench_file=bench/$(date "+%d-%m-%Y")/output.txt" >> $GITHUB_ENV
+          cp bench/$(date "+%d-%m-%Y")/output.txt output.txt
```
lexnv, line 32: nit: Could it be that the `$(date "+%d-%m-%Y")` leads to an inconsistent file between when:
 * (a) the bench file was published 
 * (b) this cronjob gets invoked?
 
 Ie, maybe we'll modify the cronjob's execution time and land one file behind?
 
 Something like the following might solve it, but didn't try it
 
 ```suggestion
           cp bench/$(ls -tA bench/ | head -n 1)/output.txt output.txt
```

I think another alternative would be for the gitlab script to also create a sym link `./bench/latest/output.txt` 😄

**Review by lexnv — APPROVED** [review]

File: .github/workflows/benchmarks_gitlab.yml (hunk 2 above)
Comment (niklasad1) [reply .github/workflows/benchmarks_gitlab.yml:32]: yeah, we could improve that but your suggestion doesn't work currently because they are other folders in bench as well

Comment (niklasad1) [reply .github/workflows/benchmarks_gitlab.yml:32]: `bench/latest/output.txt` could work and our GHA could remove that file after producing the results successfully.

no issue with the date-format yet it was some issue that `gha action` switches to `gh-pages` and that's doesn't work if we already have switched to `gh-pages` branch.

that's it :P

**Review by jsdw — APPROVED** [review]

File: .github/workflows/benchmarks_gitlab.yml [hunk 3]
```diff
@@ -2,7 +2,7 @@ name: Benchmarks gitlab
 
 on:
   schedule:
-    - cron: "0 10 * * *"
+    - cron: "0 13 * * *"
```
niklasad1, line 5:
```suggestion
    - cron: "0 14 * * *"
```

alvicsam (approved):

.github/workflows/benchmarks_gitlab.yml
niklasad1: @alvicsam https://github.com/paritytech/jsonrpsee/actions/runs/3190518964/jobs/5205730855 :tada:

PR #900, opened 2022-10-12
server: Expose the subscription ID

lexnv:
The subscription ID on the server side is required by the new [RPC Spec](https://paritytech.github.io/json-rpc-interface-spec/introduction.html),
specifically by the [archive](https://paritytech.github.io/json-rpc-interface-spec/api/archive.html) class of methods.

The subscription ID generated from the [chainHead_unstable_follow](https://paritytech.github.io/json-rpc-interface-spec/api/chainHead_unstable_follow.html#chainhead_unstable_follow) is passed to
the other archive methods (ie [chainHead_unstable_body](https://paritytech.github.io/json-rpc-interface-spec/api/chainHead_unstable_body.html#chainhead_unstable_body)). The server needs to
keep track of this ID for validity purposes.

This PR exposes the subscription ID via the `SubscriptionSink::accept`.
The `accept` method marks the point of sending the ID to the client.
Although the subscription ID exists on the server side prior to this call,
this information is unusable if it cannot be propagated to the client.

This adds a change to the main `SubscriptionSink` API without breaking 
the current behavior of `SubscriptionSink::pipe_from_stream`.
If the server is interested in the subscription ID, it can simply call the 
`accept` prior to `pipe_from_stream`.

This is part of the https://github.com/paritytech/substrate/issues/12071.

### DIFF
--- a/core/src/server/rpc_module.rs
+++ b/core/src/server/rpc_module.rs
@@ -853,10 +853,9 @@ impl SubscriptionSink {
 	}
 
 	/// Attempt to accept the subscription and respond the subscription method call.
-	/// The subscription ID is returned on success.
 	///
 	/// Fails if the connection was closed, or if called multiple times.
-	pub fn accept(&mut self) -> Result<RpcSubscriptionId<'static>, SubscriptionAcceptRejectError> {
+	pub fn accept(&mut self) -> Result<(), SubscriptionAcceptRejectError> {
 		let (id, subscribe_call) = self.id.take().ok_or(SubscriptionAcceptRejectError::AlreadyCalled)?;
 
 		let response = MethodResponse::response(id, &self.uniq_sub.sub_id, self.inner.max_response_size() as usize);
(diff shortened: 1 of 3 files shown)


core/src/server/rpc_module.rs
```diff
@@ -853,9 +853,10 @@ impl SubscriptionSink {
 	}
 
 	/// Attempt to accept the subscription and respond the subscription method call.
+	/// The subscription ID is returned on success.
 	///
 	/// Fails if the connection was closed, or if called multiple times.
-	pub fn accept(&mut self) -> Result<(), SubscriptionAcceptRejectError> {
+	pub fn accept(&mut self) -> Result<RpcSubscriptionId<'static>, SubscriptionAcceptRejectError> {
```
niklasad1, line 859: what's the benefit of returning `SubscriptionId` here instead of having `fn subscription_id(&self) -> Option<RpcSubscritionId<'static>>` as well?

It's rather edge-case to access the subscription id and I think 90% of the use cases doesn't need to know this...

lexnv: Yep indeed, this should be an edge-case and not the most common one.
It does make sense to have a dedicated method for this.
Thanks for having a look! 👍

niklasad1 (approved):

jsdw (approved):

PR #910, opened 2022-10-20
[client]: improve batch request API

### LINKED ISSUE #909: [client]: batch request API is non-ergonomic
After some debugging of batch requests it's hard to investigate why a certain batch request fails.
Currently, the assumption of the code is that all calls in the batch must succeed otherwise it's treated as an error.

For instance if one make a batch request with 100 calls inside, the 99 successful responses gets thrown away and not possible know that without using the logs, wireshark or something else. That lead me into the idea to return the entire batch regardless whether it was successful or not and let to user decide that.

It's not possible with the current API:

```rust
async fn batch_request<'a, R>(&self, batch: BatchRequestBuilder<'a>) -> Result<Vec<R>, Error>
	where
		R: DeserializeOwned + Default + Clone;
```

So I have two suggestions:

1. High-level API similar to current one but return `Vec<R, ErrorResponse>`
```rust
async fn batch_request<'a, R>(&self, batch: BatchRequestBuilder<'a>) -> Result<Vec<Result<R, ErrorResponse>, Error>
	where
		R: DeserializeOwned + Default + Clone;
```

it's then up to the user go through responses and it will be answered in the same order as inserted in the `BatchRequestBuilder`

2. Low-level API with JSON value 

```rust
enum DecodedResponse<'a> {
   Success(Response<'a, JsonValue>),
   Failed(ErrorResponse<'a>)
}

async fn batch_request<'a, R>(&self, batch: BatchRequestBuilder<'a>) -> Result<Vec<DecodedResponse>, Error>

```
the user has to deal with parsing the JsonValue to something else

niklasad1:
Close #909 

After some discussions with Kian about the API in #909 it simply quite annoying for users to deal with i.e, to iterate over all responses to check whether it was successful but useful in some contexts such as if you want to exactly which requests failed.

After some suggestions by James, I reworked the API with an iterator-like API.

The responses still doesn't expose the request ID mainly because jsonrpsee controls the request IDs and changing that quite a big change and doesn't work well with the current `BatchRequestBuilder`.

Thus, the only way to map the requests to reponses is to copy the batch request and then iterate over the `BatchResponse::into_iter` together to figure it which calls that failed such as:

```rust
	let mut batch_request = BatchRequestBuilder::new();
	batch_request.insert("say_hello", rpc_params![]).unwrap();
	batch_request.insert("say_goodbye", rpc_params![0_u64, 1, 2]).unwrap();
	batch_request.insert("get_swag", rpc_params![]).unwrap();

        let response = client.batch_request(batch_request.clone()).await.unwrap();

        for (request, response) in batch_request.zip(response.iter()) {
            match response {
               Ok(r) => {
                   println!("request: {:?} succeed with response: {}", request, r); 
               }
               Err(err) => {
                   println!("request: {:?} failed with response: {}", request, err); 
               }
            }
        }
```

Ideally, the `BatchRequestBuilder` would inject the request IDs and allow users to add their own IDs but because we use the ID range as digest for each batch request it would make things harder.

Thus,  I think this is good enough API for now.


niklasad1:
ah, I realized that this won't work for request IDs as Strings and hence those are lexicographical ordered ^^

benches/bench.rs
```diff
@@ -44,11 +44,11 @@ criterion_group!(
 	config = Criterion::default().with_profiler(PProfProfiler::new(100, Output::Flamegraph(None))).measurement_time(measurement_time_slow());
 	targets = SyncBencher::http_benches_slow, SyncBencher::websocket_benches_slow
 );
-criterion_group!(
+/*criterion_group!(
```
niklasad1, line 47: this is just some dummy stuff such that I can bench against the old batch request client.

I will revert this when the PR is reviewed properly.

jsdw: Is it time to uncomment this again before we approve/merge?

benches/bench.rs
```diff
@@ -69,7 +69,7 @@ criterion_main!(
 	sync_benches,
 	sync_benches_mid,
 	sync_benches_slow,
-	async_benches,
+	//async_benches,
```
jsdw, line 72: And this

core/src/client/mod.rs
```diff
@@ -468,14 +475,113 @@ pub enum IdKind {
...
+	pub fn num_failed_calls(&self) -> usize {
+		self.failed_calls
+	}
+
+	/// Returns `Ok(iterator)` if all responses were successful
+	/// otherwise `Err(iterator)` is returned.
+	///
+	/// If you want get all responses if an error responses occurs use [`BatchResponse::into_iter`]
+	/// instead where it's possible to implement customized logic.
+	pub fn ok(
```
jsdw, line 539: I wonder whether the names should be `ok` and `into_ok` to mirror `iter` and `into_iter`?

jsdw (approved):
Looks great to me, ncie one!

lexnv (approved):
👍  LGTM!

PR #921, opened 2022-10-31
proc macro: support camelCase & snake_case for object params

### LINKED ISSUE #920: [proc_macro] Support named RPC parameters with naming conventions
The RPC server can receive the parameters as named, or unnamed.

The spec does not enforce any naming conventions on those parameters:
- named with snakeCase:
```json
{ "jsonrpc": "2.0", "method": "subtract", "params": {"one_one": 23, "two_two": 42}, "id": 3}
```

- name with camelCase:
```json
{ "jsonrpc": "2.0", "method": "subtract", "params": {"oneOne": 23, "twoTwo": 42}, "id": 3}
```

The current implementation provides a 1-to-1 match, meaning that the definition of the method / subscription
is driving the deserialization of parameters.

The following form accepts only snake case named parameters.
```rust
#[method(name = "getString")]
async fn get_string(&self, one_one: String, two_two: String) -> String
```

* [ ] Document this behavior and keep it as the default
* [ ]  Add a new macro attribute to allow deserialization from different naming convention 

Suggestion:
The `param_convention` attribute will be applied to each parameter in order to find
the expected name for deserialization. 
Note: This applies only to named parameters.

- `camelCase(one_one) == "oneOne"` - the method expects "oneOne" and "twoTwo" parameters

```rust
#[method(name = "getString", param_convention="camelCase")]
async fn get_string(&self, one_one: String, two_two: String) -> String
```

Feature needed for: https://github.com/paritytech/substrate/pull/12544.

niklasad1:
Close #920

### DIFF
--- a/proc-macros/Cargo.toml
+++ b/proc-macros/Cargo.toml
@@ -17,7 +17,7 @@ proc-macro2 = "1.0"
 quote = "1.0"
 syn = { version = "1.0", default-features = false, features = ["extra-traits", "full", "visit", "parsing"] }
 proc-macro-crate = "1"
-convert_case = "0.6.0"
+heck = "0.4.0"
 
 [dev-dependencies]
 jsonrpsee = { path = "../jsonrpsee", features = ["full"] }
--- a/proc-macros/Cargo.toml
+++ b/proc-macros/Cargo.toml
@@ -17,6 +17,7 @@ proc-macro2 = "1.0"
 quote = "1.0"
 syn = { version = "1.0", default-features = false, features = ["extra-traits", "full", "visit", "parsing"] }
 proc-macro-crate = "1"
+convert_case = "0.6.0"
 
 [dev-dependencies]
 jsonrpsee = { path = "../jsonrpsee", features = ["full"] }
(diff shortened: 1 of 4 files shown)


proc-macros/Cargo.toml
```diff
@@ -17,6 +17,7 @@ proc-macro2 = "1.0"
 quote = "1.0"
 syn = { version = "1.0", default-features = false, features = ["extra-traits", "full", "visit", "parsing"] }
 proc-macro-crate = "1"
+convert_case = "0.6.0"
```
lexnv, line 20: Another alterantive would be [heck](https://github.com/withoutboats/heck) that we use in subxt. It has a bit more traction (git stats), No strong opinions on those, they both look fine.

lexnv (approved):

proc-macros/Cargo.toml
niklasad1: heck seems more widely used let's switch to it.

jsdw:
Code looks good, but I do wonder whether we want this level of "magic" in the macro (I commented on the backing issue)? I personally prefer it being really explicit about which method names you need to pass I guess :)

jsdw (approved):
As I said in the issue, I'm actually ok with this now I have properly understood waht it's doing :)

PR #929, opened 2022-11-04
fix(server): log connections properly

niklasad1:
This PR fixes that the connection count is read before the connection has attempted to converted in a HTTP connection which may cause the log is updated when that occurs.

In addition I refactored the connection task to separate function such that all occurs in the same connection tracing span.

### DIFF
--- a/server/src/future.rs
+++ b/server/src/future.rs
@@ -196,6 +196,7 @@ impl ServerHandle {
 }
 
 /// Limits the number of connections.
+#[derive(Debug)]
 pub(crate) struct ConnectionGuard(Arc<Semaphore>);
 
 impl ConnectionGuard {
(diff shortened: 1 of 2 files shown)


jsdw (approved):

lexnv (approved):
The test seems to fail because of the change in compile UI tests

```
note: required for `<Conf as Config>::Hash` to implement `DeserializeOwned`
```

to 
```
note: required because of the requirements on the impl of `DeserializeOwned` for `<Conf as Config>::Hash`
```

I think we could just restart the test with `TRYBUILD=overwrite`

niklasad1:
@lexnv 

> The test seems to fail because of the change in compile UI tests

That's just because the gitlab CI is using an old rustc version, I updated it to work with the latest rustc version

PR #931, opened 2022-11-08
proc-macros/tests: Update test to latest rustc v1.67

lexnv:
This PR fixes the CI check that fails for the UI testing. The failure is related
to a change in the rustc version that reports a different error than the
expected one.

This is similar to the substrate fix: https://github.com/paritytech/substrate/pull/12642.


jsdw (approved):

niklasad1:
ok, rust 1.67 is nightly right? 

because our CI runs stable and it works on 1.65.

niklasad1 (approved):

PR #835, opened 2022-07-29
[Bug Fix] - Incorrect trace caused by use of `Span::enter` in asynchronous code 

patrickkuo:
jsonrpsee is emitting incorrect trace because of incorrect use of `Span::enter` in async code

### Symtoms:
logs are very spammy as the trace is entering incorrect span 
```
[...]

PR #942, opened 2022-11-18
scripts: Fix `generate_changelog.sh` with empty printf

lexnv:
The error was generated by the following line
```
$        echo "$line" | awk -v var="$pr_md_link" '{NF--; printf "- "; printf; print var}'

- awk: cmd. line:1: (FILENAME=- FNR=1) fatal: printf: no arguments
```

To mitigate this issue, ensure AWK iterates over the correct words
from the commit title, skipping the last one.

For example, in the following commit title, the last word is `(#835)`.
```
[Bug Fix] - Incorrect trace caused by use of `Span::enter` in asynchronous code  (#835)
```
This is skipped from the output because will be replaced by

```
([#835](https://github.com/paritytech/jsonrpsee/pull/835))
```

Signed-off-by: Alexandru Vasile <alexandru.vasile@parity.io>

### DIFF
--- a/scripts/generate_changelog.sh
+++ b/scripts/generate_changelog.sh
@@ -55,8 +55,10 @@ function generate_changelog() {
         pr_link="$REMOTE_LINK$pr_number"
         # Generate the link as markdown.
         pr_md_link=" ([#$pr_number]($pr_link))"
-        # Print the changelog line as `- commit-title pr-link`.
-        echo "$line" | awk -v var="$pr_md_link" '{NF--; printf "- "; printf; print var}'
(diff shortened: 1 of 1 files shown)


jsdw (approved):
I don't know `awk` enough to comment, but if it works it works!

Would you mind doing the same in `subxt` where we use this script too iirc? :)

lexnv:
Yep, for sure, adding the fix to subxt as well. Thanks for taking a look!

niklasad1 (approved):
yeah, I'm no awk hacker myself but if it works it's good enough for me :)

PR #948, opened 2022-11-23
fix(rpc module): make async closures Clone

### LINKED ISSUE #947: rpc module: `register_async_method` only allows `Copy` closures
Im creating a method that will take the values, send it into a channel, and then await its response, by using register_async_method. However, it doesnt let me utilize the mpsc::UnboundedSender because it doesnt implement copy, which is a bound on the register_async_method callback. Is there some other way I can go about sending values down a channel? 

I tried register_method, but i can send that message into a channel just fine. But the problem is i want to await its response from the other end of the channel and i dont want to use a blocking recv method to do so.

niklasad1:
Close #947 

It's currently too restrictive with `Copy` on the closures which forces all types in the closure to be `Copy` so let's relax it to `Clone` such that it's possible to use async primitives inside there.

However, it's currently possible to move these primitives to `RpcModule` to workaround that

### DIFF
--- a/core/src/server/rpc_module.rs
+++ b/core/src/server/rpc_module.rs
@@ -577,13 +577,15 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 	where
 		R: Serialize + Send + Sync + 'static,
 		Fut: Future<Output = Result<R, Error>> + Send,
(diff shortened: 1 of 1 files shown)


lexnv (approved):
LGTM! 👍

Xanewok (approved):

PR #950, opened 2022-11-29
TowerService should be clone-able for handling concurrent request

patrickkuo:
## Summary 
`TowerService` should be cloneable like tower's `ServiceFn`

It is difficult to use the service in async block. e.g. when implementing middleware layer service, `self.inner.call(req)` cannot be called inside the async block, making it impossible to examine the request body before calling the inner service.


niklasad1 (approved):

lexnv (approved):

niklasad1:
Thanks for your PR, ideally we would have a lint against this.... but couldn't find any

PR #962, opened 2023-01-04
server: bounded channels and backpressure

niklasad1:
This PR adds a bounded channel for each connection along with backpressure i.e if there is no spot in the bounded channel buffer the server will wait until there is and then poll the underlying socket. TCP protocol will take of the slowing down the connection if the client can't keep up to speed by decreasing the TCP window size...

###  Notable changes:
- Split SubscriptionSink to `PendingSubscriptionSink` and `SubscriptionSink` again. Mainly because `accept` and `reject` needs to be async methods now because we can't ensure that these messages will go through (when the channel capacity is exceeded) when doing so `non_blocking`
- Move to `tokio::sync::mpsc` because it has this nice API called `reserve` which wait until a spot in the channel is reserved.
- Add tokio wrappers called `send - async`, `try_send - non_blocking`
- Remove `pipe_from_stream`
- Remove `resource limiting` that's somewhat redundant and `tower` supports some simple rate limiting (applies only to the HTTP part)
- Make subscription callback async which in turn is spawned by jsonrpsee under the hood.

### TODO:
- ~~Update examples especially how to deal subscriptions when the buffer is full~~
- ~~Add tests for this PR~~, I give up on this I haven't found a reliable way of testing that. I think we could run some benches to assert that :P

## Exampe subscription API:
```rust
#[tokio::main]
async fn main() -> anyhow::Result<()> {
	tracing_subscriber::FmtSubscriber::builder()
		.with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
		.try_init()
		.expect("setting default subscriber failed");

	let h = run_server().await?;

	h.stopped().await;

	Ok(())
}

async fn run_server() -> anyhow::Result<ServerHandle> {
	const LETTERS: &str = "abcdefghijklmnopqrstuvxyz";
	let server = ServerBuilder::default().set_message_buffer_capacity(10).build("127.0.0.1:9944").await?;
	module
		.register_subscription("sub_one_param", "sub_one_param", "unsub_one_param", |_, pending, _| async move {
			let mut sink = pending.accept().await?;
			let interval = interval(Duration::from_millis(200));
			let stream = IntervalStream::new(interval).map(move |_| "A".repeat(1024 * 1024));

			loop {
				tokio::select! {
					biased;
					_ = sink.closed() => {
						// Client closed connection.
						break;
					},
					maybe_item = stream.next() => {
						let item = match maybe_item {
							Some(item) => item,
							None => break,
						};

						let msg = SubscriptionMessage::from_json(&item)?;
						sink.send_timeout(msg, std::time::Duration::from_millis(100)).await;
					},
				}
			}
			Ok(())
		})
		.unwrap();

	let handle = server.start(module)?;

	Ok(handle)
}
```


niklasad1:
I switched to `tokio::sync::mpsc` because `futures::channel::mpsc` capacity is quite hard to reason about:

> The channel’s capacity is equal to buffer + num-senders

Thus, if one configures the buffer to have a buffer of 10 messages in practice it becomes "the number of pending calls" + 10.

benches/helpers.rs
```diff
@@ -146,9 +146,13 @@ pub async fn ws_server(handle: tokio::runtime::Handle) -> (String, jsonrpsee::se
 	let mut module = gen_rpc_module();
 
 	module
-		.register_subscription(SUB_METHOD_NAME, SUB_METHOD_NAME, UNSUB_METHOD_NAME, |_params, mut sink, _ctx| {
+		.register_subscription(SUB_METHOD_NAME, SUB_METHOD_NAME, UNSUB_METHOD_NAME, |_params, pending, _ctx| {
 			let x = "Hello";
-			tokio::spawn(async move { sink.send(&x) });
+			tokio::spawn(async move {
+				let sink = pending.accept().await.unwrap();
+				let msg = sink.build_message(&x).unwrap();
```
niklasad1, line 153: This is a bit awkward but I think it's easier to reason about why send failed now.

We could keep to old behavior which serializes the message to JSON-RPC notification.

niklasad1: The reason why `send` takes `JSON string` is that if the send fails we return the message and that's a JSON string.

server/src/server.rs
```diff
@@ -587,6 +558,8 @@ pub(crate) struct ServiceData<L: Logger> {
 	pub(crate) enable_http: bool,
 	/// Enable WS.
 	pub(crate) enable_ws: bool,
+	/// Bounded channel capacity.
+	pub(crate) buffer_capacity: u32,
```
lexnv, line 562: Do we need the `buffer_capacity` on this structure for the moment?

niklasad1: Yeah, it's the number of messages that server are allowed to have in memory.

we could rename this if it's unclear however we could remove `max request` and `max response` but kept that for now as the messages could be of arbitrary size.

benches/helpers.rs
jsdw: It does feel to me like this means that the "easy path" allows for a user to forget to use the `build_message` etc and send garbage back.

I feel like the two options might be:

1. If there are different message types we want the user to be able to build, then have `build_message` and `build_error_message` return some opaque type that `sink.send` requires (to enforce that the user builds a valid message), or
2. If it's just the two types, maybe having a `sink.send` and a `sink.send_error` that take a `T: Serialize` and ensure well formed messages.

jsdw: It does feel to me like this means that the "easy path" allows for a user to forget to use the `build_message` etc and send garbage back.

I feel like the two options might be:

1. If there are different message types we want the user to be able to build, then have `build_message` and `build_error_message` return some opaque type that `sink.send`/`sink.try_send` requires (to enforce that the user builds a valid message), or
2. If it's just the two types, maybe having a `sink.send` and a `sink.send_error` that take a `T: Serialize` and ensure well formed messages (though I can see try_send now, so I think option 1 makes more sense offhand).

niklasad1: Yeah, I was thinking about having a wrapper that we call `build_message() -> SubscriptionMessage(String)` which forces everyone to use `build_message` explicitly because access to the underlying String is not possible.

Then errors would return that as well but it's a bit annoying to use anyway. 

However, if we provide `encode_and_send` or something as I added then to user can select what use.

tests/tests/rpc_module.rs
```diff
@@ -340,12 +351,19 @@ async fn subscribing_without_server_bad_params() {
...
 			tokio::spawn(async move {
-				sink.pipe_from_stream(stream).await;
+				let sink = pending.accept().await.unwrap();
+
+				while let Some(item) = stream.next().await {
+					let msg = sink.build_message(&item).unwrap();
+					if sink.send(msg).await.is_ok() {
+						return;
+					}
+				}
```
jsdw, line 366: I guess whether we expose a `pipe_from_stream` or allow people to write this sort of loop manually, the end result is the same; if the sink can't consume items as quickly as the stream produces them, we'll end up with a buildup somewhere. (and if the stream happens to be bounded, then that backpressure from the slow sink will work its way all the way back and give the thing producing the stream inputs a chance to do what it needs to anyway).

All of that said, just a thought is to offer something that can function like:

```rust
// Just pipe the next item into the stream. If the last item is `Some(value)`, 
// we choose to not send it to the stream and just try sending the next (latest) value each time.
sink.pipe_from_stream_aggregating(|last: Option<Val>, next: Val| next, stream).await?;

// Until each item is consumed by the stream, we ignore next items coming in
sink.pipe_from_stream_aggregating(|last: Option<Val>, next: Val| last.unwrap_or(next), stream).await?;

// Run some function which combines the last (if present) and next value and queues that to be sent next.
sink.pipe_from_stream_aggregating(|last: Option<Val>, next: Val| aggregate(last, next), stream).await?;
```

Perhaps this would make it easy for people to use an aggregation approach when the sink is slow to consume things. (and then `pipe_from_stream_all` might exist to just buffer all messages up and allow the stream itself to do somethign with the backpressure if it is able, or else if the stream is an unbounded channel, we'll have a leak (we could document this).

benches/helpers.rs
jsdw: Mmm, how often would a user want to call `build_error_message`? I guess the advantage of keeping building and sending separate is if you want to build differently (ie Ok message or Error message) and then send differently (ie send or try_send). But I did just see those `encode_and_send` fns and they could be an addition to the API for the most common case(s) (ie building an Ok message and `send().await`ing it; I'm not really sure when one would want to use `try_send` offhand; I guess just for epheremal things that they don't care about succeeding or not..)

tests/tests/rpc_module.rs
niklasad1: yupp, sounds like a good idea have this but I think we should kind of agree on the rest of the APIs before adding this back :)

jsdw: Yup that makes sense; just wanted to get it in writing somewhere while I had it in mind :)

core/src/server/rpc_module.rs
```diff
@@ -814,277 +730,216 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
...
+		// Both are cancel-safe thus ok to use select here.
+		tokio::select! {
+			_ = self.inner.closed() => (),
+			_ = self.unsubscribe.unsubscribed() => (),
+		}
 	}
 
-	fn build_message<T: Serialize>(&self, result: &T) -> Result<String, serde_json::Error> {
+	/// ...
+	pub fn build_message<T: Serialize>(&self, result: &T) -> Result<SubscriptionMessage, serde_json::Error> {
```
niklasad1, line 927: move this

core/src/server/rpc_module.rs
```diff
@@ -814,277 +730,216 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
...
 			self.method.into(),
 			SubscriptionPayload { subscription: self.uniq_sub.sub_id.clone(), result },
 		))
+		.map(|json| SubscriptionMessage(json))
 		.map_err(Into::into)
 	}
 
-	fn build_error_message<T: Serialize>(&self, error: &T) -> Result<String, serde_json::Error> {
+	/// ...
+	pub fn build_error_message<T: Serialize>(&self, error: &T) -> Result<SubscriptionMessage, serde_json::Error> {
```
niklasad1, line 937: not public

benches/helpers.rs
niklasad1: I had a look to introduce `SubscriptionMessage::new(T) > Result<SubscriptionMessage(String), Error>` but it was quite messy to support as the the `send API` returns back the message on result 

So, I kept this for now i.e, build_message :(

tests/tests/rpc_module.rs
niklasad1: I'm still not sure whether we should use `send` or `try_send` in pipe_from_stream.

I did it with `try_send` because I don't want to "block" reading the stream,

core/src/lib.rs
```diff
@@ -58,11 +58,19 @@ cfg_client! {
...
-pub use error::Error;
+pub use error::{Error, SubscriptionAcceptRejectError, SubscriptionClosed, SubscriptionEmptyError};
 
 /// JSON-RPC result.
 pub type RpcResult<T> = std::result::Result<T, Error>;
 
+/// The return type of the subscription's method for the rpc server implementation.
+///
+/// **Note**: The error does not contain any data and is discarded on drop.
+pub type SubscriptionResult = Result<(), SubscriptionEmptyError>;
```
niklasad1, line 69: moved these to `core`, I want to keep initial idea to keep JSON-RPC spec related types in the types crate.

core/src/server/helpers.rs
```diff
@@ -107,109 +107,81 @@ impl MethodSink {
...
+		Ok(())
 	}
 
-	/// Get the maximum number of permitted subscriptions.
-	pub const fn max_response_size(&self) -> u32 {
-		self.max_response_size
+	/// Async send which will wait until there is space in channel buffer or that the subscription is disconnected.
+	pub async fn send(&self, msg: String) -> Result<(), DisconnectError> {
+		tx_log_from_str(&msg, self.max_log_length);
+		self.tx.send(msg.clone()).await?;
```
niklasad1, line 137:
```suggestion
		self.tx.send(msg).await?;
```

proc-macros/tests/ui/correct/rpc_deny_missing_docs.rs
```diff
@@ -13,7 +13,7 @@ pub trait ApiWithDocumentation {
 
 	/// Subscription docs.
 	#[subscription(name = "sub", unsubscribe = "unsub", item = String)]
-	fn sub(&self);
+	async fn sub(&self) -> jsonrpsee::core::SubscriptionResult;
```
niklasad1, line 16: I'm not sure whether we should force folks to return `SubscriptionResult` in the proc macro API.

The generated client trait returns another type anyway.

server/src/server.rs
```diff
@@ -199,6 +191,8 @@ struct Settings {
 	enable_http: bool,
 	/// Enable WS.
 	enable_ws: bool,
+	/// Number of messages that server is allowed `buffer` until backpressure kicks in.
```
niklasad1, line 194:
```suggestion
	/// Number of messages that server is allowed to `buffer` until backpressure kicks in.
```

server/src/transport/ws.rs
```diff
@@ -308,6 +260,20 @@ pub(crate) async fn background_task<L: Logger>(
...
+		let sink_permit_fut = sink.reserve();
+
+		tokio::pin!(sink_permit_fut);
+
+		// Wait until there is a slot in the bounded channel which means that
+		// the underlying TCP socket won't be read.
+		//
+		// This will force the client to read socket on the other side
+		// otherwise the socket will not be read again.
+		let sink_permit = match method_executors.select_with(Monitored::new(sink_permit_fut, &stop_handle)).await {
```
niklasad1, line 272: we need to use `select_with` here otherwise the pending futures won't be polled

niklasad1: such that if the server have a few pending RPC calls and the "bounded channel buffer is full" then `async fn MethodSink::reserve` won't return until there is a slot in the buffer but the underlying futures/rpc calls must still be polled.

core/src/server/helpers.rs
```diff
@@ -84,7 +84,7 @@ impl<'a> io::Write for &'a mut BoundedWriter {
 #[derive(Clone, Debug)]
 pub struct MethodSink {
 	/// Channel sender.
-	tx: mpsc::UnboundedSender<String>,
+	pub tx: mpsc::Sender<String>,
```
niklasad1, line 87:
```suggestion
	tx: mpsc::Sender<String>,
```

core/src/server/rpc_module.rs
```diff
@@ -416,36 +374,36 @@ impl Methods {
 
 	/// Execute a callback.
 	async fn inner_call(&self, req: Request<'_>) -> RawRpcResponse {
-		let (tx_sink, mut rx_sink) = mpsc::unbounded();
+		let (tx_sink, mut rx_sink) = mpsc::channel(u32::MAX as usize / 2);
```
jsdw, line 377: This is essentially an unbounded channel that's handed to the subscription handler; won't this defeat the purpose of bounded channels or am I missing something?

core/src/server/rpc_module.rs
```diff
@@ -762,34 +702,42 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
...
-					}
+					// The subscription callback is a future from the subscription
+					// definition and not the as same when the subscription call has been completed.
+					//
+					// This runs until the subscription callback has completed.
+					let sub_fut = callback(params.into_owned(), sink, ctx.clone());
+
+					tokio::spawn(async move {
+						if sub_fut.await.is_err() {
+							tracing::warn!("Subscribe call `{subscribe_method_name}` closed");
```
jsdw, line 728: Is it worth logging the error that caused it to close here?

core/src/server/rpc_module.rs
niklasad1: this is just used for tests :)

core/src/server/rpc_module.rs
niklasad1: we could but currently the error cause is erased by the `SubscriptionEmptyError`.

so there is no point at the moment

core/src/server/rpc_module.rs
niklasad1: but perhaps we could use this for testing some backpressure stuff without a WebSocket server.

core/src/server/rpc_module.rs
```diff
@@ -682,34 +627,30 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
...
-	///             sink.reject(err);
-	///             return Ok(());
-	///         }
-	///     };
-	///     // Sink is accepted on the first `send` call.
-	///     std::thread::spawn(move || {
-	///         let sum = x + (*ctx);
-	///         let _ = sink.send(&sum);
-	///     });
+	/// ctx.register_subscription("sub", "notif_name", "unsub", |params, pending, ctx| async move {
```
jsdw, line 630: The above comment may need updating; `callback` to mention `PendingSubscriptionSink` again :)

koute:
I was looking into perhaps working on this (in the context of our problems on RPC nodes), but seeing as you're already working on something similar I want to help out. (:

Hmmmm... this is a step in the right direction, although I feel like bounding the number of elements might not be enough. (: And we could also use this for *non* subscription endpoints (e.g. the metadata endpoint in substrate produces a lot of data and is not a subscription)

Could we instead bound the total *size* of the messages that are kept in memory, instead of their number? Since we care about the resource usage here -- or in other words, we want to bound how much memory we use -- just keeping track of the number of pending messages can be a poor proxy of that. One subscription could generate 1KB of data per message, and another could generate 10MB of data per message, so unless this is uniform across all of the RPC endpoints (which is unlikely) you can't really reliably bound the memory usage just by limiting the number of in-flight messages.

Very roughly maybe something like this could be better? (this proposal might still have some holes and need some tweaks/modifications, as I'm coming up with this on the spot)

1) After serializing the reply into JSON before pushing it into `MethodSink::tx` calculate its size, and use that as the cost of the message.
2) Have two per-connection limits - a soft memory usage limit, and a hard memory usage limit. If `pending_messages_size + new_message_size > hard_limit` then just force disconnect the client. If `pending_messages_size + new_message_size > soft_limit` then apply backpressure.
3) Have a global memory usage limit across *all* connections. If that limit is breached find the biggest memory hog that is over the `soft_limit` and forcibly disconnect it. Repeat if we're still over the limit.
4) If there are no more clients over the soft limit and we're still over the global memory usage limit then unconditionally apply backpressure, and reject any new subscriptions and requests with a "resources exceeded" or something like that (and on our load balancer we could get it to try load balance to another RPC server then or something like that)
5) Also track the in-flight memory usage for non-subscription requests. (e.g. needed for the metadata endpoint)
6) Also have a per-connection timeout that would kick in when a connection spends too much time over the soft limit and forcibly disconnect it (needed for our blocking handlers, some of which read from channels themselves; otherwise we'll just move the memory leak from jsonrpsee into those handler-side channels, and yes, we could "fix" those, but even so I'd rather have a plan B)
7) (optional) Split big JSON replies into multiple chunks and deallocate them as they're sent so that we can gradually give back memory as the data is being sent.

So this would achieve:

- The total memory usage of the RPC server would be roughly bounded.
- The most abusive/slowest clients would get punished first.
- Least abusive/fast clients under the limit would work as normal.
- The punishments are gradual based on memory usage - first throttling, then disconnection.
- If there's no one to punish/kick the server would gracefully decline new requests.
- Works both for subscriptions and non-subscriptions.

Thoughts?

jsdw:
Ooh thank you for the input!

I guess my ideal place to be would be that `jsonrpsee` provides all of the fundamentals needed to send messages and work with backpressure such that more complex schemes could be built on top of this.

We have spoken a bit about having some tower middleware for websocekt connections, and I wonder whether such a piece of middleware would be an ideal place for this sort of logic, since it would have access to all messages from/to some connection. So subscriptions and method calls could work much as they do now (and in this PR they will already be able to use backpressure to adapt to slow connections by aggregating data or whatever they choose), and then we have some middleware which tracks the cost of messages in the sort of way you describe, and possibly carries some global state to have logic around killing the worst connection and tracking global limits?

For me, this would be the best place to be; keeping `jsonrpsee` fairly simple at its core but exposing all of the tools needed to allow this sort of thing that you suggest to be built on top.

koute:
> I guess my ideal place to be would be that `jsonrpsee` provides all of the fundamentals needed to send messages and work with backpressure such that more complex schemes could be built on top of this.

For me it could work either way. I think it makes sense to provide *some* sort of a way to reliably prevent resource exhaustion (and I'd argue that it should be an essential feature), but it doesn't necessarily mean that this couldn't be e.g. kept in another crate and plugged into `jsonrpsee` at will by the user.

So all in all, I'm for it, but the question is whether it's actually possibly to implement something like that (e.g. the scheme I've described) using a tower middleware, and how much extra pieces we need to plug into `jsonrpsee` to make it work?

niklasad1:
> Could we instead bound the total size of the messages that are kept in memory, instead of their number? Since we care about the resource usage here -- or in other words, we want to bound how much memory we use -- just keeping track of the number of pending messages can be a poor proxy of that. One subscription could generate 1KB of data per message, and another could generate 10MB of data per message, so unless this is uniform across all of the RPC endpoints (which is unlikely) you can't really reliably bound the memory usage just by limiting the number of in-flight messages.

I looked into that initially but the bounded channels APIs doesn't expose anything about it so I was lazy but if I think we can write a wrapper around that track the memory usage that would be better.

> So all in all, I'm for it, but the question is whether it's actually possibly to implement something like that (e.g. the scheme I've described) using a tower middleware, and how much extra pieces we need to plug into jsonrpsee to make it work?

The tower stuff would only apply to HTTP stuff i.e, HTTP calls and HTTP upgrade request so once `WS upgrade` is completed then we have no use from tower. Either keep the resource limiting or provide some custom middleware is the way forward I think

koute:
> The tower stuff would only apply to HTTP stuff i.e, HTTP calls and HTTP upgrade request so once the `WS upgrade` is completed then we have no use from tower. Either keep the resource limiting or provide some custom middleware is the way forward I think

Hmm... good point. So if we want to make this pluggable (and keep the complexity out of core jsonrpsee crates) I think we could just make a `trait` will all appropriate customization points and then just have e.g. `.set_resource_limiter(Box::new(limiter))` on the server builder just like you have the `set_backpressure_buffer_capacity` now.

core/src/server/helpers.rs
```diff
@@ -107,109 +107,81 @@ impl MethodSink {
...
+	/// it stays closed forever and all future calls to closed will return immediately.
+	pub async fn closed(&self) {
+		self.tx.closed().await
 	}
 
-	/// Helper for sending the general purpose `Error` as a JSON-RPC errors to the client.
-	pub fn send_call_error(&self, id: Id, err: Error) -> bool {
-		self.send_error(id, err.into())
+	/// Get the max response size.
+	pub const fn max_response_size(&self) -> u32 {
```
lexnv, line 120: dq: Would it be useful for users to have a `max_log_length`?

core/src/server/helpers.rs
```diff
@@ -107,109 +107,81 @@ impl MethodSink {
...
-			.map(|p| SubscriptionPermit { _permit: p, resource: self.resource.clone() })
+	/// Helper for sending the general purpose `Error` as a JSON-RPC errors to the client.
+	pub fn send_call_error(self, id: Id, err: Error) {
+		self.send_error(id, err.into())
 	}
 
-	/// Get the maximum number of permitted subscriptions.
-	pub const fn max(&self) -> u32 {
-		self.max
+	/// Send a raw JSON-RPC message to the client, `MethodSink` does not check verify the validity
```
lexnv, line 171:
```suggestion
	/// Send a raw JSON-RPC message to the client, `MethodSink` does not check the validity
```

core/src/server/rpc_module.rs
```diff
@@ -83,16 +75,38 @@ pub type MaxResponseSize = usize;
 /// Raw response from an RPC
 /// A 3-tuple containing:
 ///   - Call result as a `String`,
-///   - a [`mpsc::UnboundedReceiver<String>`] to receive future subscription results
-///   - a [`crate::server::helpers::SubscriptionPermit`] to allow subscribers to notify their [`SubscriptionSink`] when they disconnect.
-pub type RawRpcResponse = (MethodResponse, mpsc::UnboundedReceiver<String>, SubscriptionPermit);
+///   - a [`mpsc::Receiver<String>`] to receive future subscription results
```
lexnv, line 78: nit: also a method sink?

core/src/server/rpc_module.rs
```diff
@@ -391,95 +353,107 @@ impl Methods {
...
 	///     let (sub_resp, _sub_id) = sub.next::<String>().await.unwrap().unwrap();
 	///     assert_eq!(&sub_resp, "one answer");
 	/// }
 	/// ```
-	pub async fn subscribe(&self, sub_method: &str, params: impl ToRpcParams) -> Result<Subscription, Error> {
+	pub async fn subscribe_unbounded(&self, sub_method: &str, params: impl ToRpcParams) -> Result<Subscription, Error> {
+		self.subscribe_bounded(sub_method, params, u32::MAX as usize).await
+	}
+
+	/// Similar to [`Methods::subscribe_unbounded`] but it's a using bounded channel.
```
lexnv, line 443:
```suggestion
	/// Similar to [`Methods::subscribe_unbounded`] but it's using bounded channel.
```

core/src/server/rpc_module.rs
```diff
@@ -816,98 +778,196 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
...
+		.map(SubscriptionMessage)
+		.map_err(Into::into)
 	}
 
 	/// Reads data from the `stream` and sends back data on the subscription
 	/// when items gets produced by the stream.
+	///
+	/// The underlying sink is a bounded channel which may not have room for new items
+	/// if the client can't keep up with all the messages. Thus, you need to provide a closure
+	/// with a policy what to then the underlying channel has not space to "buffer" more messages.
```
lexnv, line 969:
```suggestion
	/// with a policy what to do when the underlying channel has not space to "buffer" more messages.
```

examples/examples/proc_macro.rs
```diff
@@ -26,15 +26,22 @@
...
 use jsonrpsee::ws_client::WsClientBuilder;
 
 type ExampleHash = [u8; 32];
 type ExampleStorageKey = Vec<u8>;
 
+#[rpc(client, server)]
+pub trait DupOverride {
+	#[subscription(name = "subscribeOne" => "override", item = u8)]
+	async fn one(&self) -> jsonrpsee::core::SubscriptionResult;
+	/*#[subscription(name = "subscribeTwo" => "override", item = u8)]
```
lexnv, line 41: nit: Should we remove these?

proc-macros/src/rpc_macro.rs
```diff
@@ -295,15 +295,29 @@ impl RpcDescription {
...
-						));
-					}
-
-					if method.sig.asyncness.is_some() {
-						return Err(syn::Error::new_spanned(method, "Subscription methods must not be `async`"));
+					match method.sig.output.clone() {
+						syn::ReturnType::Type(_, ty) => {
+							if let syn::Type::Path(syn::TypePath { path, .. }) = *ty {
+								if let Some(ident) = path.get_ident() {
+									if ident != "SubscriptionResult" && ident != "Result" {
```
lexnv, line 302: nit: Could we have an `std::result::Result` here? Should we check just the last segment of the path instead of the whole ident?

examples/examples/proc_macro.rs
niklasad1: my bad :)

proc-macros/src/rpc_macro.rs
niklasad1: I think I'm favor of reverting the subscriptions to return result, it's not worth this change  to break everything.

the SubscriptionResult is different for the client and server anyway.

core/src/server/helpers.rs
niklasad1: The `MethodSink` is not really exposed to users only the `SubscriptionSink`.

Thus, I guess not.

niklasad1:
> Also track the in-flight memory usage for non-subscription requests. (e.g. needed for the metadata endpoint)

Both "subscriptions" and "plan calls" is using the same underlying channel, thus if we have "limit" of the max number of bytes that channel is allowed to "keep" we can simply just deny those calls similar to what we have right now with `max response limit` (JSON-RPC error is returned). Or am I misunderstanding something?


I think a reasonable plan is as follows:

1. merge this PR
2. add global limit in number of bytes that the bounded channel is allowed to use
3. add websocket middleware to add custom logic when to how to handle "slow clients".

jsdw:
> 

I'm up for that; this PR feels like a great step forward so I def think we should aim to merge and build on this :)

core/src/server/rpc_module.rs
```diff
@@ -238,6 +256,37 @@ impl Debug for MethodKind {
...
+	Pending,
+	/// Resolves once the future is completed.
+	Future(BoxFuture<'a, Result<(), DisconnectError>>),
+}
+
+impl<'a> std::fmt::Debug for PendingOrFuture<'a> {
+	fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+		let msg = match self {
+			Self::Pending => "pending",
+			Self::Future(_) => "future",
```
jsdw, line 271: Very much a nit: I'd prolly stick close to the default Debug and spit out "Pending" and "Future(<future>)"

core/src/server/rpc_module.rs
```diff
@@ -816,98 +846,196 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
...
+		tokio::select! {
+			_ = self.inner.closed() => (),
+			_ = self.unsubscribe.unsubscribed() => (),
+		}
+	}
+
+	/// Build message that will be serialized as JSON-RPC notification.
+	///
+	/// You need to call this method prior to send out a subscription notification.
+	pub fn build_message<T: Serialize>(&self, result: &T) -> Result<SubscriptionMessage, serde_json::Error> {
```
jsdw, line 1023: Did you decide not to have a separate `Message::build` type thing in the end?

core/src/server/rpc_module.rs
niklasad1: the default doesn't work for `BoxFuture` stuff, that's the reason why :(

koute:
> Both "subscriptions" and "plan calls" is using the same underlying channel, thus if we have "limit" of the max number of bytes that channel is allowed to "keep" we can simply just deny those calls similar to what we have right now with `max response limit` (JSON-RPC error is returned). Or am I misunderstanding something?

Yes. It's just that this does require tracking the sizes instead of the number of messages (because a normal RPC call will just generate one message).

> I think a reasonable plan is as follows:
> 
>    1. merge this PR
> 
>    2. add global limit to "max number of bytes" that the bounded channel is allowed to "buffer/keep in memory"
> 
>    3. add websocket middleware to add custom logic when to how to handle "slow clients".

Sounds good to me. Do you want to keep the message-count based backpressure support though? If we're going to be tracking the total size of buffered messages then also tracking their number should be redundant I think. (And even if we want to also limit the number of messages we could "emulate" that through the size-based mechanism: just calculate how much memory a single empty message takes and add that to the total buffered size.)

core/src/server/rpc_module.rs
```diff
@@ -816,98 +846,196 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
...
 
-impl SubscriptionSink {
-	/// Reject the subscription call from [`ErrorObject`].
-	pub fn reject(&mut self, err: impl Into<ErrorObjectOwned>) -> Result<(), SubscriptionAcceptRejectError> {
-		let (id, subscribe_call) = self.id.take().ok_or(SubscriptionAcceptRejectError::AlreadyCalled)?;
+impl PendingSubscriptionSink {
+	/// Reject the subscription call with the error from [`ErrorObject`].
+	pub async fn reject(self, err: impl Into<ErrorObjectOwned>) -> Result<(), SubscriptionAcceptRejectError> {
+		let err = MethodResponse::error(self.id, err.into());
+		let permit = self.inner.reserve().await.map_err(|_| SubscriptionAcceptRejectError::RemotePeerAborted)?;
```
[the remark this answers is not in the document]
niklasad1: initially, I had an idea to limit to sends APIs to `MethodSinkPermit` but didn't work because subscriptions may send more than one message.

Thus, you are right those "reserve" are needless.

core/src/server/rpc_module.rs
```diff
@@ -923,91 +1051,86 @@ impl SubscriptionSink {
...
 	///         }
-	///     });
+	///     }
 	///     Ok(())
 	/// });
 	/// ```
-	pub async fn pipe_from_try_stream<S, T, E>(&mut self, mut stream: S) -> SubscriptionClosed
+	pub async fn pipe_from_try_stream<'a, S, T, F, E>(&'a mut self, f: F, stream: S) -> SubscriptionClosed
 	where
+		F: Fn(Option<PendingOrFuture<'a>>, PendingOrFuture<'a>) -> PendingOrFuture<'a>,
```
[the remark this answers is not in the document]
jsdw: I also guess that this method doesn't need any "special" access to Self; it just uses the public interface to send messages. So I suppose it owuld be a separate thing in some utility module or whatever if we wanted and we don't need to complicate the "core" stuff by having it here if that makes life any easier for us!

It's also the only place `SubscriptionClosed` is returned I think? And maybe it would be nice just to pull that and this out into a separate place :)

core/src/server/rpc_module.rs
jsdw: Did we used to do that because we didn't return an `async` thing and the fn would return probably before the subscription has ended?

Does it make any sense now to allow the subscription to return a normal error object or whatever, and to send that back to the client before closing the connection if possible?

I wonder whether doing this would keep subscriptions easy to understand; if they return an error, the error will be sent to the user on close. If they return `Ok(())`, no explicit error will be sent. Perhaps we don't need an explicit `sink.close(msg)` type thing in that case; it's closed when the async block returns?

Just mulling over ideas here since right now this Ok/Err thing is sortof just ignored.

proc-macros/src/lib.rs
```diff
@@ -99,8 +99,8 @@ pub(crate) mod visitor;
 ///     async fn async_method(&self, param_a: u8, param_b: String) -> u16;
 ///     fn sync_method(&self) -> String;
 ///
-///     // Note that `subscription_sink` and `SubscriptionResult` were added automatically.
-///     fn sub(&self, subscription_sink: SubscriptionResult) -> SubscriptionResult;
+///     // Note that `pending subscription sink` was added automatically.
```
jsdw, line 102: are there meant to be underscores instead of spaces in `pending subscription sink`?

core/src/server/rpc_module.rs
jsdw: Oh yeah, I just meant the actual strings being returned from the custom impl here could look like what would normally be generated :)

core/src/server/rpc_module.rs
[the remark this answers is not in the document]
niklasad1: I would be happy to remove this API completely but `util` is probably a good compromise.

This API is still super hard to understand and use, especially with the `SubscriptionClosed` result whether to manually send a customized close message or not.

core/src/server/rpc_module.rs
niklasad1: ah gotcha

jsdw:
> Yes. It's just that this does require tracking the sizes instead of the number of messages (because a normal RPC call will just generate one message).

I wonder whether this is something I'd want us to address in a separate PR just because it might represent a bit of work to get right (and this PR by now is a good step in the right direction even if it doesn't allow for bounding things based on size at the mo).

What I mean is that we already have bounded channels that we can make use of for backpressure per message, but to apply it per-size I guess we'd probably need to write some custom channel impl that tracks message sizes as they enter/leave the channel, and returns `Poll::Pending` while over some size limit rather than some number-of-messages limit. I think this probably isn't too hard to do but would def need some care to get right and do efficiently.

I'm also wondering about this whole global limit and killing "memory hog" subscriptions a bit. It sounds quite tricky to get right, and something in general that I hope we can indeed have in some middleware and not in jsonrpsee core if it's the path we want to take.

Personally, I'd hope that once we have any form of backpressure, things will naturally balance themselves out quite nicely:
- If we have many/slow subscriptions, they will start feeling the backpressure. Subscriptions may choose to start aggregating data while they can't send messages, or maybe they can't do this and they will just close themselves.
- If we have few/fast subscriptions, they will be able to handle much higher volume and the server will be happy enough with this.

Bounding the total memory allowed is the next step on (and I guess becomes important if the server is capable of handling many more subscriptions than the amount of memory it has available to it is capable of supporting). Do we need both a soft and hard limit per connection though? If we set just one limit (the max bytes that can exist in the channel), then subscriptions will hit backpressure when they face this, and as above, some can be smart and aggregate stuff and some will just have to close. As subscriptions close, the connection falls below the limit again and we're all golden. If it's hit the limit, it will happily remain there forever if the only subscriptions in use are happy to aggregate data endlessly (but I think we can have snesible timeouts for these sorts of subscriptions because it probably doesn't make sense for this to happen endlessly).

I guess I think that we can probably satisfy these bounds wihtout any sort of global schedular that has to manage connections, but with a simple limit on the bytes allowed in the channel for each connection and the application of backpressure, which would be much simpler I think to implement/support. Am I missing something though maybe?

core/src/server/rpc_module.rs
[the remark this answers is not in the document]
niklasad1: > The function would return msg; something to try sending on the stream. This might be built be combining the two arguments together (aggregating messages if the first hasn't managed to send yet) or it might just always ignore the first message or whatever.

> Now that these PendingOrFuture args are here, there's no way to aggregate messages; only pick between ignoring one or the other

I'm not sure we should really support aggregating messages, if the subscription is defined with a `Item = usize` when aggregating would imply that subscription notification would like like:

```
# "not aggregated"
{"jsonrpc":"2.0","method":"my_sub","params":{"subscription":1882468350819392,"result":1}}
# aggregated as `Vec`, this will break the client code (potential footgun)
{"jsonrpc":"2.0","method":"my_sub","params":{"subscription":1882468350819392,"result":[1, 2]}}
# aggregated as sum
{"jsonrpc":"2.0","method":"my_sub","params":{"subscription":1882468350819392,"result":3}}
```

We could actually require` |Option<T>, T| -> T` to circumvent that

However, in order to support aggregating we are forced to use `try_send` and perhaps use `the reserve API` if there is any unsent message waiting for a slot. Instead of having a timeout to re-transmit periodically.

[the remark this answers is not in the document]
jsdw: > We could actually require |Option<T>, T| -> T to circumvent that
> 
> However, in order to support aggregating we are forced to use try_send and perhaps use the reserve API if there is any unsent message waiting for a slot. Instead of having a timeout to re-transmit periodically.

Yeah, `|Option<T>, T| -> T` is what I had in mind. And perhaps there aren't any real use cases for aggregating indeed! I'd be in favoru of just removing this API completely because it's obviously quite complicated, and just adding back what we might find useful as utils!

jsdw:
> In general though, I do agree with you that keeping this complexity out of `jsonrpsee` core is a good idea. Ideally `jsonrpsee` would only do the bare minimum (track memory, implement backpressure, allow you to loop through who's connected and how much memory they're using, trigger a disconnection, and reject new clients) and we'd have a trait with the appropriate hooks which could decide by itself whether to e.g. have a soft and hard limit or only a single limit, whether to limit the memory globally, whether to globally kick memory hogs or not, etc. I think that should be relatively doable? And then if you want you can just leave the whole shebang of implementing the logic of resource limiting/kicking/etc. to me. (:

What specifically do you think you'd need in terms of functionality to achieve what you want then? I think something like the following:
- Ability to view list of connections and current pending bytes for each one
- Ability to disconnect one of those connections
- Some hard limit for total bytes (globally or per connection?)

And when would the function that does some action fire, ideally? I guess whenever some message pushes total pending bytes above the hard limit?

Perhaps let's open a new issue perhaps for this idea and there we can flesh out some of these details and work out how we might go about adding what we need to jsonrpsee :)

koute:
> What specifically do you think you'd need in terms of functionality to achieve what you want then?

Hm.... give me some time; I'll try to whip up an API + a trait definition with the hooks that I think would work, and we can discuss it.

> Let's open a new issue for this idea and there we can flesh out some of these details and work out how we might go about adding what we need to jsonrpsee :)

Yep. Lets move this conversation into a separate issue.

core/src/server/rpc_module.rs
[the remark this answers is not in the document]
jsdw: Well, you could keep the original value around until it has sent I suppose? So you serialize, try to send, but keep the non-serialized version in case you need to provide it to the callback :)

server/src/tests/ws.rs
```diff
@@ -673,46 +671,36 @@ async fn ws_server_backpressure_works() {
...
+				let mut msg = n.clone();
+
+				loop {
+					tokio::select! {
+						biased;
+						_ = sink.closed() => {
+							// User closed connection.
+							break;
+						},
+						res = sink.send_timeout(msg.clone(), std::time::Duration::from_millis(100)) => {
```
jsdw, line 687: I'm just wondering; do we need the `send_timeout` stuff rather than using a combinator like https://docs.rs/tokio/latest/tokio/time/fn.timeout.html?

server/src/tests/ws.rs
```diff
@@ -673,46 +671,36 @@ async fn ws_server_backpressure_works() {
...
-								buf.push_back(bp.clone());
-							}
-							Ok(()) => {
-								sent += 1;
-								buf.push_back(n.clone());
-							}
-						}
+				let mut msg = n.clone();
+
+				loop {
```
jsdw, line 680: I like this change; using `send()` rather than `try_send()` feels like it would be a better true test of backpressure to me :)

core/src/server/rpc_module.rs
niklasad1: It was just super noisy to implement, lemme give it one more try.

examples/Cargo.toml
```diff
@@ -19,3 +19,4 @@ tower-http = { version = "0.3.4", features = ["full"] }
 tower = { version = "0.4.13", features = ["full"] }
 hyper = "0.14.20"
 console-subscriber = "0.1.8"
+bounded-vec-deque = "0.1.1"
```
niklasad1, line 22:
```suggestion
```

server/src/server.rs
```diff
@@ -453,6 +421,29 @@ impl<B, L> Builder<B, L> {
 		self
 	}
 
+	/// The server enforces backpressure which means that
+	/// `n` messages can be buffered and if the client
+	/// can't keep with the server.
```
niklasad1, line 426:
```suggestion
	/// can't keep with up the server.
```

server/src/server.rs
```diff
@@ -453,6 +421,29 @@ impl<B, L> Builder<B, L> {
...
+	///
+	/// This `capacity` is applied per connection and
+	/// applies globally on the connection which implies
+	/// all JSON-RPC messages.
+	///
+	/// For example if a subscription produces plenty of new items
+	/// and the client can't keep up then no new messages are handled.
+	///
+	/// If this limit is exceeded the server will "back-off"
+	/// and only accept ones the client reads pending messages.
```
niklasad1, line 436:
```suggestion
	/// and only accept new messages once the client reads pending messages.
```

core/src/server/rpc_module.rs
```diff
@@ -379,107 +411,128 @@ impl Methods {
...
+	///         format!(r#"{{"jsonrpc":"2.0","method":"hi","params":{{"subscription":{},"result":"one answer"}}}}"#, sub_resp.result),
+	///         sub_notif
 	///     );
 	/// }
 	/// ```
-	pub async fn raw_json_request(
-		&self,
-		request: &str,
-	) -> Result<(MethodResponse, mpsc::UnboundedReceiver<String>), Error> {
+	pub async fn raw_json_request(&self, request: &str, tx: mpsc::Sender<String>) -> Result<(), Error> {
```
niklasad1, line 453: changed this API, have a look

core/src/server/rpc_module.rs
```diff
@@ -671,45 +709,41 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 	/// * `unsubscription_method` - name of the method to call to terminate a subscription
 	/// * `callback` - A callback to invoke on each subscription; it takes three parameters:
 	///     - [`Params`]: JSON-RPC parameters in the subscription call.
-	///     - [`SubscriptionSink`]: A sink to send messages to the subscriber.
+	///     - [`PendingSubscriptionSink`]: A pending subscription waiting to accepted.
```
niklasad1, line 712:
```suggestion
	///     - [`PendingSubscriptionSink`]: A pending subscription waiting to be accepted, in order to send out messages on the subscription
```

core/src/server/rpc_module.rs
niklasad1: > Did we used to do that because we didn't return an async thing and the fn would return probably before the subscription has ended?

Returning an error is a hack for ergonomics i.e, to bail easily from the user defined, it's not really used except logging it here. We could remove it...

> Does it make any sense now to allow the subscription to return a normal error object or whatever, and to send that back to the client before closing the connection if possible?

Not really, because we have this `accept/reject` API which handles the JSON-RPC response to the call.
If in an error happens after that it's simply wrong to send out it as an JSON-RPC response. That's why we have a oneshot channel which indicates whether the "subscription method call" was answered or not. If it's not answered then a default JSON-RPC error response is sent out.

examples/examples/ws_pubsub_broadcast.rs
```diff
@@ -97,10 +94,48 @@ async fn run_server() -> anyhow::Result<SocketAddr> {
...
+		match future::select(closed, stream.next()).await {
+			// subscription closed.
+			Either::Left((_, _)) => break,
+
+			// received new item from the stream.
+			Either::Right((Some(Ok(item)), c)) => {
+				let notif = SubscriptionMessage::from_json(&item)?;
+
+				// NOTE: this will block until there a spot in the queue
+				// and you might want to something smarter if it's
```
niklasad1, line 116:
```suggestion
				// and you might want to do something smarter if it's
```

examples/examples/ws_pubsub_broadcast.rs
```diff
@@ -97,10 +94,48 @@ async fn run_server() -> anyhow::Result<SocketAddr> {
...
+				// critical that newer must be sent when they are produced.
```
niklasad1, line 117:
```suggestion
				// critical that "the most recent items" must be sent when they are produced.
```

core/src/server/rpc_module.rs
```diff
@@ -80,19 +71,45 @@ pub type ConnectionId = usize;
...
+pub enum TrySendError {
+	/// The channel is closed.
+	Closed(SubscriptionMessage),
+	/// The channel is full.
+	Full(SubscriptionMessage),
+}
+
+#[derive(Debug, Clone)]
+/// Represents whether a subscription was answered or not.
+pub enum SubscriptionAnswered {
```
niklasad1, line 85: I think we don't need `MethodResponse` for the subscriptions but I kept it for now to all calls to return MethodResponse.

core/src/server/rpc_module.rs
```diff
@@ -106,13 +123,103 @@ pub enum InnerSubscriptionResult {
...
 
-type Subscribers = Arc<Mutex<FxHashMap<SubscriptionKey, (MethodSink, watch::Sender<()>)>>>;
+type Subscribers = Arc<Mutex<FxHashMap<SubscriptionKey, (MethodSink, mpsc::Receiver<()>)>>>;
+
+/// This represent a response to a RPC call
+/// and `Subscribe` calls are handled differently
+/// because we want to prevent subscriptions to start
+/// before the actual subscription call has been answered.
+#[derive(Debug, Clone)]
+pub enum CallResponse {
```
niklasad1, line 163: weird name...

examples/examples/ws_pubsub_with_params.rs
```diff
@@ -63,38 +65,37 @@ async fn main() -> anyhow::Result<()> {
 
 async fn run_server() -> anyhow::Result<SocketAddr> {
 	const LETTERS: &str = "abcdefghijklmnopqrstuvxyz";
-	let server = ServerBuilder::default().build("127.0.0.1:0").await?;
+	let server = ServerBuilder::default().set_backpressure_buffer_capacity(10).build("127.0.0.1:9944").await?;
```
niklasad1, line 68:
```suggestion
	let server = ServerBuilder::default().set_backpressure_buffer_capacity(10).build("127.0.0.1:0").await?;
```

core/src/server/rpc_module.rs
niklasad1: nvm, it's required by the Logger so no way around it right now.

core/src/server/rpc_module.rs
jsdw: I do prefer this to the old API; no harder to use and gives more control to the caller over the boundedness :)

core/src/server/rpc_module.rs
```diff
@@ -762,34 +834,42 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
...
+					// The subscription callback is a future from the subscription
+					// definition and not the as same when the subscription call has been completed.
+					//
+					// This runs until the subscription callback has completed.
+					let sub_fut = callback(params.into_owned(), sink, ctx.clone());
+
+					tokio::spawn(async move {
+						if let Err(SubscriptionCallbackError::Some(msg)) = sub_fut.await {
+							tracing::warn!("Subscribe call `{subscribe_method_name}` failed: {msg}");
+						}
```
jsdw, line 861: I guess this makes sense to me; I kept mulling over what should happen if a user returns `Err(e)` in a subscription thing now, and the options are to either return it to the user as a subscription error, or not.

The nice thing about not returning the error to the user is that if you want to return some error to the user you have to do it super explicitly, otherwise no error is returned. (The downside is that if they return an error from the callback, they might expect it to be sent to the user but it won't be?).

On the other hand, if you _did_ send any error to the user here, you could remove APIs like `PendingSubscriptionSink::reject()` and `SubscriptionSink::close(error)`, and instead, returning from the subscription async callback with `Ok` would mean it closed successfully and no need to send an error, and returning with `Err(e)` means it closes and to send an error back in whatever form makes sense.

Just thinking out loud though; I guess I don't like that the error is silently disappeared from the client perspective at present, but I do like that the current approach is very explicit about things, so I'm happy either way and this doesn't block merging for me!

core/src/server/rpc_module.rs
```diff
@@ -1110,29 +1111,29 @@ impl SubscriptionSink {
 	/// }
 	/// ```
 	///
-	pub fn close(self, err: impl Into<ErrorObjectOwned>) -> bool {
+	pub fn close(self, err: impl Into<ErrorObjectOwned>) -> impl Future<Output = ()> {
```
jsdw, line 1114: Since the subscription will gracefully close if you don't call this message too, and this is really just a way to close the subscription with some error, I wonder whether it should be called eg `close_with_error`?

examples/examples/ws_pubsub_with_params.rs
```diff
@@ -63,38 +65,37 @@ async fn main() -> anyhow::Result<()> {
 
 async fn run_server() -> anyhow::Result<SocketAddr> {
 	const LETTERS: &str = "abcdefghijklmnopqrstuvxyz";
-	let server = ServerBuilder::default().build("127.0.0.1:0").await?;
+	let server = ServerBuilder::default().set_backpressure_buffer_capacity(10).build("127.0.0.1:0").await?;
```
jsdw, line 68: Nit: not a big deal but I wonder whether it's better just to call it `set_buffer_capacity` (backpressure being implied as a result of a buffer existing and the whole future poll mechanism anyway)

tests/tests/integration_tests.rs
```diff
@@ -738,128 +724,6 @@ async fn http_batch_works() {
 	assert_eq!(err_responses, vec![&ErrorObject::borrowed(UNKNOWN_ERROR_CODE, &"Custom error: err", None)]);
 }
 
-#[tokio::test]
-async fn ws_server_limit_subs_per_conn_works() {
```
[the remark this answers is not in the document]
jsdw: Ah yeah I can see it's been removed! We may need a way to have this limit down the line, but perhaps can be thought of as part of some API that Jan would hook into to control new connections etc.

jsdw (approved):
Awesome job! A couple of nits and pondering the returning result from subscrioption API but looks great!

core/src/error.rs
```diff
@@ -229,3 +185,79 @@ impl From<hyper::Error> for Error {
...
+	/// Error cause is propagated by other code or connection related.
+	None,
+	/// Some error happened to be logged.
+	Some(String),
+}
+
+// User defined error.
+impl From<anyhow::Error> for SubscriptionCallbackError {
+	fn from(e: anyhow::Error) -> Self {
+		Self::Some(format!("Other: {}", e.to_string()))
```
lexnv, line 204:
```suggestion
		Self::Some(format!("Other: {}", e))
```

examples/examples/ws_pubsub_with_params.rs
niklasad1: yeah I know but @lexnv asked what the buffer was for so I wanted it be extra verbose

Not sure, @lexnv what do you prefer?

core/src/server/rpc_module.rs
niklasad1: yeah, it's a way to indicate to the connected client that subscription is closed with a special notification.

if one doesn't call this at least our client will wait until the connection is closed unless the subscription itself has customized logic when a subscription is regarded as "closed"

examples/examples/ws_pubsub_with_params.rs
lexnv: Yep, I do remember that, I think either option is fine for me. How does `set_backpressure(capacity: u32)` feel to you 🤔 ?

lexnv (approved):

core/src/server/rpc_module.rs
niklasad1: we do send an error to the "client" if _subscription call_ fails i.e, the user wouldn't call `accept/reject`.

when we get back an error here it's just some special logic from the user code that failed such as `SubscriptionMessage` or `Sink::send_timeout` or whatever.

what I am trying to say if a subscription just executes "correctly" or is "dropped" because of some error in the async closure then the client wouldn't be notified unless one calls `SubscriptionSink::close.

examples/examples/ws_pubsub_with_params.rs
niklasad1: what about `set_message_buffer`?

jsdw: I'm ok with `set_message_buffer_capacity` or something like that? "Capacity" is nice because it isn't size (it won't use the amount unless needed)

core/src/server/rpc_module.rs
niklasad1: thus, this may be an error or success.

but maybe we should have `close_with_success` and `close_with_error` instead as you suggested.

PR #967, opened 2023-01-06
client: distinstict APIs to configure max request and response sizes

### LINKED ISSUE #959: [clients]: split/rename `max_request_body_len` to `max_request_size` and `max_response_size`
We have similar APIs for the server and let's add them to the clients as well for clarity.

`max_request_body_len` was intended to mean both request and response but having different APIs is more clear.

See #958

niklasad1:
Close https://github.com/paritytech/jsonrpsee/issues/959

### DIFF
--- a/client/http-client/src/transport.rs
+++ b/client/http-client/src/transport.rs
@@ -237,7 +237,7 @@ mod tests {
 	#[test]
 	fn https_fails_without_tls_feature() {
 		let err =
-			HttpTransportClient::new("https://localhost:9933", 80, CertificateStore::Native, 80, HeaderMap::new())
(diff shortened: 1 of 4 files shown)


jsdw (approved):

client/http-client/src/transport.rs
```diff
@@ -44,7 +44,9 @@ pub struct HttpTransportClient {
 	/// HTTP client
 	client: HyperClient,
 	/// Configurable max request body size
-	max_request_body_size: u32,
+	max_request_size: u32,
+	/// Configurable max request body size
```
lexnv, line 48:
```suggestion
	/// Configurable max response body size
```

lexnv (approved):

PR #975, opened 2023-01-19
Add HTTP(S) proxy support

allquixotic:
This seems to work with both HTTP and HTTPS sites when using the jsonrpsee_http_client crate, allowing you to put a proxy in between your client and the endpoint. Very useful for debugging. Works with Proxyman, the proxy I'm using.

### DIFF
--- a/client/http-client/Cargo.toml
+++ b/client/http-client/Cargo.toml
@@ -14,7 +14,7 @@ async-trait = "0.1"
 rustc-hash = "1"
 hyper = { version = "0.14.10", features = ["client", "http1", "http2", "tcp"] }
 hyper-rustls = { version = "0.23", optional = true }
-hyper-proxy = { version = "0", default-features = false, features = ["rustls"] }
+hyper-proxy = { version = "0.9", default-features = false }
 jsonrpsee-types = { path = "../../types", version = "0.16.2" }
 jsonrpsee-core = { path = "../../core", version = "0.16.2", features = ["client", "http-helpers"] }
 serde = { version = "1.0", default-features = false, features = ["derive"] }
@@ -22,6 +22,7 @@ serde_json = "1.0"
 thiserror = "1.0"
 tokio = { version = "1.16", features = ["time"] }
 tracing = "0.1.34"
+anyhow = "1.0"
 
 [dev-dependencies]
 tracing-subscriber = { version = "0.3.3", features = ["env-filter"] }
@@ -30,7 +31,7 @@ tokio = { version = "1.16", features = ["net", "rt-multi-thread", "macros"] }
 
 [features]
 default = ["tls"]
-tls = ["hyper-rustls/webpki-tokio"]
+tls = ["hyper-rustls/webpki-tokio", "hyper-proxy/rustls"]
 
 [package.metadata.docs.rs]
 all-features = true
--- a/client/http-client/Cargo.toml
+++ b/client/http-client/Cargo.toml
@@ -22,7 +22,6 @@ serde_json = "1.0"
 thiserror = "1.0"
 tokio = { version = "1.16", features = ["time"] }
 tracing = "0.1.34"
-anyhow = "1.0"
 
 [dev-dependencies]
 tracing-subscriber = { version = "0.3.3", features = ["env-filter"] }
--- a/client/http-client/Cargo.toml
+++ b/client/http-client/Cargo.toml
@@ -14,6 +14,7 @@ async-trait = "0.1"
 rustc-hash = "1"
 hyper = { version = "0.14.10", features = ["client", "http1", "http2", "tcp"] }
 hyper-rustls = { version = "0.23", optional = true }
+hyper-proxy = { version = "0", default-features = false, features = ["rustls"] }
 jsonrpsee-types = { path = "../../types", version = "0.16.2" }
 jsonrpsee-core = { path = "../../core", version = "0.16.2", features = ["client", "http-helpers"] }
 serde = { version = "1.0", default-features = false, features = ["derive"] }
--- a/client/http-client/Cargo.toml
+++ b/client/http-client/Cargo.toml
@@ -14,7 +14,7 @@ async-trait = "0.1"
 rustc-hash = "1"
 hyper = { version = "0.14.10", features = ["client", "http1", "http2", "tcp"] }
 hyper-rustls = { version = "0.23", optional = true }
-hyper-proxy = { version = "0.9", default-features = false }
+hyper-proxy = { version = "0.9", default-features = false, optional = true }
 jsonrpsee-types = { path = "../../types", version = "0.16.2" }
 jsonrpsee-core = { path = "../../core", version = "0.16.2", features = ["client", "http-helpers"] }
 serde = { version = "1.0", default-features = false, features = ["derive"] }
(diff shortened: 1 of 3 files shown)


client/http-client/src/client.rs
```diff
@@ -79,6 +79,7 @@ pub struct HttpClientBuilder {
 	id_kind: IdKind,
 	max_log_length: u32,
 	headers: HeaderMap,
+	proxy: Option<String>,
```
niklasad1, line 82:
```suggestion
	proxy: Option<Uri>,
```

client/http-client/Cargo.toml
```diff
@@ -14,6 +14,7 @@ async-trait = "0.1"
 rustc-hash = "1"
 hyper = { version = "0.14.10", features = ["client", "http1", "http2", "tcp"] }
 hyper-rustls = { version = "0.23", optional = true }
+hyper-proxy = { version = "0", default-features = false, features = ["rustls"] }
```
niklasad1, line 17: it should be possible to disable tls and not bring all these deps

```suggestion
hyper-proxy = { version = "0.9", default-features = false }
```

client/http-client/src/transport.rs
```diff
@@ -87,7 +92,18 @@ impl HttpTransportClient {
 						.build(),
 					_ => return Err(Error::InvalidCertficateStore),
 				};
-				HyperClient::Https(Client::builder().build::<_, hyper::Body>(connector))
+				match proxy.clone() {
+					Some(pr) => {
+						let proxy_uri = pr.parse().unwrap();
```
niklasad1, line 97:
```suggestion
						let proxy_uri = pr.parse().map_err(|e| Error::Url(format!("Invalid proxy URI: {}", e)))?;
```

client/http-client/src/client.rs
```diff
@@ -134,6 +135,14 @@ impl HttpClientBuilder {
 		self
 	}
 
+	/// Set the HTTP(S) proxy that will proxy every HTTP request (default is none).
+	/// 
+	/// The proxy should be of the form <http[s]://host_or_ip:port> (without the brackets).
+	pub fn set_proxy(mut self, proxy: String) -> Self {
```
niklasad1, line 141: would be better to return `Result<Self, Error>` and to parse proxy URI directly here.

```rust
	pub fn set_proxy(mut self, proxy: impl AsRef<str>) -> Result<Self, Error> {
      self.proxy = proxy.parse().map_err(|e| Error::Transport(....))?;
      self
   }
```

client/http-client/src/transport.rs
```diff
@@ -87,7 +92,18 @@ impl HttpTransportClient {
 						.build(),
 					_ => return Err(Error::InvalidCertficateStore),
 				};
-				HyperClient::Https(Client::builder().build::<_, hyper::Body>(connector))
+				match proxy.clone() {
```
niklasad1, line 95: needless clone

```suggestion
				match proxy {
```

client/http-client/src/transport.rs
niklasad1, line 97: if the proxy URI is parsed in the builder this hasn't to be done again.

```suggestion
```

client/http-client/src/transport.rs
```diff
@@ -87,7 +92,18 @@ impl HttpTransportClient {
...
+					Some(pr) => {
+						let proxy_uri = pr.parse().unwrap();
+						let proxy_obj = Proxy::new(Intercept::All, proxy_uri);
+						let proxy_connector = ProxyConnector::from_proxy(connector, proxy_obj).unwrap();
```
niklasad1, line 99: we probably need to add `IoError` for this to work but no unwraps are allowed here.

```suggestion
						let proxy_connector = ProxyConnector::from_proxy(connector, proxy_obj)?;
```

niklasad1 (changes requested):
Thank you, looks generally good but I have some comments/suggestion that needs to addressed before merging.

Also, this needs to be added to the `WS client` which I reckon would be useful as well.

allquixotic:
> Also, this needs to be added to the WS client which I reckon would be useful as well.

Agreed, but this is PR is for HTTP proxy for the HTTP client; at least, that's what I'm able/willing to work on (I'm not using the lib with a WebSocket server). Will you consider merging this just for http if the code is corrected?

allquixotic:
Weird; Clippy seems to be crashing in the compiler?

niklasad1:
> Agreed, but this is PR is for HTTP proxy for the HTTP client; at least, that's what I'm able/willing to work on (I'm not using the lib with a WebSocket server). Will you consider merging this just for http if the code is corrected?

yeah, we can take of the `ws client` in another PR.

> Weird; Clippy seems to be crashing in the compiler?

yeah, it's known just ignore it for now https://github.com/rust-lang/rust-clippy/issues/9922

niklasad1 (approved):

client/http-client/src/client.rs
```diff
@@ -79,6 +80,17 @@ pub struct HttpClientBuilder {
 	id_kind: IdKind,
 	max_log_length: u32,
 	headers: HeaderMap,
+	proxy: Option<Uri>,
+}
+
+/// Error that may occur when trying to parse a `proxy URL`.
+#[derive(thiserror::Error, Debug)]
+struct InvalidProxyUrl(String);
```
jsdw, line 88: It looks like this is returned in an `Error`, so I think it should be `pub` so that it can be named by people.

jsdw: (and probably it should also be moved to live alongside `Error`? It feels a bit out of place here)

jsdw: Ah I see now' it's converted into an `anyhow::Error` before being returned; ignore me then :)

jsdw (approved):

lexnv (approved):

client/http-client/src/transport.rs
```diff
@@ -32,6 +37,8 @@ impl HyperClient {
 			Self::Http(client) => client.request(req),
 			#[cfg(feature = "tls")]
 			Self::Https(client) => client.request(req),
+			#[cfg(feature = "proxy")]
```
niklasad1, line 40:
```suggestion
			#[cfg(all(feature = "proxy", feature = "tls"))]
```

client/http-client/src/client.rs
```diff
@@ -134,6 +146,17 @@ impl HttpClientBuilder {
 		self
 	}
 
+	/// Set the HTTP(S) proxy that will proxy every HTTP request (default is none).
+	///
+	/// The proxy should be of the form <http://host_or_ip:port> (without the brackets).
+	#[cfg(feature = "proxy")]
```
niklasad1, line 152:
```suggestion
	#[cfg(all(feature = "proxy", feature = "tls"))]
```

niklasad1 (changes requested):
Hey again,

I have looked into this crate `hyper-proxy` and it doesn't really seems to be actively maintained and brings in a bunch of outdated dependencies such as `hyper-rustls`, https://github.com/tafia/hyper-proxy/pull/28/files has been opened over a year.

I think we should find another library and or write something ourselves using tower middleware.
Sorry, figured this out before I was going to merge it.

allquixotic:
Created [issue hyper#3124](https://github.com/hyperium/hyper/issues/3124) to ask the hyper maintainers what they would suggest for HTTP proxy support. I'd rather not reinvent it within jsonrpsee. 

Supporting an http proxy within an HTTP client library is a generally useful feature that can and should be either part of the hyper crate itself, or part of a maintained crate focused on that feature. This is basic stuff that exists everywhere else I've done HTTP requests (JavaScript, .NET, Java, GLib/C, Qt/C++, Python, Ruby, etc.) It's not jsonrpsee's responsibility to code this up.

PR #977, opened 2023-01-19
Custom errors

### LINKED ISSUE #973: Allow custom RPC method errors
I'd like to do something like this:

```rust
#[rpc(server, client, namespace = "myrpc")]
pub trait MyRpc {
    // an explicit error for this method
    #[method(name = "myMethod1")]
    async fn my_method1(&self) -> Result<MyData1, MyError1>;

    // an explicit error for this method
    #[method(name = "myMethod2")]
    async fn my_method2(&self) -> Result<MyData2, MyError2>;

    // backward-compatible generic error for this method
    #[method(name = "myMethod3")]
    async fn my_method3(&self) -> Result<MyData3, jsonrpsee::core::Error>;
}

pub enum MyError1 {
  // some variants here
}

pub enum MyError2 {
  // some variants here
}

impl From<MyError1> for jsonrpsee::core::Error {
    fn from(val: MyError1) -> jsonrpsee::core::Error {
        // my custom implementation for this method's errors
        jsonrpsee::core::Error::Other(ErrorObjectOwned { /*  ... */ }}
    }
}

impl From<MyError2> for jsonrpsee::core::Error {
    fn from(val: MyError2) -> jsonrpsee::core::Error {
        // my custom implementation for this method's errors
        jsonrpsee::core::Error::Other(ErrorObjectOwned { /*  ... */ }}
    }
}
```

The idea here is to allow methods to return an error type that can be converted into `jsonrpsee::core::Error` instead of requiring it explicitly.

This would allow us to have type-safe and clearly expressed application-level errors for the RPC calls - something we have had trouble with for quite some time.

We want to supply the data for certain errors, and our current implementation is a mess; people tend to manually construct the `jsonrpsee::core::Error` instead of going through the helper functions (which are not really discoverable, at least not in a way the type-safe errors enable).

Our frontend guys are also complaining it's hard to understand all possible application-level error cases as the code evolves, and we figured having an explicit error enum for every method would work very well for us.

---

This can be implemented in a backward-compatible way if the error is made to be `impl Into<jsonrpsee::core::Error>` instead of an explicit `jsonrpsee::core::Error`.

Alternatively, `Into<jsonrpsee::types:error::ErrorObjectOwned>` could be used - this is more direct, but doesn't allow invoking the baked-in errors, which is kind of a limitation.

---

This does not make much sense from the client-side perspective, so I propose the client trait to rewrite the errors to still be `jsonrpsee::core::Error`.

MOZGIII:
Closes #973.

---

To do:

- [x] permit other paths to `Result` in the error type rewrite macro code

### DIFF
--- a/core/src/server/rpc_module.rs
+++ b/core/src/server/rpc_module.rs
@@ -576,7 +576,7 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 	) -> Result<MethodResourcesBuilder, Error>
 	where
 		R: Serialize + Send + Sync + 'static,
-		E: for<'a> Into<Error>,
+		E: Into<Error>,
 		Fut: Future<Output = Result<R, E>> + Send,
 		Fun: (Fn(Params<'static>, Arc<Context>) -> Fut) + Clone + Send + Sync + 'static,
 	{
@@ -615,7 +615,7 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 	where
 		Context: Send + Sync + 'static,
 		R: Serialize,
-		E: for<'a> Into<Error>,
+		E: Into<Error>,
 		F: Fn(Params, Arc<Context>) -> Result<R, E> + Clone + Send + Sync + 'static,
 	{
 		let ctx = self.ctx.clone();
--- a/core/src/server/rpc_module.rs
+++ b/core/src/server/rpc_module.rs
@@ -569,14 +569,15 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 	}
 
 	/// Register a new asynchronous RPC method, which computes the response with the given callback.
-	pub fn register_async_method<R, Fun, Fut>(
+	pub fn register_async_method<R, E, Fun, Fut>(
 		&mut self,
 		method_name: &'static str,
 		callback: Fun,
 	) -> Result<MethodResourcesBuilder, Error>
 	where
 		R: Serialize + Send + Sync + 'static,
-		Fut: Future<Output = Result<R, Error>> + Send,
+		E: for<'a> Into<Error>,
+		Fut: Future<Output = Result<R, E>> + Send,
 		Fun: (Fn(Params<'static>, Arc<Context>) -> Fut) + Clone + Send + Sync + 'static,
 	{
 		let ctx = self.ctx.clone();
@@ -589,7 +590,7 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 				let future = async move {
 					let result = match callback(params, ctx).await {
 						Ok(res) => MethodResponse::response(id, res, max_response_size),
-						Err(err) => MethodResponse::error(id, err),
+						Err(err) => MethodResponse::error(id, err.into()),
 					};
 
 					// Release claimed resources
@@ -606,7 +607,7 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 
 	/// Register a new **blocking** synchronous RPC method, which computes the response with the given callback.
 	/// Unlike the regular [`register_method`](RpcModule::register_method), this method can block its thread and perform expensive computations.
-	pub fn register_blocking_method<R, F>(
+	pub fn register_blocking_method<R, E, F>(
 		&mut self,
 		method_name: &'static str,
 		callback: F,
@@ -614,7 +615,8 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 	where
 		Context: Send + Sync + 'static,
 		R: Serialize,
-		F: Fn(Params, Arc<Context>) -> Result<R, Error> + Clone + Send + Sync + 'static,
+		E: for<'a> Into<Error>,
+		F: Fn(Params, Arc<Context>) -> Result<R, E> + Clone + Send + Sync + 'static,
 	{
 		let ctx = self.ctx.clone();
 		let callback = self.methods.verify_and_insert(
@@ -626,7 +628,7 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 				tokio::task::spawn_blocking(move || {
 					let result = match callback(params, ctx) {
 						Ok(result) => MethodResponse::response(id, result, max_response_size),
-						Err(err) => MethodResponse::error(id, err),
+						Err(err) => MethodResponse::error(id, err.into()),
 					};
 
 					// Release claimed resources
--- a/proc-macros/src/render_client.rs
+++ b/proc-macros/src/render_client.rs
@@ -87,6 +87,11 @@ impl RpcDescription {
 			_ => unreachable!("Unexpected Result structure"),
 		};
 
+		if args.len() != 2 {
+			// Unexpected number of type arguments, just leave it as-is.
+			return syn::Type::Path(path);
+		}
+
 		let error = args.last_mut().unwrap();
 		let error_type = match error {
 			GenericArgument::Type(error_type) => error_type,
--- a/proc-macros/src/render_client.rs
+++ b/proc-macros/src/render_client.rs
@@ -24,11 +24,12 @@
 // IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 // DEALINGS IN THE SOFTWARE.
 use crate::attributes::ParamKind;
-use crate::helpers::{generate_where_clause, path_segment_mut};
+use crate::helpers::generate_where_clause;
 use crate::rpc_macro::{RpcDescription, RpcMethod, RpcSubscription};
 use proc_macro2::TokenStream as TokenStream2;
-use quote::quote;
-use syn::{AngleBracketedGenericArguments, FnArg, GenericArgument, Pat, PatIdent, PatType, PathArguments, TypeParam};
+use quote::{quote, quote_spanned};
+use syn::spanned::Spanned;
+use syn::{AngleBracketedGenericArguments, FnArg, Pat, PatIdent, PatType, PathArguments, TypeParam};
 
 impl RpcDescription {
 	pub(super) fn render_client(&self) -> Result<TokenStream2, syn::Error> {
@@ -68,44 +69,44 @@ impl RpcDescription {
 		Ok(trait_impl)
 	}
 
-	/// Alter the error varaint (`Err`) type of the [`Result`], or return the type as-is if it is not a [`Result`].
-	/// Intended for rewriting the retuirn type of RPC methods for clients, where the return type is defined
-	/// as a [`Result`] with a custom error. Clients can not actually benefit from custom errors right now, so
-	/// we just rewrtie the type to be the good old [`core::Error`].
-	fn patch_result_error(&self, ty: &syn::Type) -> syn::Type {
-		let mut type_path = match ty {
-			syn::Type::Path(path) => path.clone(),
-			_ => panic!("Client only supports bare or Result values: {:?}", ty),
+	/// Verify and rewrite the return type (for methods).
+	fn return_result_type(&self, ty: &syn::Type) -> TokenStream2 {
+		// We expect a valid type path.
+		let syn::Type::Path(type_path) = ty else  {
+			return quote_spanned!(ty.span() => compile_error!("Expecting something like 'Result<Foo, Err>' here. (1)"));
 		};
 
-		let valids_paths = [&["Result"][..], &["std", "result", "Result"][..], &["core", "result", "Result"][..]];
-		let Some(result_segment) = path_segment_mut(&mut type_path.path, valids_paths) else {
-			return syn::Type::Path(type_path);
+		// The path (eg std::result::Result) should have a final segment like 'Result'.
+		let Some(type_name) = type_path.path.segments.last() else {
+			return quote_spanned!(ty.span() => compile_error!("Expecting this path to end in something like 'Result<Foo, Err>'"));
 		};
 
-		if result_segment.ident != "Result" {
-			return syn::Type::Path(type_path);
+		// Get the generic args eg the <T, E> in Result<T, E>.
+		let PathArguments::AngleBracketed(AngleBracketedGenericArguments { args, .. }) = &type_name.arguments else {
+			return quote_spanned!(ty.span() => compile_error!("Expecting something like 'Result<Foo, Err>' here, but got no generic args (eg no '<Foo,Err>')."));
+		};
+
+		if type_name.ident == "Result" {
+			// Result<T, E> should have 2 generic args.
+			if args.len() != 2 {
+				return quote_spanned!(args.span() => compile_error!("Expecting two generic args here."));
+			}
+
+			// The first generic arg is our custom return type.
+			// Return a new Result with that custom type and jsonrpsee::core::Error:
+			let custom_return_type = args.first().unwrap();
+			let error_type = self.jrps_client_item(quote! { core::Error });
+			quote!(std::result::Result<#custom_return_type, #error_type>)
+		} else if type_name.ident == "RpcResult" {
+			// RpcResult<T> (an alias we export) should have 1 generic arg.
+			if args.len() != 1 {
+				return quote_spanned!(args.span() => compile_error!("Expecting two generic args here."));
+			}
+			quote!(#ty)
+		} else {
+			// Any other type name isn't allowed.
+			quote_spanned!(type_name.span() => compile_error!("The response type should be Result or RpcResult"))
 		}
-
-		let args = match result_segment.arguments {
-			PathArguments::AngleBracketed(AngleBracketedGenericArguments { ref mut args, .. }) => args,
-			_ => unreachable!("Unexpected Result structure"),
-		};
-
-		if args.len() != 2 {
-			// Unexpected number of type arguments, just leave it as-is.
-			return syn::Type::Path(type_path);
-		}
-
-		let error = args.last_mut().unwrap();
-		let error_type = match error {
-			GenericArgument::Type(error_type) => error_type,
-			_ => unreachable!("Unexpected Result structure"),
-		};
-
-		*error_type = syn::Type::Verbatim(self.jrps_client_item(quote! { core::Error }));
-
-		syn::Type::Path(type_path)
 	}
 
 	fn render_method(&self, method: &RpcMethod) -> Result<TokenStream2, syn::Error> {
@@ -123,7 +124,7 @@ impl RpcDescription {
 		// `returns` represent the return type of the *rust method* (`Result< <..>, jsonrpsee::core::Error`).
 		let (called_method, returns) = if let Some(returns) = &method.returns {
 			let called_method = quote::format_ident!("request");
-			let returns = self.patch_result_error(returns);
+			let returns = self.return_result_type(returns);
 			let returns = quote! { #returns };
 
 			(called_method, returns)
--- a/proc-macros/src/render_client.rs
+++ b/proc-macros/src/render_client.rs
@@ -89,7 +89,7 @@ impl RpcDescription {
 		if type_name.ident == "Result" {
 			// Result<T, E> should have 2 generic args.
 			if args.len() != 2 {
-				return quote_spanned!(args.span() => compile_error!("Result must be have two arguments));
+				return quote_spanned!(args.span() => compile_error!("Result must be have two arguments"));
 			}
 
 			// Force the last argument to be `jsonrpsee::core::Error`:
--- a/proc-macros/src/render_client.rs
+++ b/proc-macros/src/render_client.rs
@@ -70,19 +70,19 @@ impl RpcDescription {
 	}
 
 	/// Verify and rewrite the return type (for methods).
-	fn return_result_type(&self, ty: &syn::Type) -> TokenStream2 {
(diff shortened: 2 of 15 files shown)


dmitrylavrenov (approved):
Useful PR and good test example!

niklasad1 (commented):
Generally looks good and nice improvement.

The only thing I worry about is the readability of client trait i.e if one provides a custom type then 
we would modify the error type to `jsonrpsee::core::Error` under the hood which is hard to understand.

We are already doing this anyway for subscriptions to some extent and generates a slightly different trait.

Give us a couple days to think about it, I reach out again soon.

core/src/server/rpc_module.rs
```diff
@@ -569,14 +569,15 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
...
-	pub fn register_async_method<R, Fun, Fut>(
+	pub fn register_async_method<R, E, Fun, Fut>(
 		&mut self,
 		method_name: &'static str,
 		callback: Fun,
 	) -> Result<MethodResourcesBuilder, Error>
 	where
 		R: Serialize + Send + Sync + 'static,
-		Fut: Future<Output = Result<R, Error>> + Send,
+		E: for<'a> Into<Error>,
```
jsdw, line 579: The `for<'a>` here doesn't seem to do anything (ie the `'a` isn't used)?

core/src/server/rpc_module.rs
```diff
@@ -606,15 +607,16 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
...
+	pub fn register_blocking_method<R, E, F>(
 		&mut self,
 		method_name: &'static str,
 		callback: F,
 	) -> Result<MethodResourcesBuilder, Error>
 	where
 		Context: Send + Sync + 'static,
 		R: Serialize,
-		F: Fn(Params, Arc<Context>) -> Result<R, Error> + Clone + Send + Sync + 'static,
+		E: for<'a> Into<Error>,
```
jsdw, line 618: Unneeded `for<'a>`?

proc-macros/src/render_client.rs
```diff
@@ -83,6 +113,7 @@ impl RpcDescription {
 		// `returns` represent the return type of the *rust method* (`Result< <..>, jsonrpsee::core::Error`).
 		let (called_method, returns) = if let Some(returns) = &method.returns {
 			let called_method = quote::format_ident!("request");
+			let returns = self.patch_result_error(returns);
```
jsdw, line 116: I don't understand what this `patch_result_error` function is doing now offhand; could you explain it to me?

core/src/server/rpc_module.rs
```diff
@@ -569,14 +569,15 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 	}
 
 	/// Register a new asynchronous RPC method, which computes the response with the given callback.
-	pub fn register_async_method<R, Fun, Fut>(
+	pub fn register_async_method<R, E, Fun, Fut>(
```
jsdw, line 572: I am wondering whether we'd gain a bit more flexibility ultimately by copying Axum a bit and having a trait which is something like

```rust
trait IntoResponse {
    fn into_response(self) -> MethodResponse;
}
```

This could then be impl'd for `Result<V, E> where V: Serialize, E: Into<Error>` for instance:

```rust
impl <V, E> IntoResponse for Result<V, E> 
where
    V: Serialize,
    E: Into<Error>
{
    fn into_response(self) -> MethodResponse {
        // look at impl MethodResponse::response/error and use this here
    }
} 
```

fns like this `register_method` would expect a callback which was like `Fn(Params, &Context) ->  R` (where `R: IntoResponse`) and would call `into_response()` on the result (perhaps passing in any specific bits they need to as args to that).

Then we could impl things like `From<Infallible> for Error` to allow `Result<V, Infallible>` responses for instance, and perhaps have more room then to consider supporting other response types than Result in the future.

Just thinking out loud really..

proc-macros/src/render_client.rs
```diff
@@ -68,6 +68,36 @@ impl RpcDescription {
...
+		let mut path = match ty {
+			syn::Type::Path(path) => path.clone(),
+			_ => panic!("Client only supports bare or Result values: {:?}", ty),
+		};
+
+		let Some(first_segment) = path.path.segments.first_mut() else {
+			return syn::Type::Path(path);
+		};
+
+		if first_segment.ident != "Result" {
```
lexnv, line 81: I'm wondering would this work as well for `std::result::Result`?

proc-macros/src/render_client.rs
```diff
@@ -68,6 +68,36 @@ impl RpcDescription {
...
+			return syn::Type::Path(path);
+		}
+
+		let args = match first_segment.arguments {
+			PathArguments::AngleBracketed(AngleBracketedGenericArguments { ref mut args, .. }) => args,
+			_ => unreachable!("Unexpected Result structure"),
+		};
+
+		let error = args.last_mut().unwrap();
```
lexnv, line 90: nit: Should we check here that we only have 2 arguments? Tho, I'm wondering how this would play into some syntax sugar like `type Result<T> = std::Result<T, MyErr>`?

proc-macros/src/render_client.rs
MOZGIII: Sure! I'll also add the comment documenting it since it's not trivial!

It tries to rewrite the return type from `Result<_, E>` (generic `E`) to `Result<_, Error>` (concrete `Error`) - which is required because the client can't be expected to parse the custom error type back from JSON.

core/src/server/rpc_module.rs
MOZGIII: Oh yes, I considered it, but ultimately decided against doing huge changes in a single PR. This is something that, I think, can be added later iteratively - and, also, maybe someone else would like to tackle that instead of me. :D I'm fine with looking into it, but I'm also happy to share the fun.

proc-macros/src/render_client.rs
MOZGIII: This is a very good find! In fact, this logic has to be altered to permit a number of paths here, including `std::result::Result` - however I need more research into proc macros to determine exactly what things to cover, so I left it for now; this is one of the TODOs before the merge.

core/src/server/rpc_module.rs
MOZGIII: I propose that we split the error layers into jsonrpsee's and application level, and, further, split the jsonrpsee's error from one single struct into separate `ClientError` and `ServerError` as there are very distinctly different in handling. I think it would significantly improve the ergonomics, and unlock more explicit and sensible error customization for the client side, as this PR only addresses the server side really.

proc-macros/src/render_client.rs
```diff
@@ -68,6 +68,41 @@ impl RpcDescription {
 		Ok(trait_impl)
 	}
 
+	fn patch_result_error(&self, ty: &syn::Type) -> syn::Type {
```
jsdw, line 71: I had a play and wonder whether this would be a clearer approach to take:

```rust
fn return_result_type(&self, ty: &syn::Type) -> TokenStream2 {
	// We expect a valid type path.
	let syn::Type::Path(type_path) = ty else  {
		return quote_spanned!(ty.span() => compile_error!("Expecting something like 'Result<Foo, Err>' here. (1)"));
	};

	// The path (eg std::result::Result) should have a final segment like 'Result'.
	let Some(type_name) = type_path.path.segments.last() else {
		return quote_spanned!(ty.span() => compile_error!("Expecting this path to end in something like 'Result<Foo, Err>'"));
	};

	// Get the generic args eg the <T, E> in Result<T, E>.
	let PathArguments::AngleBracketed(AngleBracketedGenericArguments { args, .. }) = &type_name.arguments else {
		return quote_spanned!(ty.span() => compile_error!("Expecting something like 'Result<Foo, Err>' here, but got no generic args (eg no '<Foo,Err>')."));
	};

	if type_name.ident == "Result"{
		// Result<T, E> should have 2 generic args.
		if args.len() != 2 {
			return quote_spanned!(args.span() => compile_error!("Expecting two generic args here."));
		}
	} else if type_name.ident == "RpcResult" {
		// RpcResult<T> (an alias we export) should have 1 generic arg.
		if args.len() != 1 {
			return quote_spanned!(args.span() => compile_error!("Expecting two generic args here."));
		}
	} else {
		// Any other type name isn't allowed.
		return quote_spanned!(type_name.span() => compile_error!("The response type should be Result or RpcResult"));
	}

	// The first generic arg is our custom return type.
	// Return a new Result with that custom type and jsonrpsee::core::Error:
	let custom_return_type = args.first().unwrap();
	let error_type = self.jrps_client_item(quote! { core::Error });
	quote!(std::result::Result<#custom_return_type, #error_type>)
}
```

Aside from being documented, it checks for both Result and RpcResult (and verified the number of generics), and returns nicer compile errors which highlight the offending return type when it's wrong.

Instead of modifying the type I just looked for the generic param and then return a new Result type which uses it and the standard error.

Needs testing properly though but it compiles ok!

MOZGIII: Yeah, it is a different approach; when choosing between the two I figured I'd rather impose as few restrictions as possible, and thus tried to surgically target the `Result`'s error. My rationale for going this way is, for instance, I'm not certain at all that types other than `RpcResult` and `Result<_, _>` are not allowed... But after a quick look through the tests, I can tell it is the case.
So, yeah, there is definitely no such restriction explicitly anywhere just yet, and I wasn't sure it would be the right way of introducing it.

There's also a concern with supporting multiple paths to [`RpcResult`] - but I've implemented a nice helper for handling paths for [`Result`] that would work with [`RpcResult`] too.

MOZGIII: Tried it! One unexpected artifact is this UI warning:

```
┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈
warning: unused import: `core::RpcResult`
 --> .../jsonrpsee/proc-macros/tests/ui/correct/only_client.rs:3:17
  |
3 | use jsonrpsee::{core::RpcResult, proc_macros::rpc};
  |                 ^^^^^^^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default
```

This comes from this code:

```rust
//! Example of using proc macro to generate working client and server.

use jsonrpsee::{core::RpcResult, proc_macros::rpc};

#[rpc(client)]
pub trait Rpc {
	#[method(name = "foo")]
	async fn async_method(&self, param_a: u8, param_b: String) -> RpcResult<u16>;

	#[method(name = "bar")]
	fn sync_method(&self) -> RpcResult<u16>;

	#[subscription(name = "subscribe", item = String)]
	fn sub(&self);
}

fn main() {}
```

If we rewrite the code to use `Result` explicitly it might actually make `RpcResult` unused! Good thing there are UI tests in this project, here...

MOZGIII: What if user has a custom `Result` type for whatever reason? Shall we replace it with `std::result::Result`? It is a pretty edge case, yet I'd suggest that we pass through that type too, and just alter the last argument. That would be a bit less surprising for someone injecting custom `Result`s.

MOZGIII: I think I have a version that covers all concerns now, please take a look. It does error rewrite again to be more permissive but provides those nice compile errors (for which I need to add tests actually!).

proc-macros/src/render_client.rs
MOZGIII: This was addressed by using the last argument

core/src/server/rpc_module.rs
MOZGIII: If you like the direction I proposed I'll create an issue for splitting the errors and mark this discussion as resolved.

core/src/server/rpc_module.rs
MOZGIII: Good find, I removed this one and the other one! It was a leftover from my experiments with directly using the `ErrorObject<'a>`

proc-macros/src/render_client.rs
jsdw: > What if user has a custom `Result` type for whatever reason? Shall we replace it with `std::result::Result`? It is a pretty edge case, yet I'd suggest that we pass through that type too, and just alter the last argument. That would be a bit less surprising for someone injecting custom `Result`s.

I think that is the user has some custom type that isn't something we expect (ie a Result or an RpcResult) then offhand the safest option is to complain, since in the general case we can't possibly know what the structure of said custom type is. @niklasad1 do you have any opinions on this?

core/src/server/rpc_module.rs
jsdw: @niklasad1 whatcha reckon? I think you were less keen on splitting the errors?

But in any case, I think for now I'm fine with just allowing the custom error and not being more general; perhaps in some future iteration we can add the above sort of thing if it makes sense :)

proc-macros/src/render_client.rs
jsdw: FYI I had a look at your latest version; it doesn't look any more permissive to me since it still requires the name to be Result or RpcResult (and the Result type still requires 2 args, so it's highly unlikely somebody will be importing a "Result" type alias of that kind anyway). So I still prefer my approach (and avoiding mutating things for clarity) but I'm not super bothered.

niklasad1: > I think that is the user has some custom type that isn't something we expect (ie a Result or an RpcResult) then offhand the safest option is to complain, since in the general case we can't possibly know what the structure of said custom type is. @niklasad1 do you have any opinions on this?

Yeah, I agree I think it's better to throw a compile-time error if the error type isn't Result/RpcResult.

core/src/server/rpc_module.rs
niklasad1: I think it will be weird to keep the proc macro API i.e, to share the same definition with different error types for the client and server. It's still possible to do it in the proc macro code as you have changed it this PR.

Example both client and server
```rust
struct CustomError;

#[rpc(server, client)]
pub trait Rpc
{
	// The client will return `ClientError` here.....
	async fn f(&self) -> Result<String, CustomError>;
}
```

Example both cliet
```rust

#[rpc(client)]
pub trait Rpc
{
	// I can enter whatever type I want here but it will return ClientError anyway.....
	async fn f(&self) -> Result<String, ClientError>;
}
```

I mostly care about having the client and server to share the API definition which I like to avoid a bunch of boiler plate and so on.

proc-macros/src/render_client.rs
MOZGIII: The case where this solution is more permissive is when the user provides a type called `Result` that has the same shape as the std `Result`, but is a different type. For instance, a struct instead of enum (a case that is unlikely to happen), or a wrapper (much more common, although maybe not for `Result`), for instance for implementing some traits on it that conflict otherwise.

core/src/server/rpc_module.rs
MOZGIII: I'm thinking about it more like this:

```
struct CustomError;

#[rpc(server, client)]
pub trait Rpc {
	// The client will return `ClientError` here.....
	async fn f(&self) -> Result<String, CustomError>;
}


struct Server;

#[async_trait]
impl RpcServer for Server {
    async fn f(&self) -> Result<String, ServerError<CustomError>>; // <------ see the `ServerError`
}
```

And the client would have a `ClientError` or `ClientError<CustomError>` type.

niklasad1: Sure, that idea is interesting but it complicates things in other crates such as `core` but I would prefer to keep it out of this PR.

That alone would cause plenty of changes

The benefit of splitting it up is that is easier to reason about the error scenarios but forces users to import a few different error types (fine though)

proc-macros/src/render_client.rs
```diff
@@ -68,6 +69,46 @@ impl RpcDescription {
...
+
+		// Get the generic args eg the <T, E> in Result<T, E>.
+		let PathArguments::AngleBracketed(AngleBracketedGenericArguments { args, .. }) = &mut type_name.arguments else {
+			return quote_spanned!(ty.span() => compile_error!("Expecting something like 'Result<Foo, Err>' here, but got no generic args (eg no '<Foo,Err>')."));
+		};
+
+		if type_name.ident == "Result" {
+			// Result<T, E> should have 2 generic args.
+			if args.len() != 2 {
+				return quote_spanned!(args.span() => compile_error!("Expecting two generic args here."));
```
niklasad1, line 92:
```suggestion
				return quote_spanned!(args.span() => compile_error!("std::result::Result<T, E> must be used, no type aliases are allowed));
```

proc-macros/src/render_client.rs
```diff
@@ -68,6 +69,46 @@ impl RpcDescription {
...
+
+			// Force the last argument to be `jsonrpsee::core::Error`:
+			let error_arg = args.last_mut().unwrap();
+			*error_arg = syn::GenericArgument::Type(syn::Type::Verbatim(self.jrps_client_item(quote! { core::Error })));
+
+			quote!(#ty)
+		} else if type_name.ident == "RpcResult" {
+			// RpcResult<T> (an alias we export) should have 1 generic arg.
+			if args.len() != 1 {
+				return quote_spanned!(args.span() => compile_error!("Expecting two generic args here."));
```
niklasad1, line 103:
```suggestion
				return quote_spanned!(args.span() => compile_error!("RpcResult must have one argument"));
```

proc-macros/src/render_client.rs
```diff
@@ -68,6 +69,46 @@ impl RpcDescription {
...
+			}
+			quote!(#ty)
+		} else {
+			// Any other type name isn't allowed.
+			quote_spanned!(type_name.span() => compile_error!("The response type should be Result or RpcResult"))
```
niklasad1, line 108:
```suggestion
			quote_spanned!(type_name.span() => compile_error!("The return type must be Result or RpcResult"))
```

proc-macros/src/render_client.rs
```diff
@@ -68,6 +69,46 @@ impl RpcDescription {
...
+		let Some(type_name) = type_path.path.segments.last_mut() else {
+			return quote_spanned!(ty.span() => compile_error!("Expecting this path to end in something like 'Result<Foo, Err>'"));
+		};
+
+		// Get the generic args eg the <T, E> in Result<T, E>.
+		let PathArguments::AngleBracketed(AngleBracketedGenericArguments { args, .. }) = &mut type_name.arguments else {
+			return quote_spanned!(ty.span() => compile_error!("Expecting something like 'Result<Foo, Err>' here, but got no generic args (eg no '<Foo,Err>')."));
+		};
+
+		if type_name.ident == "Result" {
```
niklasad1, line 89: technically I'm not sure that we need these guards anymore as long the error type impls Into<ErrorObjectOwned>

one would get some weird compile error once it doesn't I guess.

core/src/server/rpc_module.rs
```diff
@@ -569,14 +569,15 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
...
 		&mut self,
 		method_name: &'static str,
 		callback: Fun,
 	) -> Result<MethodResourcesBuilder, Error>
 	where
 		R: Serialize + Send + Sync + 'static,
-		Fut: Future<Output = Result<R, Error>> + Send,
+		E: Into<Error>,
```
niklasad1, line 579: It didn't work with `E: Into<ErrorObjectOwned>`? 

Would be great skip the conversion to `jsonrpsee::core::Error`

proc-macros/src/render_client.rs
jsdw: The only reason we have these guards etc now is to check what sort of type we're looking at in order to trust that the `T` param is the return type so that we can rewrite the error properly to a `Result<T, core::Error>`.

Without them things would work still but be more permissive and may lead to weird compile errors with custom types rather than our nicer "use a Result or RpcResult" type ones.

niklasad1: gotcha

core/src/server/rpc_module.rs
MOZGIII: The whole purpose of using `jsonrpsee::core::Error` there is to enable the apps to easier construct the errors that are [`jsonrpsee::core::Error`].

niklasad1: gotcha, I think we should break it for ergonomics but let's do it in another PR

niklasad1 (approved):

MOZGIII:
I think we might want to add a corrsponding code to check for `Result`/`RpcResult` to the server side as well. No rewrites just yet (to be added in the subsequest PRs), only the checking that the type must be explicitly `Result` or `RpcResult`.

What do you think?

niklasad1:
> I think we might want to add a corrsponding code to check for Result/RpcResult to the server side as well. No rewrites just yet (to be added in the subsequest PRs), only the checking that the type must be explicitly Result or RpcResult.

What do you think?

AFAIU, the only reason we need that for the client is because we are trying to rewrite to return type as `Result<T, jsonrpsee::core::Error>` and for the server that is not needed.

Thus, the "ordinary" compile error should good enough then.

jsdw (approved):

lexnv (approved):

PR #981, opened 2023-01-23
feat(http client): add tower middleware

niklasad1:


client/http-client/src/transport.rs
```diff
@@ -98,35 +134,30 @@ impl HttpTransportClient {
 			}
 		};
 
-		// Cache request headers: 2 default headers, followed by user custom headers.
-		// Maintain order for headers in case of duplicate keys:
-		// https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.2
-		let mut cached_headers = HeaderMap::with_capacity(2 + headers.len());
-		cached_headers.insert(hyper::header::CONTENT_TYPE, HeaderValue::from_static(CONTENT_TYPE_JSON));
-		cached_headers.insert(hyper::header::ACCEPT, HeaderValue::from_static(CONTENT_TYPE_JSON));
```
lexnv, line 106: Dumb question: Would we need a http::middleware in substrate to insert those two headers?

core/src/http_helpers.rs
```diff
@@ -47,33 +54,45 @@ pub async fn read_body(
...
-		_ => return Err(GenericTransportError::Malformed),
-	};
+		if data.chunk().len() > max_request_body_size as usize {
+			return Err(GenericTransportError::TooLarge);
+		}
 
-	let mut received_data = Vec::with_capacity(body_size as usize);
-	received_data.extend_from_slice(&first_chunk);
+		if received_data.is_empty() {
+			let first_non_whitespace = data.chunk().iter().find(|byte| !byte.is_ascii_whitespace());
```
lexnv, line 70: Nit: If we were to remove the previous check we could optimize for malformed input with an `.take(32)`, if we don't see the `{` / `[` character we could bail out sooner

lexnv (approved):
LGTM! Amazing job 👍

core/src/http_helpers.rs
```diff
@@ -47,33 +54,45 @@ pub async fn read_body(
...
-	let first_non_whitespace = first_chunk.iter().find(|byte| !byte.is_ascii_whitespace());
+	while let Some(d) = body.data().await {
+		let data = d.map_err(|e| GenericTransportError::Inner(anyhow!(e.into())))?;
 
-	let single = match first_non_whitespace {
-		Some(b'{') => true,
-		Some(b'[') => false,
-		_ => return Err(GenericTransportError::Malformed),
-	};
+		if data.chunk().len() > max_request_body_size as usize {
```
lexnv, line 65: Nit: since we also validate below that `body_size > max_request`, would it be safe to remote this check from here? 🤔

niklasad1: yeah, good catch this check is useless

client/http-client/src/transport.rs
niklasad1: I can move that back, I didn't like the `cached headers fn` but did it when I was experimenting with the conditional compilation

niklasad1: Not sure but these can be overwritten by custom headers or middleware :)

lexnv: I like it how it is now 👍 I missed it during my first review pass and then it made sense for me in the second run 😄 thanks!

examples/examples/http.rs
```diff
@@ -41,7 +45,20 @@ async fn main() -> anyhow::Result<()> {
...
+				|request: &hyper::Request<hyper::Body>, _span: &tracing::Span| tracing::info!(request = ?request, "on_request"),
+			)
+			.on_body_chunk(|chunk: &Bytes, latency: Duration, _: &tracing::Span| {
+				tracing::info!(size_bytes = chunk.len(), latency = ?latency, "sending body chunk")
+			})
+			.make_span_with(DefaultMakeSpan::new().include_headers(true))
+			.on_response(DefaultOnResponse::new().include_headers(true).latency_unit(LatencyUnit::Micros)),
+	);
+
+	let client = HttpClientBuilder::default().set_middleware(middleware).build(url)?;
```
jsdw, line 61: Very cool!

core/src/http_helpers.rs
niklasad1: let's do it another PR, we could also fetch the position here skip the bytes with whitespaces

niklasad1: I changed my mind, I used `take(128)` but not sure what is reasonable limit

lexnv: That sounds reasonable to me. I'm not sure if the spec says anything about it, but I would expect most cases to be a few characters away

niklasad1: it's just that it's opinionated and we might reject valid HTTP responses that contains more than 128 whitespaces but it's valid JSON-RPC response :P 

however, I think it's fair trade-off because the JSON decoding happens later.

client/http-client/src/transport.rs
```diff
@@ -6,43 +6,75 @@
...
+	B: HttpBody + Send + 'static,
+	B::Data: Send,
+	B::Error: Into<Box<dyn StdError + Send + Sync>>,
+{
+	type Response = hyper::Response<Body>;
+	type Error = Error;
+	type Future = Pin<Box<dyn Future<Output = Result<Self::Response, Self::Error>> + Send>>;
+
+	fn poll_ready(&mut self, _: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+		Poll::Ready(Ok(()))
```
jsdw, line 57: Just to thread this through properly, could this be:

```rust
let resp = match self {
	Self::Http(inner) => inner.poll_ready(ctx),
	#[cfg(feature = "tls")]
	Self::Https(inner) => inner.poll_ready(ctx),
};
```

core/src/http_helpers.rs
```diff
@@ -27,18 +27,25 @@
...
-) -> Result<(Vec<u8>, bool), GenericTransportError<hyper::Error>> {
+) -> Result<(Vec<u8>, bool), GenericTransportError>
+where
+	B: HttpBody + Send + 'static,
+	B::Data: Send,
+	B::Error: Into<Box<dyn StdError + Send + Sync>>,
+{
 	// NOTE(niklasad1): Values bigger than `u32::MAX` will be turned into zero here. This is unlikely to occur in
 	// practice and for that case we fallback to allocating in the while-loop below instead of pre-allocating.
 	let body_size = read_header_content_length(headers).unwrap_or(0);
```
jsdw, line 51: Just a nit: what if somebody sets a `Content-Length: u32::MAX` type header just to DoS this with huge allocations?

Maybe it's best to do a `.max(1024 * 1024 * 4)` or something to set a sane limit on how much to pre-allocate?

jsdw: (could also take `max_request_body_size` into account?)

jsdw (approved):
Looks great! Just a couple of small things :)

core/src/http_helpers.rs
niklasad1: we have this check before allocating:

```rust
	if body_size > max_request_body_size {
		return Err(GenericTransportError::TooLarge);
	}
```

Thus, we can't allocate more than `max_request_body_size`. Probably sufficient?

niklasad1: we could take it one stop further though as you said with `min(max_request_body_size, 1024 * 16)`

jsdw: Yeah maybe both is good; that way something can't force us to pre-allocate `max_body_size` every time even unless it actually sends that much data! But with that check you mentioned in place this is def much less important so whatever you think best :)

niklasad1: that's probably fair, done

core/src/http_helpers.rs
```diff
@@ -27,53 +27,84 @@
...
 	}
 
-	let first_chunk =
-		body.next().await.ok_or(GenericTransportError::Malformed)?.map_err(GenericTransportError::Inner)?;
-
-	if first_chunk.len() > max_request_body_size as usize {
-		return Err(GenericTransportError::TooLarge);
+	futures_util::pin_mut!(body);
+
+	// don't allocate `max_body_size` if someone sets the `content length` to some huge number.
```
niklasad1, line 59:
```suggestion
	// only allocate up to 16KB initially
```

PR #992, opened 2023-01-30
fix: disable clippy for tests

niklasad1:
Because of https://github.com/rust-lang/rust-clippy/issues/9922 let's disable clippy for tests until that is fixed.

### DIFF
--- a/proc-macros/Cargo.toml
+++ b/proc-macros/Cargo.toml
@@ -15,7 +15,7 @@ proc-macro = true
 [dependencies]
 proc-macro2 = "1.0"
 quote = "1.0"
-syn = { version = "1.0", default-features = false, features = ["extra-traits", "full", "visit", "parsing"] }
+syn = { version = "1.0", default-features = false, features = ["extra-traits", "full", "visit", "parsing", "printing", "clone-impls", "proc-macro"] }
 proc-macro-crate = "1"
 heck = "0.4.0"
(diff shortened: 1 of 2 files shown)


alvicsam (approved):

lexnv (approved):
LGTM! I'm wondering if we were to apply `clippy --fix` is it would still crash the clippy? 🤔

jsdw (approved):

proc-macros/Cargo.toml
```diff
@@ -15,7 +15,7 @@ proc-macro = true
 [dependencies]
 proc-macro2 = "1.0"
 quote = "1.0"
-syn = { version = "1.0", default-features = false, features = ["extra-traits", "full", "visit", "parsing"] }
+syn = { version = "1.0", default-features = false, features = ["extra-traits", "full", "visit", "parsing", "printing", "clone-impls", "proc-macro"] }
```
jsdw, line 18: Interesting!

niklasad1:
> LGTM! I'm wondering if we were to apply clippy --fix is it would still crash the clippy? thinking

Yes, I tried earlier :(

proc-macros/Cargo.toml
niklasad1: These features leaked from the workspace but not sure why it's was catched now

PR #994, opened 2023-01-30
client: feature gate tls cert store

niklasad1:
Currently, the clients will include dependencies for both `webpki` and `native` certificate store which can be configured at runtime but that leads to a bigger dependency tree.

IIRC, I did it like this to avoid having "mutually exclusive feature" but I looked at the reqwest API and having these APIs behind a feature flag looks clean to me.

No default features implies ->  no TLS


lexnv (approved):

client/http-client/Cargo.toml
```diff
@@ -28,8 +28,11 @@ jsonrpsee-test-utils = { path = "../../test-utils" }
 tokio = { version = "1.16", features = ["net", "rt-multi-thread", "macros"] }
 
 [features]
-default = ["tls"]
-tls = ["hyper-rustls/webpki-tokio"]
+default = ["native-tls"]
+native-tls = ["hyper-rustls/native-tokio", "__tls"]
```
niklasad1, line 32: we can bikeshed on the naming of these features.

technically it only configures native vs webpki certificate store but without them no tls is configured.
I don't want to call them `native-cert` or something.

more accurate would be native-rustls and webpki-rustls but it's a bit hard to understand.

client/http-client/src/transport.rs
```diff
@@ -71,9 +55,10 @@ impl HttpTransportClient {
 		}
 
 		let client = match target.scheme_str() {
-			Some("http") => HyperClient::Http(Client::new()),
-			#[cfg(feature = "tls")]
-			Some("https") => {
+			#[cfg(not(feature = "__tls"))]
+			Some("http") => Client::new(),
+			#[cfg(all(feature = "__tls", feature = "native-tls", feature = "webpki-tls"))]
+			Some("https") | Some("http") => {
```
niklasad1, line 61: for simplicity, just use connector based on the feature flag and yes the https connector is a bit slower on plain http than http connector.

client/http-client/Cargo.toml
```diff
@@ -28,8 +28,11 @@ jsonrpsee-test-utils = { path = "../../test-utils" }
...
+webpki-tls = ["hyper-rustls/webpki-tokio", "__tls"]
+
+__tls = ["hyper-rustls"]
```
jsdw, line 35: I'm happy with this for clarity, but just noting that as it stands, `hyper-rustls`will automatically be bought in when `native-tls` or `webpki-tls` is enabled without needing it :)

client/http-client/src/transport.rs
```diff
@@ -71,9 +55,10 @@ impl HttpTransportClient {
 		}
 
 		let client = match target.scheme_str() {
-			Some("http") => HyperClient::Http(Client::new()),
-			#[cfg(feature = "tls")]
-			Some("https") => {
+			#[cfg(not(feature = "__tls"))]
+			Some("http") => Client::new(),
+			#[cfg(all(feature = "__tls", feature = "native-tls", feature = "webpki-tls"))]
```
jsdw, line 60: Maybe I'm misunderstanding but this looks like it requires _both_ of those `-tls` features to be enabled (`all`)?

jsdw: Ooohhh no ok, you handle the individual oens below, got it!

jsdw (approved):
Nice, LGTM! (personally I have no opinion or issue with the feature name choices!)

PR #995, opened 2023-01-30
[ci] Change runners for CI

alvicsam:
PR changes ci runners for dynamically created ones. Also it changes runner for benchmark, so the next run will have different results and a new baseline.

cc https://github.com/paritytech/ci_cd/issues/730

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -14,11 +14,10 @@ stages:
   - check_benchmark
   - publish
 
-variables:                         &default-vars
+variables:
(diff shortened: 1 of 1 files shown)


lexnv (approved):

niklasad1:
> Also it changes runner for benchmark, so the next run will have different results and a new baseline.

That's a bummer but I guess we have to build a new baseline

niklasad1 (approved):

mateo-moon (approved):

PR #1012, opened 2023-02-13
client: rename `max_notifs_per_subscription` to `max_buffer_capacity_per_subscription`

niklasad1:
I had a look again at client after the recent changes in the server and concluded that `max_notifs_per_subscription` is a bad name. It only applies to the capacity so I think this is a good change

### DIFF
--- a/core/src/client/mod.rs
+++ b/core/src/client/mod.rs
@@ -217,7 +217,7 @@ pub enum SubscriptionKind {
 ///
 /// It will try to `unsubscribe` in the drop implementation
 /// but it may fail if the underlying buffer is full.
-/// Thus, if you want to ensure it's actually unsubscribe then
+/// Thus, if you want to ensure it's actually unsubscribed then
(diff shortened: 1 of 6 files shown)


jsdw (approved):

lexnv (approved):

core/src/client/mod.rs
```diff
@@ -216,8 +215,10 @@ pub enum SubscriptionKind {
 
 /// Active subscription on the client.
 ///
-/// It will automatically unsubscribe in the [`Subscription::drop`] so no need to explicitly call
-/// the `unsubscribe method` if it is an an subscription based on [`SubscriptionId`].
+/// It will try to `unsubscribe` in the drop implementation
+/// but it may fail if the underlying buffer is full.
+/// Thus, if you want to ensure it's actually unsubscribe then
```
niklasad1, line 220:
```suggestion
/// Thus, if you want to ensure it's actually unsubscribed then
```

PR #1013, opened 2023-02-13
rpc_module: `SubscriptionSink::close` -> `SubscriptionSink::close_with_error` 

### LINKED ISSUE #1009: rpc module: split SubscriptionSink::close to Subscription::close_with_error and Subscription::close_with_success

niklasad1:
Close #1009

### DIFF
--- a/core/src/server/rpc_module.rs
+++ b/core/src/server/rpc_module.rs
@@ -1114,7 +1114,7 @@ impl SubscriptionSink {
 	///
 	/// This can be used to signal that an subscription was closed because of some particular state
 	/// and doesn't imply that subscription was closed because of an error occurred. Just
-	/// a custom way to indicate the client that subscription was closed.
+	/// a custom way to indicate to the client that the subscription was closed.
 	///
 	/// If you'd like to to close the subscription without sending an extra notification,
 	/// just drop it and don't call this method.
--- a/core/src/server/rpc_module.rs
+++ b/core/src/server/rpc_module.rs
@@ -39,7 +39,7 @@ use crate::{SubscriptionCallbackError, SubscriptionResult};
 use futures_util::future::Either;
 use futures_util::{future::BoxFuture, FutureExt};
 use jsonrpsee_types::error::{CallError, ErrorCode, ErrorObject, ErrorObjectOwned};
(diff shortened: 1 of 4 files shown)


jsdw (approved):
Makes sense to me!

I'm not entirely sure we need `close_with_result_notif` (and as a nit I might have gone for `close_with_error` instead of `close_with_error_notif`) but all good :)

niklasad1:
> I'm not entirely sure we need close_with_result_notif (and as a nit I might have gone for close_with_error instead of close_with_error_notif) but all good :)

The reason I did it like that is `close_with_error` isn't necessarily an error it may be like the subscription could be defined like without no message when it's "completed" for some reason. However, I would happy to remove those annoying names if that's clear enough now

core/src/server/rpc_module.rs
```diff
@@ -1081,55 +1095,78 @@ impl SubscriptionSink {
...
-		))
-		.map_err(Into::into)
-	}
-
-	/// Close the subscription, sending a notification with a special `error` field containing the provided error.
+	/// Close the subscription, sending a notification with a special `error` field containing the provided close reason.
+	///
+	/// This can be used to signal that an subscription was closed because of some particular state
+	/// and doesn't imply that subscription was closed because of an error occurred. Just
+	/// a custom way to indicate the client that subscription was closed.
```
niklasad1, line 1117:
```suggestion
	/// a custom way to indicate to the client that subscription was closed.
```

lexnv (approved):

examples/examples/ws_pubsub_with_params.rs
```diff
@@ -111,40 +111,42 @@ async fn run_server() -> anyhow::Result<SocketAddr> {
...
 					Err(TrySendError::Full(_)) => (),
 				}
 			}
 		}
-	}
-
-	sink.close(ErrorObject::owned(1, "Ok", None::<()>)).await;
+	};
+
+	// NOTE: we are using `close_with_err_notifor` for the jsonrpsee client to terminate the subscription
```
niklasad1, line 143:
```suggestion
	// NOTE: we are using `close_with_error` for the jsonrpsee client to terminate the subscription
```

PR #1015, opened 2023-02-15
[RFC] proc-macros: Support custom parameter names

### LINKED ISSUE #1007: Overriding name of parameter in param_kind=map for proc macro
Hi,

In my program which uses the jsonrpsee http client, I have a problem: my JSON-RPC server expects an input param with the name of, literally, `type`. Unfortunately, a variable cannot be named `type` in Rust because `type` is a language keyword.

See [here](https://github.com/allquixotic/encuum/blob/apps/src/applications.rs#L24) the open source code I'm working on with the problem. On line 24 here, I have to name the parameter `_type`, but that isn't what needs to be sent in the request: the request should send `type` as normal.

Is there a way to continue using the rpc proc macro with this naming convention for an input parameter, or do I need to use the client "late binding" (via serde_json::Value) and not benefit from the convenience of the proc macro?

The solution I envision would be something like specifying some kind of macro in front of the parameter that needs to be renamed. Like:

`#[custom_serial(name="type")]` (also would work with `name="fn"`, etc.) 

However, `#[serde(rename="something")]` is only for fields on structs from what I can tell...

allquixotic:
Should resolve #1007.

I've tested this with the HTTP client and it works for me. 

Comments:

 - I'm not sure if this is the best way to expose the interface in the `method` and `subscription` macros. I don't feel confident that this will be well-received. I initially prototyped a different version that accepted a single String parameter with a comma-separated list of name=value, but that seemed even more error-prone and harder to understand.
 - I didn't implement this yet for `render_server.rs` because it seems like the code is pretty hairy (I don't understsand it, to be honest) and I didn't want to put in all the effort to do it if my approach isn't viable to begin with.

# Status

 - [x] Implementation for `method` macro
 - [x] Implementation for `subscription` macro
 - [x] Implementation for `render_client`
 - [ ] Implementation for `render_server`
 - [x] Documentation update
 - [ ] Unit tests update
 - [ ] Examples update (in examples/ dir, not just in the docs)

### DIFF
--- a/proc-macros/src/render_client.rs
+++ b/proc-macros/src/render_client.rs
@@ -1,5 +1,3 @@
-use std::collections::HashMap;
-
 // Copyright 2019-2021 Parity Technologies (UK) Ltd.
 //
 // Permission is hereby granted, free of charge, to any
@@ -138,8 +136,7 @@ impl RpcDescription {
 		};
 
 		// Encoded parameters for the request.
-		let parameter_builder =
-			self.encode_params(&method.params, &method.param_kind, &method.signature, &method.name_map);
+		let parameter_builder = self.encode_params(&method.params, &method.param_kind, &method.signature);
 		// Doc-comment to be associated with the method.
 		let docs = &method.docs;
 		// Mark the method as deprecated, if previously declared as so.
@@ -175,7 +172,7 @@ impl RpcDescription {
 		let returns = quote! { Result<#sub_type<#item>, #jrps_error> };
 
 		// Encoded parameters for the request.
-		let parameter_builder = self.encode_params(&sub.params, &sub.param_kind, &sub.signature, &sub.name_map);
+		let parameter_builder = self.encode_params(&sub.params, &sub.param_kind, &sub.signature);
 		// Doc-comment to be associated with the method.
 		let docs = &sub.docs;
 
@@ -194,7 +191,6 @@ impl RpcDescription {
 		params: &[(syn::PatIdent, syn::Type)],
 		param_kind: &ParamKind,
 		signature: &syn::TraitItemMethod,
-		name_map: &HashMap<String, String>,
 	) -> TokenStream2 {
 		let jsonrpsee = self.jsonrpsee_client_path.as_ref().unwrap();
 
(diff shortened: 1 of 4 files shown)


niklasad1:
All right, I agree that the proc macros are complex enough as it is currently.

It would be easier to understand if we would have attributes on the parameters itself I think.
However, as it this is quite uncommon usecase it should be sufficient to have raw identifiers as workaround which could strip in macros.

Thus, "r#type" -> "type"

allquixotic:
Okay; I went ahead and updated my branch with an extremely simple implementation of removing the `r#` of raw types when it's detected in the client.

proc-macros/src/render_client.rs
```diff
@@ -206,7 +206,10 @@ impl RpcDescription {
 				let param_names = extract_param_names(&signature.sig);
 				// Combine parameter names and values to pass them as parameters.
 				let params_insert = param_names.iter().zip(params).map(|pair| {
-					let name = pair.0;
+					let mut name = pair.0.to_owned();
+					if name.starts_with("r#") {
```
niklasad1, line 210: you could use https://doc.rust-lang.org/std/string/struct.String.html#method.strip_prefix here instead

niklasad1:
Just add it for the server as well and it should good to go :)

allquixotic:
How would this be done for the server? I'm having a tough time seeing where in the server code this would belong. I guess it's most important in the _response_ data, right? We can assume that inbound data from the _request_ is correct, and just strip `r#` from any struct members in the response. But it's still not clear to me how to walk the "tree" of structs of structs (of structs, etc.) in the return value of a given call in order to look for things to strip the `r#` off of.

Also, I realized this is a breaking change. If someone literally wanted a parameter or result member to be called `r#type`, they would have to call it `r#r#type` (if Rust even supports that). I doubt this would break anyone, but there's always that chance ( https://xkcd.com/1172/ ).

niklasad1:
> How would this be done for the server? I'm having a tough time seeing where in the server code this would belong. I guess it's most important in the response data, right? We can assume that inbound data from the request is correct, and just strip r# from any struct members in the response. But it's still not clear to me how to walk the "tree" of structs of structs (of structs, etc.) in the return value of a given call in order to look for things to strip the r# off of.

The server has a similar trait where the identifiers from their are used for object params. So `type` won't work there as well similar to what you did for the client.

> Also, I realized this is a breaking change. If someone literally wanted a parameter or result member to be called r#type, they would have to call it r#r#type (if Rust even supports that). I doubt this would break anyone, but there's always that chance ( https://xkcd.com/1172/ ).

Yeah, I'm fine breaking that :) Happy to argue that this is limitation to rust identifiers and technically dept to this library....

> Do we also need to support going from "result": { "type": "a"} to struct Foo { r#type: String } in the client?

I don't get this question, it's already handled by your code in proc macros or am I missing something?

niklasad1:
Superseeded by #1365

PR #1018, opened 2023-02-16
refactor rpc_module

### LINKED ISSUE #1014: Get connection ID in rpc handler
So that it is possible to rate limit message per seconds per connection

xlc:
Breakdown this 1200+ LOC file into multiple smaller file to make it easier to maintain.

Also makes it is possible to construct `Methods` directly without using `RpcModule`, which closes #1014. 

I don't understand why it was coded this way that `RpcModule` only expose limited parameters from `MethodCallback` and make `Methods` not directly usable.


core/src/server/rpc_module/types.rs
```diff
@@ -0,0 +1,389 @@
...
+	Async(AsyncMethod<'static>),
+	/// Subscription method handler.
+	Subscription(SubscriptionMethod<'static>),
+	/// Unsubscription method handler.
+	Unsubscription(UnsubscriptionMethod),
+}
+
+/// Method callback wrapper that contains a sync or async closure,
+#[derive(Clone, Debug)]
+pub struct MethodCallback {
```
xlc, line 271: What's the purpose of this? It adds no functionality over `MethodKind`

xlc:
the test failing seems like due to upstream changes and unrelated to this PR?

niklasad1:
yes, https://github.com/rust-lang/rust/issues/107950 I thought it would be fixed by now but just ignore it

niklasad1 (commented):
I don't understand what this PR accomplishes, can you please add a doc test in the code or example in the PR description. I could only see that Methods have `pub(crate) callbacks` but how does that help?

In addition, as this review comes up with 1300 LoC can you please highlight the code you have changed?

In the future it would much easier if your break up your PR in different commits and save us lots of time to see what is pure refactoring vs code changes.

core/src/server/rpc_module/types.rs
niklasad1: it's simplify a convenience wrapper of methods calls, it could be removed if it helps but it's quite readable is it I reckon.

niklasad1:
> I don't understand why it was coded this way that RpcModule only expose limited parameters from MethodCallback and make Methods not directly usable.

This abstraction is written to be used for the jsonrpsee server for RPC method calls and not really exposed in the server APIs so I think it's expected if some things are private etc, as it's not used elsewhere what I'm aware of.

xlc:
The main goal is to make the code easier to deal with. Large file is usually a code smell. 

The secondary goal is to expose Methods, which is useful and I want to use it.

core/src/server/rpc_module/methods.rs
```diff
@@ -0,0 +1,297 @@
...
+		if self.callbacks.contains_key(name) {
+			return Err(Error::MethodAlreadyRegistered(name.into()));
+		}
+
+		Ok(())
+	}
+
+	/// Inserts the method callback for a given name, or returns an error if the name was already taken.
+	/// On success it returns a mut reference to the [`MethodCallback`] just inserted.
+	pub fn verify_and_insert(
```
xlc, line 68: there are no code change at all. this is the only public visibility change. other pub(crate) are only to make it compile.

core/src/server/rpc_module/types.rs
xlc: I consider it is unnecessary abstraction and only makes the code harder to reason. The less code, the easier to understand.

jsdw:
@xlc It would be significantly better to discuss this sort of change as an issue before opening up a PR in, so that we can actually discuss it and reach a consensus on whether we want this change in the first place (eg in this case, we all have different opinions on what constitutes good code, and as the maintainers of the library it is ultimately up to us to decide which style we want to use). By jumping straight to a PR you run the risk that we will disagree and your time will be wasted.

xlc:
This PR only takes half hour so no hard feelings if you don’t want to accept it. It is just moving code around and adjust imports and fix visibility. It sometimes quicker to just show some code instead of trying to explain it. 

Anyway but at least to me personally, the rpc_module code isn’t easy to understand due to complex coupling between struct s (RpcModule and Methods). Separate them will help catch code smells (eg all the pub crate) and allow better modulation of code. As the library user, I will the. be able to access lower level features (Methods) when higher level code doesn’t meet my requirements.

xlc:
Let me know what concerns you have about this PR. I plan to make more large-ish refactors in future and I am trying to do them in an incremental way to reduce review work. But for changes like this just moving code around, it is not easy to not make it be a big diff.

niklasad1 (changes requested):
I'm not against splitting up the rpc_module into smaller files but this code organization doesn't make sense to me.

methods: isn't coupled to the `rpc module` so it shouldn't be part of the rpc module if we want this exposed. would be better to have in the server mod instead.

also, we could probably move the subscription related types into a separate mod and the same with error related types.

types: move out the subscription and error types then maybe merge those with server/helpers.

then some small documentation on the modules to explain how these relates to each other.

xlc:
I did this way to minimize the required changes. But yeah I agree the file structure in this PR isn't the best.
Also Methods is coupled to the RpcModule with current code and `pub(crate)` indicates it.

I see few path forwards:
- Close this one. Let you guys do this refactor.
- Close this one. Do other refactor first. (e.g. remove `MethodCallback`)
- Merge this one. Then do other refactor. This shouldn't break anything (as test indicates) and IMO an improvement. It is not perfect but that doesn't need to be a blocker.

Let me know what do you think.

niklasad1:
I think it's better to close this PR and split it into two different PRs:

1. Make make things public on Methods and keep everything in the same file (i.e, no refactoring)
2. Do pure refactoring into smaller modules/files

This way it will be easier to review and you don't have create such big PRs

PR #1021, opened 2023-02-16
fix server: don't send error on JSON-RPC notifications

### LINKED ISSUE #1017: JSON RPC notification
Not that I need this feature at all but reading the code makes me think the notification feature implementation isn't really correct.

Firstly, notifications are completely ignored (if handled). I don't see any interface exposed by jsonrpsee to allow subscription of notifications.

Secondly, the websocket server handles notifications inconsistently. For batch request, it will parse notifications and suppress responses. However, for single request, it simply error out with parse error, which I don't think is not spec confirming.

niklasad1:
After some refactoring I introduced a bug where the notifications were not handled properly and were treated as JSON-RPC errors.

Also in the batch responses `Notification<T>` was used such that `missing params<T>` became a serde decoding error when it was missing and needs to be handled as `Option<serde_json::RawValue>` instead.

Close #1017

### DIFF
--- a/server/src/tests/ws.rs
+++ b/server/src/tests/ws.rs
@@ -650,6 +650,22 @@ async fn batch_with_mixed_calls() {
 }
 
 #[tokio::test]
+async fn batch_notif_without_params_works() {
+	init_logger();
+
+	let addr = server().with_default_timeout().await.unwrap();
+	let mut client = WebSocketTestClient::new(addr).with_default_timeout().await.unwrap().unwrap();
+	// mixed notifications, method calls and valid json should be valid.
+	let req = r#"[
+			{"jsonrpc": "2.0", "method": "add", "params": [1,2,4], "id": "1"},
+			{"jsonrpc": "2.0", "method": "add"}
+		]"#;
+	let res = r#"[{"jsonrpc":"2.0","result":7,"id":"1"}]"#;
+	let response = client.send_request_text(req.to_string()).with_default_timeout().await.unwrap().unwrap();
+	assert_eq!(response, res);
+}
+
+#[tokio::test]
 async fn ws_server_backpressure_works() {
 	init_logger();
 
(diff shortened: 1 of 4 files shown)


server/src/transport/ws.rs
```diff
@@ -151,12 +153,19 @@ pub(crate) async fn process_batch_request<L: Logger>(b: Batch<'_, L>) -> Option<
...
+pub(crate) async fn process_single_request<L: Logger>(
+	data: Vec<u8>,
+	call: CallData<'_, L>,
+) -> Option<CallOrSubscription> {
+	tracing::info!("process_request: {:?}", data);
+
 	if let Ok(req) = serde_json::from_slice::<Request>(&data) {
-		execute_call_with_tracing(req, call).await
+		Some(execute_call_with_tracing(req, call).await)
+	} else if let Ok(_) = serde_json::from_slice::<Notif>(&data) {
```
niklasad1, line 164:
```suggestion
	} else if serde_json::from_slice::<Notif>(&data).is_ok() {
```

jsdw (approved):

server/src/tests/ws.rs
```diff
@@ -745,3 +761,12 @@ async fn ws_server_backpressure_works() {
...
 	assert!(seen_item_after_backpressure);
 }
+
+#[tokio::test]
+async fn notif_is_ignored() {
+	init_logger();
+	let addr = server().await;
+	let mut client = WebSocketTestClient::new(addr).with_default_timeout().await.unwrap().unwrap();
+
+	assert!(client.send_request_text(r#"{"jsonrpc":"2.0","method":"bar"}"#).with_default_timeout().await.is_err());
```
niklasad1, line 771: this is weird but should not be "answered" :)

lexnv (approved):

server/src/tests/ws.rs
jsdw: Maybe this is worth a quick comment about why it can be ignored since it looks like a method call that should be responded to?

PR #1025, opened 2023-02-17
rpc module: remove `SubscriptionAnswer`

niklasad1:
This abstraction is not very nice and I think it's more readable with Result<MethodResult, Id> anyway so let's remove it.


server/src/transport/ws.rs
```diff
@@ -153,12 +153,10 @@ pub(crate) async fn process_batch_request<L: Logger>(b: Batch<'_, L>) -> Option<
 	}
 }
 
-pub(crate) async fn process_single_request<L: Logger>(
+pub(crate) async fn process_single_request<'a, L: Logger>(
```
niklasad1, line 156:
```suggestion
pub(crate) async fn process_single_request<L: Logger>(
```

lexnv (approved):

jsdw (approved):
LGTM, one less type :)

PR #1030, opened 2023-02-24
Update format rule

boundless-forest:
The current import items are a bit of a mess.  Updating the format rule as the substrate repo.

- https://github.com/paritytech/substrate/blob/master/rustfmt.toml

No logic update, just format.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,11 +25,11 @@ jobs:
 
       - uses: actions/checkout@v3.3.0
 
-      - name: Install Rust stable toolchain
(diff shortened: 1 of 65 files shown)


niklasad1 (changes requested):
Hey, 

"The format rules" for imports are like this for a reason and what substrate is using a style-guide doesn't apply for this repo.

So, yes we are opinionated here and as a maintainer of repo this is the style is what I prefer.

I guess `wrap_comments = true` could be useful though.

In the future please open issues to discuss changes as many things may be rejected.

PR #1034, opened 2023-02-28
rpc module: refactor subscriptions to return `impl IntoSubscriptionResponse`

niklasad1:
The goal with this PR is to simplify the subscription API from the server-side of things.

Previously the subscriptions were changed to return `Result` for ergonomics but the errors were never used and only logged after the latest refactoring. 

One problem with the current API is that "result error bailing" is hard to get right and might cause the error not to sent the client.

This PR adds a new trait `IntoSubscriptionCloseResponse` which can be implemented for customized types and it is implemented for `Result<(), StringError> and ()` by jsonrpsee

### Result

Once an error occurs it's sent out a subscription error notification

```rust
    module
        .register_subscription("my_sub", "my_sub", "my_unsub", |_, pending, _| async move {
            let mut sink = pending.accept().await?;
            sink.send("foo").await?;
            Ok(())
        })
        .unwrap();
```

The outcome is that the proc macro APIs can't assume that the return type will be `Result<(), Error>` anymore instead it could be any type that implement `IntoSubscriptionResponse`


core/src/server/rpc_module.rs
```diff
@@ -838,14 +866,28 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 					let sub_fut = callback(params.into_owned(), sink, ctx.clone());
 
 					tokio::spawn(async move {
-						if let Err(SubscriptionCallbackError::Some(msg)) = sub_fut.await {
-							tracing::warn!("Subscribe call `{subscribe_method_name}` failed: {msg}");
+						match sub_fut.await {
+							Some(Err(msg)) if tx.is_closed() => {
```
niklasad1, line 870: disable to send out notifications on subscription that hasn't been accepted.

tests/tests/rpc_module.rs
```diff
@@ -424,10 +419,8 @@ async fn reject_works() {
 	module
 		.register_subscription("my_sub", "my_sub", "my_unsub", |_, pending, _| async move {
 			let err = ErrorObject::borrowed(PARSE_ERROR_CODE, &"rejected", None);
-			let res = pending.reject(err.into_owned()).await;
-			assert!(matches!(res, Ok(())));
-
-			Ok(())
+			pending.reject(err.into_owned()).await;
+			None
```
jsdw, line 423: If you return an error instead of calling `pending.reject`, what will happen? And if you return an error as well as calling it, what happens?

I guess I'd hope that `pending.reject()` can actually go away in favour of returning an error, so that there's exactly one way to return an error from a subscription which also guarantees it's ended (the callback has returned).

niklasad1: it will deny the subscription will a default error and the returned error will be ignored as it's not valid to send out notifications before the subscription call has been answered.

core/src/server/mod.rs
```diff
@@ -32,3 +32,48 @@ pub mod helpers;
...
+/// Extension trait that converts errors into an optional error message.
+///
+/// Internally it converts an `Result<T, Error>` into `Result<T, Option<Error>`
+/// which forces users to convert `Error` to `Option<SubscriptionMessage>`
+/// where `Some(msg)` represents that the message is sent as subscription notification error
+/// and `None` doesn't do anything.
+///
+/// This is implemented for the types in jsonrpsee where the behavior is not application dependent.
+/// For other types you have to implement this trait or deal with it manually.
+pub trait MapSubscriptionError<T> {
```
niklasad1, line 52: I'm not sure that we should have this trait, it works good for user-defined types but for other types from jsonrpsee such as `TrySendError`, `SendTimeoutError` then folks has to move them into a new type to not bump orphan rule issues.

niklasad1: also to `map_err` works now so perhaps better just to have that

tests/tests/rpc_module.rs
niklasad1: hmm, good point we could actually take the error and call reject on the hood.

that's a nice API improvement IMHO

niklasad1: it's a bit tricky to implement with the middleware but I'll like to tackle it in another PR for sure.

core/src/server/mod.rs
jsdw: Mmm I'm not super keen on this trait; if I was going for any trait it'd be something like:

```rust
trait IntoResponse {
    fn into_response(self) -> SubscriptionCloseResponse
}

impl IntoResponse for Option<T> where T: Into<SubscriptionMessage> {
    fn into_response(self) ->  SubscriptionCloseResponse {
         // None means nothing returned, Some(msg) means a non-error msg is returned
    }
}

impl IntoResponse for Result<T, E> {
    fn into_response(self) ->  SubscriptionCloseResponse {
          // Err(e) means error returned, Ok(m) means a non-error msg is returned
    }
}
```

But should `Ok(())` and `None` mean that no message is returned at all, or with `Ok(())` would you get back `[]` as a message? It feels like it would be safer if no message came back (otherwise using `Result` there is no way _not_ to return some message if all goes ok), and you just have to explicitly send one yourself (or register one to be sent on the sink).

So I guess I'd only support `Result<(),IntoError>` and `Option<()>`, and then we can ergonomically return errors using `Result` and `?` or ergonomically return but emit no errors with `Option` and `None`.

So one could write:

```rust
module.register_subscription("my_sub", "my_sub", "my_unsub", |_, pending, _| async move {
	// Do stuff; can use ? to return errors here
    pending.accept().await?;

	Ok(())
})
```

Or one can opt to not return errors like this:

```rust
module.register_subscription("my_sub", "my_sub", "my_unsub", |_, pending, _| async move {
	// Due to return type, need to `.ok()` explicitly to throw away errors.
    pending.accept().await.ok()?;

	None
})
```

And maybe if you want to send a default closing message some api like:

```rust
module.register_subscription("my_sub", "my_sub", "my_unsub", |_, pending, _| async move {
    let mut sink = pending.accept().await.ok()?;

    // register message to be sent when closes (presumably if an error is returned because we
    // allow this via `Result`, we will either send this just prior to the error, or not send it at all 
    // (don't know that it matters which one!):
    sink.send_on_close(|| SomeMessage);

	None
})
```

But if all of that is sortof annoying or too much effort, then the alternative which I also think is perfectly reasonable is to only return `()` (and force error handling to be done by the user), or only return `Result<(), SomeError>` to force the user to do `pending.accept().await? else { /* handle error manually each time */ }` but make it easy if the user wants to return errors by default.

Really, the only thing I don't like about the currently released interface is that you can return the error and then it's silently discarded, which I think may be surprising.

jsdw: I guess the current API you've settled on without the trait is pretty much equivalent to returning `()` in terms of usability, except for allowing the user to return some message (or error I think?) if they like. So I'm also happy with that offhand if that's what you prefer!

niklasad1: The reason why I didn't go down the `into_response` route is because it didn't work with `?` but maybe I did something wrong when I tried it.

niklasad1: so yeah, I got it working as well but as the type inference forces the `async block` to return `Result or Option` with the design which makes it not possible to ignore some error but in another scenario return an error if that makes sense :)

[example here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d7429031e971510c4ffd00f347df433b)

jsdw: > which makes it not possible to ignore some error but in another scenario return an error in the same async block

I think you mean that you want some errors to be ignored (eg connection closed) when `?` is applied on them, while letting other errors propagate to the user?

`IntoResponse` could always expect an error enum to be returned and then "connection closed" style errors can be manually filtered out if they are seen? Or errors can impl `Into<ResponseError>` which would have the error info and also note whether it should be sent to the user or not (http libraries sometimes do similar stuff; sending 4xx errors to users and only logging 5xx errors on the server while sending the user some generic message)

But I might be missing the point so apologies if so!

niklasad1: Yeah, that's true maybe I'm a bit narrow minded but initially the idea was that `return Option == ignore error` and `return Err == always propagate the error` to make it explicit for the user.

Having what you suggested is nice but isn't really what I had in mind and easy to "bail early with ?`....

jsdw: > but initially the idea was that return Option == ignore error and return Err == always propagate the error to make it explicit for the user.

Yup that was definitely the thinking but I thought you were raising that not all errors should be returned by default, but I may have misunderstood!

niklasad1: yeah sorry I was more picking hair on the that you can't really return `Result and Option` from the substration callback and tried to elaborate that it's ok but yeah we could ignore some error but it doesn't really matter if the connection is closed anyway :P

I think I prefer your suggestion anyway, I'll give it a try again

server/src/tests/ws.rs
```diff
@@ -687,9 +695,9 @@ async fn ws_server_backpressure_works() {
 			"n",
 			"unsubscribe_with_backpressure_aggregation",
 			move |_, pending, mut backpressure_tx| async move {
-				let sink = pending.accept().await?;
-				let n = SubscriptionMessage::from_json(&1).unwrap();
-				let bp = SubscriptionMessage::from_json(&2).unwrap();
+				let sink = pending.accept().await.map_err(|e| anyhow::anyhow!("{:?}", e))?;
```
niklasad1, line 698:
```suggestion
				let sink = pending.accept().await?;
```

proc-macros/src/lib.rs
```diff
@@ -293,23 +293,21 @@ pub(crate) mod visitor;
...
 ///         }
 ///
 ///         // Send out two values on the subscription.
 ///         async fn sub(&self, pending: PendingSubscriptionSink) -> SubscriptionResult {
-///             let sink = pending.accept().await?;
+///             let sink = pending.accept().await.map_err(|_| None)?;
 ///
-///             let msg1 = SubscriptionMessage::from_json(&"Response_A")?;
-///             let msg2 = SubscriptionMessage::from_json(&"Response_B")?;
+///             let msg1 = SubscriptionMessage::from_json(&"Response_A").expect("Infallible");
```
jsdw, line 305: I don't get this yet; wouldn't `from_json` reject the invalid json "Response_A"?

proc-macros/src/lib.rs
```diff
@@ -293,23 +293,21 @@ pub(crate) mod visitor;
...
+///             // As &str is quite common and the serialization is infallible this is also possible to use.
+///             let msg2 = SubscriptionMessage::from("Response_B");
 ///
-///             sink.send(msg1).await?;
-///             sink.send(msg2).await?;
+///             sink.send(msg1).await.map_err(|_| None)?;
+///             sink.send(msg2).await.map_err(|_| None)?;
```
jsdw, line 310: Since we're just throwing away the message, does it make sense for the example to `map_err(|_| None)` everywhere insetad of returning some `anyhow::Error` or something?

benches/helpers.rs
```diff
@@ -151,11 +151,11 @@ pub async fn ws_server(handle: tokio::runtime::Handle) -> (String, jsonrpsee::se
 			SUB_METHOD_NAME,
 			UNSUB_METHOD_NAME,
 			|_params, pending, _ctx| async move {
-				let sink = pending.accept().await?;
+				let sink = pending.accept().await.unwrap();
 				let msg = SubscriptionMessage::from("Hello");
-				sink.send(msg).await?;
+				sink.send(msg).await.unwrap();
```
jsdw, line 156: Nit; is there any harm in using `?` here?

core/src/server/error.rs
```diff
@@ -0,0 +1,96 @@
...
+	#[error("Timed out waiting on send operation")]
+	Timeout(SubscriptionMessage),
+	/// The channel is full.
+	#[error("Closed")]
+	Closed(SubscriptionMessage),
+}
+
+/// The error returned while accepting or rejecting a subscription.
+#[derive(Debug, Copy, Clone, thiserror::Error)]
+pub struct SubscriptionAcceptRejectError;
```
jsdw, line 60: Nit: `SubscriptionAcceptRejectError` is a bit confusing to me; something like `PendingSubscriptionAcceptError` maybe? (it's an error accepting a pending subscription)

proc-macros/src/lib.rs
niklasad1: I haven't updated this file yet :)

core/src/server/error.rs
```diff
@@ -0,0 +1,96 @@
...
+
+impl std::fmt::Display for SubscriptionAcceptRejectError {
+	fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+		f.write_str("The remote peer closed the connection")
+	}
+}
```
jsdw, line 66: Since `thiserror` is being used, I think you could also add an `#[error("The remote peer closed the connection")]` to the type for the same effect (and for DisconnectError)

core/src/server/rpc_module.rs
```diff
@@ -721,39 +561,73 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
...
-	/// for a more ergonomic API and is not used (except logged for user-related caused errors).
-	/// By default jsonrpsee doesn't send any special close notification,
-	/// it can be a footgun if one wants to send out a "special notification" to indicate that an error occurred.
+	/// An async block which returns something that implements [`crate::server::IntoSubscriptionResponse`] which
+	/// decides what action to take once the subscription callback returns i.e, whether to sent out another message
+	/// on the subscription stream before closing down it.
+	///
+	/// This is implemented for `Result<T, E>` and `Option<T>`.
+	///
+	/// It's recommended to use `Result` if you want to propagate the error as special error notification
```
jsdw, line 570: I haven't checked yet, but if the subscription returns an error before `pending.accept()` does that also come back (but as an error to the subscription request rather than an error notification). Maybe the text should be more general and just say something like "Use `Result` if you want the error to be sent back to the user. Use `Option` if you'd like to avoid sending any errors back to the user."?

core/src/server/rpc_module.rs
```diff
@@ -721,39 +561,73 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
...
+	/// and `Option` if no error's should be propagated.
 	///
-	/// If you want to a special error notification use `SubscriptionSink::close` or
-	/// `SubscriptionSink::send` before returning from the async block.
+	/// Another option is to implement [`crate::server::IntoSubscriptionResponse`] if you want customized behaviour.
+	///
+	/// The error notification has the following format:
```
jsdw, line 575: If I'm right above, maybe also add a note on what the error looks like if the Subscription wasn't accepted yet?

core/src/server/rpc_module.rs
```diff
@@ -721,39 +561,73 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
...
 	///
 	/// let mut ctx = RpcModule::new(99_usize);
 	/// ctx.register_subscription("sub", "notif_name", "unsub", |params, pending, ctx| async move {
-	///     let x = params.one::<usize>()?;
 	///
-	///     // mark the subscription is accepted after the params has been parsed successful.
+	///     let x = match params.one::<usize>() {
+	///         Ok(x) => x,
+	///         Err(e) => {
+	///            pending.reject(ErrorObjectOwned::from(e)).await;
```
jsdw, line 603: I wondered whether a user could return an error for this too, but realised that you _have_ to return some sort of error for this, so yeah I see why it's separate and it makes sense to me!

(I guess 99% of the time people will just `pending.accept()` anyway!)

core/src/server/rpc_module.rs
jsdw: Ah I see; it is a different thing!

core/src/server/rpc_module.rs
jsdw: Ah nevermind; it's a different thing and the user has to `reject()` with an error, which makes sense, so ignore this :)

core/src/server/subscription.rs
```diff
@@ -0,0 +1,475 @@
...
+pub enum SubscriptionCloseResponse {
+	/// No further message will be sent.
+	None,
+	/// Send a ordinary subscription response.
+	Some(SubscriptionMessage),
+	/// Send a subscription error response.
+	Err(SubscriptionMessage),
+}
+
+impl<T> IntoSubscriptionResponse for Option<T>
```
jsdw, line 61: Nit: I feel like this name, `IntoSubscriptionResponse`, should be consistent with the name of the thing that's it converts stuff into (ie maybe `SubscriptionResponse` rather than `SubscriptionCloseResponse`)

core/src/server/subscription.rs
```diff
@@ -0,0 +1,475 @@
...
+pub struct SubscriptionKey {
+	pub(crate) conn_id: ConnectionId,
+	pub(crate) sub_id: SubscriptionId<'static>,
+}
+
+#[derive(Debug, Clone, Copy)]
+pub(crate) enum SubNotifResultOrError {
+	Result,
+	Error,
+}
```
jsdw, line 142: Nit: since `Result` is a thing with an `Ok` or `Err` variant already, maybe `Response, Error` makes more sense here than `Result, Error` for the variant names?

core/src/server/subscription.rs
```diff
@@ -0,0 +1,475 @@
...
+	pub(crate) subscribers: Subscribers,
+	/// Unique subscription.
+	pub(crate) uniq_sub: SubscriptionKey,
+	/// ID of the `subscription call` (i.e. not the same as subscription id) which is used
+	/// to reply to subscription method call and must only be used once.
+	pub(crate) id: Id<'static>,
+	/// Sender to answer the subscribe call.
+	pub(crate) subscribe: mpsc::Sender<MethodResponse>,
+	/// Subscription permit.
+	pub(crate) permit: SubscriptionPermit,
```
jsdw, line 203: Nit: I'm wondering how many of these need to be `pub(crate)` and whether we can hide a bit more

core/src/server/rpc_module.rs
niklasad1: This applies only to subscription notifications i.e, subscription messages and before that the subscription has been accepted we don't allow any messages to be sent. So those are ignored unless the `reject` is called explicitly. 

However, it would be nice to incorporate that in the callback and let remove `reject API`

TLDR: Returning stuff before `accept` has no effect.

niklasad1: So, I want to keep it out of this PR but indeed this is hairy stuff to understand.

core/src/server/subscription.rs
```diff
@@ -0,0 +1,475 @@
...
+	/// Reject the subscription call with the error from [`jsonrpsee_types::error::ErrorObject`].
+	pub async fn reject(self, err: impl Into<ErrorObjectOwned>) {
+		let err = MethodResponse::error(self.id, err.into());
+		_ = self.inner.send(err.result.clone()).await;
+		_ = self.subscribe.send(err).await;
+	}
+
+	/// Attempt to accept the subscription and respond the subscription method call.
+	///
+	/// Returns `None` if the connection is already closed, `Some(SubscriptionSink)` otherwise.
```
jsdw, line 216: Comment out of date

server/src/tests/ws.rs
```diff
@@ -404,11 +404,13 @@ async fn register_methods_works() {
 	assert!(module.register_method("say_hello", |_, _| Ok("lo")).is_ok());
 	assert!(module.register_method("say_hello", |_, _| Ok("lo")).is_err());
 	assert!(module
-		.register_subscription("subscribe_hello", "subscribe_hello", "unsubscribe_hello", |_, _, _| async { Ok(()) })
+		.register_subscription("subscribe_hello", "subscribe_hello", "unsubscribe_hello", |_, _, _| async {
+			Ok::<_, String>(())
```
jsdw, line 408: I guess these things could also be `Some(())` now to avoid type inference on error stuff?

core/src/server/rpc_module.rs
jsdw: Yeah makes sense; I can imagine that if the user returns an IntoResponse thing before calling `pending.accept()`, that maybe it could return the error they gave back (if a `Result` type was used) or just return some default error. And then maybe keep `pending.reject(explicit_error)` around so that even if the user wants to return `Option` from the subscription, they can still send a non-default error back here if they want.

But yeah, definitely can be a separate PR if wanted :)

benches/helpers.rs
niklasad1: no :)

I didn't change it after playing around with different approaches.

core/src/server/error.rs
niklasad1: nice catch, didn't know that it worked on structs

core/src/server/rpc_module.rs
```diff
@@ -474,7 +316,7 @@ impl Methods {
 			Some(MethodCallback::Async(cb)) => (cb)(id.into_owned(), params.into_owned(), 0, usize::MAX).await,
 			Some(MethodCallback::Subscription(cb)) => {
 				let conn_state =
-					ConnState { conn_id: 0, id_provider: &RandomIntegerIdProvider, subscription_permit: p1 };
+					SubscriptionState { conn_id: 0, id_provider: &RandomIntegerIdProvider, subscription_permit: p1 };
```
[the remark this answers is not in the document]
niklasad1: it's just to `mock` a `SubscriptionPermit` when it's used for tests/raw calls on RPC module without a "real transport".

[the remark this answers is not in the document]
niklasad1: hopefully more readable in https://github.com/paritytech/jsonrpsee/pull/1034/commits/438371b99873f691cd848ceb0b307cc68dbcb905

core/src/server/subscription.rs
niklasad1: it's instantiated from the `RpcModule` so needs to be `pub(crate)` or having `pub(crate) fn new`

core/src/server/helpers.rs
```diff
@@ -191,59 +189,6 @@ pub fn prepare_error(data: &[u8]) -> (Id<'_>, ErrorCode) {
...
-	/// Get the handle to [`tokio::sync::Notify`].
-	pub fn handle(&self) -> Arc<Notify> {
-		self.resource.clone()
-	}
-}
-
-/// Wrapper over [`tokio::sync::Notify`] with bounds check.
-#[derive(Debug, Clone)]
-pub struct BoundedSubscriptions {
-	resource: Arc<Notify>,
```
niklasad1, line 211: these `Arc<Notify>` are not used anymore

proc-macros/tests/ui/incorrect/sub/sub_unsupported_field.rs
```diff
@@ -4,7 +4,7 @@ use jsonrpsee::proc_macros::rpc;
 #[rpc(client, server)]
 pub trait UnsupportedField {
 	#[subscription(name = "sub", unsubscribe = "unsub", item = u8, magic = true)]
-	async fn sub(&self);
+	async fn sub(&self) -> RpcResult<()>;
```
jsdw, line 7: I guess you could `impl IntoSubscriptionResponse` for `()` too (ie no message or error sent back) if you want to allow empty return types.

niklasad1: yeah I was thinking the same, why not

core/src/server/error.rs
```diff
@@ -0,0 +1,91 @@
...
+// ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED
+// TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+// PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT
+// SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+// CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR
+// IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+// DEALINGS IN THE SOFTWARE.
+
+use crate::server::subscription::SubscriptionMessage;
```
niklasad1, line 27:
```suggestion
use crate::server::SubscriptionMessage;
```

core/src/server/rpc_module.rs
```diff
@@ -495,7 +336,7 @@ impl Methods {
 
 		tracing::trace!("[Methods::inner_call] Method: {}, response: {:?}", req.method, response);
 
-		(response, rx, p2)
+		(response, rx)
```
niklasad1, line 339: this permit was never used

core/src/server/error.rs
```diff
@@ -0,0 +1,91 @@
...
+	/// The data could not be sent because the timeout elapsed
+	/// which most likely is that the channel is full.
+	#[error("Timed out waiting on send operation")]
+	Timeout(SubscriptionMessage),
+	/// The channel is full.
+	#[error("Closed")]
+	Closed(SubscriptionMessage),
+}
+
+/// The error returned while accepting or rejecting a subscription.
```
niklasad1, line 58:
```suggestion
/// The error returned while accepting a subscription.
```

core/src/server/subscription.rs
niklasad1, line 1: subscription related types moved to this file

jsonrpsee/src/lib.rs
```diff
@@ -90,7 +90,7 @@ cfg_types! {
 }
 
 cfg_server! {
-	pub use jsonrpsee_core::server::rpc_module::{RpcModule, SubscriptionSink, PendingSubscriptionSink, SubscriptionMessage, DisconnectError, TrySendError, SendTimeoutError};
+	pub use jsonrpsee_core::server::*;
```
niklasad1, line 93: this is already re-exported by the server maybe not needed

proc-macros/src/lib.rs
```diff
@@ -292,27 +314,40 @@ pub(crate) mod visitor;
...
+///             let msg2 = SubscriptionMessage::from("Response_B");
 ///
 ///             sink.send(msg1).await?;
 ///             sink.send(msg2).await?;
 ///
 ///             Ok(())
 ///         }
+///
+///         // If one doesn't want sent out a close message when a subscription terminates or treat
+///         // error as subscription error notifications then it's possible to implement
```
niklasad1, line 337:
```suggestion
///         // errors as subscription error notifications then it's possible to implement
```

proc-macros/tests/ui/incorrect/sub/sub_unsupported_field.rs
niklasad1: implemented now

core/src/server/subscription.rs
```diff
@@ -0,0 +1,444 @@
...
+impl<T, E> IntoSubscriptionCloseResponse for Result<T, E>
+where
+	T: serde::Serialize,
+	E: ToString,
+{
+	fn into_response(self) -> SubscriptionCloseResponse {
+		match self {
+			Ok(msg) => match SubscriptionMessage::from_json(&msg) {
+				Ok(m) => SubscriptionCloseResponse::Some(m),
+				Err(e) => SubscriptionCloseResponse::Err(e.to_string().into()),
```
jsdw, line 72: Just checking; is this Err from attempting to serialize the message? I wonder whether we should send nothing back to the user and just log it in that case, since it's more of an internal error?

niklasad1: yeah, that error is useless anyway I think `serialization failed` :P

core/src/server/subscription.rs
```diff
@@ -0,0 +1,444 @@
...
+
+impl IntoSubscriptionCloseResponse for () {
+	fn into_response(self) -> SubscriptionCloseResponse {
+		SubscriptionCloseResponse::None
+	}
+}
+
+impl IntoSubscriptionCloseResponse for SubscriptionCloseResponse {
+	fn into_response(self) -> SubscriptionCloseResponse {
+		SubscriptionCloseResponse::None
```
jsdw, line 87: Shouldn't this return `self`?

core/src/server/error.rs
```diff
@@ -0,0 +1,91 @@
...
+/// The error returned while accepting a subscription.
+#[derive(Debug, Copy, Clone, thiserror::Error)]
+#[error("The remote peer closed the connection")]
+pub struct PendingSubscriptionAcceptError;
+
+impl std::fmt::Display for DisconnectError {
+	fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+		f.write_str("Closed")
+	}
+}
```
jsdw, line 67: Could add `#[error("Closed")]` to DisconnectError instead?

core/src/server/subscription.rs
```diff
@@ -0,0 +1,444 @@
...
+	}
+}
+
+impl Drop for Subscription {
+	fn drop(&mut self) {
+		self.close();
+	}
+}
+
+/// Wrapper [`tokio::sync::Semaphore`] for subscriptions.
```
jsdw, line 390:
```suggestion
/// Wrap [`tokio::sync::Semaphore`] for subscriptions.
```

examples/examples/ws_pubsub_broadcast.rs
```diff
@@ -116,20 +112,19 @@ async fn pipe_from_stream_with_bounded_buffer(
...
 
 				closed = c;
 			}
 
-			// stream is closed or some error, just quit.
-			Either::Right((_, _)) => {
-				break;
-			}
+			// Send back back the error.
+			Either::Right((Some(Err(e)), _)) => break Err(e.into()),
```
jsdw, line 122: It's nice that it's easy to return an error like this now!

jsdw (commented):
Looks awesome; just a couple of small comments/questions :)

server/src/tests/ws.rs
```diff
@@ -421,8 +431,12 @@ async fn register_methods_works() {
...
 	let mut module = RpcModule::new(());
 	assert!(matches!(
-		module
-			.register_subscription("subscribe_hello", "subscribe_hello", "subscribe_hello", |_, _, _| async { Ok(()) }),
+		module.register_subscription::<SubscriptionResult, _, _>(
+			"subscribe_hello",
+			"subscribe_hello",
+			"subscribe_hello",
+			|_, _, _| async { Ok(()) }
+		),
```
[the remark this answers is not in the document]
niklasad1: perhaps your idea is slightly better but as long we don't have `Ok::<_, Error>::(())` then I'm happy

core/src/server/subscription.rs
```diff
@@ -0,0 +1,444 @@
...
+	pub(crate) subscribers: Subscribers,
+	/// Unique subscription.
+	pub(crate) uniq_sub: SubscriptionKey,
+	/// ID of the `subscription call` (i.e. not the same as subscription id) which is used
+	/// to reply to subscription method call and must only be used once.
+	pub(crate) id: Id<'static>,
+	/// Sender to answer the subscribe call.
+	pub(crate) subscribe: mpsc::Sender<MethodResponse>,
+	/// Subscription permit.
+	pub(crate) permit: OwnedSemaphorePermit,
```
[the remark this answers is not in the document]
niklasad1: I could make it `pub(crate) new -> PendingSubscriptionSink` but I like this better with named initialization

core/src/server/subscription.rs
```diff
@@ -0,0 +1,444 @@
...
+	///
+	/// # Panics
+	///
+	/// Panics if the subscription response exceeded the `max_response_size`.
+	pub async fn accept(self) -> Result<SubscriptionSink, PendingSubscriptionAcceptError> {
+		let response =
+			MethodResponse::response(self.id, &self.uniq_sub.sub_id, self.inner.max_response_size() as usize);
+		let success = response.success;
+		self.inner.send(response.result.clone()).await.map_err(|_| PendingSubscriptionAcceptError)?;
+		self.subscribe.send(response).await.map_err(|_| PendingSubscriptionAcceptError)?;
```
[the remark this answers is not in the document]
niklasad1: Opened an issue for it and added some inline comments, hopefully more understandable now

core/src/lib.rs
```diff
@@ -58,16 +58,11 @@ cfg_client! {
...
-pub use error::{Error, SubscriptionAcceptRejectError, SubscriptionCallbackError};
+pub use error::Error;
 
 /// JSON-RPC result.
 pub type RpcResult<T> = std::result::Result<T, Error>;
 
-/// The return type of the subscription's method for the rpc server implementation.
-///
-/// **Note**: The error does not contain any data and is discarded on drop.
-pub type SubscriptionResult = Result<(), SubscriptionCallbackError>;
```
niklasad1, line 69: So, I removed this because folks have to convert their subscriptions it may be possible that they sent out subscription close errors if they doesn't read the docs carefully enough and just keep `Result<T, E>` in the subscriptions.

we could keep it as well.

core/src/server/subscription.rs
```diff
@@ -0,0 +1,486 @@
...
+	///    "subscription": "<subscriptionID>",
+	///    "error": <your msg>
+	///    }
+	///  }
+	/// }
+	/// ```
+	NotifErr(SubscriptionMessage),
+}
+
+impl<T, E> IntoSubscriptionCloseResponse for Result<T, E>
```
niklasad1, line 91: I think we `()` will be serialized as `null` which is annoying

examples/examples/ws_pubsub_broadcast.rs
niklasad1: yeah, the only annoyance is if you want a JSON value here.

Then you would need to `serde_json::to_string(&v)` and the `Subscription::from(String)` would add an extra `""kakak""`.... but that's sacrificed with the StdError/ToString here :(

core/src/server/subscription.rs
niklasad1: we could also implement that for `Result<(), Error>` instead

jsdw: As we said in the call, that sounds like a good idea to me!

I even wonder whether it makes sense for `Error` to be a concrete type (`?` will ".into()" other error types into it if possible anyway). This would remove the need to specify any generic types anywhere :)

core/src/server/subscription.rs
jsdw: Maybe the change isn't on github yet; just want to check that this is not missed :)

core/src/server/subscription.rs
niklasad1: What I had in mind with this is that you can also inject your own JSON here such as

```
result: { "my": 120, "hehe": { "1":1} }
```

if we put `Error` there that is not possible it would just be:

```
result: "<err.to_string()>"
```

however, `ToString` here doesn't work super well either I think we would add extra string quotes ("")

core/src/server/subscription.rs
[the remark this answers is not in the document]
niklasad1: right, we could limit that to the `rpc_module` to be even more strict

core/src/server/subscription.rs
niklasad1: sorry, I didn't realize that I wrote `Error` above I meant `E: ToString` :)

jsdw: > What I had in mind with this is that you can also inject your own JSON here such as
> 
> ```
> result: { "my": 120, "hehe": { "1":1} }
> ```
> 
> if we put `Error` there that is not possible it would just be:
> 
> ```
> result: "<err.to_string()>"
> ```
> 
> however, `ToString` here doesn't work super well either I think we would add extra string quotes ("") such if one does:
> 
> ```
> sink.send().await.map_err(|_| serde_json::to_string(&MyError).unwrap())?;
> ```
> 
> lemme create a test for that

`E` impl's `ToString` at the moment, but for instance if we had a concrete `StringError` type for instance that serialized to a string correctly, and add `impl <T: ToString> From<T> for StringError` then all `?`'s would continue to work (as long as the errors impl Display which they generally always do) and no generics would be needed.

Dunno; maybe that doesn't make sense?

core/src/server/error.rs
```diff
@@ -0,0 +1,85 @@
...
+// DEALINGS IN THE SOFTWARE.
+
+use crate::server::SubscriptionMessage;
+use tokio::sync::mpsc;
+
+/// Error that may occur during [`crate::server::SubscriptionSink::try_send`].
+#[derive(Debug, thiserror::Error)]
+pub enum TrySendError {
+	/// The channel is closed.
+	#[error("Closed")]
```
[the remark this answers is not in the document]
niklasad1: maybe I can get rid of those `SubscriptionMessage` there looks a bit weird and just convert them in the `SubscriptionSink impls` instead

core/src/server/subscription.rs
```diff
@@ -0,0 +1,475 @@
...
+		tracing::trace!("[Subscription::close] Notifying");
+		self.rx.close();
+	}
+
+	/// Get the subscription ID
+	pub fn subscription_id(&self) -> &SubscriptionId {
+		&self.sub_id
+	}
+
+	/// Returns `Some((val, sub_id))` for the next element of type T from the underlying stream,
```
lexnv, line 394: nit: Do we need to update the comment here? This returns Some(Ok((val, sub_id)) and does not panic anymore. Also, we could remove the `let res` below and return directly, I guess that was used for debugging :D

core/src/server/subscription.rs
```diff
@@ -0,0 +1,475 @@
...
+
+impl IntoSubscriptionCloseResponse for () {
+	fn into_response(self) -> SubscriptionCloseResponse {
+		SubscriptionCloseResponse::None
+	}
+}
+
+impl IntoSubscriptionCloseResponse for SubscriptionCloseResponse {
+	fn into_response(self) -> Self {
+		Self::None
```
jsdw, line 111:
```suggestion
		self
```

Surely? :D

proc-macros/tests/ui/incorrect/sub/sub_no_item.rs
```diff
@@ -4,7 +4,7 @@ use jsonrpsee::proc_macros::rpc;
 #[rpc(client, server)]
 pub trait NoSubItem {
 	#[subscription(name = "sub")]
-	async fn sub(&self);
+	async fn sub(&self) -> RpcResult<()>;
```
lexnv, line 7: nit: Should we add a `sub_no_item.stderr` for this and the others missing the `.stderr`?

proc-macros/src/render_server.rs
```diff
@@ -76,11 +76,6 @@ impl RpcDescription {
 			let subscription_sink: syn::FnArg = syn::parse_quote!(subscription_sink: #subscription_sink_ty);
 			let mut sub_sig = sub.signature.clone();
 
-			// For ergonomic reasons, the server's subscription method should return `SubscriptionResult`.
-			let return_ty = self.jrps_server_item(quote! { core::SubscriptionResult });
-			let output: ReturnType = parse_quote! { -> #return_ty };
-			sub_sig.sig.output = output;
```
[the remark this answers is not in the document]
niklasad1: No, if anything we should check if the type returns `impl IntoSubscriptionResponse`  but that is so tricky/impossible so why even bother?

The compile error is not that bad:

```bash
error[E0277]: the trait bound `std::option::Option<usize>: IntoSubscriptionCloseResponse` is not satisfied
  --> examples/examples/proc_macro.rs:38:1
   |
38 | #[rpc(server, client, namespace = "state")]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `IntoSubscriptionCloseResponse` is not implemented for `std::option::Option<usize>`
   |
   = help: the following other types implement trait `IntoSubscriptionCloseResponse`:
             ()
             Result<(), E>
             SubscriptionCloseResponse

```

core/src/server/subscription.rs
niklasad1: ah, nice catch

core/src/server/subscription.rs
niklasad1: no, it won't compile otherwise some lifetime issue without the let res binding ^^

proc-macros/tests/ui/incorrect/sub/sub_no_item.rs
niklasad1: we already got one and I think this happens before the `RpcResult` is checked but let's revert these UI tests

it should work for both `Result and ()` anyway

core/src/server/subscription.rs
niklasad1: ok, it won't work with the proc macro code basically it's not possible to have the assumption that `?` will work it could be `()` or some other type.

Thus, I could just think of something like:

```rust
  if ret_ty == "Result" {
                #into_sub_response::into_response(context.as_ref().#rust_method_name(subscription_sink, #params_seq).await.map_err(Into::into))
  } else {
    #into_sub_response::into_response(context.as_ref().#rust_method_name(subscription_sink, #params_seq).await
  }
```

niklasad1: Implemented something similar as we have for the client to check the return type ident but I'm not super happy with it.

It works but a little hacky but probably fine here.

jsdw: > ok, it won't work with the proc macro code basically it's not possible to have the assumption that ? will work it could be () or some other type.

Ah I didn't think it would impact the proc macro; I'd assuemd that any IntoSubscriptionCloseResponse thing would "just work" as a return type so it didn't matter. Just happens that for user facing code if they want to return a Result type then they can use "?"'s and it'll work out?

proc-macros/src/helpers.rs
```diff
@@ -193,9 +208,52 @@ pub(crate) fn extract_doc_comments(attrs: &[syn::Attribute]) -> TokenStream2 {
...
+
+	// Get the generic args eg the <T, E> in Result<T, E>.
+	let PathArguments::AngleBracketed(AngleBracketedGenericArguments { args, .. }) = &mut type_name.arguments else {
+		return false;
+	};
+
+	if type_name.ident == "Result" {
+		// Result<T, E> should have 2 generic args.
+		if args.len() != 2 {
+			panic!("Result type aliases are not allowed only `std::result::Result<T, E>` and `jsonrpsee::core::RpcResult` are supported");
```
lexnv, line 239: nit: Would it be possible to also include the `Span` of the return type in the error?

Maybe returing the `TokenStream` or even panicking with the span info 
```rust
return Err(syn::Error::new(ty.span(), "msg").to_compile_error());
```

We don't necessarily the return types path `std::result::Result<T, E>` maybe renaming to `Result<T, E>` or similar?

niklasad1: thanks, I'll revert his and return `Result<(), StirngError` instead I don't like this hacky code

niklasad1: it was more to show James what I meant

core/src/server/subscription.rs
niklasad1: fixed now

tests/tests/rpc_module.rs
```diff
@@ -502,3 +495,36 @@ async fn bounded_subscription_works() {
 		assert_eq!(item, exp);
 	}
 }
+
+#[tokio::test]
+async fn serialize_sub_error_adds_extra_string_quotes() {
```
niklasad1, line 500: @jsdw @lexnv 

This is a test to illustrate that `StringError` will be serialized with extra string quotes such as `"error":"{...}" even if the value is JSON

We could actually parse the string before adding the `"<string>"` but I came to the solution that is too opinionated and likely better for folks to implement their own type then.

lexnv: Yep, that does make sense! In the end, we will be wrapping the message (may it be a string or already a valid json) with "{ msg }". I'm guessing this is not the common behavior and we don't need to provide a sort of wrapped error to ensure we don't add the extra quotes.
I do wonder, do we have in substrate subscriptions that return directly a valid json for the error? I'd expect that to be unlikely

lexnv (approved):
LGTM! Amazing work here!

jsdw (approved):
Awesome stuff! :)

core/src/server/subscription.rs
```diff
@@ -0,0 +1,484 @@
...
+use jsonrpsee_types::{response::SubscriptionError, ErrorObjectOwned, Id, SubscriptionId, SubscriptionResponse};
+use parking_lot::Mutex;
+use rustc_hash::FxHashMap;
+use serde::{de::DeserializeOwned, Serialize};
+use std::{sync::Arc, time::Duration};
+use tokio::sync::{mpsc, oneshot, OwnedSemaphorePermit, Semaphore};
+
+/// Type-alias for subscribers.
+pub type Subscribers = Arc<Mutex<FxHashMap<SubscriptionKey, (MethodSink, mpsc::Receiver<()>)>>>;
+/// Subscription permit
```
niklasad1, line 42:
```suggestion
/// Subscription permit.
```

proc-macros/src/lib.rs
```diff
@@ -86,7 +86,7 @@ pub(crate) mod visitor;
 ///     fn sync_method(&self) -> String;
 ///
 ///     #[subscription(name = "subscribe", item = "String")]
-///     async fn sub(&self);
+///     async fn sub(&self) -> Result<(), Error>;
```
niklasad1, line 89:
```suggestion
///     async fn sub(&self) -> SubscriptionResult;
```

proc-macros/src/lib.rs
```diff
@@ -99,8 +99,8 @@ pub(crate) mod visitor;
 ///     async fn async_method(&self, param_a: u8, param_b: String) -> u16;
 ///     fn sync_method(&self) -> String;
 ///
-///     // Note that `pending_subscription_sink` and `SubscriptionResult` were added automatically.
-///     async fn sub(&self, pending: PendingSubscriptionSink) -> SubscriptionResult;
+///     // Note that `pending_subscription_sink` was added automatically.
+///     async fn sub(&self, pending: PendingSubscriptionSink) -> Result<(), Error>;
```
niklasad1, line 103:
```suggestion
///     async fn sub(&self, pending: PendingSubscriptionSink) -> SubscriptionResult;
```

PR #1035, opened 2023-03-01
rustfmt: wrap comments

niklasad1:


niklasad1:
ah, let's wait until this one is in stable then

PR #1036, opened 2023-03-01
core: remove unsed dependency arrayvec

niklasad1:


jsdw (approved):
I keep forgetting abotu this in other crates, but is it worth adding `#![deny(unused_crate_dependencies)]` to the top of the `lib.rs` on each of the jsonrpsee crates to make it obvious if a dep isn't used?

niklasad1:
Yupp, opened an issue for it :)

lexnv (approved):

jsdw:
> Yupp, opened an issue for it :)

(I said that, and then I tried making use of it on one of the `scale-*` crates and immediately hit an issue where it was unhappy about a dependency used in tests (trybuild); turns out it's not all that clever in some cases!)

PR #1040, opened 2023-03-06
Add `unused_crate_dependencies` 

### LINKED ISSUE #1038: add `deny unused crates lint` to the crates in this repo
I keep forgetting abotu this in other crates, but is it worth adding `#![deny(unused_crate_dependencies)]` to the top of the `lib.rs` on each of the jsonrpsee crates to make it obvious if a dep isn't used?

_Originally posted by @jsdw in https://github.com/paritytech/jsonrpsee/pull/1036#pullrequestreview-1320112805_

boundless-forest:
Try fix #1038

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -32,7 +32,7 @@ tokio-stream = { version = "0.1", optional = true }
 
 [features]
 default = []
-http-helpers = ["hyper", "futures-util", "tokio/macros"]
+http-helpers = ["hyper", "futures-util", "tokio/macros", "tokio/rt"]
 server = [
 	"futures-util/alloc",
 	"globset",
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -25,10 +25,9 @@ rand = { version = "0.8", optional = true }
 soketto = { version = "0.7.1", optional = true }
 parking_lot = { version = "0.12", optional = true }
 tokio = { version = "1.16", optional = true }
-wasm-bindgen-futures = { version = "0.4.19", optional = true }
+# wasm-bindgen-futures = { version = "0.4.19", optional = true }
 futures-timer = { version = "3", optional = true }
 globset = { version = "0.4", optional = true }
-http = { version = "0.2.7", optional = true }
 tokio-stream = { version = "0.1", optional = true }
 
 [features]
@@ -58,7 +57,7 @@ async-client = [
 async-wasm-client = [
 	"async-lock",
 	"client",
-	"wasm-bindgen-futures",
+	# "wasm-bindgen-futures",
 	"rustc-hash/std",
 	"futures-timer/wasm-bindgen",
 	"tokio-stream",
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -25,7 +25,7 @@ rand = { version = "0.8", optional = true }
 soketto = { version = "0.7.1", optional = true }
 parking_lot = { version = "0.12", optional = true }
 tokio = { version = "1.16", optional = true }
-# wasm-bindgen-futures = { version = "0.4.19", optional = true }
+wasm-bindgen-futures = { version = "0.4.19", optional = true }
 futures-timer = { version = "3", optional = true }
 globset = { version = "0.4", optional = true }
 tokio-stream = { version = "0.1", optional = true }
@@ -57,7 +57,7 @@ async-client = [
 async-wasm-client = [
 	"async-lock",
 	"client",
-	# "wasm-bindgen-futures",
+	"wasm-bindgen-futures",
 	"rustc-hash/std",
 	"futures-timer/wasm-bindgen",
 	"tokio-stream",
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -32,7 +32,7 @@ tokio-stream = { version = "0.1", optional = true }
 
 [features]
 default = []
-http-helpers = ["hyper", "futures-util", "tokio/rt", "tokio/macros"]
+http-helpers = ["hyper", "futures-util", "tokio/macros"]
 server = [
 	"futures-util/alloc",
 	"globset",
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -32,7 +32,7 @@ tokio-stream = { version = "0.1", optional = true }
 
 [features]
 default = []
-http-helpers = ["hyper", "futures-util"]
+http-helpers = ["hyper", "futures-util", "tokio/rt", "tokio/macros"]
 server = [
 	"futures-util/alloc",
 	"globset",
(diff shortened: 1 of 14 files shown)


core/Cargo.toml
```diff
@@ -28,12 +28,11 @@ tokio = { version = "1.16", optional = true }
 wasm-bindgen-futures = { version = "0.4.19", optional = true }
```
niklasad1, line 28:
```suggestion
```

niklasad1: remove in below in async-client to make the CI happy

core/Cargo.toml
```diff
@@ -28,12 +28,11 @@ tokio = { version = "1.16", optional = true }
...
 futures-timer = { version = "3", optional = true }
 globset = { version = "0.4", optional = true }
-http = { version = "0.2.7", optional = true }
 tokio-stream = { version = "0.1", optional = true }
 
 [features]
 default = []
-http-helpers = ["hyper", "futures-util"]
+http-helpers = ["hyper", "futures-util", "tokio/rt", "tokio/macros"]
```
niklasad1, line 35: `tokio/rt` is not needed here, it just used for tests

niklasad1, line 35:
```suggestion
http-helpers = ["hyper", "futures-util", "tokio/macros"]
```

core/Cargo.toml
```diff
@@ -66,8 +65,6 @@ async-wasm-client = [
 
 [dev-dependencies]
 serde_json = "1.0"
-tokio = { version = "1.16", features = ["macros", "rt"] }
```
niklasad1, line 69: revert this

core/src/http_helpers.rs
```diff
@@ -136,31 +136,26 @@ pub fn read_header_values<'a>(
 	headers.get_all(header_name)
 }
 
-#[cfg(test)]
```
niklasad1, line 139: Can you revert this? 

I prefer to have the tests in a test module

core/Cargo.toml
boundless-forest: I noticed the `wasm-bindgen-futures` is used under `#[cfg(all(feature = "async-wasm-client", target_arch = "wasm32"))]` this feature.

If I remove it from `async-client` as you suggest, will this break any functionality? 

```sh
$ cargo check --release --features async-wasm-client --target wasm32-unknown-unknown
    Checking jsonrpsee-core v0.16.2 (/home/bear/coding/jsonrpsee/core)
error[E0433]: failed to resolve: use of undeclared crate or module `wasm_bindgen_futures`
   --> core/src/client/async_client/mod.rs:210:3
    |
210 |         wasm_bindgen_futures::spawn_local(async move {
    |         ^^^^^^^^^^^^^^^^^^^^ use of undeclared crate or module `wasm_bindgen_futures`

For more information about this error, try `rustc --explain E0433`.
error: could not compile `jsonrpsee-core` due to previous error

```

niklasad1 (commented):
Nice

If this works it's w.r.t to the UI tests then definitely this is useful.

I'm a little concerned about `deny` though as it might break downstream users build if the rustc gets smarter and detects more unused deps. Similar to the reason `deny(warnings)` is a anti-pattern

Thus, I suggest running the CI with `RUSTFLAGS="-D unused_crate_dependencies"` instead and downgrade `#![deny(unused_crate_dependencies)]` to  `#![warn(unused_crate_dependencies)]`

core/Cargo.toml
boundless-forest: Okay. Let me have a try.

core/Cargo.toml
boundless-forest: If we revert this, after adding `unused_crate_dependencies`:

```rs
bear@cypress ~/c/j/core (bear-fix-unused-deps)> cargo test
   Compiling tokio v1.24.2
warning: unused macro definition: `cfg_async_client`
  --> core/src/macros.rs:29:14
   |
29 | macro_rules! cfg_async_client {
   |              ^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_macros)]` on by default

warning: `jsonrpsee-core` (lib) generated 1 warning
   Compiling jsonrpsee-core v0.16.2 (/home/bear/coding/jsonrpsee/core)
error: external crate `tokio` unused in `jsonrpsee_core`: remove the dependency or add `use tokio as _;`
  --> core/src/lib.rs:27:1
   |
27 | //! Shared utilities for `jsonrpsee`.
   | ^
   |
note: the lint level is defined here
  --> core/src/lib.rs:30:9
   |
30 | #![deny(unused_crate_dependencies)]
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^

warning: `jsonrpsee-core` (lib test) generated 1 warning (1 duplicate)
error: could not compile `jsonrpsee-core` due to previous error; 1 warning emitted
```

That's why I add the `tokio/macros` to the `http-helpers` feature.

core/src/http_helpers.rs
boundless-forest: Done

core/Cargo.toml
niklasad1: tokio/macros is correct but `rt` is just a dev-dependency IIRC

core/Cargo.toml
niklasad1: don't do it, it's a mess

lemme find a branch for it instead.

boundless-forest: Nice, I'll revert these changes after your branch merged.

core/Cargo.toml
niklasad1, line 69: sorry I meant that you could have `tokio = { version = "1.16", features = ["rt"] }`here and remove `tokio/rt` from `http-helpers`

boundless-forest: https://www.reddit.com/r/rust/comments/r5munx/comment/hmpgec2/?utm_source=share&utm_medium=web2x&context=3 Maybe I should give [cargo machete](https://blog.benj.me/2022/04/27/cargo-machete) a try.

niklasad1: nah, I looked at your code and it's `dev-dependencies` to leaks in `cargo hack` pushing a fix soon :)

niklasad1: Ok, I think `cargo machete` will give us less headache.

It looks like `dev-dependencies` doesn't work properly for us, basically we would have to `#[cfg(test)] pub use crate as _` for that to work.

> This lint is "allow" by default because it can provide false positives depending on how the build system is configured. For example, when using Cargo, a "package" consists of multiple crates (such as a library and a binary), but the dependencies are defined for the package as a whole. If there is a dependency that is only used in the binary, but not the library, then the lint will be incorrectly issued in the library.

:point_up: I guess that regards `tests` is treated as another binary and maybe it doesn't work

niklasad1:
Sorry @boundless-forest I looked at the conditional compilation stuff and that change is so big, better suited for another PR.

I think we can conclude that `deny_unused_crate_dependencies` doesn't work properly for us.
We could accept "the removed dependencies" as a standalone PR or if you want to add a CI check for `cargo machete` up to you.

Again sorry, I should I have investigated it further before you started work on it so TIL.

PR #1043, opened 2023-03-14
refactor: ErrorResponse::error_object API

### LINKED ISSUE #1042: Allow taking ownership of `ErrorObject` from `ErrorResponse`
Feature request:

Client-side users of `ErrorResponse` typically don't need the ID or jsonrpc2.0 string. Allow the user to take ownership of the error object and drop the ID and 2.0 string

The current alternative is `err_resp.error_object().to_owned().into_owned()`, which is obviously bad code

```
impl ErrorResponse {
    // consume self, drop id and jsonrpc2.0, return only the object
    pub fn take_object(self) -> ErrorObject {
        self.error
    }
}
```

niklasad1:
Close #1042

### DIFF
--- a/core/src/server/rpc_module.rs
+++ b/core/src/server/rpc_module.rs
@@ -36,8 +36,8 @@ use crate::id_providers::RandomIntegerIdProvider;
 use crate::server::helpers::{BoundedSubscriptions, MethodSink};
 use crate::traits::{IdProvider, ToRpcParams};
 use crate::{SubscriptionCallbackError, SubscriptionResult};
-use futures_util::future::Either;
-use futures_util::{future::BoxFuture, FutureExt};
+use futures_util::future::{BoxFuture, Either};
+use futures_util::FutureExt;
 use jsonrpsee_types::error::{CallError, ErrorCode, ErrorObject, ErrorObjectOwned};
 use jsonrpsee_types::response::SubscriptionError;
 use jsonrpsee_types::{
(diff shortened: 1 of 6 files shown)


core/src/server/rpc_module.rs
```diff
@@ -732,7 +738,7 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 	/// # Examples
 	///
 	/// ```no_run
-	///
+	///
```
lexnv, line 741:
```suggestion
	///
```

types/src/error.rs
```diff
@@ -63,11 +62,16 @@ impl<'a> ErrorResponse<'a> {
...
+	/// Get a reference to the [`ErrorObject`] of the error response.
+	pub fn as_error_object(&self) -> &ErrorObject {
 		&self.error
 	}
 
+	/// Consume the error response and extract the [`ErrorObject`].
+	pub fn into_error_object(self) -> ErrorObject<'a> {
+		self.error
+	}
+
```
jsdw, line 74: I wonder whether it's better to return `ErrorObject<'_>` or `&'_ ErrorObject<'a>` here; one could do this for instance:

```rust
/// Get the [`ErrorObject`] of the error response with a lifetime tied to self.
pub fn as_error_object(&self) -> ErrorObject<'_> {
	ErrorObject {
		code: self.error.code,
		message: StdCow::Borrowed(self.error.message.borrow()),
		data: self.error.data.as_ref().map(|d| StdCow::Borrowed(d.borrow())),
	}
}
```

But yeah I don't know! Looks good as is; just pondering :)

jsdw (approved):

lexnv (approved):

types/src/error.rs
niklasad1: good idea, I like it

niklasad1: We could just clone the `Cow` instead to get `'a` but maybe a bit weird in `fn as_error_object`

I think this should be what most users want to anyway,

jsdw: Cloning has the downside of _maybe_ allocating and maybe not depending on the original error state (if we knew the error was always borrowed I'd probably go for that). With your API it's explicit at least, so yeah I think it's all good the way you did it!

PR #1046, opened 2023-03-15
Make `jsonrpc` protocol version field in `Response` as `Option`

### LINKED ISSUE #1045: RPC `Response`: error while deserializing response for missing `jsonrpc` version field
I was using the lib for interacting with a JSON-RPC server that does not provide the `jsonrpc` protocol version in its responses (actually it sets this field to `null`) and received an error in deserializing them.
Looking at sources, I saw that the `Response` struct defines the `jsonrpc` version field as required, while checking in the JSON-RPC 2.0 specification, this field is not mandatory but is actually optional.
Here the spec: [JSON-RPC 2.0 Spec](https://www.jsonrpc.org/specification#response_object)

gnapoli23:
# Problem
JSON-RPC 2.0 specification does not require `jsonrpc` version in responses as a mandatory field: [JSON-RPC 2.0 Spec](https://www.jsonrpc.org/specification#response_object)
For RPC servers that put this value to `null`, the lib can't deserialize this responses, 'cause it actually requires this field to be present

# Solution
Make `jsonrpc` field in `Response` as `Option`.

Fixes #1045

### DIFF
--- a/types/src/response.rs
+++ b/types/src/response.rs
@@ -36,7 +36,7 @@ use serde::{Deserialize, Serialize};
 #[derive(Serialize, Deserialize, Debug)]
 pub struct Response<'a, T> {
 	/// JSON-RPC version.
-	pub jsonrpc: TwoPointZero,
+	pub jsonrpc: Option<TwoPointZero>,
 	/// Result.
 	pub result: T,
 	/// Request ID
@@ -47,7 +47,7 @@ pub struct Response<'a, T> {
 impl<'a, T> Response<'a, T> {
 	/// Create a new [`Response`].
 	pub fn new(result: T, id: Id<'a>) -> Response<'a, T> {
-		Response { jsonrpc: TwoPointZero, result, id }
+		Response { jsonrpc: Some(TwoPointZero), result, id }
 	}
 
 	/// Create an owned [`Response`].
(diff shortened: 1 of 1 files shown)


types/src/response.rs
```diff
@@ -93,14 +93,14 @@ mod tests {
...
 	#[test]
 	fn serialize_call_response() {
-		let ser = serde_json::to_string(&Response { jsonrpc: TwoPointZero, result: "ok", id: Id::Number(1) }).unwrap();
+		let ser = serde_json::to_string(&Response { jsonrpc: Some(TwoPointZero), result: "ok", id: Id::Number(1) }).unwrap();
 		let exp = r#"{"jsonrpc":"2.0","result":"ok","id":1}"#;
 		assert_eq!(ser, exp);
 	}
 
 	#[test]
 	fn deserialize_call() {
```
niklasad1, line 102: Can you add a similar test without the `jsonrpc version field`? 

I'm quite certain that it will work but doesn't hurt to have a test for it :)

niklasad1 (approved):

jsdw (approved):

types/src/response.rs
```diff
@@ -93,17 +93,34 @@ mod tests {
...
+		let ser =
+			serde_json::to_string(&Response { jsonrpc: Some(TwoPointZero), result: "ok", id: Id::Number(1) }).unwrap();
 		let exp = r#"{"jsonrpc":"2.0","result":"ok","id":1}"#;
 		assert_eq!(ser, exp);
 	}
 
+	#[test]
+	fn serialize_call_response_missing_version_field() {
+		let ser = serde_json::to_string(&Response { jsonrpc: None, result: "ok", id: Id::Number(1) }).unwrap();
+		let exp = r#"{"jsonrpc":null,"result":"ok","id":1}"#;
```
niklasad1, line 105:
```suggestion
		let exp = r#"{"result":"ok","id":1}"#;
```

jsdw: Skipping sounds good to me :)

gnapoli23:
Btw don't know why the CI failed for last commit but I tested in local and everything goes well. Maybe a re-run of the CI can be a better shot :)

niklasad1:
> Btw don't know why the CI failed for last commit but I tested in local and everything goes well. Maybe a re-run of the CI can be a better shot :)

Don't worry the gitlab pipeline has an old rustc version running, we should update it soon that's why it fails

PR #1057, opened 2023-03-27
 add `IntoResponse` trait for method calls

niklasad1:
This PR adds a similar trait as we introduced for RPC subscriptions but for RPC method calls.

```rust
/// Something that can be converted into a JSON-RPC method call response.
pub trait IntoResponse {
	type Output: serde::Serialize + ToOwned<Owned = Output>;

	/// Something that can be converted into a JSON-RPC method call response.
	fn into_response(self) -> ResponsePayload<'static, Output>;
}

impl<T: serde::Serialize, E: IntoErrorObject> IntoResponse for Result<T, E> {
	type Output = T;

	fn into_response(self) -> PartialResponse<'static, T> {
		match self {
			Ok(val) => PartialResponse::Result(val),
			Err(e) => PartialResponse::Error(e.into_error_object()),
		}
	}
}
```

The benefit is that the return types are more flexible and any type that implements `IntoResponse` can be defined in the RPC proc macro API after this change but it only works for the `proc macro API for the server-only (see #1067 for futher information)


tests/tests/proc_macros.rs
```diff
@@ -104,11 +104,11 @@ mod rpc_impl {
 			Ok(format!("Zero copy params: {}, {}", matches!(a, std::borrow::Cow::Borrowed(_)), b.is_borrowed()))
 		}
 
-		#[method(name = "blocking_call", blocking)]
+		/*#[method(name = "blocking_call", blocking)]
```
niklasad1, line 107: todo :)

core/src/client/async_client/mod.rs
```diff
@@ -339,7 +339,7 @@ impl ClientT for Client {
 			Err(_) => return Err(self.read_error_from_backend().await),
 		};
 
-		rx_log_from_json(&Response::new(&json_value, id), self.max_log_length);
+		rx_log_from_json(&Response::new(PartialResponse::Result(json_value.clone()), id), self.max_log_length);
```
jsdw, line 342: I wonder what the cost of cloning is here?

niklasad1: we can provide a some borrowing here, I didn't bother just did naive thing :P

core/src/server/mod.rs
```diff
@@ -42,3 +42,74 @@ pub use helpers::{BatchResponseBuilder, BoundedWriter, MethodResponse, MethodSin
...
+/// Something that can be converted into a JSON-RPC method call response.
+///
+/// If the value couldn't be serialized/encoded, jsonrpsee will sent out an error
+/// to the client `response could not be serialized`.
+pub trait IntoResponse {
+	/// Output.
+	type Output: serde::Serialize;
+
+	/// Something that can be converted into a JSON-RPC method call response.
+	fn into_response(self) -> PartialResponse<Self::Output>;
```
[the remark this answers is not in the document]
jsdw: I guess I'm wondering ; can we justify the differences between this and 

```rust
pub trait IntoSubscriptionCloseResponse {
	fn into_response(self) -> SubscriptionCloseResponse;
}
```

Right now `IntoResponse` is really quite different (more generic, for instance with its associated param, and so on). Should these traits be aiming to be as consistent with eachother as makes sense, or are there good reasons for the differences that we can explain?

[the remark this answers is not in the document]
niklasad1: I would say that they are fundamentally different mainly because `IntoResponse` is something that converts a response to a JSON-RPC response or JSON-RPC error object and `IntoSubscriptionClose` is really just something that is converted into a subscription notification.

In practice, users would not need to implement `IntoResponse` except some rare edge-cases for custom types that are not `Result<T, impl IntoErrorObject`. For example you could return `enum MyResponse { Success(String), Error1, Error }` where one would need to implement this trait

Thus, is in practice the only trait that folks needs to implement is `IntoErrorObject` which a specific error code and so on.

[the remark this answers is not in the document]
jsdw: > I would say that they are fundamentally different mainly because IntoResponse is something that converts a response to a JSON-RPC response or JSON-RPC error object (a response with the result or error entry). IntoSubscriptionClose is really just something that is converted into a subscription notification.

I guess what I meant was that they both basically take some type and return some part of a JSON response, but this one comes with an associated type and is more generic, and this isn't. It's not a huge deal for me; just wondering whether there is anything that should be made more consistent between these traits, but happy if not :)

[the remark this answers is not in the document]
niklasad1: we want to enforce these formats and be flexible enough for folks to use:

```json
{"jsonrpc": "2.0", "result": -19, "id": 2}
```

or 

```json
{"jsonrpc": "2.0", "error": {"code": -32601, "message": "Method not found"}, "id": "1"}
```

so we could serialize the response in the `IntoResponse impl` to remove the generic stuff which probably is better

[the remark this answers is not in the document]
jsdw: I guess efficiency is the main thing; if we can be less generic ie have something like:

```rust
pub trait IntoResponse {
	fn into_response(self) -> Response; // returns some hardcoded Response type
}
```

Where `into_response` is serializing into some non-generic type, and _if_ that doesn't lose out in efficiency (because we have to serializie stuff anyway or whatever) then I geuss it'd be easier to udnerstand and more consistent with the subscription version.

But if having the generic `PartialResponse<Self::Output>` type thing makes stuff more efficient then I get it :)

benches/helpers.rs
```diff
@@ -169,23 +169,17 @@ pub async fn ws_server(handle: tokio::runtime::Handle) -> (String, jsonrpsee::se
 fn gen_rpc_module() -> jsonrpsee::RpcModule<()> {
 	let mut module = jsonrpsee::RpcModule::new(());
 
-	module.register_method(SYNC_FAST_CALL, |_, _| Ok("lo")).unwrap();
-	module
-		.register_async_method(ASYNC_FAST_CALL, |_, _| async { Result::<_, jsonrpsee::core::Error>::Ok("lo") })
-		.unwrap();
+	module.register_method(SYNC_FAST_CALL, |_, _| "lo").unwrap();
+	module.register_async_method(ASYNC_FAST_CALL, |_, _| async { "lo" }).unwrap();
```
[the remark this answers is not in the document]
niklasad1: it's cheating I implemented the `IntoResponse trait` for the primitive types so that can go away :P

types/src/response.rs
```diff
@@ -88,40 +132,297 @@ pub struct SubscriptionPayloadError<'a, T> {
...
+/// Represent the result or error field of the JSON-RPC response object
+///
+/// It can be:
+///
+/// ```json
+/// "result":<value>
+/// "error":{"code":<code>,"message":<msg>,"data":<data>}
+/// ```
+#[derive(Debug, PartialEq)]
+pub enum PartialResponseSer<'a, T> {
```
jsdw, line 144: Nit: Would `ResponsePayloadSer` make more sense? (I guess I see the "result" or "error" as the payload of the response)

core/src/server/mod.rs
```diff
@@ -42,3 +42,65 @@ pub use helpers::{BatchResponseBuilder, BoundedWriter, MethodResponse, MethodSin
...
+
+				fn into_response(self) -> ResponsePayload<'static, Self::Output> {
+					ResponsePayload::result(self)
+				}
+			}
+		)+
+	}
+}
+
+impl_into_response!(
```
niklasad1, line 88: we could remove these because it doesn't work to implement these for `[T], Vec<T>` and other sequence types because of the result impl above

types/src/response.rs
```diff
@@ -26,40 +26,77 @@
...
+	}
+}
+
+/// JSON-RPC response object as defined in the [spec](https://www.jsonrpc.org/specification#response_object)
+/// but differs from [`Response`] as it only represent a successful response.
+#[derive(Debug)]
+pub struct Success<'a, T> {
+	/// JSON-RPC version.
+	pub jsonrpc: Option<TwoPointZero>,
+	/// Result or error.
```
jsdw, line 86: No error here now I guess :)

core/src/server/mod.rs
```diff
@@ -42,3 +42,65 @@ pub use helpers::{BatchResponseBuilder, BoundedWriter, MethodResponse, MethodSin
...
+
+use jsonrpsee_types::{ErrorObjectOwned, ResponsePayload};
+
+/// Something that can be converted into a JSON-RPC method call response.
+///
+/// If the value couldn't be serialized/encoded, jsonrpsee will sent out an error
+/// to the client `response could not be serialized`.
+pub trait IntoResponse {
+	/// Output.
+	type Output: serde::Serialize + ToOwned<Owned = Self::Output>;
```
jsdw, line 54: Aah I see; it's a shame about the `ToOwned` but I see why!

`ToOwned` is impl for all `T where T: Clone`, so I wonder whether there `type Output: serde::Serialize + Clone` would be possible also and maybe a touch cleaner looking?. Hmm but the current approach is more "correct" though. Will have to ponder this, but yeah maybe it's best as it is here :)

proc-macros/src/helpers.rs
```diff
@@ -87,7 +87,7 @@ fn find_jsonrpsee_crate(crate_names: &[&str]) -> Result<proc_macro2::TokenStream
 ///  use jsonrpsee::core::{RpcResult, SubscriptionResult};
 ///
 ///  #[rpc(client, server)]
-///  pub trait RpcTrait<A, B, C> {
+///  pub trait RpcTrait<A, B: Clone, C> {
```
jsdw, line 90: Out of interest, why did this need adding?

jsdw: (some question for some other `Clone`'s. Is it to do with the `ToOwned` bound on the response? Will this make using the macro more painful?)

niklasad1: Every response for the server needs to implement Clone/ToOwned

such as

```
fn call() -> Result<T, Erro>;
```

T will be used in `IntoResponse trait` and thus must implement Clone.

Yes, it will be more annoying to use but Alex added these custom bounds on the traits which is okay I guess.

core/src/server/rpc_module.rs
```diff
@@ -386,15 +384,12 @@ impl Methods {
...
-			Ok(r) => r,
-			Err(_) => match serde_json::from_str::<ErrorResponse>(&resp.result) {
-				Ok(err) => return Err(Error::Call(CallError::Custom(err.into_error_object()))),
-				Err(err) => return Err(err.into()),
-			},
-		};
+		// TODO: hack around the lifetime on the `SubscriptionId` by deserialize first to serde_json::Value.
+		let as_success: ResponseSuccess<serde_json::Value> = serde_json::from_str::<Response<_>>(&resp.result)?
+			.try_into()
+			.map_err(|e| Error::Call(CallError::Custom(e)))?;
```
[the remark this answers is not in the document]
niklasad1: so if I deserialize to to `SubscriptionId<'a>` then later calls `sub_id.into_owned()` then serde complains about static lifetime issues

core/src/server/mod.rs
niklasad1: I don't know so I was strict and went with `ToOwned` and I don't really care.

Seems like it's equal to `Clone` in practice anyway.

types/src/response.rs
```diff
@@ -88,40 +125,314 @@ pub struct SubscriptionPayloadError<'a, T> {
...
+	fn eq(&self, other: &Self) -> bool {
+		match (self, other) {
+			(ResponsePayload::Error(a), ResponsePayload::Error(b)) => a == b,
+			(ResponsePayload::Result(a), ResponsePayload::Result(b)) => a == b,
+			_ => false,
+		}
+	}
+}
+
+impl<'a> ResponsePayload<'a, ()> {
```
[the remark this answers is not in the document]
niklasad1: I did it like that because specifying the `()` all over the place for error stuff is annoying

proc-macros/src/helpers.rs
niklasad1: it's a trade-off all these annoyance can away with separate serialize/deserialize types :)

That's why I skipped these complicated stuff in the first place but reduces some technical debt with duplicate types and hard to read.

jsdw (approved):

lexnv (approved):

proc-macros/src/helpers.rs
niklasad1, line 90:
```suggestion
///  pub trait RpcTrait<A, B, C> {
```

PR #1058, opened 2023-03-28
feat: add Methods::merge_replace

### LINKED ISSUE #1053: Support overriding existing Methods Callbacks
Currently, it's not possible to override `MethodCallback` in `Methods` because `Methods::merge` fails early if a callback with the name already exists

https://github.com/paritytech/jsonrpsee/blob/a0ce8d4c5c0dc1f4c65bd1efbe34dd16cc060b7a/core/src/server/rpc_module.rs#L355-L357

This makes it impossible to disable various callbacks dynamically when `RpcModule`s (via server proc macro's `into_rpc`) are merged.

now that `verify_method_name` is public this could be solved by:

* support `Methods::iter` and `Methods::into_iter` so callbacks can be inserted individually.
* support `Methods::merge_replace` which replaces existing handlers

mattsse:
Closes #1053

add `merge_replace` that replaces callbacks

### DIFF
--- a/core/src/server/rpc_module.rs
+++ b/core/src/server/rpc_module.rs
@@ -207,7 +207,7 @@ impl Methods {
 
 	/// Merge two [`Methods`]'s by adding all [`MethodCallback`]s from `other` into `self`.
 	///
-	/// Unlike [Self::merge], existing callbacks are replaced.
+	/// Unlike [`Methods::merge`], existing callbacks are replaced.
 	pub fn merge_replace(&mut self, other: impl Into<Methods>)  {
 		let mut other = other.into();
 
--- a/core/src/server/rpc_module.rs
+++ b/core/src/server/rpc_module.rs
@@ -205,6 +205,18 @@ impl Methods {
 		Ok(())
 	}
 
+	/// Merge two [`Methods`]'s by adding all [`MethodCallback`]s from `other` into `self`.
+	///
+	/// Unlike [Self::merge], existing callbacks are replaced.
+	pub fn merge_replace(&mut self, other: impl Into<Methods>)  {
+		let mut other = other.into();
+
+		let callbacks = self.mut_callbacks();
+		for (name, callback) in other.mut_callbacks().drain() {
+			callbacks.insert(name, callback);
+		}
+	}
+
 	/// Returns the method callback.
 	pub fn method(&self, method_name: &str) -> Option<&MethodCallback> {
 		self.callbacks.get(method_name)


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ This release adds `Clone` and `Copy` implementations.
 
 ### [Added]
 
+- add `Methods::merge_replace` ([#1058](https://github.com/paritytech/jsonrpsee/pull/1058))
```
niklasad1, line 18: revert this or add a new `UNRELEASED section`.

we already got a script that generates the CHANGELOG for us so easiest just to remove this and deal with it in a release PR.

lexnv, line 18: nit: This is not necessarily needed now. We'll generate the changelog with the next release, since its not part of the v0.16.2, but of the next release

core/src/server/rpc_module.rs
```diff
@@ -205,6 +205,18 @@ impl Methods {
 		Ok(())
 	}
 
+	/// Merge two [`Methods`]'s by adding all [`MethodCallback`]s from `other` into `self`.
+	///
+	/// Unlike [Self::merge], existing callbacks are replaced.
```
niklasad1, line 210: would be good with a rustdoc links here, not sure if `Self::merge` works then

```suggestion
	/// Unlike [`Methods::merge`], existing callbacks are replaced.
```

core/src/server/rpc_module.rs
```diff
@@ -205,6 +205,18 @@ impl Methods {
...
 
+	/// Merge two [`Methods`]'s by adding all [`MethodCallback`]s from `other` into `self`.
+	///
+	/// Unlike [`Methods::merge`], existing callbacks are replaced.
+	pub fn merge_replace(&mut self, other: impl Into<Methods>)  {
+		let mut other = other.into();
+
+		let callbacks = self.mut_callbacks();
+		for (name, callback) in other.mut_callbacks().drain() {
+			callbacks.insert(name, callback);
```
niklasad1, line 216: So this is bit tricky to use I reckon now as it's easy to override something and not realize it.

As I see it we could have `merge_by` as well

```rust
enum Insert {
 Yes,
 No,
}

fn merge_by<F: Fn(&str, &str) -> Insert)(&mut self, other: impl Into<Methods>, f: F) { ...  }

// example
methods.merge_by(other, |method,  other)| {
   // method names are equal
   if method == other {
      tracing::warn!("Overriding method: `{method}`");
   }
   Insert::Yes
}); 

```

jsdw: I could get behind that (I might be tempted to call it `.merge_if` and have it return a bool, true to merge)!.

If that existed, could merge be then a simple wrapper around it? 
Would this function also need the `self.verify_method_name(name)?;` check that `merge` has?

niklasad1: > Would this function also need the self.verify_method_name(name)?; check that merge has?

no, that only checks if a method with the same name exist and this API leaves that to the user to decide i.e, if two methods with the same name occurs then either replace it with the new one or something else.

lexnv: If we had `|existing: &str, other: &str|`, what would be the value of `existing` if the method did not prior to merging exist?

Another idea would be to have the callback fire on collisions only:

```rust

fn merge_if<F: Fn(&str) -> bool)(&mut self, other: impl Into<Methods>, f: F) { ...  }

methods.merge_if(other, |collision_name|| {
   if collision_name == "CoreFn" {
       warn!("Don't want to replace `{collision_name}`;
       return false;
   }
   true
```



Indeed we should fire the callback only if:
- `Entry::Occupied(..)` similar to `verify_method_name`
- name collides


I feel that both solutions limit the user to pick either the existing or the new method.
For that contributor's use case, the function should always be `|_collision_name| { true }`.

As a crazy thought, we could probably chain methods, but Id have a hard time arguing the complexity of the code is worth this kind of runtime sorcery:

```rust

fn map<F: Fn(&str, &mut MethodCallback )(&mut self, f: F) { ...  }

methods.map(other, |(existing_name, existing_method)| {
   if existing_name == "CoreFn" {
       // Modify existing_method: 
       // Chain it with some other callback, or replace it directly
       info!("Modifying `{existing_name}`;
   }
```

PR #1062, opened 2023-03-30
fix: tokio v1.27

### LINKED ISSUE #1059: CI: broken after tokio v1.27.0
The CI is broken after the latest tokio release v1.27.0

niklasad1:
Tokio v1.27 changed how interval/timers are woken up.

Because we already have `async fn StopHandle::shutdown` we could just select on that along with the other future.
Thus, these periodically wake-ups are not needed anymore which should fix this properly.

Close #1059

### DIFF
--- a/server/src/future.rs
+++ b/server/src/future.rs
@@ -132,9 +132,6 @@ where
 
 /// Represent a stop handle which is a wrapper over a `multi-consumer receiver`
 /// and cloning [`StopHandle`] will get a separate instance of the underlying receiver.
-///
-/// This means that `shutdown_requested` and `shutdown` can't be called
-/// on the same [`StopHandle`]
 #[derive(Debug, Clone)]
 pub(crate) struct StopHandle(watch::Receiver<()>);
 
@@ -144,8 +141,7 @@ impl StopHandle {
 	}
 
 	/// A future that resolves when server has been stopped
-	/// it consumes the stop handle such that [`StopHandle::shutdown_requested`]
-	/// can't be called on the same [`StopHandle`].
+	/// it consumes the stop handle.
 	pub(crate) async fn shutdown(mut self) {
 		let _ = self.0.changed().await;
 	}
--- a/server/src/future.rs
+++ b/server/src/future.rs
@@ -34,10 +34,6 @@ use std::task::{Context, Poll};
 use futures_util::future::FutureExt;
 use jsonrpsee_core::Error;
 use tokio::sync::{watch, OwnedSemaphorePermit, Semaphore, TryAcquireError};
-use tokio::time::{self, Duration, Interval};
-
-/// Polling for server stop monitor interval in milliseconds.
-const STOP_MONITOR_POLLING_INTERVAL: Duration = Duration::from_millis(1000);
 
 /// This is a flexible collection of futures that need to be driven to completion
 /// alongside some other future, such as connection handlers that need to be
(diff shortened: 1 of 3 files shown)


server/src/transport/ws.rs
```diff
@@ -519,3 +467,71 @@ async fn send_task(
...
+	receiver: &mut Receiver,
+	mut data: &mut Vec<u8>,
+	method_executor: &mut FutureDriver<F>,
+	stopped: S,
+) -> Receive<S>
+where
+	F: Future + Unpin,
+	S: Future<Output = ()> + Unpin,
+{
+	// Need the extra scope to drop this pinned future and reclaim access to `data`
```
niklasad1, line 487:
```suggestion
```

server/src/transport/ws.rs
```diff
@@ -519,3 +467,70 @@ async fn send_task(
...
+	stopped: S,
+) -> Receive<S>
+where
+	F: Future + Unpin,
+	S: Future<Output = ()> + Unpin,
+{
+	let receive = async {
+		// Identical loop to `soketto::receive_data` with debug logs for `Pong` frames.
+		loop {
+			match receiver.receive(&mut data).await? {
```
niklasad1, line 490:
```suggestion
			match receiver.receive(data).await? {
```

server/src/transport/ws.rs
```diff
@@ -519,3 +467,70 @@ async fn send_task(
...
+
+enum Receive<S> {
+	Shutdown,
+	Err(SokettoError, S),
+	Ok(S),
+}
+
+async fn try_recv<F, S>(
+	receiver: &mut Receiver,
+	mut data: &mut Vec<u8>,
```
niklasad1, line 479:
```suggestion
	data: &mut Vec<u8>,
```

server/src/future.rs
```diff
@@ -94,12 +85,6 @@ where
 			}
 		}
 	}
-
-	fn poll_stop_monitor_heartbeat(&mut self, cx: &mut Context) {
-		// We don't care about the ticks of the heartbeat, it's here only
-		// to periodically wake the `Waker` on `cx`.
-		let _ = self.stop_monitor_heartbeat.poll_tick(cx);
```
lexnv, line 101: IIUC we relied on this to make sure that we wake up:
- when there's no scheduled tasks that would yield to the executor (call waker.wake())
- for at most one second 

To remove `tokio::time::Interval` we are polling the `FutureDriver` alongside other operations that we do in the server:
- try_accept_conn: Await incoming connection
- wait_for_permit: Await backpressure for some internal buffer space
- try_recv: Can wait for `soketto::Pong`, `soketto::Data`,  and `soketto::Closed`

Could we still get in a situation where we have no other competing tasks to drive the `FutureDriver`?
(ie, we don't have `soketto::Pong` to wake us up -- or is configured at 100seconds -- and we don't receive `soketto::Data`)

jsdw: After convincing myself that this DriverSelect thing (ie `select_with`) will poll the futures in itself as needed, maybe I'm missing something but why does it matter if this thing doesn't wake up for a while if none of the futures in it need to progress?

niklasad1: so, that's good question/observation but I found that quite unlikely because both batch and calls are executed as one future/task.

Then in each loop iteration both `try_recv` and `wait_for_permit` are awaited on which checks "future vec" but not sure if there is possible to spawn huge futures and then "never send some data again/backpressure kicks in" then those would never woken up again but I don't see it as an issue because it's to the client enforce that.

If it's idle fine but we could have a few ready tasks in the vec.

server/src/server.rs
```diff
@@ -167,7 +168,7 @@ where
 			}
 		}
 
-		connections.await;
+		while connections.next().await.is_some() {}
```
jsdw, line 171: Ah ok, so if I understand this right, `FuturesUnordered` won't poll anything until you tell it to, so you are spawning tasks into it and then this line just makes sure that all tasks are completed before the function ends?

niklasad1: yeah, exactly

jsdw (approved):
From what I understand of this, it looks good to me!

It'd be interesting to see what, if any, effect this has on benchmarks :)

lexnv (approved):
LGTM! 👍

PR #1073, opened 2023-04-11
feat(server): configurable limit for batch requests.

### LINKED ISSUE #987: Configure batch support max items
We would like to enable batch support but with a relatively low limits. So it will be good if we can configure the max items in a batch request.

niklasad1:
Close #987

### DIFF
--- a/server/src/server.rs
+++ b/server/src/server.rs
@@ -285,7 +285,7 @@ impl<B, L> Builder<B, L> {
 	/// Configure how [batch requests](https://www.jsonrpc.org/specification#batch) shall be handled
 	/// by the server.
 	///
-	/// Default: batch requests are allowed and can be arbitrary big but the maximum response size is limited.
+	/// Default: batch requests are allowed and can be arbitrary big but the maximum payload size is limited.
 	pub fn set_batch_request_config(mut self, cfg: BatchRequestConfig) -> Self {
 		self.settings.batch_requests_config = cfg;
 		self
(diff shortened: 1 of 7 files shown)


jsdw (approved):

server/src/server.rs
```diff
@@ -272,10 +282,12 @@ impl<B, L> Builder<B, L> {
...
 	}
 
-	/// Enables or disables support of [batch requests](https://www.jsonrpc.org/specification#batch).
-	/// By default, support is enabled.
-	pub fn batch_requests_supported(mut self, supported: bool) -> Self {
-		self.settings.batch_requests_supported = supported;
+	/// Configure how [batch requests](https://www.jsonrpc.org/specification#batch) shall be handled
+	/// by the server.
+	///
+	/// Default: batch requests are allowed and can be arbitrary big but the maximum response size is limited.
```
niklasad1, line 288:
```suggestion
	/// Default: batch requests are allowed and can be arbitrary big but the maximum payload size is limited.
```

lexnv (approved):

server/src/server.rs
```diff
@@ -207,6 +207,16 @@ struct Settings {
 	message_buffer_capacity: u32,
 }
 
+#[derive(Debug, Copy, Clone)]
+pub enum BatchRequestConfig {
+	/// Batch requests are disabled.
+	Disabled,
+	/// The batch request is limited to `len`.
```
niklasad1, line 214:
```suggestion
	/// Each batch request is limited to `len` and any batch request bigger than `len` will not be processed.
```

PR #1077, opened 2023-04-12
docs: introduce workspace attributes and add keywords

### LINKED ISSUE #970: Add keywords to crates.io page / to crate
Hi,

it seems like this crate is not easily discoverable. Adding the jsonrpc keyword to it in the Cargo.toml and therefore also on https://crates.io/crates/jsonrpsee could maybe help.

Kind regards
Moritz

niklasad1:
Close #970


Cargo.toml
```diff
@@ -16,3 +16,14 @@ members = [
...
+[workspace.package]
+authors = ["Parity Technologies <admin@parity.io>", "Pierre Krieger <pierre.krieger1708@gmail.com>"]
+edition = "2021"
+rust-version = "1.64.0"
+license = "MIT"
+repository = "https://github.com/paritytech/jsonrpsee"
+documentation = "https://docs.rs/jsonrpsee"
+homepage = "https://www.parity.io/"
+keywords = ["jsonrpc", "json", "http", "websocket", "WASM"]
+readme = "README.md"
```
niklasad1, line 29:
```suggestion
readme = "README.md"

```

jsdw, line 29: Any idea how it resolves this in dependencies; does it use the path relative to this workspace toml?

client/transport/Cargo.toml
```diff
@@ -1,13 +1,17 @@
...
 name = "jsonrpsee-client-transport"
-version = "0.16.2"
-authors = ["Parity Technologies <admin@parity.io>", "Pierre Krieger <pierre.krieger1708@gmail.com>"]
 description = "JSON-RPC client transport"
-edition = "2021"
-license = "MIT"
-repository = "https://github.com/paritytech/jsonrpsee"
-homepage = "https://github.com/paritytech/jsonrpsee"
-documentation = "https://docs.rs/jsonrpsee-client-transport"
+version = "0.16.2"
```
jsdw, line 4: I'm not sure why version wasn't inherited in Subxt; if we want to keep it the same it seems to make sense to just inherit it too

jsdw (approved):
LGTM; I guess you didn't want to inherit authors for all because those other crates were post-Pierre?

Not sure how README path inherits; maybe it just works; would be good to know :)

niklasad1:
> LGTM; I guess you didn't want to inherit authors for all because those other crates were post-Pierre?

I was just lazy with the test crates, I'll fix it tomorrow.

niklasad1:
> Not sure how README path inherits; maybe it just works; would be good to know :)

According to https://rust-lang.github.io/rfcs/2906-cargo-workspace-deduplicate.html#placeholder-values it should work

Cargo.toml
niklasad1: According to https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table it should work

Cargo.toml
```diff
@@ -16,3 +16,25 @@ members = [
...
+
+[workspace.dependencies]
+jsonrpsee-types = { path = "types", version = "0.16.2" }
+jsonrpsee-core = { path = "core", version = "0.16.2" }
+jsonrpsee-server = { path = "server", version = "0.16.2" }
+jsonrpsee-ws-client = { path = "client/ws-client", version = "0.16.2" }
+jsonrpsee-http-client = { path = "client/http-client", version = "0.16.2" }
+jsonrpsee-wasm-client = { path = "client/wasm-client", version = "0.16.2" }
+jsonrpsee-client-transport = { path = "client/transport", version = "0.16.2" }
+jsonrpsee-proc-macros = { path = "proc-macros", version = "0.16.2" }
```
niklasad1, line 40:
```suggestion
jsonrpsee-proc-macros = { path = "proc-macros", version = "0.16.2" }

```

tests/wasm-tests/Cargo.toml
```diff
@@ -1,7 +1,8 @@
 [package]
 name = "wasm-test"
 version = "0.1.0"
-edition = "2021"
+edition.workspace = true
+license.workspace = true
```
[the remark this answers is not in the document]
niklasad1: yes

lexnv (approved):

Cargo.toml
jsdw: Ah; readme and license paths are relative to the workspace root, ok cool! Not so useful if you want to have crate specific readme's but good if you just want to use the same one for all crates, so makes sense to me!

PR #1087, opened 2023-04-19
remove `CallError`

### LINKED ISSUE #1085: remove `CallError`
The CallError is not necessarily needed anymore it should be replaced by `ErrorObject`.

- `CallError::Custom => ErrorObject`
- `CallError::InvalidParams => new type InvalidParams`
- `CallError::Failed => remove`

niklasad1:
Close https://github.com/paritytech/jsonrpsee/issues/1085

One downside of this is that `jsonrpsee::core::Error` can't be used in the proc macro API and I worked around that by changing `type RpcResult<T> = Result<T, ErrorObjectOwned>`

The reason why :point_up: is because the `IntoResponse` requires the error to implement `Into<ErrorObjectOwned>` which ends up being really weird to implement from `Error variants` that isn't really a JSON-RPC v2 spec error but something else like I/O etc.

Thus, it's more explicit for users to implement `Into<ErrorObjectOwned>` themselves which is explicit and more understandable.

### DIFF
--- a/core/src/error.rs
+++ b/core/src/error.rs
@@ -46,7 +46,7 @@ impl<T: fmt::Display> fmt::Display for Mismatch<T> {
 /// Error type.
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
-	/// Error that occurs when a call failed.
+	/// JSON-RPC error which can occur when a JSON-RPC call fails.
 	#[error("{0}")]
 	Call(#[from] ErrorObjectOwned),
 	/// Networking error or error on the low-level protocol layer.
--- a/core/src/error.rs
+++ b/core/src/error.rs
@@ -26,9 +26,7 @@
 
 use std::fmt;
 
-use jsonrpsee_types::error::{
-	CallError, ErrorObject, ErrorObjectOwned, CALL_EXECUTION_FAILED_CODE, INVALID_PARAMS_CODE, UNKNOWN_ERROR_CODE,
-};
+use jsonrpsee_types::error::ErrorObjectOwned;
 
 /// Convenience type for displaying errors.
 #[derive(Clone, Debug, PartialEq, Eq)]
@@ -45,20 +43,12 @@ impl<T: fmt::Display> fmt::Display for Mismatch<T> {
 	}
 }
 
-// NOTE(niklasad1): this `From` impl is a bit opinionated to regard all generic errors as `CallError`.
-// In practice this should be the most common use case for users of this library.
-impl From<anyhow::Error> for Error {
-	fn from(err: anyhow::Error) -> Self {
-		Error::Call(CallError::Failed(err))
-	}
-}
-
 /// Error type.
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
 	/// Error that occurs when a call failed.
 	#[error("{0}")]
-	Call(#[from] CallError),
+	Call(#[from] ErrorObjectOwned),
 	/// Networking error or error on the low-level protocol layer.
 	#[error("Networking or low-level protocol error: {0}")]
 	Transport(#[source] anyhow::Error),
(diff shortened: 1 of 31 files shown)


core/src/error.rs
```diff
@@ -45,20 +43,12 @@ impl<T: fmt::Display> fmt::Display for Mismatch<T> {
...
-impl From<anyhow::Error> for Error {
-	fn from(err: anyhow::Error) -> Self {
-		Error::Call(CallError::Failed(err))
-	}
-}
-
 /// Error type.
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
 	/// Error that occurs when a call failed.
```
niklasad1, line 49:
```suggestion
	/// JSON-RPC error which can occur when a JSON-RPC call fails.
```

core/src/lib.rs
```diff
@@ -62,7 +62,7 @@ pub use async_trait::async_trait;
 pub use error::{Error, GenericTransportError, StringError};
 
 /// JSON-RPC result.
-pub type RpcResult<T> = std::result::Result<T, Error>;
+pub type RpcResult<T> = std::result::Result<T, ErrorObjectOwned>;
```
niklasad1, line 65:
```suggestion
pub type RpcResult<T> = std::result::Result<T, jsonrpee_types::ErrorObjectOwned>;
```

core/src/lib.rs
```diff
@@ -77,6 +77,7 @@ pub mod __reexports {
 }
 
 pub use beef::Cow;
+use jsonrpsee_types::ErrorObjectOwned;
```
niklasad1, line 80:
```suggestion
```

types/src/params.rs
```diff
@@ -382,6 +384,10 @@ impl<'a> Id<'a> {
 	}
 }
 
+fn invalid_params(e: impl ToString) -> ErrorObjectOwned {
+	ErrorObject::owned(ErrorCode::InvalidParams.code(), e.to_string(), None::<()>)
```
niklasad1, line 388: strictly from the jsonrpc spec:

```
code     | message          | meaning
-32602  | Invalid params | Invalid method parameter(s).
```

we are using -32602 => `message == e: impl ToString`

niklasad1: but it was like that before this PR as well, it's picking hair I'm fine having a slightly better error message than "Invalid Params" though we could modify it in another and add this extra data in the data field instead of the message.

proc-macros/tests/ui/correct/only_client.rs
```diff
@@ -1,14 +1,14 @@
 //! Example of using proc macro to generate working client.
 
-use jsonrpsee::{core::RpcResult, proc_macros::rpc};
+use jsonrpsee::{core::Error, proc_macros::rpc};
```
lexnv, line 3:
```suggestion
use jsonrpsee::proc_macros::rpc;
```

lexnv: Git seems to complain about this, although we do end up using it in the macro

lexnv (approved):

proc-macros/tests/ui/correct/only_client.rs
niklasad1: it's probably because we overwrite the error type in the client and rustc understands that :P

server/src/tests/helpers.rs
```diff
@@ -102,21 +100,21 @@ pub(crate) async fn server_with_handles() -> (SocketAddr, ServerHandle) {
...
 			RpcResult::Ok("ok")
 		})
 		.unwrap();
 	module
-		.register_async_method("should_ok_async", |_p, ctx| async move {
-			ctx.ok().map_err(CallError::Failed)?;
-			Result::<_, Error>::Ok("ok")
+		.register_async_method::<Result<&str, MyAppError>, _, _>("should_ok_async", |_p, ctx| async move {
+			ctx.ok()?;
+			Ok("ok")
```
jsdw, line 117: Nit: maybe consistify these generic ones with what you did below with §Result::<_, MyAppError>::Ok(..)`

types/src/error.rs
```diff
@@ -265,30 +256,6 @@ impl serde::Serialize for ErrorCode {
...
-
-impl CallError {
-	/// Create `CallError` from a generic error.
-	pub fn from_std_error<E>(err: E) -> Self
-	where
-		E: std::error::Error + Send + Sync + 'static,
-	{
-		CallError::Failed(err.into())
-	}
-}
```
jsdw, line 290: Ah ok, so previously user errors were being turned into thie `Failed` variant, and now the user describes how to turn them into an `ErrorObjectOwned` basically? This makes a lot of sense to me; give users the ability to set the error code etc :)

types/src/params.rs
jsdw: Doesn't feel like a big deal (I'd expect people to match on the code), but having some more detailed thing inthe data field sounds more "correct" (would it be annoying for us to do this?)

jsdw (approved):
Looks good and if I understand right, I like the extra control that users get here

types/src/error.rs
niklasad1: Yeah, this was a way to easily convert any other StdError into a `ErrorObject` and let's jsonrpsee pick the error code.

This was a bad idea it's better that folks take care of the error code etc and explicitly.

PR #1090, opened 2023-04-20
server: fix http graceful shutdown

### LINKED ISSUE #1089: http server: graceful shutdown doesn't work
```rust
// Attempts to create a HTTP connection from a socket.
async fn to_http_service<S, B>(socket: TcpStream, service: S, stop_handle: StopHandle)
where
	S: Service<hyper::Request<hyper::Body>, Response = hyper::Response<B>> + Send + 'static,
	S::Error: Into<Box<dyn std::error::Error + Send + Sync>>,
	S::Future: Send,
	B: HttpBody + Send + 'static,
	<B as HttpBody>::Error: Send + Sync + StdError,
	<B as HttpBody>::Data: Send,
{
	let conn = hyper::server::conn::Http::new().serve_connection(socket, service).with_upgrades();

	tokio::pin!(conn);

	tokio::select! {
		res = &mut conn => {
			if let Err(e) = res {
				tracing::warn!("HTTP serve connection failed {:?}", e);
			}
		}
		_ = stop_handle.shutdown() => {
			conn.graceful_shutdown();
		}
	}
}
```

So yeah I'm surprised by this behavior but I have a closer look

niklasad1:
Close #1089

### DIFF
--- a/server/src/server.rs
+++ b/server/src/server.rs
@@ -812,7 +812,8 @@ where
 	let res = match future::select(conn, stopped).await {
 		Either::Left((conn, _)) => conn,
 		Either::Right((_, mut conn)) => {
-			// The connection should continue to be polled until shutdown can finish.
+			// NOTE: the connection should continue to be polled until shutdown can finish.
+			// Thus, both lines below are needed and not a nit.
 			Pin::new(&mut conn).graceful_shutdown();
 			conn.await
 		}
--- a/server/src/server.rs
+++ b/server/src/server.rs
@@ -812,6 +812,10 @@ where
 	let res = match future::select(conn, stopped).await {
 		Either::Left((conn, _)) => conn,
 		Either::Right((_, mut conn)) => {
+			// Warning:
+			//
+			// These two lines are required for doing graceful shutdown.
+			// See https://github.com/hyperium/hyper/issues/2939#issuecomment-1213481266 for further information.
 			Pin::new(&mut conn).graceful_shutdown();
 			conn.await
 		}
--- a/server/src/server.rs
+++ b/server/src/server.rs
@@ -815,7 +815,8 @@ where
 			}
 		}
 		_ = stop_handle.shutdown() => {
-			conn.graceful_shutdown();
+			conn.as_mut().graceful_shutdown();
+			let _ = conn.await;
 		}
 	}
 }
--- a/server/src/server.rs
+++ b/server/src/server.rs
@@ -812,7 +812,6 @@ where
 	let res = match future::select(conn, stopped).await {
 		Either::Left((conn, _)) => conn,
 		Either::Right((_, mut conn)) => {
-			// Warning:
 			//
 			// The connection should continue to be polled until shutdown can finish.
 			Pin::new(&mut conn).graceful_shutdown();
--- a/server/src/server.rs
+++ b/server/src/server.rs
@@ -36,7 +36,7 @@ use crate::future::{ConnectionGuard, ServerHandle, StopHandle};
 use crate::logger::{Logger, TransportProtocol};
 use crate::transport::{http, ws};
 
-use futures_util::future::{Either, FutureExt};
+use futures_util::future::{self, Either, FutureExt};
 use futures_util::io::{BufReader, BufWriter};
 
 use futures_util::stream::{FuturesUnordered, StreamExt};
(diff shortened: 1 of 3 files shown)


server/src/server.rs
```diff
@@ -815,7 +815,8 @@ where
 			}
 		}
 		_ = stop_handle.shutdown() => {
-			conn.graceful_shutdown();
+			conn.as_mut().graceful_shutdown();
+			let _ = conn.await;
```
niklasad1, line 819: this is the fix

server/src/server.rs
```diff
@@ -805,18 +805,20 @@ where
...
-				tracing::warn!("HTTP serve connection failed {:?}", e);
-			}
-		}
-		_ = stop_handle.shutdown() => {
-			conn.graceful_shutdown();
+	let res = match future::select(conn, stopped).await {
+		Either::Left((conn, _)) => conn,
+		Either::Right((_, mut conn)) => {
+			Pin::new(&mut conn).graceful_shutdown();
+			conn.await
```
niklasad1, line 816: this is the fix, we need to await for the conn future anyway not just call `graceful_shutdown` my bad

niklasad1: See https://github.com/hyperium/hyper/issues/2939#issuecomment-1213481266

jsdw: Ah interesting! (I found https://docs.rs/hyper/latest/hyper/server/conn/struct.Connection.html#method.graceful_shutdown more useful)

Might be worth a comment next to these lines to note that since it's a slightly odd API?

tests/tests/integration_tests.rs
```diff
@@ -1206,4 +1206,72 @@ async fn graceful_shutdown_works() {
...
+	// Assert that no calls have been answered yet
+	// The assumption is that the server should be able to read 10 messages < 10s.
+	assert!(!call_answered.load(std::sync::atomic::Ordering::SeqCst));
+
+	// Stop the server.
+	handle.stop().unwrap();
+	handle.stopped().await;
+
+	// The pending calls should be answered before shutdown.
+	assert_eq!(res.await.unwrap(), 10);
```
jsdw, line 1273: So If I understand right, the core of this test is that it makes some requests to the server which will take time to respond, and then it asks to stop the server (after all calls have definitely made it to the server but before they will have resolved) and all calls should resolve before the server stops?

Sounds good to me offhand! I assume that without your fix above this fails?

I wonder whether some comment might make it easier to understand the aim, like:
```rust
/// - Make 10 calls to the server.
/// - Once they have all reached the server but before they have responded, stop the server.
/// - All calls should be responded to before the server shuts down.
```

Out of interest what happens if, while it's shutting down (ie while waiting for calls to be responded to), you make some new calls? I'd expect any new calls to to rejected or something, but would they slow down the shutting down process?

niklasad1: Yeah, I did one call after the `server.stopped returned` to ensure that the server is closed and doesn't accept new connections/calls

jsdw: Ah but what if new calls come in after you call `handle.stop()` but before `handle.stopped().await;` finishes?

Eg if the server is quite busy and new connections keep coming in; will it still shutdown?

niklasad1: it's not tested here indeed but I think it should be possible to add

niklasad1: > Sounds good to me offhand! I assume that without your fix above this fails?

Yes, it fails without my fix

jsdw: > it's not tested here indeed but I think it should be possible to add

It'd be good to check this somehow I think just to make sure that continuing new connections won't stop the server from ever shutting down :)

> Yes, it fails without my fix

Woop!

jsdw (approved):
Small suggestion re the test but looks good to me!

server/src/server.rs
```diff
@@ -805,18 +805,23 @@ where
...
-		_ = stop_handle.shutdown() => {
-			conn.graceful_shutdown();
+	let stopped = stop_handle.shutdown();
+
+	tokio::pin!(stopped);
+
+	let res = match future::select(conn, stopped).await {
+		Either::Left((conn, _)) => conn,
+		Either::Right((_, mut conn)) => {
+			// Warning:
```
niklasad1, line 815:
```suggestion
```

server/src/server.rs
```diff
@@ -805,18 +805,21 @@ where
...
-			if let Err(e) = res {
-				tracing::warn!("HTTP serve connection failed {:?}", e);
-			}
-		}
-		_ = stop_handle.shutdown() => {
-			conn.graceful_shutdown();
+	let res = match future::select(conn, stopped).await {
+		Either::Left((conn, _)) => conn,
+		Either::Right((_, mut conn)) => {
+			// The connection should continue to be polled until shutdown can finish.
```
niklasad1, line 815:
```suggestion
			// NOTE: the connection should continue to be polled until shutdown can finish.
			// Thus, both lines below are needed and not a nit.
```

lexnv (approved):

PR #1094, opened 2023-04-24
fix(proc macros): support parsing params !Result

### LINKED ISSUE #1093: proc macros: IntoResponse doesn't work for parameters that isn't `Result`
https://github.com/paritytech/jsonrpsee/pull/1057 introduced new a trait for method call responses however there is a an edge-case where the parameter parsing in the `proc macro code` assumes it can't return [Err(e.into)](https://github.com/paritytech/jsonrpsee/blob/master/proc-macros/src/render_server.rs#L306-#L316) this doesn't work for a type which isn't a `Result` and the removal of `CallError/jsonrpsee::core::Error` needs an extra `From` impl.

These needs to be fixed before releasing in v0.18 and ideally the params parsing could just return `PartialResponse` directly as handled for subscriptions as well.

### Example with the custom error type

```rust
pub struct MyError;

impl From<MyError> for ErrorObjectOwned {
	fn from(_: MyError) -> Self {
		ErrorObject::owned(1, "my_error", None::<()>)
	}
}

// This is annoying and shouldn't be needed..
impl From<ErrorObjectOwned> for MyError {
	fn from(_: ErrorObjectOwned) -> MyError {
		MyError
	}
}

#[rpc(server, client, namespace = "t")]
pub trait Api {
	#[method(name = "getKeys")]
	async fn f(&self, x: usize) -> Result<usize, MyError>;
}
```

### Example with the custom return type

```rust
// Serialize impl is not used as the responses are sent out as error.
#[derive(Serialize, Clone)]
pub enum CustomError {
	One,
	Two { custom_data: u32 },
}

impl IntoResponse for CustomError {
	type Output = Self;

	fn into_response(self) -> ResponsePayload<'static, Self::Output> {
           todo!()
        }
	
}

#[rpc(server)]
pub trait Rpc {
        // This won't compile as it would return `Result<usize, ErrorObject>` when parsing `x` in the call
	#[method(name = "method1")]
	async fn method1(&self, x: usize) -> CustomError;
}
```

niklasad1:
Fixes #1093

### DIFF
--- a/core/src/server/mod.rs
+++ b/core/src/server/mod.rs
@@ -104,6 +104,25 @@ where
 	}
 }
 
+impl<T> IntoResponse for ResponsePayload<'static, T>
+where
+	T: serde::Serialize + Clone,
+{
+	type Output = T;
+
+	fn into_response(self) -> ResponsePayload<'static, Self::Output> {
+		self
+	}
+}
+
+impl IntoResponse for ErrorObjectOwned {
+	type Output = ErrorObjectOwned;
+
+	fn into_response(self) -> ResponsePayload<'static, Self::Output> {
+		ResponsePayload::Error(self)
+	}
+}
+
 macro_rules! impl_into_response {
 	($($n:ty),*) => {
 		$(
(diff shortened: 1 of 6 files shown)


tadeohepperle (approved):
Looks good to me :)

lexnv (approved):

PR #1098, opened 2023-04-25
rpc module: fix race in subscription close callback

niklasad1:
It's technically possible that `subscription future` finishes after the subscription callback is executed (most likely just in tests) anyway this PR ensures that the `subscription future` completes before the subscription callback is terminated.

### DIFF
--- a/tests/tests/helpers.rs
+++ b/tests/tests/helpers.rs
@@ -36,7 +36,7 @@ use jsonrpsee::server::{
 	AllowHosts, PendingSubscriptionSink, RpcModule, ServerBuilder, ServerHandle, SubscriptionMessage, TrySendError,
 };
 use jsonrpsee::types::{ErrorObject, ErrorObjectOwned};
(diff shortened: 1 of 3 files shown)


jsdw (approved):
Looks good, and the code is easier to follow too :)

tests/tests/helpers.rs
```diff
@@ -105,25 +105,8 @@ pub async fn server_with_subscription_and_handle() -> (SocketAddr, ServerHandle)
...
-				}
-			}
-
-			let Ok(_sink) = pending.accept().await else {
-				return Response::Nothing;
-			};
-
-			Response::Closed
+			let _ = pending.accept().await;
+			SubscriptionCloseResponse::None
```
niklasad1, line 109: flaky test, it shouldn't return anything such that it triggers a timeout..

tadeohepperle (approved):

PR #1099, opened 2023-04-26
Added Default Generic Param to HttpClient

### LINKED ISSUE #1096: http client: add default backend type
After we added tower middleware support for the HttpClient one has to do:

```rust
type HttpClient = jsonrpsee::http_client::HttpClient<jsonrpsee::http_client::transport::HttpBackend>;

struct MyWrapper(HttpClient);
```

It would be good if folks doesn't have to do that when the default backend is used i.e, not tower middleware enabled on the client.

Such as:

```rust
/// JSON-RPC HTTP Client that provides functionality to perform method calls and notifications.
#[derive(Debug, Clone)]
pub struct HttpClient<S = HttpBackend> {
	/// HTTP transport client.
	transport: HttpTransportClient<S>,
	/// Request timeout. Defaults to 60sec.
	request_timeout: Duration,
	/// Request ID manager.
	id_manager: Arc<RequestIdManager>,
}
``` 

Code to change/fix https://github.com/paritytech/jsonrpsee/blob/master/client/http-client/src/client.rs#L236-#L245

tadeohepperle:
fixes #1096 

Added Default Generic Param like so:
```rs
/// JSON-RPC HTTP Client that provides functionality to perform method calls and notifications.
#[derive(Debug, Clone)]
pub struct HttpClient<S = HttpBackend> {
	/// HTTP transport client.
	transport: HttpTransportClient<S>,
	/// Request timeout. Defaults to 60sec.
	request_timeout: Duration,
	/// Request ID manager.
	id_manager: Arc<RequestIdManager>,
}
```

### DIFF
--- a/client/http-client/src/client.rs
+++ b/client/http-client/src/client.rs
@@ -30,7 +30,7 @@ use std::fmt;
 use std::sync::Arc;
 use std::time::Duration;
 
-use crate::transport::{self, Error as TransportError, HttpTransportClient};
+use crate::transport::{self, Error as TransportError, HttpBackend, HttpTransportClient};
 use crate::types::{NotificationSer, RequestSer, Response};
 use async_trait::async_trait;
 use hyper::body::HttpBody;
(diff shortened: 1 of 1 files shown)


niklasad1 (approved):
:+1:

jsdw (approved):

PR #1103, opened 2023-04-26
fix(ws server): fix shutdown on connection closed

niklasad1:


server/src/transport/ws.rs
```diff
@@ -330,14 +330,33 @@ pub(crate) async fn background_task<L: Logger>(
...
 	// but it's preferred that the `stop_handle.stopped()` should not return until all methods has been
 	// executed and the connection has been closed.
-	tokio::select! {
-		// All pending calls executed.
-		_ = pending_calls.for_each(|_| async {}) => {
-			_ = conn_tx.send(());
+	match result {
+		Ok(Shutdown::Stopped) | Err(_) => {
+			// Soketto doesn't have a way to signal when the connection is closed
+			// thus just throw the data and terminate the stream once the connection has
```
niklasad1, line 336:
```suggestion
			// thus just throw away the data and terminate the stream once the connection has
```

server/src/transport/ws.rs
```diff
@@ -330,14 +330,33 @@ pub(crate) async fn background_task<L: Logger>(
...
-		// All pending calls executed.
-		_ = pending_calls.for_each(|_| async {}) => {
-			_ = conn_tx.send(());
+	match result {
+		Ok(Shutdown::Stopped) | Err(_) => {
+			// Soketto doesn't have a way to signal when the connection is closed
+			// thus just throw away the data and terminate the stream once the connection has
+			// been terminated.
+			//
+			// The receiver is not cancel-safe such that it used in stream to enforce that.
```
niklasad1, line 339:
```suggestion
			// The receiver is not cancel-safe such that it used in a stream to enforce that.
```

server/src/transport/ws.rs
```diff
@@ -330,14 +330,33 @@ pub(crate) async fn background_task<L: Logger>(
...
+			let disconnect_stream = futures_util::stream::unfold((receiver, data), |(mut receiver, mut data)| async {
+				if let Err(SokettoError::Closed) = receiver.receive(&mut data).await {
+					None
+				} else {
+					Some(((), (receiver, data)))
+				}
+			});
+
+			let pending = pending_calls.for_each(|_| async {});
+			let disconnect = disconnect_stream.for_each(|_| async {});
```
niklasad1, line 349: it's possible that the connection is terminated during the graceful shutodown and this ensures that it is aborted once the close message is sent.

however, this is slightly error prone if the no proper close is sent but yeah would neat if soketto had something to indicate when the connection is closed

server/src/tests/ws.rs
```diff
@@ -819,7 +819,13 @@ async fn drop_client_with_pending_calls_works() {
 	init_logger();
 
 	let (handle, addr) = {
-		let server = ServerBuilder::default().build("127.0.0.1:0").with_default_timeout().await.unwrap().unwrap();
+		let server = ServerBuilder::default()
+			.ping_interval(std::time::Duration::from_secs(60 * 60))
```
jsdw, line 823: Is the point here that we send some requests that will never be responded to, and then expect that the client will never get a response back but that the server will shutdown fully?

I guess the big ping time prevents pings from helping move the server along? (is there a reason for 60 * 60 secs specifically; would any large enough time work?)

Might be worth a comment in any case here so future us knows why the piong interval was changed!

niklasad1: yupp, if the graceful shutdown doesn't resolve in 60 seconds the timeout below will fail the test

server/src/transport/ws.rs
```diff
@@ -272,11 +272,11 @@ pub(crate) async fn background_task<L: Logger>(
 				stopped = stop;
 				permit
 			}
-			None => break Ok(()),
+			None => break Ok(Shutdown::ConnectionClosed),
```
jsdw, line 275: Am I right in thinking that:
- `Stopped` means the user has manually stopped the server, so connection was closed
- `ConnectionClosed` means the connection was closed for some other reason (eg network issue or whatever)

niklasad1: yes, this is only fails if the `send_task` has been completed and the receiver of the bounded channel has been dropped.

probably some other variant is in appropriate  here then

server/src/tests/ws.rs
jsdw: Gotcha,. thanks! I think it'd be worth making this explicit via comments because "with_default_timeout" below etc doesn't make it super obvious :)

server/src/transport/ws.rs
```diff
@@ -330,14 +330,33 @@ pub(crate) async fn background_task<L: Logger>(
...
+				}
+			});
+
+			let pending = pending_calls.for_each(|_| async {});
+			let disconnect = disconnect_stream.for_each(|_| async {});
+
+			tokio::select! {
+				_ = pending => (),
+				_ = disconnect => (),
+			}
```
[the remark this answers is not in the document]
jsdw: I find it a bit hard to follow the logic here!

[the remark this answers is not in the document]
niklasad1: yeah, it's way to make it possible to detect whether connection has terminated while we try to wait for the pending futures to complete

jsdw (approved):

server/src/tests/ws.rs
niklasad1: added a constant for that

server/src/transport/ws.rs
```diff
@@ -330,14 +330,33 @@ pub(crate) async fn background_task<L: Logger>(
...
+				} else {
+					Some(((), (receiver, data)))
+				}
+			});
+
+			let graceful_shutdown = pending_calls.for_each(|_| async {});
+			let disconnect = disconnect_stream.for_each(|_| async {});
+
+			tokio::select! {
+				_ = pending => (),
```
niklasad1, line 352:
```suggestion
				_ = graceful_shutdown => (),
```

server/src/tests/ws.rs
```diff
@@ -847,5 +835,51 @@ async fn drop_client_with_pending_calls_works() {
...
+	let (handle, addr) = server_with_infinite_call(MAX_TIMEOUT.checked_mul(10).unwrap()).await;
+
+	{
+		let mut client = WebSocketTestClient::new(addr).with_default_timeout().await.unwrap().unwrap();
+
+		for _ in 0..10 {
+			let req = r#"{"jsonrpc":"2.0","method":"infinite_call","id":1}"#;
+			client.send(req).with_default_timeout().await.unwrap().unwrap();
+		}
+		tokio::time::sleep(std::time::Duration::from_secs(10)).await;
```
niklasad1, line 855:
```suggestion
       // Assumption: the calls would be ACK:ed by server after 10 seconds (no way knowing that)
		tokio::time::sleep(std::time::Duration::from_secs(10)).await;
```

server/src/tests/ws.rs
```diff
@@ -847,5 +835,52 @@ async fn drop_client_with_pending_calls_works() {
...
+
+	{
+		let mut client = WebSocketTestClient::new(addr).with_default_timeout().await.unwrap().unwrap();
+
+		for _ in 0..10 {
+			let req = r#"{"jsonrpc":"2.0","method":"infinite_call","id":1}"#;
+			client.send(req).with_default_timeout().await.unwrap().unwrap();
+		}
+		// Assumption: the calls would be ACK:ed by server after 10 seconds (no way knowing that except parsing CLI output)
+		tokio::time::sleep(std::time::Duration::from_secs(10)).await;
```
lexnv, line 856: dq: I remember from the substrate CI that these constants could be exceeded on overcommited days. Should we increase the timeout a bit here, or this should be sufficient?

lexnv (approved):

server/src/tests/ws.rs
niklasad1: I don't know what I was thinking, I could embed `mpsc::Sender` in the method callback and await for 10 messages

Nice to get rid of sleeps, those shouldn't be used if possible :D

niklasad1:
@jsdw can you have another look if it's easier to read now? Did some minor refactoring

Issue #1095, opened 2023-04-24
RequestMiddlewareAction 


aniketfuryrocks:
Hi I am porting the old jsonrpc implementation of solana to new jsonrpsee.

JsonRpc RequestMiddlewareAction implementation.
```rust
/// Action undertaken by a middleware.
pub enum RequestMiddlewareAction {
    /// Proceed with standard RPC handling
    Proceed {
        /// Should the request be processed even if invalid CORS headers are detected?
        /// This allows for side effects to take place.
        should_continue_on_invalid_cors: bool,
        /// The request object returned
        request: hyper::Request<Body>,
    },
    /// Intercept the request and respond differently.
    Respond {
        /// Should standard hosts validation be performed?
        should_validate_hosts: bool,
        /// a future for server response
        response: Pin<Box<dyn Future<Output = hyper::Result<hyper::Response<Body>>> + Send>>,
    },
}
```

What's is the recommended alternative in jsonrpsee 🤔

[Current implementation using jsonrpc](https://github.com/solana-labs/solana/blob/1b08d01a80f5aa9e9e6326c87ed6f6356a9a66d0/rpc/src/rpc_service.rs#L247)

[JsonRpc Request Middleware](https://github.com/paritytech/jsonrpc/blob/38af3c9439aa75481805edf6c05c6622a5ab1e70/ws/src/session.rs#L27)

niklasad1:
Closing this a duplicate to #804

aniketfuryrocks:
```rust
use std::{
    error::Error,
    future::Future,
    pin::Pin,
    sync::Arc,
    task::{Context, Poll},
};
use tower::{Layer, Service};

/// Action undertaken by a middleware.
pub enum RequestMiddlewareAction {
    /// Proceed with standard RPC handling
    Proceed,
    /// Intercept the request and respond differently.
    Respond(Pin<Box<dyn Future<Output = hyper::Result<hyper::Response<hyper::Body>>> + Send>>),
}

/// Allows to intercept request and handle it differently.
pub trait RequestMiddleware: Send + Sync + 'static {
    /// Takes a request and decides how to proceed with it.
    fn on_request(&self, request: &hyper::Request<hyper::Body>) -> RequestMiddlewareAction;
}

/// Layer that applies [`ProxyGetRequest`] which proxies the `GET /path` requests to
/// specific RPC method calls and that strips the response.
///
/// See [`ProxyGetRequest`] for more details.
#[derive(Clone)]
[...]

PR #1111, opened 2023-05-08
RequestMiddleware

aniketfuryrocks:
Implementation for #1095.

### DIFF
--- a/examples/examples/request_middleware.rs
+++ b/examples/examples/request_middleware.rs
@@ -30,7 +30,8 @@ impl RequestMiddleware for MyMiddleware {
 		if req.uri() == "/foo" {
 			RequestMiddlewareAction::Respond(Box::pin(async { Ok(hyper::Response::new(hyper::Body::from("bar"))) }))
 		} else {
-			RequestMiddlewareAction::Proceed(req)
+			// same as RequestMiddlewareAction::Proceed(req)
+			req.into()
 		}
 	}
 }
@@ -80,10 +81,7 @@ async fn main() -> anyhow::Result<()> {
 	let client = hyper::Client::new();
 
 	assert_eq!("bar", &request_foo_path(&client, &base_uri).await?);
-	assert_eq!(
-		r#"{"jsonrpc":"2.0","result":"hi","id":1}"#,
-		&request_say_hi_method(&client, &base_uri).await?
-	);
+	assert_eq!(r#"{"jsonrpc":"2.0","result":"hi","id":1}"#, &request_say_hi_method(&client, &base_uri).await?);
 
 	Ok(())
 }
--- a/examples/examples/request_middleware.rs
+++ b/examples/examples/request_middleware.rs
@@ -1,0 +1,89 @@
+use hyper::{client::HttpConnector, Client};
+use jsonrpsee::{
+	proc_macros::rpc,
+	server::{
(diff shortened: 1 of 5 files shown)


server/src/middleware/request_middleware.rs
```diff
@@ -0,0 +1,92 @@
...
+	error::Error,
+	future::Future,
+	pin::Pin,
+	sync::Arc,
+	task::{Context, Poll},
+};
+use tower::{Layer, Service};
+
+/// Action undertaken by a middleware.
+pub enum RequestMiddlewareAction {
```
aniketfuryrocks, line 11: The clippy policy requires this to be debug. But it's not possible due to pin.

server/src/middleware/request_middleware.rs
```diff
@@ -0,0 +1,101 @@
...
+impl<F> RequestMiddleware for F
+where
+	F: Fn(hyper::Request<hyper::Body>) -> RequestMiddlewareAction + Sync + Send + 'static,
+{
+	fn on_request(&self, request: hyper::Request<hyper::Body>) -> RequestMiddlewareAction {
+		(*self)(request)
+	}
+}
+
+/// Layer that applies [`RequestMiddlewareService`] which proxies the `GET /path` requests to
```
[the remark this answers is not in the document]
niklasad1: not just `GET /path` if I understand this correctly

server/src/middleware/mod.rs
```diff
@@ -2,3 +2,5 @@
 
 /// Proxy `GET /path` to internal RPC methods.
 pub mod proxy_get_request;
+/// Proxy switch requset
```
niklasad1, line 5: I think `request_middleware` is too vague and might confuse users.

Perhaps something like a `proxy_http_request` and write some better documentation that it 
exposes the HTTP request and it's possible to override the default jsonrpsee handler response.

examples/examples/request_middleware.rs
```diff
@@ -0,0 +1,87 @@
...
+		ServerBuilder,
+	},
+	types::ErrorObject,
+};
+use std::net::SocketAddr;
+
+#[rpc(server)]
+pub trait MyRpc {
+	#[method(name = "sayHi")]
+	fn say_hi(&self) -> Result<String, ErrorObject<'static>>;
```
niklasad1, line 15:
```suggestion
	fn say_hi(&self) -> String;
```

examples/examples/request_middleware.rs
```diff
@@ -0,0 +1,87 @@
...
+}
+
+pub struct MyRpc;
+
+impl MyRpcServer for MyRpc {
+	fn say_hi(&self) -> Result<String, ErrorObject<'static>> {
```
niklasad1, line 21:
```suggestion
	fn say_hi(&self) -> String {
```

niklasad1 (commented):
Thanks for the PR.

The idea is that users can just write their own middleware and but this more generic
than our existing `proxy HTTP get request` so I'm good with it general.

We could even base the `proxy HTTP get request` on top in the middleware you introduced in this PR.

But it's quite hard to understand the difference between `proxy_get_request` and `request_middleware`, so to have this in better documentation is required.

//cc @lexnv @jsdw thoughts?

niklasad1:
We have JSON-RPC specific call middleware which provide something similar and works for both WebSocket and HTTP.

PR #1116, opened 2023-05-11
shutdown server while there is connection does not work

xlc:
I am upgrading jsonrpsee and identified a regression from our tests that shutdown server while there is existing connection doesn't work.

Shouldn't the server automatically terminate the connections when it is being told to shutdown?

This is a failed test to demo this regression.

This is our test that got broken after upgrade https://github.com/AcalaNetwork/subway/blob/8021a32ec385607976597b4c49f09360903a91c4/src/client/tests.rs#L158

### DIFF
--- a/server/src/tests/ws.rs
+++ b/server/src/tests/ws.rs
@@ -837,9 +837,6 @@ async fn close_client_with_pending_calls_works() {
 		assert!(rx.recv().await.is_some());
 	}
 
-	client.close().await.unwrap();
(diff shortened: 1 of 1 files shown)


niklasad1:
There was a bug before that the server didn't really gracefully shutdown i.e, regardless whether it had pending calls or not.
Thus, we changed that if there are pending calls the server will stop (async fn stopped will return then) after those has been completed.

Please open an issue instead

PR #1135, opened 2023-05-26
fix: include error code in RpcLogger

### LINKED ISSUE #1129: Feature Request: have `jsonrpsee::server::logger` return result as RpcResult instead of a string?
I have some prometheus metrics set up on a logger that implements `jsonrpsee::server::logger`. All errors currently increment a single `errors_by_route` metric, but now I'd like to increment different error metrics depending on the response error code. However, the logger only provides a `success: bool` from `on_result`, or a `result: &str` from `on_response`. Could we have `on_response` return the actual RpcResult instead?

niklasad1:
Close #1129 

In addition it contains a bug fix that were introduced by the `IntoResponse` trait and I forgot to check the outcome from the `ResponsePayload` where the user could inject there own JSON-RPC error codes etc.

Thanks to @wlmyng for spotting the bug.

### DIFF
--- a/core/src/server/helpers.rs
+++ b/core/src/server/helpers.rs
@@ -179,7 +179,51 @@ pub struct MethodResponse {
 	/// Serialized JSON-RPC response,
 	pub result: String,
 	/// Indicates whether the call was successful or not.
-	pub success: bool,
+	pub success_or_error: MethodResponseResult,
+}
+
+impl MethodResponse {
+	/// Returns whether the call was successful.
+	pub fn is_success(&self) -> bool {
+		self.success_or_error.is_success()
+	}
+
+	/// Returns whether the call failed.
+	pub fn is_error(&self) -> bool {
+		self.success_or_error.is_success()
+	}
+}
+
+/// Represent the outcome of method call i.e, success or failed.
+///
+/// If the call failed then the JSON-RPC error code is embedded
+/// to be used for logging and metrics.
+#[derive(Debug, Copy, Clone)]
+pub enum MethodResponseResult {
+	/// The method call was successful.
+	Success,
+	/// The method call failed and the error code can used in the RpcLogger.
+	Failed(i32),
+}
+
+impl MethodResponseResult {
+	/// Returns whether the call was successful.
+	pub fn is_success(&self) -> bool {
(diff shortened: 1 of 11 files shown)


tests/tests/metrics.rs
```diff
@@ -154,12 +159,16 @@ async fn ws_server_logger() {
 	let res: Result<String, Error> = client.request("unknown_method", rpc_params![]).await;
 	assert!(res.is_err());
 
+	let res: Result<String, Error> = client.request("err", rpc_params![]).await;
```
niklasad1, line 162: Expanded to test for failed RPC calls too

wlmyng:
Thank you for looking into this!

core/src/server/helpers.rs
```diff
@@ -224,7 +274,7 @@ impl MethodResponse {
 	pub fn error<'a>(id: Id, err: impl Into<ErrorObject<'a>>) -> Self {
 		let err = ResponsePayload::error_borrowed(err);
 		let result = serde_json::to_string(&Response::new(err, id)).expect("JSON serialization infallible; qed");
-		Self { result, success: false }
+		Self { result, success_or_error: MethodResponseResult::Failed(ErrorCode::InternalError.code()) }
```
jsdw, line 277: maybe a dumb question but why does this set an error code which is different from the error code in the `ErrorObject`?

niklasad1: it's a bug, good catch

jsdw (approved):
looks good!

core/src/server/helpers.rs
```diff
@@ -173,13 +173,56 @@ pub fn prepare_error(data: &[u8]) -> (Id<'_>, ErrorCode) {
...
+		self.success_or_error.is_success()
+	}
+}
+
+/// Represent the outcome of a method call success or failed.
+#[derive(Debug, Copy, Clone)]
+pub enum MethodResponseResult {
+	/// The method call was successful.
+	Success,
+	/// The method call with error code.
```
niklasad1, line 202:
```suggestion
	/// The method call failed with error code.
```

lexnv (approved):
LGTM!

PR #1137, opened 2023-05-26
make `Server::start` infallible and add `fn builder()`

niklasad1:
`Server::start` was infallible to let's change the API to reflect that.

### DIFF
--- a/benches/helpers.rs
+++ b/benches/helpers.rs
@@ -125,7 +125,7 @@ pub async fn http_server(handle: tokio::runtime::Handle) -> (String, jsonrpsee::
 	let module = gen_rpc_module();
 
 	let addr = server.local_addr().unwrap();
-	let handle = server.start(module).unwrap();
+	let handle = server.start(module);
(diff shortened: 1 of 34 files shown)


jsdw (approved):
LGTM! The only reason to keep the result would be if we thought it might become fallible again, but best not to anticipate such things anyway :)

niklasad1:
Yeah, I'm quite tempted of removing the entire start API then just start the server from the Builder

lexnv (approved):

jsdw:
Having a guide gander, it looks like you may as well keep a separate thing to start it offhand; I tend to feel that breaking functionality apart like this is a good thing (and then providing higher level stuff to do the common cases where applicable)

(I'd probably add a `Server::builder()` method though as a random thought because that might be a little cleaner than importing a separate `Builder` type to use to build it)

niklasad1:
I will remove the `start API` in another PR.

It's quite messy to migrate all tests.

Mmagkooo (approved):

PR #1138, opened 2023-05-29
refactor(server): remove start API and start the server by `ServerBuilder::build`

niklasad1:
This PR removes the `start API` by removing the old `Server type` with `ServerHandle`

Such that one has to do to start the server now:

 ```rust
    let api = {
        module.register_method("say_hello", |_, _| "hello").unwrap();
        module
    };

    let server = Server::builder().build("127.0.0.1:0", api).await.unwrap();
 ```

Before this commit one had to do:

```rust
    let server = Server::builder().build("127.0.0.1:0").await?;

    let api = {
        module.register_method("say_hello", |_, _| "hello").unwrap();
        module
    };
    
    let handle = server.start(module);

```

Then it's a bit annoying to have the addr and rpc module in the build fn, so we could add it in the builder but then we would need to either force default values or panic if those are missing. I figured it's better to enforce but open to better suggestions.

It would be sufficient to review the changed server code and the rest of the changes are noise.

### DIFF
--- a/benches/helpers.rs
+++ b/benches/helpers.rs
@@ -110,10 +110,10 @@ pub async fn ws_server(handle: tokio::runtime::Handle) -> (String, jsonrpc_ws_se
 
 /// Run jsonrpsee HTTP server for benchmarks.
 #[cfg(not(feature = "jsonrpc-crate"))]
-pub async fn http_server(handle: tokio::runtime::Handle) -> (String, jsonrpsee::server::ServerHandle) {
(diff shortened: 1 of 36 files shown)


jsdw:
So the difference is really just:

```rust
// old
let handle = Server::builder().build("127.0.0.1:0").await?.start(module);
// new
let server = Server::builder().build("127.0.0.1:0", api).await?;
```

Personally I like the separation from building a server to starting it running; it feels good to keep the two things separate and doesn't really make life any easier for people?

PR #1145, opened 2023-06-21
refactor(async client): refactor background task

niklasad1:
Split send and receive to separate tasks to support multiplexing reads and writes

Summary of what was wrong before:

1. A request from frontend came into this task
2. We try to send it off to backend, .awaiting that to complete
4. But! The backend is applying backpressure, oh no! so it won't accept any new messages until some messages are read from it to clear that queue.
3. So this send keeps .awaiting because reading happened in the same task, no reads will happen until the send completes, so essentially we are in a deadlock situation.

### Benches

```bash
group                                           client-multiplex                        master
-----                                           ----------------                        ------
async/ws_batch_requests/fast_call/10            1.00     42.2±4.86µs 231.5 KElem/sec    1.30     54.6±0.85µs 178.7 KElem/sec
async/ws_batch_requests/fast_call/100           1.00   317.8±20.38µs 307.3 KElem/sec    1.12   356.3±19.86µs 274.1 KElem/sec
async/ws_batch_requests/fast_call/2             1.56     52.9±3.86µs 36.9 KElem/sec     1.00     33.8±5.25µs 57.7 KElem/sec
async/ws_batch_requests/fast_call/5             1.00     37.3±5.52µs 130.8 KElem/sec    1.08     40.5±5.66µs 120.7 KElem/sec
async/ws_batch_requests/fast_call/50            1.11    192.4±4.70µs 253.8 KElem/sec    1.00   173.3±33.70µs 281.7 KElem/sec
async/ws_concurrent_conn_calls/fast_call/128    1.05      2.1±0.05ms        ? ?/sec     1.00  1966.4±11.49µs        ? ?/sec
async/ws_concurrent_conn_calls/fast_call/16     1.00  1091.3±89.12µs        ? ?/sec     1.06  1160.0±134.30µs        ? ?/sec
async/ws_concurrent_conn_calls/fast_call/2      1.00  1092.4±112.38µs        ? ?/sec    1.07  1174.0±86.38µs        ? ?/sec
async/ws_concurrent_conn_calls/fast_call/256    1.06      3.9±0.03ms        ? ?/sec     1.00      3.7±0.04ms        ? ?/sec
async/ws_concurrent_conn_calls/fast_call/32     1.00  1204.2±135.04µs        ? ?/sec    1.00  1209.4±143.03µs        ? ?/sec
async/ws_concurrent_conn_calls/fast_call/4      1.02  1241.3±71.76µs        ? ?/sec     1.00  1212.0±104.79µs        ? ?/sec
async/ws_concurrent_conn_calls/fast_call/64     1.05  1217.5±15.00µs        ? ?/sec     1.00  1159.5±10.37µs        ? ?/sec
async/ws_concurrent_conn_calls/fast_call/8      1.01  1186.8±123.04µs        ? ?/sec    1.00  1178.7±73.72µs        ? ?/sec
```


core/src/client/async_client/mod.rs
```diff
@@ -20,8 +20,9 @@ use helpers::{
 	build_unsubscribe_message, call_with_timeout, process_batch_response, process_notification,
 	process_single_response, process_subscription_response, stop_subscription,
 };
-use jsonrpsee_types::{ResponseSuccess, TwoPointZero};
+use jsonrpsee_types::{InvalidRequestId, ResponseSuccess, TwoPointZero};
 use manager::RequestManager;
+use std::sync::Arc;
 
 use async_lock::Mutex;
```
jsdw, line 27: I wonder whether its worth doing `use async_lock::Mutex as AsyncMutex` to make it clear in other places that this is not a sync Mutex?

core/src/client/async_client/manager.rs
```diff
@@ -259,10 +261,11 @@ impl RequestManager {
...
 		match (self.requests.entry(request_id), self.subscriptions.entry(subscription_id)) {
-			(Entry::Occupied(request), Entry::Occupied(subscription))
+			(Entry::Occupied(mut request), Entry::Occupied(subscription))
 				if matches!(request.get(), Kind::Subscription(_)) =>
 			{
-				let (_req_id, kind) = request.remove_entry();
+				// Mark the request ID as pending unsubscription.
+				let kind = std::mem::replace(request.get_mut(), Kind::PendingMethodCall(None));
 				let (sub_id, _req_id) = subscription.remove_entry();
 				if let Kind::Subscription((unsub_req_id, send_back, unsub)) = kind {
```
jsdw, line 270: Nit: we could remove the unreachable by doing something like:

```rust
match (self.requests.entry(request_id), self.subscriptions.entry(subscription_id)) {
    (Entry::Occupied(mut request), Entry::Occupied(subscription))
        if matches!(request.get(), Kind::Subscription(_)) =>
    {
        let Kind::Subscription((unsub_req_id, send_back, unsub)) = request.get() else {
            return None
        };
    }
```

Maybe a bit cleaner :)

core/src/client/async_client/mod.rs
```diff
@@ -726,95 +797,111 @@ async fn background_task<S, R>(
...
 
-	// Place frontend and backend messages into their own select.
-	// This implies that either messages are received (both front or backend),
-	// or the submitted ping timer expires (if provided).
-	let mut message_fut = future::select(frontend.next(), backend_event.next());
-	let mut exit_or_ping_fut = future::select(on_exit, ping_fut(ping_interval));
+	let mut exit_or_ping_fut = future::select(closed, ping_fut(ping_interval));
 
-	loop {
-		match future::select(message_fut, exit_or_ping_fut).await {
```
jsdw, line 746: So this is wherre the issue popped up before then? Offhand this all looks ok because the `select` would poll send and receive to try and make progress on each regardless of whether one of them was blocked.

The way I understood it so far, there's some global lock on send or receive in soketto or something which means that this isn't actually working though, but I don't quite follow how splitting into separate tasks would suddenly allow one to progress when the other is stuck in the same way?

core/src/client/async_client/helpers.rs
```diff
@@ -124,42 +123,42 @@ pub(crate) fn process_subscription_response(
...
-	let request_id = match manager.get_request_id_by_subscription_id(&sub_id) {
-		Some(request_id) => request_id,
+	match manager.get_request_id_by_subscription_id(&sub_id) {
+		Some(request_id) => {
+			manager.remove_subscription(request_id, sub_id).expect("Both request ID and sub ID in RequestManager; qed");
+		}
 		None => {
-			tracing::error!("The server tried to close an invalid subscription: {:?}", sub_id);
-			return Err(Error::InvalidSubscriptionId);
+			tracing::debug!("The server tried to close an non-pending subscription: {:?}", sub_id);
```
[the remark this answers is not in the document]
niklasad1: It's possible that someone "unsubscribes" but the server sends a notification "at the same time", so did this instead of killing the connection if that occurs...

core/src/client/async_client/mod.rs
niklasad1: The "old task would block" until the send went through and no receive would be processed.

So if it's backpressured the receive stuff won't get the change as the background task would be stuck on a send operation.

The new code doesn't hold the mutex over `send operations` just before/after or in the receive task. Otherwise, it would deadlock at the some point.

We could have a separate task for the request manager but it's bit awkward to have two different messages to accomplish that.

niklasad1: No, soketto has a BiLock where the send and receiver can make progress independently on the socket.

core/src/client/async_client/mod.rs
```diff
@@ -513,46 +559,56 @@ impl SubscriptionClientT for Client {
...
 		raw: &[u8],
-		manager: &mut RequestManager,
-		sender: &mut S,
+		manager: &ThreadSafeRequestManager,
+		sender: &mpsc::Sender<FrontToBack>,
 		max_buffer_capacity_per_subscription: usize,
 	) -> Result<(), Error> {
 		let first_non_whitespace = raw.iter().find(|byte| !byte.is_ascii_whitespace());
 
+		tracing::trace!(
```
niklasad1, line 577: remove...

core/src/client/async_client/mod.rs
jsdw: > The new code doesn't hold the mutex over send operations

Where abouts does the old code hold a mutex over send operations? Offhand I can't find it :)

core/src/client/async_client/helpers.rs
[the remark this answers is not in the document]
jsdw: That makes sense, thanks!

core/src/client/async_client/mod.rs
niklasad1: No, it was just the background task that got stuck because of backpressure in the old background task, nothing really "deadlocked". That mutex is only used by the "frontend/client" and the background task doesn't care about it to make progress.

For example [this](https://github.com/paritytech/jsonrpsee/blob/master/core/src/client/async_client/mod.rs#L654-#L660) got stuck when it was backpressured because it's single-threaded and background task would block on that.

jsdw: Aha perfect; thanks! That link was what I needed; I was missing that we are awaiting there :D

jsdw (approved):
Code looks good to me overall, and the separate taks looks pretty clean!

If the `ThreadSafeRequestManager` didn't need to lock over await points then maybe a standard mutex would be a better choice (a safe way to do this would be to make the mutex an integral part of `RequestManager` so that each function only locks internall for as long as it needs and it's impossible to hold a lock over any await points or whatever).

I'd be interested to know whether this has any impact on benchmarks in either direction :)

core/src/client/async_client/helpers.rs
```diff
@@ -124,42 +123,42 @@ pub(crate) fn process_subscription_response(
...
-/// Returns Ok() if the response was successfully handled
-/// Returns Err() if there was no handler for the method
-pub(crate) fn process_notification(manager: &mut RequestManager, notif: Notification<JsonValue>) -> Result<(), Error> {
+/// If the notification is not found it's just logged as a warning and the connection
+/// will continue.
+///
+/// It's possible that user close down the subscription before this notification is received.
+pub(crate) fn process_notification(manager: &mut RequestManager, notif: Notification<JsonValue>) {
 	match manager.as_notification_handler_mut(notif.method.to_string()) {
 		Some(send_back_sink) => match send_back_sink.try_send(notif.params) {
```
lexnv, line 153: style nit (can ignore): `if let Err(err) = send_back_sink.try_send`

core/src/client/async_client/manager.rs
```diff
@@ -95,6 +95,7 @@ pub(crate) struct RequestManager {
 
 impl RequestManager {
 	/// Create a new `RequestManager`.
+	#[allow(unused)]
```
niklasad1, line 98:
```suggestion
```

core/src/client/async_client/mod.rs
```diff
@@ -740,81 +668,188 @@ async fn background_task<S, R>(
...
+
+	tokio::pin!(closed);
 
-	loop {
-		match future::select(message_fut, exit_or_ping_fut).await {
+	let res = loop {
+		match future::select(message_fut, closed).await {
 			// Message received from the frontend.
-			Either::Left((Either::Left((frontend_value, backend)), exit_or_ping)) => {
+			Either::Left((Either::Left((frontend_value, backend)), c)) => {
```
lexnv, line 678: nit: Maybe something like `next_closed_fut` instead of `c`?

core/src/client/async_client/mod.rs
```diff
@@ -726,6 +622,38 @@ async fn background_task<S, R>(
...
+					break Err(Error::Transport(e.into()));
+				}
+			}
+			_ = ping_fut(ping_interval) => {
+				if let Err(e) = sender.send_ping().await {
+					break Err(Error::Transport(e.into()));
+				}
+			}
+			_ = close_tx.closed() => {
+				break Ok(())
```
lexnv, line 642: Could some tracing debug logs be useful on the break branches to know the reason of the `send_task` exit?

core/src/client/async_client/mod.rs
```diff
@@ -740,81 +668,188 @@ async fn background_task<S, R>(
...
-					&mut manager,
-					&mut sender,
-					max_buffer_capacity_per_subscription,
-				)
-				.await;
+				match frontend_value {
+					FrontToBack::Batch(batch) => {
+						if let Err(send_back) = manager.insert_pending_batch(batch.ids.clone(), batch.send_back) {
+							tracing::warn!("[backend]: Batch request already pending: {:?}", batch.ids);
+							let _ =
```
lexnv, line 692: Would a tracing log be of use here? Or could we deduct elsewhere that the batch channel was dropped?

lexnv (approved):

core/src/client/async_client/mod.rs
niklasad1: The channel is only closed if the batch request was dropped but sure we could add some debug logs for that

niklasad1: ah, I'll ignore it in this PR because I would like to do it for all operations or none :P

core/src/client/async_client/mod.rs
niklasad1: I added `Client dropped` logs when` either the send or receive task returns `Ok(())` ok?

These channels should be closed if send/receive tasks are closed, if something terminates with an error that much occur before this and we don't add confusing logs.

However, if it's somewhere that could be tricky detect by looking at the logs.

core/src/client/async_client/mod.rs
```diff
@@ -174,22 +174,19 @@ impl ClientBuilder {
...
-			)
-			.await;
-		});
+		// NOTE: this is a an unbounded channel to ensure that sends never blocks
+		// if the send task is got backpressured or something else.
+		let (msg_tx, msg_rx) = mpsc::unbounded_channel();
+		let (close_tx, close_rx) = mpsc::channel(1);
+
+		tokio::spawn(send_task(sender, msg_rx, ping_interval, close_tx.clone()));
+		tokio::spawn(read_task(receiver, msg_tx, from_front, max_buffer_capacity_per_subscription, close_tx));
```
jsdw, line 183: To check my understanding, it looks like:

- `read_task` reads messages from the frontend _or_ the backend. If they come from the frontend, the messagesa re passed on to `send_task` via an unbounded channel
- `send_task` receives messages from `read_task` via the unbounded channel and sends them to the backend.

So, if I make lots of `request`'s with the client, and the backend isn't ready to receive them yet, then they will just accumulate in the unbounded channel. If we were to use a bounded channel there, then the `read_task` would block because it's the one sending messages to this channel, and we'd end up at square one again where we'd deadlock on reading messages from the backend because `read_task` is blocked trying to send something.

So in this PR it looks like:

```
client ----> read_task <---- backend
                 +
              manager
                 |
                 | (unbounded channel)
                 v
             send_task ----> backend
``` 

Ie client sends request to `read_task`, which stores it in manager to wait for a response, and sends it on to the backend via `send_task`.
When response comes back to `read_task`, manager checked and value given back to user via oneshot.

Here, I think backpressure is lost in the client because an unlimited amount of `requests` can be made and the client will take them all happily and put them into the unbounded channel.

I think the earlier iteration of this with the mutex looked more like this:

```
client ----> from_front ----> backend
                 |
                 v
client <----- manager
                 ^
                 |
             from_back <---- backend
```

In this, both sides have to briefly lock the manager to insert requests or remove requests to respond to, but `from_front` and `from_back` can never block each other; if one is stuck, the other will always try to keep working, and backpressure will go back to the client calling `request` or whatever as needed.

So basically, I think what you did earlier is the "cleaner" approach, and avoids coupling the two tasks (or having an unbounded channel)

jsdw: (and if the mutexed request manager was to become a bottleneck, it should be quite easy to shard the map (ie use soemthing like `dashmap`) so that accesses don't block eachother on each side so much :))

niklasad1: I think your understanding of the code is correct however the old code had still had a coupling between the send/receive task because if the receive task couldn't propagate the subscription to the user i.e, if the subscription was dropped.

Then the client would an unsubscribe call to the server, it was done here: https://github.com/paritytech/jsonrpsee/blob/8478e142d557543627f5cab38101b23c51f1233c/core/src/client/async_client/mod.rs#L861-#L863

But yeah that would be much more unlikely to occur because after the refactoring all goes via that channel...

jsdw: > however the old code had still had a coupling between the send/receive task because if the receive task couldn't propagate the subscription to the user i.e, if the subscription was dropped.

Aah I see, so in the mutex version the receive task would still try sending things to the send task.

I guess I'm wondering whether it makes sense to havea  flow more like what I assumed it was anyway:

```
client ----> from_front ----> backend
                 |
                 v
client <----- manager
                 ^
                 |
             from_back <---- backend
```

That way, there is for sure no way that one of the tasks can ever block the other one from proceeding (at most they will lock the manager very briefly when they need to update things in it).

But maybe I'm missing something :)

niklasad1: I have reverted the latest stuff but still it's not as clean as having everything in the manager. Still the receive task can send unsubscribe messages via the front/back channel but it's not an unbounded channel at least.

In order to have something as clean as you suggested we would need to have shared queue in the RequestManager where stuff can inserted and pop:ed from to send back unsubscribe calls.

It's quite to hard to do unless we split up the RequestManager to use several locks for the data and that will become even harder to maintain I think.....

core/src/client/async_client/manager.rs
niklasad1: yeah but I think I did in this way because the `None` should be unreachable i.e, a logic bug lemme have a look if I broke something or simply remembering this wrong...

niklasad1: ah, the unreachable can be removed done...

core/src/client/async_client/mod.rs
```diff
@@ -726,95 +771,157 @@ async fn background_task<S, R>(
...
+		self.futs.push(fut);
+	}
+}
+
+impl<Fut: Future> Stream for MaybePendingFutures<Fut> {
+	type Item = Fut::Output;
+
+	fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+		if self.futs.is_empty() {
+			return Poll::Pending;
```
jsdw, line 922: At this point, the waker is not handed anywhere and so if any other futures are added, they will be ignored.

I _think_ that this is ok though because you just call `next()` on this in a loop, so it'll get polled at least once per loop anyway.

niklasad1: I tried using pure `FutureUnordered` and then the select loop will be busy that it returns `Poll::Ready i.e None` when it's empty and steal CPU cycles from more important stuff, that's the reason why I added it...

niklasad1: Yeah, if it's empty and futures are added those will read on the next iteration

core/src/client/async_client/mod.rs
```diff
@@ -726,95 +771,157 @@ async fn background_task<S, R>(
...
+	let closed = close_tx.closed();
+	let pending_unsubscribes = MaybePendingFutures::new();
+
+	tokio::pin!(backend_event, closed, pending_unsubscribes);
+
+	let rx_item = backend_event.next();
+	let mut rx_or_closed = future::select(rx_item, closed);
+
+	let res = loop {
+		match future::select(rx_or_closed, pending_unsubscribes.next()).await {
```
jsdw, line 860: I wonder whether it's worth rearranging a little so that we poll, in order, `closed`, then `pending_unsubscribes.next()`, then `rx_item`, to ensure that if lots of items are coming in, we still prioritise handing any close message or pending unsub first?

core/src/client/async_client/mod.rs
jsdw: Ah I see, so when it's empty it'll spam `Poll::Ready(None)` whenever polled; yeah that sucks. Your version looks good to me offhand in the current code now, though it's a little scary that if used wrong it'll just `.await` forever, so a small footgun to be wary of for future us!

core/src/client/async_client/mod.rs
niklasad1: fair enough, maybe we should use select_biased

core/src/client/async_client/mod.rs
lexnv: I believe this should be fine for now, because we use the read_task to drive this:

```
read_task
	select (server_receiver, closed, pending_unsub)
		server_receiver -> if Ok(Some(m)) = handle_backend_messages
			pending_unsub.push( to_send_task.send(m) )
```

One case where this might be problematic is when:
- server_receiver and closed futures are not ready
 - the pending_unsub is empty and thus will return `Poll::Pending` and I believe will never woke up for the current iteration loop because there's nothing that will wake up the `cx.waker` again of the future
- then pending_unsub has some elements added to it

But since`pending_unsub` having elements added is dependent on `server_receiver` being triggered first; it should be fine here as long as we don't save the current `pending_unsub` future that returned `Pending` and instead call next directly in the loop `future::select(rx_or_closed, pending_unsubscribes.next())`

Let me know if my reasoning is right 🙏 

Indeed there might be dragons if we change something in the `read_task` (maybe a comment to not save `pending_unsub.next()` future is enough here).

We may be able to handle this by:
 - if futs.is_empty(), then we could save the cx
 - when adding a new future, if we have `Some(cx)` we could wake up the waker
 - preserve the same pending_unsub.next() between iterations similar to `let rx_item = backend_event.next();`
 
 Not quite sure if worth going that way

core/src/client/async_client/mod.rs
```diff
@@ -726,95 +771,157 @@ async fn background_task<S, R>(
...
+{
+	let ReadTaskParams { receiver, close_tx, to_send_task, manager, max_buffer_capacity_per_subscription } = params;
+
+	let backend_event = futures_util::stream::unfold(receiver, |mut receiver| async {
+		let res = receiver.receive().await;
+		Some((res, receiver))
+	});
+
+	let closed = close_tx.closed();
+	let pending_unsubscribes = MaybePendingFutures::new();
```
lexnv, line 852: I do wonder, is this only used for `unsubscribes`? Wouldn't the `receiver: R` also provide pings and plain messages?

lexnv:
This looks good to me! I believe the backpressure would work fine, nice job here!
Happy to merge this

core/src/client/async_client/mod.rs
niklasad1: No, this is really an edge-case if a subscription fails (such as request ID is already occupied) or that a subscription can't keep it with server all other things are just handled independently of the send and receive tasks by modifying the RequestManager.  

Thus, this really just is a matter to propagate it to the frontend properly.

core/src/client/async_client/mod.rs
niklasad1: yeah, you are correct.

niklasad1: did what you suggested it didn't require much code but instead I had to clone the Waker because `Context` is !Send

core/src/client/async_client/mod.rs
niklasad1: I switched to tokio::select, I was forced to drop the unsubscribe streams in order to push to stream and tokio::select makes it more readable.

Everything is cancel-safe anyway but added a warning in the code.

core/src/client/async_client/mod.rs
jsdw: Ah that's what you did!

PR #1159, opened 2023-07-27
RFC: feat: flatten request

polachok:
I would like to receive request as a single struct instead of a parameter list, because I'd like to do validation on it and pass it further, but it's annoying to write `let req = Request { a, b, c, ... }` every time.

So I propose a new attribute: flatten (like serde(flatten)).
```
#[method(name = "echo", param_kind = map, flatten)]
async fn echo(&self, req: EchoRequest) -> Result<(), Error>
```
makes Request consume entire `params` field.

### DIFF
--- a/proc-macros/src/render_server.rs
+++ b/proc-macros/src/render_server.rs
@@ -127,7 +127,7 @@ impl RpcDescription {
 				// provided `Params` object.
 				// `params_seq` is the comma-delimited sequence of parameters we're passing to the rust function
 				// called..
-				let (parsing, params_seq) = self.render_params_decoding(&method.params, None);
+				let (parsing, params_seq) = self.render_params_decoding(&method.params, None, method.flatten);
 
 				let into_response = self.jrps_server_item(quote! { IntoResponse });
 
@@ -170,7 +170,7 @@ impl RpcDescription {
 				// provided `Params` object.
 				// `params_seq` is the comma-delimited sequence of parameters.
 				let pending = proc_macro2::Ident::new("subscription_sink", rust_method_name.span());
-				let (parsing, params_seq) = self.render_params_decoding(&sub.params, Some(pending));
+				let (parsing, params_seq) = self.render_params_decoding(&sub.params, Some(pending), false);
 				let into_sub_response = self.jrps_server_item(quote! { IntoSubscriptionCloseResponse });
 
 				check_name(&rpc_sub_name, rust_method_name.span());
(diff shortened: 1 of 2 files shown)


niklasad1:
Hey @polachok 

I don't really follow why the `Request` is the annoyance because your fix is in the "proc macro API" and that shouldn't expose the request just the params but most likely you are doing something more complicated that I don't follow...

Can you elaborate with an example what you want to do?

polachok:
Imagine you have many params in api call, so the request looks like this:

```json
{ 
"jsonrpc": "2.0", 
"id": "12345", 
"method": "doTheThing", 
"params": { 
 "first": "AAA", 
 "second": 5884105, 
 "third": "2000", 
 "fourth": 1, 
 "fifth": "BBB",
 "sixth": "CCC", 
 "seventh": "SSS", 
 } 
}
```

And the server method would look like 
```rust
#[method(name = "doTheThing", param_kind = map)]
fn do_the_thing(&self, first: String, second: u64, third: String, fourth: i32, fifth: String, sixth: String, seventh: String) -> ... {
   let thing_request = ThingRequest { first, second, third, fourth, ... }; // kinda annoying!
   thing_request.validate()?;
   do_something(thing_request);
   ...
}
```

I would like to write this instead:
```rust
#[method(name = "doTheThing", param_kind = map, flatten)]
fn do_the_thing(&self, thing_request: ThingRequest) -> ... {
   thing_request.validate()?;
   do_something(thing_request);
   ...
}
```

jsdw:
I think with the current logic, having

```rust
#[method(name = "echo", param_kind = map, flatten)]
async fn echo(&self, req: EchoRequest) -> Result<(), Error>
```

Would mean that _every_ parameter is "flattened" into the params object (or really, un-flattened; `flatten` is really talking about how to serialize nested obejcts into a flattened object).

So eg:

```rust
struct Foo {a: u8, b: u8}
struct Bar{c: bool, d: String}

#[method(name = "echo", param_kind = map, flatten)]
async fn echo(&self, foo: Foo, bar: Bar) -> Result<(), Error>
```

I think this would work too and would expect params like `{ "a": 1, "b": 2, "c": true, "d": "hello" }`.

Probably not a big deal though; normally you'd only flatten one param anyway (I don't see why you'd want a mixture). _possibnly_ we could deliberately restrict to one param for clarity. 

This would also allow eg `HashMap<String, u8>` to be a parameter that would decode any valid numbers; I wonder if that poses any sort of issue or whether it's just a nice thing that's enabled!

PR #1160, opened 2023-07-28
refactor: make ErrorObject::borrowed accept `&str`

aj3n:
If you pass DST reference type as message to `ErrorObject::borrowed`, you'll get error like this:
```
error[E0277]: the size for values of type `str` cannot be known at compilation time
  --> src/main.rs:2:50
   |
2  |     jsonrpsee::types::ErrorObjectOwned::borrowed(0, "ok", None);
   |     --------------------------------------------    ^^^^ doesn't have a size known at compile-time
   |     |
   |     required by a bound introduced by this call
   |
   = help: the trait `Sized` is not implemented for `str`
```
It could be more intuitive to accept such usage.

### DIFF
--- a/types/src/error.rs
+++ b/types/src/error.rs
@@ -73,7 +73,7 @@ impl<'a> ErrorObject<'a> {
 	}
 
 	/// Create a new [`ErrorObject`] with optional data.
-	pub fn borrowed(code: i32, message: &'a impl AsRef<str>, data: Option<&'a RawValue>) -> ErrorObject<'a> {
+	pub fn borrowed(code: i32, message: &'a (impl AsRef<str> + ?Sized), data: Option<&'a RawValue>) -> ErrorObject<'a> {
 		ErrorObject { code: code.into(), message: StdCow::Borrowed(message.as_ref()), data: data.map(StdCow::Borrowed) }
 	}
 
--- a/types/src/error.rs
+++ b/types/src/error.rs
@@ -74,7 +74,7 @@ impl<'a> ErrorObject<'a> {
 
 	/// Create a new [`ErrorObject`] with optional data.
 	pub fn borrowed(code: i32, message: &'a str, data: Option<&'a RawValue>) -> ErrorObject<'a> {
-		ErrorObject { code: code.into(), message: StdCow::Borrowed(message.as_ref()), data: data.map(StdCow::Borrowed) }
+		ErrorObject { code: code.into(), message: StdCow::Borrowed(message), data: data.map(StdCow::Borrowed) }
 	}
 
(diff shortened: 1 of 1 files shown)


types/src/error.rs
```diff
@@ -73,7 +73,7 @@ impl<'a> ErrorObject<'a> {
 	}
 
 	/// Create a new [`ErrorObject`] with optional data.
-	pub fn borrowed(code: i32, message: &'a impl AsRef<str>, data: Option<&'a RawValue>) -> ErrorObject<'a> {
+	pub fn borrowed(code: i32, message: &'a (impl AsRef<str> + ?Sized), data: Option<&'a RawValue>) -> ErrorObject<'a> {
```
jsdw, line 76: Good catch; `&'a AsRef<str>` means that `str == AsRef<str>` if you pass in `&str`, so indeed we need the `?Sized` here. (oftentimes we can just accept `AsRef<str>` ie no reference, if we don't care about the lifetime).

I do think the whole thing should probably just be simpler and accept `&'a str` though; I suspect doing so would make it no harder for users and make the type simpler to follow here!

niklasad1: yeah, please make this `message: &'a str` or just `message: impl AsRef<str>` but unfortunately a breaking change

jsdw: We can't do the latter (` message: impl AsRef<str>`) because we need to capture/use the lifetime from the `str` (I tried it first :D) but yeah just `message: &'a str` makes sense to me!

niklasad1, line 76:
```suggestion
	pub fn borrowed(code: i32, message: &'a str, data: Option<&'a RawValue>) -> ErrorObject<'a> {
```

types/src/error.rs
```diff
@@ -73,7 +73,7 @@ impl<'a> ErrorObject<'a> {
 	}
 
 	/// Create a new [`ErrorObject`] with optional data.
-	pub fn borrowed(code: i32, message: &'a impl AsRef<str>, data: Option<&'a RawValue>) -> ErrorObject<'a> {
+	pub fn borrowed(code: i32, message: &'a str, data: Option<&'a RawValue>) -> ErrorObject<'a> {
 		ErrorObject { code: code.into(), message: StdCow::Borrowed(message.as_ref()), data: data.map(StdCow::Borrowed) }
```
niklasad1, line 77:
```suggestion
		ErrorObject { code: code.into(), message: StdCow::Borrowed(message), data: data.map(StdCow::Borrowed) }
```

niklasad1 (approved):

jsdw (approved):

PR #1163, opened 2023-08-02
feat: get connection id from PendingSubscriptionSink

bmuddha:
adds public associated method to PendingSubscriptionSink, which allows for retrieval of connection identifier, which was used to perform subscription request

motivation: sometimes it's useful to have a link between websocket subscription and websocket connection, e.g. for preventing duplicate subscriptions on the same connection. Currently there's no API exposing underlying connection identifier, so this PR adds a public method just for this purpose.


niklasad1 (approved):

jsdw (approved):
Looks non-contentious to me :)

niklasad1:
yeah, we should probably the same to the `SubscriptionSink` as well for a uniform API :)

PR #1170, opened 2023-08-07
refactor(ws client): remove port from host header

### LINKED ISSUE #1169: Bug: Handshake fails using TLS webSocket handshake: http version was not 1.1
I'm using [subxt](https://github.com/paritytech/subxt) for communication with some parachains, then I got unable to connect to Astar at `wss://rpc.astar.network:443`, when I attempt to connect to it, I get this error:
```
Error in the WebSocket handshake: http version was not 1.1
```

I think this is a bug because astar node is using http 1.1, and I'm able to connect to it using [tokio-tungstenite](https://github.com/snapview/tokio-tungstenite), so probably the issue is somewhere when decoding the packets.

Tracking down the error, it starts here:
https://github.com/paritytech/jsonrpsee/blob/v0.16.2/client/transport/src/ws/mod.rs#L298-L299
the function above uses [soketto](https://github.com/paritytech/soketto/tree/v0.7.1) calling this function:
https://github.com/paritytech/soketto/blob/v0.7.1/src/handshake/client.rs#L100-L115
the error happens when calling self.decode_response()? , which returns the error here:
https://github.com/paritytech/soketto/blob/v0.7.1/src/handshake/client.rs#L171-L173

niklasad1:
Remove optional `port` from host header because some servers doesn't seem to parse it properly. 
See  https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Host for further information.

Close #1169

### DIFF
--- a/client/transport/src/ws/mod.rs
+++ b/client/transport/src/ws/mod.rs
@@ -318,7 +318,7 @@ impl WsTransportClientBuilder {
 
 				let mut client = WsHandshakeClient::new(
 					BufReader::new(BufWriter::new(tcp_stream)),
-					&target.host_header,
+					&target.host,
 					&target.path_and_query,
 				);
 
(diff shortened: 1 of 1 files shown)


niklasad1:
I'm not convinced about this change but want to hear other opinions :)

niklasad1:
Superseeded by #1172

PR #1171, opened 2023-08-07
fix(server): host filtering `*` allow default port

niklasad1:
If one defined `mydomain.com:*` and the client sent `Host: mydomain.com` instead of `Host: mydomain.com:9999` then the request was denied which this PR fixes.

### DIFF
--- a/core/src/server/host_filtering/host.rs
+++ b/core/src/server/host_filtering/host.rs
@@ -32,12 +32,12 @@ use crate::Error;
 const SPLIT_PROOF: &str = "split always returns non-empty iterator.";
 
 /// Port pattern
-#[derive(Clone, Hash, PartialEq, Eq, Debug)]
+#[derive(Clone, Copy, Hash, PartialEq, Eq, Debug)]
 pub enum Port {
 	/// No port specified (default port)
-	None,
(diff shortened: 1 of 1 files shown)


core/src/server/host_filtering/host.rs
```diff
@@ -162,20 +119,42 @@ impl AllowHosts {
...
+fn parse_host(input: &str) -> (String, Port) {
+	// Remove possible protocol definition such as `wss://<hostname>:<port>/<route>`.
+	let with_route = input.split("://").last().expect(SPLIT_PROOF);
+
+	// Remove everything beyond `/` which is part of the route such as `<hostname>:<port>/<route>`.
+	let mut it = with_route.split('/');
+	let host_and_port = it.next().expect(SPLIT_PROOF).to_lowercase();
+
+	// The rest is `host:<optional port>`.
+	let mut it = host_and_port.split(":");
```
jsdw, line 131: Just to check, is `username:password@hostname:port/route` possible also and should the parsing account for that?

niklasad1: I think you are correct, I'll try to refactor this code and I never liked this and I think we are better off using `Uri::parse` here instead.

PR #1172, opened 2023-08-07
refactor(client): support default port number

### LINKED ISSUE #1048: Allow default port in clients
Currently users have to pass a `impl AsRef<str>` to the client builder that includes a port. 

I know there have been other issues (https://github.com/paritytech/jsonrpsee/issues/554, https://github.com/paritytech/subxt/issues/644) about this, and it is marked as blocked on https://github.com/hyperium/http/issues/509. But I would still like to advocate once more for jsonrpsee handling of a default port when passing a URL to it.

As a library user, I expect that I could pass a valid URL to jsonrpsee. The [URL standard](https://url.spec.whatwg.org/#port-state) says that the port should be omitted if it is the default port. This is also what users expect.

The problem is further amplified by rust `url` and `http` crates. If an application takes user input and parses it to an `url::Url`, the default port gets lost, and there is no nice way to force the crate to add it back when converting to string. `http::Uri` does not remove the default port, but is not able to provide the default port for different schemes (`Url` has at least [port_or_known_default](https://docs.rs/url/latest/url/struct.Url.html#method.port_or_known_default)) if none was provided.

**Ideas:**
- Maybe jsonrpsee could do manual port handling that does not rely on https://github.com/hyperium/http/issues/509 (e.g. extracting it with a regex).
- Alternatively we could maybe have a builder method `build_with_components(scheme, host, port, path ...)` which would add some convenience

niklasad1:
Close #1048 and https://github.com/paritytech/jsonrpsee/issues/1169

I switched to the `url` crate which supports default port numbers and once the `Url::parse` fails with `RelativeUrErrorl` on a redirection then it's treated as well-formed. If it fails then connection will be dropped anyway.

### DIFF
--- a/client/transport/src/ws/mod.rs
+++ b/client/transport/src/ws/mod.rs
@@ -349,7 +349,7 @@ impl WsTransportClientBuilder {
 					}
 					Ok(ServerResponse::Redirect { status_code, location }) => {
 						tracing::debug!("Redirection: status_code: {}, location: {}", status_code, location);
-						match url::Url::parse(&location) {
+						match Url::parse(&location) {
 							// redirection with absolute path => need to lookup.
 							Ok(uri) => {
 								// Absolute URI.
--- a/client/transport/src/ws/mod.rs
+++ b/client/transport/src/ws/mod.rs
@@ -27,7 +27,7 @@
 mod stream;
 
 use std::io;
-use std::net::{SocketAddr, ToSocketAddrs};
+use std::net::SocketAddr;
 use std::time::Duration;
 
 use futures_util::io::{BufReader, BufWriter};
@@ -44,6 +44,7 @@ use tokio::net::TcpStream;
 
 pub use http::{uri::InvalidUri, HeaderMap, HeaderValue, Uri};
 pub use soketto::handshake::client::Header;
+pub use url::Url;
 
 /// Sending end of WebSocket transport.
 #[derive(Debug)]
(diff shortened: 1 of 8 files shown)


client/transport/src/ws/mod.rs
```diff
@@ -348,54 +349,49 @@ impl WsTransportClientBuilder {
 					}
 					Ok(ServerResponse::Redirect { status_code, location }) => {
 						tracing::debug!("Redirection: status_code: {}, location: {}", status_code, location);
-						match location.parse::<Uri>() {
+						match url::Url::parse(&location) {
```
niklasad1, line 352:
```suggestion
						match Url::parse(&location) {
```

tests/tests/integration_tests.rs
```diff
@@ -339,27 +339,11 @@ async fn https_works() {
...
+	let client = WsClientBuilder::default().build("wss://kusama-rpc.polkadot.io").await.unwrap();
 	let response: String = client.request("system_chain", rpc_params![]).await.unwrap();
 	assert_eq!(&response, "Kusama");
 }
 
-#[tokio::test]
-async fn ws_with_non_ascii_url_doesnt_hang_or_panic() {
-	init_logger();
-
-	let err = WsClientBuilder::default().build("wss://♥♥♥♥♥♥∀∂").await;
```
niklasad1, line 351: The URL crate regards `non-ascii` chars as a valid URL

jsdw:
Woooo, looks good to me (It's also well tested which is reassuring!)

lexnv (approved):

tadeohepperle (approved):

PR #1174, opened 2023-08-08
refactor: server host filtering

### LINKED ISSUE #1075: host filtering: support `ipv6 addresses`
Currently ipv6 addresses such as `[::1]:8080` won't work because it tries to split the port based on `:` so need to be fixed to work.

niklasad1:
Cleans up the host filtering by supporting `ipv6 addresses`, fixes a bug when host filtering with `*` is configured when a request is missing the default port.

In addition the API on the server is stricter and hosts filtering with invalid authorities are now rejected which wasn't the case before.

Close #1075


core/src/server/host_filtering.rs
```diff
@@ -0,0 +1,243 @@
+// Copyright 2019-2021 Parity Technologies (UK) Ltd.
```
niklasad1, line 1:
```suggestion
// Copyright 2019-2023 Parity Technologies (UK) Ltd.
```

tests/tests/integration_tests.rs
```diff
@@ -1021,10 +1022,14 @@ async fn http_health_api_works() {
 #[tokio::test]
 async fn ws_host_filtering_wildcard_works() {
 	use jsonrpsee::server::*;
+	use std::str::FromStr;
```
niklasad1, line 1025:
```suggestion
```

core/src/server/host_filtering.rs
```diff
@@ -0,0 +1,243 @@
...
+
+impl From<u16> for Port {
+	fn from(port: u16) -> Port {
+		Port::Fixed(port)
+	}
+}
+
+/// Represent the http URI scheme that is returned by the HTTP host header
+///
+/// <http-URI = "http:" "//" authority path-abempty [ "?" query ][ "#" fragment ]>
```
jsdw, line 53: `path-abempty`? :)

niklasad1: Copied it from the spec, I assume it means non-empty :P

core/src/server/host_filtering.rs
```diff
@@ -0,0 +1,243 @@
...
+}
+
+impl FromStr for Authority {
+	type Err = String;
+
+	fn from_str(s: &str) -> Result<Self, Self::Err> {
+		let uri: Uri = s.parse().map_err(|e: InvalidUri| e.to_string())?;
+		let authority = uri.authority().ok_or_else(|| "HTTP Host must contain authority".to_owned())?;
+		let hostname = authority.host();
+		let maybe_port = &authority.as_str()[hostname.len()..];
```
jsdw, line 69: Ah; can't do `authority.port()` because port may bve eg `*` I guess!

core/src/server/host_filtering.rs
```diff
@@ -0,0 +1,243 @@
...
+	hostname: String,
+	port: Port,
+}
+
+impl FromStr for Authority {
+	type Err = String;
+
+	fn from_str(s: &str) -> Result<Self, Self::Err> {
+		let uri: Uri = s.parse().map_err(|e: InvalidUri| e.to_string())?;
+		let authority = uri.authority().ok_or_else(|| "HTTP Host must contain authority".to_owned())?;
```
jsdw, line 67: Nit: a proper `enum` error would prolly be nicer than strings here, since there don't look like many error cases to handle (and lets people push something nicer into their higher level errors when configuring this stuff)

jsdw (approved):
Couple of small nits but looks cleaner to me :)

server/src/server.rs
```diff
@@ -419,9 +419,19 @@ impl<B, L> Builder<B, L> {
 		self
 	}
 
-	/// Sets host filtering.
-	pub fn set_host_filtering(mut self, allow: AllowHosts) -> Self {
-		self.settings.allow_hosts = allow;
+	/// Enables host filtering and allow only the specified hosts.
+	///
+	/// Default: allow all.
+	pub fn host_filter<T: IntoIterator<Item = Authority>>(mut self, allow_only: T) -> Self {
```
jsdw, line 425: Just to copy my element comment:

It could accept anything that can `TryFrom` into the `Authority` type and then it'd have to return the error when `host_filter` is called if there is one. Prolly a bit nicer cos then you can just do

```rust
.host_filter([
    "foo.com",
    "something_else.com:*"
])?
```

for instance which would be quite cool.

(I think it should return an `AuthorityError` enum if the conversion fails personally rather than a `String`)

core/src/server/host_filtering.rs
niklasad1: yes and also `http::Uri::port` won't tell us whether it has a default port or not as well, so we need to do some manual stuff here.

so a little bit hacky but this was cleanest I could come up with

lexnv (approved):

server/src/server.rs
```diff
@@ -419,9 +419,27 @@ impl<B, L> Builder<B, L> {
 		self
 	}
 
-	/// Sets host filtering.
-	pub fn set_host_filtering(mut self, allow: AllowHosts) -> Self {
-		self.settings.allow_hosts = allow;
+	/// Enables host filtering and allow only the specified hosts.
+	///
+	/// Default: allow all.
```
niklasad1:
```suggestion
	/// Default: no host filtering is enabled.
```

server/src/server.rs
```diff
@@ -419,9 +419,27 @@ impl<B, L> Builder<B, L> {
...
+		U: TryInto<Authority, Error = AuthorityError>,
+	{
+		let allow_only: Result<Vec<_>, _> = allow_only.into_iter().map(|a| a.try_into()).collect();
+		self.settings.allow_hosts = AllowHosts::Only(WhitelistedHosts::from(allow_only?));
+		Ok(self)
+	}
+
+	/// Disable host filtering and allow all.
+	///
+	/// Default: allow all.
```
niklasad1, line 440:
```suggestion
	/// Default: no host filtering is enabled.
```

PR #1175, opened 2023-08-09
feat(server): add `SubscriptionSink::connection_id`

niklasad1:
Similar to #1163, added to provide uniform APIs

### DIFF
--- a/core/src/server/subscription.rs
+++ b/core/src/server/subscription.rs
@@ -322,6 +322,11 @@ impl SubscriptionSink {
 		self.method
 	}
 
+	/// Get the connection ID.
+	pub fn connection_id(&self) -> ConnectionId {
+		self.uniq_sub.conn_id
+	}
+
 	/// Send out a response on the subscription and wait until there is capacity.
 	///
 	///


jsdw (approved):

core/src/server/subscription.rs
```diff
@@ -322,6 +322,11 @@ impl SubscriptionSink {
 		self.method
 	}
 
+	/// Get the connection ID.
+	pub fn connection_id(&self) -> ConnectionId {
```
lexnv, line 326: Amazing!

Different topic: Would it also be possible to identify the client connected by this `ConnectionID`? (https://github.com/paritytech/substrate/issues/14734)

lexnv (approved):

core/src/server/subscription.rs
niklasad1: Yes that should work but the connection IDs may be reused once, the connection is dropped and the counter wraps.

I guess that could handled when the subscription is dropped then just remove it from a HashMap or something

PR #1177, opened 2023-08-09
perf: improve subscription message serialization

niklasad1:
Replace `format!` with `push_str`

### DIFF
--- a/core/src/server/subscription.rs
+++ b/core/src/server/subscription.rs
@@ -36,7 +36,7 @@ use jsonrpsee_types::{
 use parking_lot::Mutex;
 use rustc_hash::FxHashMap;
 use serde::{de::DeserializeOwned, Serialize};
-use std::{sync::Arc, time::Duration};
+use std::{fmt::Write, sync::Arc, time::Duration};
 use tokio::sync::{mpsc, oneshot, OwnedSemaphorePermit, Semaphore};
 
 /// Type-alias for subscribers.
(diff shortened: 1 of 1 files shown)


core/src/server/subscription.rs
```diff
@@ -480,15 +480,42 @@ pub(crate) fn sub_message_to_json(
...
+			s.push_str(SUB_HEADER);
+			s.push_str(&sub_id);
+			s.push(',');
+			s.push('"');
+			s.push_str(result_or_err.as_str());
+			s.push('"');
+			s.push(':');
+			s.push_str(&result);
+			s.push('}');
+			s.push('}');
```
jsdw, line 516: I guess you could use `write!(&mut s, "{JSON_RPC_VERSION}{METHOD_HEADER}...")` rather than a lot of pushes to achieve the same; the benefit here is really only that you are doing `String::with_capacity` which is more precise than the one that `format!` can do (I had a look and that will look at the "static" pieces of the string and then double the capacity needed for those if any args (`{}`'s) are given)

Overall though I'd be surprised if there was a big difference doing manual vs `format`?

niklasad1: I got

```bash

Gnuplot not found, using plotters backend
ser_push_str            time:   [277.91 µs 278.53 µs 279.14 µs]
                        change: [-6.7155% -6.4778% -6.2319%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (2.00%) high mild
  1 (1.00%) high severe

ser_format              time:   [441.84 µs 444.11 µs 446.11 µs]
                        change: [+16.180% +16.649% +17.140%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 18 outliers among 100 measurements (18.00%)
  18 (18.00%) high mild


```

niklasad1: but yeah could `with_capacity` that helps lemme try `write!`

jsdw: > I got

Ooh interesting! I had a look at `write!` and it looks like it basically boils down to `push_str` (with other bits) but yeah I'd be interested to see if there was a difference!

niklasad1: ah damn, I had a typo in my format! comparison the string was much bigger :see_no_evil: 


```bash
Gnuplot not found, using plotters backend
ser_push_str            time:   [283.71 µs 285.40 µs 287.71 µs]
                        change: [-6.4897% -6.1258% -5.7465%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 16 outliers among 100 measurements (16.00%)
  9 (9.00%) low severe
  1 (1.00%) low mild
  2 (2.00%) high mild
  4 (4.00%) high severe

ser_format              time:   [295.12 µs 295.24 µs 295.37 µs]
                        change: [-1.0393% -0.9361% -0.8382%] (p = 0.00 < 0.05)
                        Change within noise threshold.
  3 (3.00%) high mild
  2 (2.00%) high severe

ser_write               time:   [36.317 µs 36.339 µs 36.363 µs]
                        change: [+3.4460% +3.9117% +4.3517%] (p = 0.00 < 0.05)
                        Performance has regressed.

```


weird to the current `write!` version is 10x faster .... https://github.com/niklasad1/bench-push-str-format-write/blob/master/bench.rs#L53-#L115

Maybe you can try it on your machine @jsdw? :)

niklasad1:
My bad my benches were faulty and this make very little different; closing

PR #1178, opened 2023-08-10
fix(server): host filter via URI read authority

niklasad1:
It is possible that the HOST is sent via the URI and when that is the case the server should read the `Authority (host:port)` not only the host

### DIFF
--- a/server/src/transport/http.rs
+++ b/server/src/transport/http.rs
@@ -9,7 +9,7 @@ use futures_util::future::Either;
 use futures_util::stream::{FuturesOrdered, StreamExt};
 use hyper::Method;
 use jsonrpsee_core::error::GenericTransportError;
-use jsonrpsee_core::http_helpers::read_body;
+use jsonrpsee_core::http_helpers::{self, read_body};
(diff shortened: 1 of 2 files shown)


server/src/transport/http.rs
```diff
@@ -50,6 +50,20 @@ pub(crate) async fn reject_connection(socket: tokio::net::TcpStream) {
 	}
 }
 
+/// The `Authority` can be sent by the client in the `Host header` or in the `URI`
+/// such that we must check both.
+pub(crate) fn fetch_authority(request: &hyper::Request<hyper::Body>) -> Option<&str> {
+	let host_header = http_helpers::read_header_value(request.headers(), hyper::header::HOST);
+	let uri = request.uri().authority();
```
niklasad1, line 57: This is fix and mainly regards HTTP2 but could be some edge-case as well.

Thus, if one would enable a filter for `port X` and the request is targets port `X` but it's not read and would regarded as the default port if just the `host` is read i.e, not the port

jsdw (approved):

lexnv (approved):

PR #1179, opened 2023-08-11
refactor(server): move host filtering to tower middleware

### LINKED ISSUE #868: Move `host filtering` to a tower layer
We have an access control filtering that verifies both the `host` and `origin`. 

ACL requires the `host` to be present in the headers, while the `origin` is optional.
Both values are checked against an allow list, if they are not present in the allow list the
request is denied.

However, for HTTP 2.0 the `host` is part of the request's `uri`, see  https://github.com/paritytech/jsonrpsee/pull/866.

Move the access control filtering to a tower layer to make this feature optional.

niklasad1:
This PR moves the `HTTP host filter` code to tower middleware and it's optional to use. 

It breaks the functionality somewhat as the old code checked that `host` was included in the request even when host filtering were disabled and now jsonrpsee doesn't check anything in the host header or URI when host filtering is not enabled.

It breaks the current API and the host filtering can only be enabled as middleware in this PR.

Close https://github.com/paritytech/jsonrpsee/issues/868


server/src/middleware/host_filter.rs
```diff
@@ -24,13 +24,79 @@
...
+		let allow_only: Result<Vec<_>, _> = allow_only.into_iter().map(|a| a.try_into()).collect();
+		Ok(Self(WhitelistedHosts::from(allow_only?)))
+	}
+}
+
+impl<S> Layer<S> for HostFilterLayer {
+	type Service = HostFilter<S>;
+
+	fn layer(&self, inner: S) -> Self::Service {
+		HostFilter { inner, filter: self.0.clone() }
```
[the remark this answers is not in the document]
lexnv: I wonder if `HostFilterLayer(Arc<WhitelistedHosts>);` would work here to make this clone more efficient.

I think the `route_recognizer::Router<_>` has a few vectors into it coming from inner structures ([Router](https://github.com/http-rs/route-recognizer/blob/14b94114a83a4e0d4398bce2e77dc501f0f7ab1a/src/lib.rs#L214-L218) and [NFA](https://github.com/http-rs/route-recognizer/blob/14b94114a83a4e0d4398bce2e77dc501f0f7ab1a/src/nfa.rs#L206C1-L210C2))

lexnv (approved):

server/src/middleware/host_filter.rs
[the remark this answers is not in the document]
niklasad1: thanks, good suggestion as it's not mutated after initialization

server/src/middleware/mod.rs
```diff
@@ -1,4 +1,9 @@
 //! Various middleware implementations for RPC specific purposes.
 
+/// HTTP Host filtering middleware.
+mod host_filter;
 /// Proxy `GET /path` to internal RPC methods.
-pub mod proxy_get_request;
+mod proxy_get_request;
+
+pub use host_filter::*;
+pub use proxy_get_request::*;
```
jsdw, line 9: Nit: Given that `host_filter` has types like `Authority` etc in, maybe it's best to expose each one in its own module so it's obvious which types are for what (especially if we ever add any more)?

server/src/transport/http.rs
```diff
@@ -52,14 +53,22 @@ pub(crate) async fn reject_connection(socket: tokio::net::TcpStream) {
 
 /// The `Authority` can be sent by the client in the `Host header` or in the `URI`
 /// such that we must check both.
-pub(crate) fn fetch_authority(request: &hyper::Request<hyper::Body>) -> Option<&str> {
-	let host_header = http_helpers::read_header_value(request.headers(), hyper::header::HOST);
-	let uri = request.uri().authority();
+pub(crate) fn authority(request: &hyper::Request<hyper::Body>) -> Option<Authority> {
```
jsdw, line 56: Is this only used in the middleware now; I wonder whether it should move there?

jsdw (approved):
Looks awesome!

Is it work adding an example for this new middleware to go along with the other middleware examples? :)

tadeohepperle (approved):
Looks good to me.

server/src/transport/http.rs
niklasad1: I moved it to `Authority::from_http_request` I think that's cleaner

server/src/middleware/mod.rs
niklasad1: Added an `uri mod` and simplified the `host_filter mod`, lemme know if you are happy with it? :)

jsdw: Def looks better now that the name `Host` is in all of the exposed things! Personally I liked that the `Authority` stuff lived here (because it's only used in this filter so keeping the code next to it felt good) but I'm easy either way :)

Issue #1146, opened 2023-07-03
regression: `jsonrpsee > v0.16.x` allocates more memory for subscriptions


niklasad1:
`jsonrpsee versions > 0.16.x` allocates more memory because the subscriptions were changed to be async to avoid users to spawn the subscriptions themselves.

The reason is that the `params` in the JSON-RPC method call were forced to be cloned to be able to be sent across thread boundaries i.e, requires it to be `Send + 'static` 

So if one has a subscription with params as input it may allocate lots of more memory when I tried it allocated ~30% more memory.

The idea is to spawn each connection on a single thread such as [tokio::LocalSet](https://docs.rs/tokio/latest/tokio/task/struct.LocalSet.html) to avoid enforcing each future to be `Send + 'static` and hence avoid these annoying needless allocations.

The same applies to `async methods` which is `BoxedFuture<'static, T>`....

niklasad1:
This is fixed by the low-level subscription API which can be used be marking the subscription as "non-async" in proc macros as well

PR #1182, opened 2023-08-14
feat(server): add sync subscription API `register_subscription_raw`

niklasad1:
The motivation is behind this is because this is https://github.com/paritytech/jsonrpsee/issues/1146 if a subscription is having params that are big or happen very frequently it may allocate.

Thus, we now exposes two different APIs for that and our proc macros API will use the lower-level/sync API to parse the params outside async block/callback to avoid that allocation.

After this PR we will have:

- register_method (doesn't allocate for the params)
- register_async_method
- register_subscription_raw (doesn't not allocate for the params)
- register_subscription

### DIFF
--- a/core/src/server/rpc_module.rs
+++ b/core/src/server/rpc_module.rs
@@ -269,7 +269,7 @@ impl Methods {
 	///     use futures_util::StreamExt;
 	///
 	///     let mut module = RpcModule::new(());
-	///     module.register_async_subscription("hi", "hi", "goodbye", |_, pending, _| async {
+	///     module.register_subscription("hi", "hi", "goodbye", |_, pending, _| async {
 	///         let sink = pending.accept().await?;
 	///
 	///         // see comment above.
@@ -352,7 +352,7 @@ impl Methods {
 	///     use jsonrpsee::core::{Error, EmptyServerParams, RpcResult};
 	///
 	///     let mut module = RpcModule::new(());
-	///     module.register_async_subscription("hi", "hi", "goodbye", |_, pending, _| async move {
+	///     module.register_subscription("hi", "hi", "goodbye", |_, pending, _| async move {
 	///         let sink = pending.accept().await?;
 	///         sink.send("one answer".into()).await?;
 	///         Ok(())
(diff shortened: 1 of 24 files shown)


core/src/server/rpc_module.rs
```diff
@@ -754,6 +754,183 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
...
+						id: id.clone().into_owned(),
+						subscribe: tx,
+						permit: conn.subscription_permit,
+					};
+
+					// The subscription callback is a future from the subscription
+					// definition and not the as same when the subscription call has been completed.
+					//
+					// This runs until the subscription callback has completed.
+					let res = callback(params, sink, ctx.clone());
```
lexnv, line 889: To make sure I get this right, this avoids an allocation because we pass the parameters directly to the callback instead of calling `into_owned`, which can allocate quite a bit since the params might either be frequent or large?

core/src/server/rpc_module.rs
```diff
@@ -754,6 +754,183 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
...
+		notif_method_name: &'static str,
+		unsubscribe_method_name: &'static str,
+		callback: F,
+	) -> Result<&mut MethodCallback, Error>
+	where
+		Context: Send + Sync + 'static,
+		F: (Fn(Params, PendingSubscriptionSink, Arc<Context>) -> R) + Send + Sync + Clone + 'static,
+		R: IntoSubscriptionCloseResponse + Send + 'static,
+	{
+		if subscribe_method_name == unsubscribe_method_name {
```
lexnv, line 815: nit: Since the beginning of this is similar to `register_async_subscription` would it be possible to move some code into a shared function? Maybe we could have a `internal_verify_and_unsubscribe` which does the initial validation and registers the unsubscribe?

niklasad1: sure, I just want know what you think about the API but unfortunately a breaking change :(

lexnv: Yep, that makes sense! From my perspective, this gives a nice boost in terms of performance, users that may value more ergonomics still the old method 👍

jsdw (approved):
Looks good to me!

Personally I'd swap the names around sortof, and have something like:
- `register_subscription` - the original one as it was
- `register_subscription_sync`/`register_subscription_raw` - the lower level synchronous/raw method (you still need to `.await` things inside it but can decide how.

I might be tempted to also `#[doc(hidden)]` the `sync`/`raw` one because maybe better if people don't rely on it in the interface and then we can tune it however we need to work best with the macro without worrying about any breakages.

lexnv (approved):

jsdw:
That's cool; `register_subscription_raw` and `register_method_raw` could be the "low level" interfaces then if people want to customise how to do the async stuff, and the current ones stay the same :)

niklasad1:
Yeah, I think that makes sense because this added API isn't really ergonomic to use

PR #1185, opened 2023-08-14
client: downgrade log for unknown subscription to `DEBUG`

xlc:
Why this is a warning? It is not possible to completely eliminate such case and there is not really anything to be worried about when this happens.


niklasad1 (approved):

niklasad1:
Fair enough, it can occur if the client terminates the subscription and until the unsubscription call has been received additonal notifications/subscription messages can be sent out but if this occurs frequently I'm a little worried that is some issue somewhere that's why it was a `warn`

Fine, it's annoying `debug` is better

lexnv (approved):

niklasad1 (approved):

PR #1188, opened 2023-08-18
refactor(server): less allocs per method call

niklasad1:
This PR removes a few needless clones for shared params which are now put in an `Arc` to be able to be shared
across `tokio::spawn`

Seems to improve things slightly as this occurs on every call.


jsdw (approved):
Makes sense to me; I guess none of the htings in params were very big (mostly other Arc-y type things I think) but one arc clone is always going to win :)

niklasad1:
yeah it made things slightly better, as you noted I think it's like doing one Arc::clone instead of five Arc::clones per request.

lexnv (approved):

PR #1192, opened 2023-08-22
feat: max_inactive_interval option for Websocket server

bmuddha:
adds extra option to Websocket server, which can be used to configure forceful disconnection of clients, which are not submitting any requests (including pongs).

### DIFF
--- a/server/src/server.rs
+++ b/server/src/server.rs
@@ -152,7 +152,7 @@ where
 						max_subscriptions_per_connection,
 						batch_requests_config,
 						id_provider: id_provider.clone(),
-						ping_interval: self.cfg.ping_interval,
+						ping_config: self.cfg.ping_config,
 						stop_handle: stop_handle.clone(),
 						conn_id: id,
 						logger: logger.clone(),
@@ -210,14 +210,14 @@ struct Settings {
 	batch_requests_config: BatchRequestConfig,
 	/// Custom tokio runtime to run the server on.
 	tokio_runtime: Option<tokio::runtime::Handle>,
-	/// The interval at which `Ping` frames are submitted.
-	ping_interval: Duration,
 	/// Enable HTTP.
 	enable_http: bool,
 	/// Enable WS.
 	enable_ws: bool,
 	/// Number of messages that server is allowed to `buffer` until backpressure kicks in.
 	message_buffer_capacity: u32,
+	/// Ping settings.
+	ping_config: PingConfig,
 }
 
 /// Configuration for batch request handling.
@@ -231,6 +231,54 @@ pub enum BatchRequestConfig {
 	Unlimited,
 }
 
+/// Configuration for WebSocket ping's.
+///
+/// If the server sends out a ping then remote peer must reply with a corresponding pong message.
+///
+/// It's possible to just send out pings then don't care about response
+/// or terminate the connection if the ping isn't replied to the configured `max_inactivity` limit.
+///
+/// NOTE: It's possible that a `ping` may be backpressured and if you expect a connection
+/// to be reassumed after interruption it's not recommended to enable the activity check.
+#[derive(Debug, Copy, Clone)]
+pub enum PingConfig {
+	/// The server pings the connected clients continuously at the configured interval but
+	/// doesn't disconnect them if no pongs are received from the client.
+	WithoutInactivityCheck(Duration),
+	/// The server pings the connected clients continuously at the configured interval
+	/// and terminates the connection if no websocket messages received from client
+	/// after the max limit is exceeded.
+	WithInactivityCheck {
+		/// Time interval between consequent pings from server
+		ping_interval: Duration,
+		/// Max allowed time for connection to stay idle
+		inactive_limit: Duration,
+	},
+}
+
+impl PingConfig {
+	pub(crate) fn ping_interval(&self) -> Duration {
+		match self {
+			Self::WithoutInactivityCheck(ping_interval) => *ping_interval,
+			Self::WithInactivityCheck { ping_interval, .. } => *ping_interval,
+		}
+	}
+
+	pub(crate) fn inactive_limit(&self) -> Option<Duration> {
+		if let Self::WithInactivityCheck { inactive_limit, .. } = self {
+			Some(*inactive_limit)
+		} else {
+			None
+		}
+	}
+}
+
+impl Default for PingConfig {
+	fn default() -> Self {
+		Self::WithoutInactivityCheck(Duration::from_secs(60))
+	}
+}
+
 impl Default for Settings {
 	fn default() -> Self {
 		Self {
@@ -241,10 +289,10 @@ impl Default for Settings {
 			max_subscriptions_per_connection: 1024,
 			batch_requests_config: BatchRequestConfig::Unlimited,
 			tokio_runtime: None,
-			ping_interval: Duration::from_secs(60),
 			enable_http: true,
 			enable_ws: true,
 			message_buffer_capacity: 1024,
+			ping_config: PingConfig::WithoutInactivityCheck(Duration::from_secs(60)),
 		}
 	}
 }
@@ -368,25 +416,31 @@ impl<B, L> Builder<B, L> {
 		self
 	}
 
-	/// Configure the interval at which pings are submitted.
+	/// Configure the interval at which pings are submitted,
+	/// and optionally enable connection inactivity check
 	///
-	/// This option is used to keep the connection alive, and is just submitting `Ping` frames,
-	/// without making any assumptions about when a `Pong` frame should be received.
+	/// This option is used to keep the connection alive, and can be configured to just submit `Ping` frames or with extra parameter, configuring max interval when a `Pong` frame should be received
 	///
-	/// Default: 60 seconds.
+	/// Default: ping interval is set to 60 seconds and the inactivity check is disabled
 	///
 	/// # Examples
 	///
 	/// ```rust
 	/// use std::time::Duration;
-	/// use jsonrpsee_server::ServerBuilder;
+	/// use jsonrpsee_server::{ServerBuilder, PingConfig};
 	///
-	/// // Set the ping interval to 10 seconds.
-	/// let builder = ServerBuilder::default().ping_interval(Duration::from_secs(10));
+	/// // Set the ping interval to 10 seconds but terminate the connection if a client is inactive for more than 2 minutes
+	/// let builder = ServerBuilder::default().ping_interval(PingConfig::WithInactivityCheck { ping_interval: Duration::from_secs(10), inactive_limit: Duration::from_secs(2 * 60) }).unwrap();
 	/// ```
-	pub fn ping_interval(mut self, interval: Duration) -> Self {
-		self.settings.ping_interval = interval;
-		self
+	pub fn ping_interval(mut self, config: PingConfig) -> Result<Self, Error> {
+		if let PingConfig::WithInactivityCheck { ping_interval, inactive_limit } = config {
+			if ping_interval >= inactive_limit {
+				return Err(Error::Custom("`inactive_limit` must be bigger than `ping_interval` to work".into()));
+			}
+		}
+
+		self.settings.ping_config = config;
+		Ok(self)
 	}
 
 	/// Configure custom `subscription ID` provider for the server to use
@@ -576,8 +630,8 @@ pub(crate) struct ServiceData<L: Logger> {
 	pub(crate) batch_requests_config: BatchRequestConfig,
 	/// Subscription ID provider.
 	pub(crate) id_provider: Arc<dyn IdProvider>,
-	/// Ping interval
-	pub(crate) ping_interval: Duration,
+	/// Ping configuration.
+	pub(crate) ping_config: PingConfig,
 	/// Stop handle.
 	pub(crate) stop_handle: StopHandle,
 	/// Connection ID
@@ -700,8 +754,8 @@ struct ProcessConnection<L> {
 	batch_requests_config: BatchRequestConfig,
 	/// Subscription ID provider.
 	id_provider: Arc<dyn IdProvider>,
-	/// Ping interval
-	ping_interval: Duration,
+	/// Ping config.
+	ping_config: PingConfig,
 	/// Stop handle.
 	stop_handle: StopHandle,
 	/// Max connections,
@@ -769,7 +823,7 @@ fn process_connection<'a, L: Logger, B, U>(
 			max_subscriptions_per_connection: cfg.max_subscriptions_per_connection,
 			batch_requests_config: cfg.batch_requests_config,
 			id_provider: cfg.id_provider,
-			ping_interval: cfg.ping_interval,
+			ping_config: cfg.ping_config,
 			stop_handle: cfg.stop_handle.clone(),
 			conn_id: cfg.conn_id,
 			logger: cfg.logger,
--- a/server/src/transport/ws.rs
+++ b/server/src/transport/ws.rs
@@ -1,13 +1,14 @@
 use std::sync::Arc;
-use std::time::Duration;
+use std::time::{Duration, Instant};
 
 use crate::logger::{self, Logger, TransportProtocol};
 use crate::server::{BatchRequestConfig, ServiceData};
+use crate::PingConfig;
 
-use futures_util::future::{self, Either};
+use futures_util::future::{self, Either, Fuse};
 use futures_util::io::{BufReader, BufWriter};
 use futures_util::stream::{FuturesOrdered, FuturesUnordered};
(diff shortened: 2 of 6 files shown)


niklasad1 (commented):
Thanks for your PR and overall I think this feature is useful but this code could be significantly simplified.

1. Remove the `fn max_inactive_duration` and refactor `ServerBuilder::ping_interval` to take a `ping configuation struct` with an option to enable or disable disconnection if a client doesn't reply to the ping's as input. Similar to what we have for the `BatchRequestConfig` (I reckon that that makes is easier configure this with conflicting intervals for the ping vs disconnect)
2. This is `Arc::new(Notify::new())` shouldn't be needed once the fn try_recv returns SokettoError::Closed then the receiver will be dropped and the send task will be closed because the send operations will fail.
3. Refactor `try_recv` and break up the `recv async block` to get rid of the Mutex in there. The Mutex shouldn't be needed there.

Let us know if you need help or something doesn't work

server/src/transport/ws.rs
```diff
@@ -426,16 +444,26 @@ where
...
+) -> Receive<S>
 where
 	S: Future<Output = ()> + Unpin,
 {
+	let last_active = Mutex::new(Instant::now());
 	let receive = async {
 		// Identical loop to `soketto::receive_data` with debug logs for `Pong` frames.
 		loop {
 			match receiver.receive(data).await? {
 				soketto::Incoming::Data(d) => break Ok(d),
```
niklasad1, line 462: I think it makes sense to take `data messages` into account as well because it possible if some client attempts to send lots of messages the pings may be delayed....

bobs4462:
> Let us know if you need help or something doesn't work

Ok, I think I'm stuck, @niklasad1 can you please elaborate on 

> break up the recv async block to get rid of the Mutex in there

How are we supposed to borrow something mutably in async block (to update last_active variable), and borrow the same variable immutably (to check current value) while mutable borrow has not ended yet? I don't think there's a way to convince the borrow checker without resorting to unsafe

niklasad1:
> How are we supposed to borrow something mutably in async block (to update last_active variable), and borrow the same variable immutably (to check current value) while mutable borrow has not ended yet? I don't think there's a way to convince the borrow checker without resorting to unsafe

I mean that you could "break" up the async block and update the instant what that returns something like:


```rust
/// Attempts to read data from WebSocket fails if the server was stopped.
async fn try_recv<S>(
	receiver: &mut Receiver,
	data: &mut Vec<u8>,
	mut stopped: S,
	max_inactive_interval: Duration,
	disconnect_notify: Arc<Notify>,
) -> Receive<S>
where
	S: Future<Output = ()> + Unpin,
{
	let mut last_active = Instant::now();

	enum Op {
		Data,
		Ping,
	}

	let receive = async {
		// Identical loop to `soketto::receive_data` with debug logs for `Pong` frames.

		match receiver.receive(data).await? {
			soketto::Incoming::Data(_) => Ok(Op::Data),
			soketto::Incoming::Pong(_) => Ok(Op::Ping),
			soketto::Incoming::Closed(_) => {
				// The closing reason is already logged by `soketto` trace log level.
				// Return the `Closed` error to avoid logging unnecessary warnings on clean shutdown.
				Err(SokettoError::Closed)
			}
		}
	};

	let mut inactivity_check = Box::pin(tokio::time::sleep(max_inactive_interval));

	tokio::pin!(receive);

	let mut futs = futures_util::future::select(receive, inactivity_check);

	loop {
		match futures_util::future::select(futs, stopped).await {
			Either::Left((Either::Left((Ok(Op::Data), _)), s)) => break Receive::Ok(s),
			Either::Left((Either::Left((Ok(Op::Ping), inactive)), s)) => {
				last_active = Instant::now();
				stopped = s;
				futs = futures_util::future::select(receive, inactive);
			}
			Either::Left((Either::Right((_instant, rcv)), s)) => {
				if last_active.elapsed() > max_inactive_interval {
					disconnect_notify.notify_one();
					break Receive::Err(SokettoError::Closed, s);
				}
				inactivity_check = Box::pin(tokio::time::sleep(max_inactive_interval));

				stopped = s;
				futs = futures_util::future::select(rcv, inactivity_check);
			}
			Either::Left((Either::Left((Err(e), _)), s)) => break Receive::Err(e, s),
			Either::Right(_) => break Receive::Shutdown,
		}
	}
}
```

bobs4462:
https://github.com/paritytech/jsonrpsee/pull/1192#issuecomment-1690398405 
```rust
// receive is borrowed here for the entire lifetime of the loop below
let mut futs = futures_util::future::select(receive.next(), inactivity_check); 

	loop {
		match futures_util::future::select(futs, stopped).await {
			Either::Left((Either::Left((None, _)), s)) => break Receive::Ok(s),
			Either::Left((Either::Left((Some(Ok(_)), inactive)), s)) => {
				last_active = Instant::now();
				stopped = s;
				// it's impossible to re-borrow it inside the loop 
				futs = futures_util::future::select(receive.next(), inactive); 
			}
```
there's a problem with this code as well, borrow checker is not smart enough to understand that receiver is no longer borrowed if the corresponding future completes, and refuses to compile such a code

niklasad1:
it compiles just fine on my machine, which rustc toolchain are you using?

bobs4462:
tried it on 1.67.1, it compiles on 1.7.1 though, borrow checker got smarter it seems :)

server/src/server.rs
```diff
@@ -218,6 +219,9 @@ struct Settings {
 	enable_ws: bool,
 	/// Number of messages that server is allowed to `buffer` until backpressure kicks in.
 	message_buffer_capacity: u32,
+	/// Maximum allowed time duration for connection to be inactive,
+	/// after which it will be forcefully terminated
+	max_inactive_duration: Duration,
```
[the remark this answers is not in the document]
bobs4462: Done

bobs4462:
All required changes were implemented, but tests keep failing, rustdoc uses older version of rustc, and fails to compile the code. Also there's some kind of ui tests which fail as well.

server/src/server.rs
```diff
@@ -231,6 +235,26 @@ pub enum BatchRequestConfig {
 	Unlimited,
 }
 
+/// Configuration for batch request handling.
```
niklasad1, line 238:
```suggestion
/// Configuration for WebSocket ping's a.k.a "alive-check"
///
/// If the server sends out a ping then remote peer must reply with a corresponding pong message.
///
/// It's possible to disable pings completely, just send out pings then don't care about response or terminate the connection if the ping isn't replied to in `max_inactivity limit.
```

server/src/server.rs
```diff
@@ -231,6 +235,26 @@ pub enum BatchRequestConfig {
...
+#[derive(Debug, Copy, Clone)]
+pub enum PingConfig {
```
niklasad1, line 240: please add `Disabled` here as well

server/src/server.rs
```diff
@@ -382,10 +408,17 @@ impl<B, L> Builder<B, L> {
 	/// use jsonrpsee_server::ServerBuilder;
 	///
 	/// // Set the ping interval to 10 seconds.
-	/// let builder = ServerBuilder::default().ping_interval(Duration::from_secs(10));
+	/// let builder = ServerBuilder::default().ping_interval(PingConfig::Periodic(Duration::from_secs(60)));
 	/// ```
-	pub fn ping_interval(mut self, interval: Duration) -> Self {
-		self.settings.ping_interval = interval;
+	pub fn ping_interval(mut self, config: PingConfig) -> Self {
```
niklasad1, line 413: handle the error here instead of `fn build` below.

```suggestion
	pub fn ping_interval(mut self, config: PingConfig) -> Result<Self, Error> {
```

server/src/server.rs
niklasad1: nah let's do in another PR

niklasad1:
> All required changes were implemented, but tests keep failing, rustdoc uses older version of rustc, and fails to compile the code. Also there's some kind of ui tests which fail as well.

I pushed two commits to your branch but yeah we are using an old rustc version for the rustdoc check so don't worry about that one. It will be fixed by https://github.com/paritytech/jsonrpsee/pull/1194

server/src/server.rs
bobs4462: Should we make ping_interval optional then?

niklasad1: What do you mean by "optional"?

`ping_interval` has a default value and anyone that wants another behavior can just call `ping_interval(PingConfig)` but I guess you mean disable it by default?

niklasad1:
@bobs4462 looks good now but please revert the UI tests changes and merge to master then I believe the CI should be happy.

proc-macros/tests/ui/incorrect/rpc/rpc_deprecated_method.stderr
```diff
@@ -1,11 +1,11 @@
-error: use of deprecated method `DeprecatedClient::async_method`: please use `new_method` instead
-  --> $DIR/rpc_deprecated_method.rs:63:20
+error: use of deprecated associated function `DeprecatedClient::async_method`: please use `new_method` instead
```
niklasad1, line 1: revert

server/src/server.rs
```diff
@@ -231,6 +231,50 @@ pub enum BatchRequestConfig {
...
+/// If the server sends out a ping then remote peer must reply with a corresponding pong message.
+///
+/// It's possible to just send out pings then don't care about response
+/// or terminate the connection if the ping isn't replied to the configured `max_inactivity limit.
+#[derive(Debug, Copy, Clone)]
+pub enum PingConfig {
+	/// The server pings the connected clients continuously at configured interval but
+	/// doesn't disconnect if no pongs are received from the client.
+	WithoutInactivityCheck(Duration),
+	/// The server pings the connected clients with continuously at configured interval
```
niklasad1, line 245:
```suggestion
	/// The server pings the connected clients continuously at configured interval
```

bobs4462:
Ok, I've rebased to master, reverted ui tests, fixed the comments, tests are passing now.

server/src/transport/ws.rs
```diff
@@ -426,31 +427,61 @@ where
...
+			Ok(soketto::Incoming::Closed(_)) => Some((Err(SokettoError::Closed), (receiver, data))),
+			// The closing reason is already logged by `soketto` trace log level.
+			// Return the `Closed` error to avoid logging unnecessary warnings on clean shutdown.
+			Err(e) => Some((Err(e), (receiver, data))),
 		}
-	};
+	});
+
+	let mut inactivity_check =
+		Box::pin(ping_config.inactive_limit().map(|d| tokio::time::sleep(d).fuse()).unwrap_or_else(Fuse::terminated));
```
niklasad1, line 448: I'm not super excited about `Box::pin(tokio::time::sleep(...))`, would be nicer with `tokio::time::Interval (which measures the time between ticks)` then we could just `reset` the interval if a message is received and once a `tick` expired then we know that `max_limit` has expired and just terminate without any extra checks.

However, that would require to set `tokio::time::interval(Duration::MAX)` when the inactivity check is disabled which I'm fine with

server/src/transport/ws.rs
```diff
@@ -278,14 +279,16 @@ pub(crate) async fn background_task<L: Logger>(sender: Sender, mut receiver: Rec
...
-		// TCP retransmission mechanism will take of the rest and adjust the window size accordingly.
+		// TCP retransmission mechanism will take care of the rest and adjust the window size accordingly.
 		let Some(stop) = wait_until_connection_buffer_has_capacity(&sink, stopped).await else {
 			break Ok(Shutdown::ConnectionClosed);
 		};
 
 		stopped = stop;
 
-		match try_recv(&mut receiver, &mut data, stopped).await {
+		let receive = try_recv(&mut receiver, &mut data, stopped, ping_config);
```
niklasad1, line 289: please revert this extra line, it doesn't make the code more readable

server/src/server.rs
```diff
@@ -231,6 +231,51 @@ pub enum BatchRequestConfig {
...
 
+/// Configuration for WebSocket ping's.
+///
+/// If the server sends out a ping then remote peer must reply with a corresponding pong message.
+///
+/// It's possible to just send out pings then don't care about response
+/// or terminate the connection if the ping isn't replied to the configured `max_inactivity` limit.
+#[derive(Debug, Copy, Clone)]
+pub enum PingConfig {
+	/// The server pings the connected clients continuously at configured interval but
```
niklasad1, line 242:
```suggestion
	/// The server pings the connected clients continuously at the configured interval but
```

server/src/server.rs
```diff
@@ -231,6 +231,51 @@ pub enum BatchRequestConfig {
...
+	/// doesn't disconnect if no pongs are received from the client.
+	WithoutInactivityCheck(Duration),
+	/// The server pings the connected clients continuously at configured interval
```
niklasad1, line 245:
```suggestion
	/// The server pings the connected clients continuously at the configured interval
```

server/src/server.rs
```diff
@@ -231,6 +231,51 @@ pub enum BatchRequestConfig {
 	Unlimited,
 }
 
+/// Configuration for WebSocket ping's.
+///
+/// If the server sends out a ping then remote peer must reply with a corresponding pong message.
+///
+/// It's possible to just send out pings then don't care about response
+/// or terminate the connection if the ping isn't replied to the configured `max_inactivity` limit.
+#[derive(Debug, Copy, Clone)]
```
niklasad1, line 240:
```suggestion
///
/// NOTE: It's possible that a `ping` may be backpressured and if you except a connection to reassumed 
/// after interruption it's not recommended to enable the activity check.
#[derive(Debug, Copy, Clone)]
```

server/src/transport/ws.rs
bobs4462: Well, yes, that would have been more ergonomic way of doing things, but with Interval::tick, one encounters problems with borrow checker, as `interval` would be borrowed in `futures::select`, it would be impossible to call `reset` on it. I've rewritten it using `tokio::time::interval`, but still with extra checks because of borrow checker problems.

server/src/server.rs
bobs4462: Done

server/src/server.rs
bobs4462: Done

server/src/server.rs
bobs4462: Done

bobs4462:
Now tests are failing, because we cannot use Duration::MAX when initializing interval, using smaller value helps (like 10 years), but I'm not sure whether this arbitrary duration is a good solution for this issue...

server/src/transport/ws.rs
```diff
@@ -379,7 +380,8 @@ async fn send_task(
 			}
 
 			// Handle timer intervals.
-			Either::Right((Either::Left((Some(_instant), stop)), next_rx)) => {
+			Either::Right((Either::Left((_instant, _stopped)), next_rx)) => {
+				stop = _stopped;
```
lexnv, line 384: nit: Are we adding this line for better readability?

lexnv (commented):
Lovely PR! Thanks for contributing!🙏 

From my side, everything looks solid. Since the tests are failing, it might be a good time to add another variant: `PingConfig::Disable`.

This would change the `PingConfig::ping_interval -> Option<Duration>` API, or similar. Knowing this we could utilize `Fuse::<Delay>::terminated()` for `Disable` states similarly as before

niklasad1:
> Now tests are failing, because we cannot use Duration::MAX when initializing interval, using smaller value helps (like 10 years), but I'm not sure whether this arbitrary duration is a good solution for this issue...

Yeah, you are running into the 2038 problem whereas tokio is using `Instant::now` under the hood sorry missed that.
Just revert that last commit and `tokio::time::sleep` is good enough for now.

FWIW, `tokio::time::interval` yields a tick immediately and you should use `tokio::time::interval_at` instead :)

niklasad1:
> From my side, everything looks solid. Since the tests are failing, it might be a good time to add another variant: PingConfig::Disable.

> This would change the PingConfig::ping_interval -> Option<Duration> API, or similar. Knowing this we could utilize Fuse::<Delay>::terminated() for Disable states similarly as before

It's a good a idea but requires a bunch of changes as the send task is also using, so let's do it another PR
Created an issue for it yesterday https://github.com/paritytech/jsonrpsee/issues/1195

bobs4462:
All is set it seems, if nothing else should be fixed/improved, I suppose we can merge this PR.

niklasad1 (approved):

server/src/transport/ws.rs
niklasad1: no, ideally it should be removed :)

lexnv (approved):
Nice one! Thanks for contributing! 👍

PR #1203, opened 2023-09-15
refactor(server): remove needless clone in `ws background task`

niklasad1:


server/src/transport/ws.rs
```diff
@@ -492,18 +489,15 @@ struct ExecuteCallParams<L: Logger> {
 	max_log_length: u32,
 	sink: MethodSink,
 	logger: L,
+	bounded_subscriptions: BoundedSubscriptions,
 }
 
-async fn execute_unchecked_call<L: Logger>(
-	params: Arc<ExecuteCallParams<L>>,
-	data: Vec<u8>,
-	bounded_subscriptions: BoundedSubscriptions,
```
lexnv, line 500: Nice catch 👍 

Would be interesting to see how big the performance improvement is, since I would expect the `BoundedSubscriptions` clone to be cheap (contains an `Arc<..>` and a u64)

lexnv (approved):

server/src/transport/ws.rs
niklasad1: yeah, shouldn't matter in practice but found it when I was debugging stuff :P

jsdw (approved):

PR #1204, opened 2023-09-15
fix(server): fix leak in FuturesUnordered

### LINKED ISSUE #1200: High memory consumption after 0.16.x
It look like #1150 .

It will use 5MB memory, when I send every 3000 message by a WebSocket connection.

Test way:
```js
var rpc_data = {
	"jsonrpc": "2.0",
	"method": "hello",
	"params": [],
	"id": 2
};
var ws = new WebSocket("ws://192.168.2.3:3030");

ws.onopen = ()=>{
	console.log("open ws");
	ws.send(JSON.stringify(rpc_data));
};
ws.onmessage = (data)=>{
	console.log("recv");
	ws.send(JSON.stringify(rpc_data));
}
```

niklasad1:
FuturesUnordered was not getting polled until shutdown, so it was accumulating tasks forever. The change uses a channel instead to wait until all senders are closed before shutdown can happen.

Reduces the memory usage significantly but still slightly worse than v0.16.x, around 2x but that should be okay.

Close #1200

### DIFF
--- a/server/src/transport/ws.rs
+++ b/server/src/transport/ws.rs
@@ -250,8 +250,8 @@ pub(crate) async fn background_task<L: Logger>(sender: Sender, mut receiver: Rec
 	let bounded_subscriptions = BoundedSubscriptions::new(max_subscriptions_per_connection);
 
 	// On each method call the `pending_calls` is cloned
-	// then all pending_calls are dropped then a graceful shutdown
-	// has occured.
+	// then when all pending_calls are dropped
+	// a graceful shutdown can has occur.
 	let (pending_calls, pending_calls_completed) = mpsc::channel::<()>(1);
 
 	// Spawn another task that sends out the responses on the Websocket.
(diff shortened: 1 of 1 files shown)


jsdw (approved):
LGTM; nice catch! FuturesUnordered was not getting polled until shutdown, so it was accumulating tasks forever. The change uses a channel instead to wait until all senders are closed before shutdown can happen.

server/src/transport/ws.rs
```diff
@@ -550,6 +559,8 @@ async fn execute_unchecked_call<L: Logger>(params: Arc<ExecuteCallParams<L>>, da
 			_ = params.sink.send_error(Id::Null, ErrorCode::ParseError.into()).await;
 		}
 	};
+
+	drop(drop_on_completion);
```
niklasad1, line 563:
```suggestion
    // NOTE: This channel is only used to indicate that a method call was completed
    // thus the drop here tells the main task that method call was completed.
	drop(drop_on_completion);
```

lexnv (approved):

PR #1213, opened 2023-10-11
feat(middleware): add `HostFilterLater::disable`

niklasad1:
This is just a convenience method to make it easier to use in some scenarios because when enabled it will generate another type

Such as this example won't compile:

 ```rust
    let middleware = if host_filter {
        tower::ServiceBuilder::new()
            .layer(HostFilterLayer::new(["example.com"]).unwrap())
            .layer(ProxyGetRequestLayer::new("/health", "system_health").unwrap())
    } else {
        tower::ServiceBuilder::new()
    };
```

### DIFF
--- a/server/src/middleware/host_filter.rs
+++ b/server/src/middleware/host_filter.rs
@@ -39,7 +39,7 @@ use tower::{Layer, Service};
 
 /// Middleware to enable host filtering.
 #[derive(Debug)]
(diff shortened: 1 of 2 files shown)


lexnv (approved):
LGTM!

jsdw (approved):
Maybe an argument for not exposing the ServiceBuilder stuff at all and accepting layers instead or something, but yup lgtm :)

Issue #871, opened 2022-09-07
[ws client]: doesn't wait for `Close ACK` from server during close handshake


niklasad1:
From what I understand [soketto::close](https://docs.rs/soketto/0.7.1/soketto/connection/struct.Sender.html#method.close) doesn't 
wait for `CLOSE ACK` from the server which causes  the `ACK` to fail from the server-side.

That certainly explains error such as:
`2022-09-07T10:18:15.621704Z ERROR jsonrpsee_server::transport::ws: WS transport error: i/o error: Transport endpoint is not connected (os error 107); terminate connection: 0` which I thought was some race-condition is soketto.

The downside is that soketto doesn't have an API to do that on the client-side AFAIU.

//cc @lexnv

niklasad1:
Should be fixed by https://github.com/paritytech/soketto/pull/115

PR #1216, opened 2023-10-13
server: graceful shutdown check `Incoming::Closed`

niklasad1:


server/src/transport/ws.rs
```diff
@@ -564,10 +565,13 @@ async fn graceful_shutdown(
...
 			let disconnect_stream = futures_util::stream::unfold((receiver, data), |(mut receiver, mut data)| async {
-				if let Err(SokettoError::Closed) = receiver.receive(&mut data).await {
-					None
-				} else {
-					Some(((), (receiver, data)))
+				match receiver.receive(&mut data).await {
+					Ok(Incoming::Closed(_)) | Err(SokettoError::Closed) => None,
+					Ok(Incoming::Data(_) | Incoming::Pong(_)) => Some(((), (receiver, data))),
+					Err(e) => {
+						tracing::warn!("Graceful shutdown got WebSocket error: {e} but polling until the connection closed or all pending calls has been executed");
```
niklasad1, line 572:
```suggestion
						tracing::warn!("Graceful shutdown got WebSocket error: {e} but polling until the connection is closed or all pending calls has been executed");
```

lexnv (approved):

server/src/transport/ws.rs
```diff
@@ -564,10 +565,13 @@ async fn graceful_shutdown(
...
 			// The receiver is not cancel-safe such that it's used in a stream to enforce that.
 			let disconnect_stream = futures_util::stream::unfold((receiver, data), |(mut receiver, mut data)| async {
-				if let Err(SokettoError::Closed) = receiver.receive(&mut data).await {
-					None
-				} else {
-					Some(((), (receiver, data)))
+				match receiver.receive(&mut data).await {
+					Ok(Incoming::Closed(_)) | Err(SokettoError::Closed) => None,
+					Ok(Incoming::Data(_) | Incoming::Pong(_)) => Some(((), (receiver, data))),
+					Err(e) => {
```
jsdw, line 571: Just to check; do you no longer need to return `None` when `Err(SokettoError::Closed)` is a thing now?

niklasad1: We will keep the stream alive until either `SokettoError::Closed` or `Incoming::Closed` is returned back because some implementations may not do the proper `WebSocket close handshake` message such as the soketto client https://github.com/paritytech/jsonrpsee/issues/871. Thus, we need to check both

However, soketto will eventually return SokettError::Closed after this https://github.com/paritytech/soketto/blob/master/src/connection.rs#L386-L388 so this extra check in this PR shouldn't matter in "practice" but we need to call receive one more time instead then also I like us to be explicit here which makes it easier to understand.

FWIW, it may also a bunch of I/O errors that could terminate the connection as well but those are super-tricky to detect so we are relying on the `close stuff`...

jsdw: Oohh I actually missed that you were still paying attention to `SokettoError::Closed` anyway, so yup all makes sense to me now :)

jsdw (approved):

PR #1218, opened 2023-10-18
graceful shutdown: terminate on error

niklasad1:
This fixes a `soketto::Receiver` drop issue where two streams were created and this PR uses the same stream for both receiving messages and shutdown.

In addition, this PR reverts the behavior to keep the socket alive despite errors and the server will only perform the graceful shutdown if a shutdown was requested. 

master:

```bash
2023-10-19T08:46:52.456649Z  INFO jsonrpsee_server::transport::ws: server is closed because: Ok(Stopped)
// This is not a message that is sent by the client i.e, read as empty......
2023-10-19T08:46:52.456664Z TRACE soketto::connection: 5627ceca: recv: (Continue (fin 0) (rsv 000) (mask (0 0)) (len 0))
2023-10-19T08:46:52.456671Z DEBUG soketto::connection: 5627ceca: continue frame while not processing message fragments
2023-10-19T08:46:52.456678Z  WARN jsonrpsee_server::transport::ws: Graceful shutdown got WebSocket error: unexpected opcode: Continue but polling until the connection closed or all pending calls has been executed
// Websocket Close handshake .... 
```

This branch:

```bash
2023-10-19T08:52:19.060822Z  INFO jsonrpsee_server::transport::ws: server is closed because: Ok(Stopped)
2023-10-19T08:52:19.162134Z TRACE method_call{method="sleep_20s"}: jsonrpsee_core::tracing: send="{\"jsonrpc\":\"2.0\",\"id\":10,\"method\":\"sleep_20s\"}"
2023-10-19T08:52:19.162269Z TRACE jsonrpsee_client_transport::ws: send: {"jsonrpc":"2.0","id":10,"method":"sleep_20s"}
2023-10-19T08:52:19.162301Z TRACE soketto::connection: 2425d827: send: (Text (fin 1) (rsv 000) (mask (1 1e4e0b66)) (len 46))
2023-10-19T08:52:19.162341Z TRACE soketto::connection: 2425d827: Sender flushing connection
2023-10-19T08:52:19.162486Z TRACE soketto: read 2 bytes
2023-10-19T08:52:19.162511Z TRACE soketto: read 4 bytes
2023-10-19T08:52:19.162524Z TRACE soketto::connection: 33584b71: recv: (Text (fin 1) (rsv 000) (mask (1 1e4e0b66)) (len 46))
// Websocket Close handshake happend but no empty message is read above....
```

On errors we just terminate the connection for simplicity as it did before.

PR #1220, opened 2023-10-19
graceful shutdown: distinguish between stopped and conn closed

niklasad1:
Follow-up on #1218 

Basically, I missed that we must ensure that the stream is not polled after `Poll::Ready(None)` which can happen
if the stream is closed but since `Receive::Shutdown was mapped to Shutdown::Stopped` where the connection could already have been closed and thus the stream had already return `Poll::Ready(None)`


lexnv (approved):

jsdw (approved):
I think I follow; I wonder whether `fuse()`-ing the stream is a safe way to ensure it won't be polled again or whatever, rather than any extra types, but code looks good to me!

PR #1224, opened 2023-10-23
feat: jsonrpsee as `service` and `low-level API for more fine-grained API to disconnect peers etc`

### LINKED ISSUE #1000: Be able to drop connection in RPC handler
So we can implement rate limiting feature to drop connection if client is sending too many requests

### LINKED ISSUE #1014: Get connection ID in rpc handler

### LINKED ISSUE #1016: server: expose RPC handler/JsonrpseeTowerService
Currently the server has its own `TowerService` which isn't exposed to external users which makes it not possible to fetch connection related details along with access to the JSON-RPC methods stuff.

Thus, the idea is to have a builder on or something similar which makes it possible to just get this JsonrpseeTowerService such that users can start their own hyper server without having to write tower middleware.

Such as

```rust
    // This API needs to be added
    let service = jsonrpsee::server::Builder::service();

   let wrapper_service = make_service_fn(move |conn: &AddrStream| {
       // write your logic here and utilize service
       
   });

   let listener = TcpListener::bind(&addr)?;

   loop {
        let (socket, _) = listener.accept().await?;
        tokio::spawn(try_accept_connection(socket, wrapper_service));
    }
```

niklasad1:
This PR adds two new APIs (in order for users to manage the state themselves):

1. "low-level API to jsonrpsee": to give the ability to terminate connections
2. "jsonrpsee as a service": to expose the jsonrpsee tower service that may be used to get access to the HTTP request and to implement full functionality of the old RpcLogger, this API can be used.

### Limitations

The API to manage the websocket connection is quite high-level, such that this design doesn't allow to disconnect peers because of lower-level WebSocket details such as if the ping wasn't answered in time, send took longer than some hard limit, message buffer was exceeded x times, and similar.

### Annoyance

As demonstrated in the added examples, it's a bit annoying for folks to bother with "server handle and stop handles" but I don't want to make such big refactor for that.

To implement "grafana metrics or something equivalent to the old RpcLogger" then the jsonrpsee as service must be used which is a bit harder to implement I reckon.

### Examples

Two examples of how to use the `hyper::service fn` to utilize the jsonrpsee TowerService

1. Utilize the jsonrpsee::TowerService to get access to the HTTP request (example/jsonrpsee_service.rs)  
2. Utilize the jsonrpsee low-level API to disconnect misbehaving peers. A few helper functions are exposed from the jsonrpsee::server to help with this.

Basically, we want people to be able to disconnect misbehaving peers, and this is probably the cleanest way to keep jsonrpsee  
as simple as possible.

Resolves https://github.com/paritytech/jsonrpsee/issues/1016, resolves https://github.com/paritytech/jsonrpsee/issues/1014, resolves https://github.com/paritytech/jsonrpsee/issues/1000


examples/examples/jsonrpsee_server_disconnect_conn.rs
```diff
@@ -0,0 +1,264 @@
+// Copyright 2019-2021 Parity Technologies (UK) Ltd.
```
niklasad1, line 1: //cc @xlc 

This is the suggested API to disconnect peers etc, it's still just a PoC and I need to polish it.
Is it sufficient for proxy implementation?

This example just disconnect that don't comply with the "dummy rate limit middleware" and blacklists the IP addr.

server/src/server.rs
```diff
@@ -559,6 +540,16 @@ impl<HttpMiddleware, RpcMiddleware> Builder<HttpMiddleware, RpcMiddleware> {
 		self
 	}
 
+	/// Extract the server configuration to use jsonrpsee as service.
+	pub fn to_service(self, methods: impl Into<Methods>) -> ServiceConfig<RpcMiddleware, HttpMiddleware> {
```
niklasad1, line 544: we may omit methods from this API...

server/src/server.rs
```diff
@@ -691,31 +680,31 @@ where
...
 			let response = match server.receive_request(&request) {
 				Ok(response) => {
-					let (tx, rx) = mpsc::channel::<String>(this.message_buffer_capacity as usize);
+					let (tx, rx) = mpsc::channel::<String>(this.cfg.message_buffer_capacity as usize);
 					let sink = MethodSink::new(tx);
 
 					// On each method call the `pending_calls` is cloned
 					// then when all pending_calls are dropped
-					// a graceful shutdown can occur.
+					// a graceful shutdown can has occur.
```
niklasad1, line 695:
```suggestion
					// a graceful shutdown can occur.
```

server/src/transport/ws.rs
```diff
@@ -334,3 +329,120 @@ async fn graceful_shutdown<S>(
...
+					other: params,
+					ws_sender: sender,
+					ws_receiver: receiver,
+					rpc_service,
+					sink,
+					rx,
+					pending_calls_completed,
+				};
+
+				background_task(params).await;
```
niklasad1, line 438: Perhaps we should improve this to send a message to background task instead of just dropping the future.

examples/examples/jsonrpsee_service.rs
```diff
@@ -0,0 +1,133 @@
...
+
+	let client = HttpClientBuilder::default().build("http://127.0.0.1:9944").unwrap();
+
+	let x: String = client.request("say_hello", rpc_params!()).await.unwrap();
+	tracing::info!("response: {x}");
+
+	Ok(())
+}
+
+fn run_server() -> ServerHandle {
```
jsdw, line 74: Aah ok, so this is the "minimal" example of using `jsonrpsee` as a tower service; this is interesting!

We are exposing a bunch of things I guess in order to make this possible; I wonder whether it's possible to hide any of this stuff without reducing the flexibility of this approach in general! Eg maybe a builder struct or something to make constructing the `TowerService` easier (or a `TowerServiceBuilder` that can be turned into a new `TowerService` on each connection or something). 

Partly I'm just wondering whether we _want_ to expose some of the things that we have in jsonrpsee, or whether it's ok that we exposed the extra bits? And partly just thinking how to make this super easy in the "simple" case but allowing for flexibility when needed!

niklasad1: I realized that it won't play that nicely with our http middleware such that one would need to use the low-level API in for hyper that.

Thus, I believe that this would require less code https://github.com/hyperium/hyper/blob/master/examples/service_struct_impl.rs

niklasad1: Fair point, I made everything public because it was an easy way to extract all settings from the ServerBuilder to avoid having 10 parameters for the user to inject.

Lemme a think about it

niklasad1: I have tried to implement something cleaner and I removed the jsonrpsee::to_service from the low-level api example

examples/examples/jsonrpsee_server_low_level_api.rs
```diff
@@ -0,0 +1,286 @@
...
+{
+	async fn call(&self, req: Request<'a>, t: TransportProtocol) -> MethodResponse {
+		let mut lock = self.count.lock().await;
+
+		if *lock >= 10 {
+			let _ = self.state.try_send(());
+			MethodResponse::error(req.id, ErrorObject::borrowed(-32000, "RPC rate limit", None))
+		} else {
+			let rp = self.service.call(req, t).await;
+			*lock = *lock + 1;
```
jsdw, line 81: Maybe worth a comment like "In reality, this number should be decremented again at some point"?

examples/examples/jsonrpsee_server_low_level_api.rs
```diff
@@ -0,0 +1,286 @@
...
+use std::net::SocketAddr;
+use std::sync::atomic::AtomicU32;
+use std::sync::{Arc, Mutex};
+use std::time::Duration;
+
+use futures::FutureExt;
+use jsonrpsee::core::{async_trait, client::ClientT};
+use jsonrpsee::http_client::HttpClientBuilder;
+use jsonrpsee::proc_macros::rpc;
+use jsonrpsee::server::middleware::rpc::*;
```
jsdw, line 49: How many htings come from here? would it be easy to remove the `*`?

examples/examples/jsonrpsee_server_low_level_api.rs
```diff
@@ -0,0 +1,286 @@
...
+#[async_trait]
+impl<'a, S> RpcServiceT<'a> for DummyRateLimit<S>
+where
+	S: Send + Sync + RpcServiceT<'a>,
+{
+	async fn call(&self, req: Request<'a>, t: TransportProtocol) -> MethodResponse {
+		let mut lock = self.count.lock().await;
+
+		if *lock >= 10 {
+			let _ = self.state.try_send(());
```
jsdw, line 77: Maybe worth a comment about what this is doing?

examples/examples/jsonrpsee_server_low_level_api.rs
```diff
@@ -0,0 +1,286 @@
...
+	// And a MakeService to handle each connection...
+	let make_service = make_service_fn(move |conn: &AddrStream| {
+		// Connection state.
+		let conn_id = conn_id.fetch_add(1, std::sync::atomic::Ordering::SeqCst);
+		let remote_addr = conn.remote_addr();
+		let stop_handle = stop_handle2.clone();
+		let conn_guard = conn_guard.clone();
+		let methods = methods.clone();
+		let blacklisted_peers = blacklisted_peers.clone();
+		let http_rate_limit = http_rate_limit.clone();
```
jsdw, line 196: Maybe worth a note that the HTTP rate limit is global (because new connection every time) but the WS rate limit is per connection?

examples/examples/jsonrpsee_server_low_level_api.rs
```diff
@@ -0,0 +1,286 @@
...
+			let methods = methods.clone();
+			let stop_handle = stop_handle.clone();
+			let blacklisted_peers = blacklisted_peers.clone();
+			let http_rate_limit = http_rate_limit.clone();
+
+			Ok::<_, Infallible>(service_fn(move |req| {
+				// Connection number limit exceeded.
+				let Some(conn_permit) = conn_guard.try_acquire() else {
+					return async { Ok::<_, Infallible>(http::response::too_many_requests()) }.boxed();
+				};
```
jsdw, line 210: Maybe worth a comment that this bit isn't realted to the rate limiting service; here we are separately showing that we can also limit the number of concurrent connections (to 100 in this example)?

examples/examples/jsonrpsee_server_low_level_api.rs
```diff
@@ -0,0 +1,286 @@
...
+								Ok(rp)
+							}
+							Err(rp) => Ok(rp),
+						}
+					}
+					.boxed()
+				} else if !ws::is_upgrade_request(&req) {
+					// In this example it doesn't make sense to disconnect the peer in the middleware rate limit
+					// is triggered as it will just reply to HTTP request i.e, the "mpsc::Sender disconnect"
+					// is not used.
```
jsdw, line 254: But this means that an IP address used for HTTP connections will never be blacklisted, unlike WS?

examples/examples/jsonrpsee_as_service.rs
```diff
@@ -0,0 +1,200 @@
...
+			Ok::<_, Box<dyn StdError + Send + Sync>>(service_fn(move |req| {
+				let metrics = metrics.clone();
+				let svc_builder = svc_builder.clone();
+				let methods = methods.clone();
+				let stop_handle = stop_handle.clone();
+
+				async move {
+					let headers = req.headers().clone();
+
+					// NOTE, the rpc middleware must be initialized here to be able to created once per connection.
```
niklasad1, line 170:
```suggestion
					// NOTE, the rpc middleware must be initialized here to be able to created once per connection
					// with data from the connection such as the headers in this example
```

examples/examples/jsonrpsee_as_service.rs
```diff
@@ -0,0 +1,201 @@
...
+				let svc_builder = svc_builder.clone();
+				let methods = methods.clone();
+				let stop_handle = stop_handle.clone();
+
+				async move {
+					let headers = req.headers().clone();
+
+					// NOTE, the rpc middleware must be initialized here to be able to created once per connection
+					// with data from the connection such as the headers in this example
+					let rpc_middleware = RpcServiceBuilder::new()
```
niklasad1, line 172: I don't like this but that's the trade-off right now to get state to the middleware

server/src/future.rs
```diff
@@ -33,16 +33,17 @@ use tokio::sync::{watch, OwnedSemaphorePermit, Semaphore, TryAcquireError};
 /// Represent a stop handle which is a wrapper over a `multi-consumer receiver`
 /// and cloning [`StopHandle`] will get a separate instance of the underlying receiver.
 #[derive(Debug, Clone)]
-pub(crate) struct StopHandle(watch::Receiver<()>);
+pub struct StopHandle(watch::Receiver<()>);
 
 impl StopHandle {
```
jsdw, line 38: So here, the user has to create the tx/rx themselves, and then construct the stop handle and server handle. I wonder if there's a better way to hide the implementation details around this shutting down stuff so that the user isn't making txs/rxs etc. Or does the user benefit from being able to make these themselves?

examples/examples/jsonrpsee_server_low_level_api.rs
niklasad1: It will be blacklisted but the channel is not used to "kill the connection" because for HTTP it's only one call to complete
and it makes no sense to be cancel a RPC on HTTP.

This example blacklisting per HTTP request won't work in practice because one would need a HashMap or something to just ban if the same IP peer is making many requests.

This will blacklist the request which fires trigger the rate limit but it could in fact be the first HTTP request it sends ^^

server/src/transport/http.rs
```diff
@@ -1,39 +1,117 @@
-use std::convert::Infallible;
+use http::Method;
+use jsonrpsee_core::{http_helpers::read_body, GenericTransportError};
+
+use crate::{
+	middleware::rpc::{RpcService, RpcServiceBuilder, RpcServiceCfg, RpcServiceT, TransportProtocol},
+	server::{handle_rpc_call, Params, Settings},
```
jsdw, line 6: Seeing the two together; how do we explain the difference between `Params` and `Settings` (both sound like the same thing); maybe some naming tweak can better differentiate?

jsdw (commented):
Mostly looks good to me, though we make a bunch of fields and typees public, which leads to quite a big "surface area" that the user needs to know/care about, and I do wonder whether we can make that surface area smaller (to make it easier for a user to understand but also expose fewer implementation details that we become committed to supporting)

niklasad1:
> Mostly looks good to me, though we make a bunch of fields and typees public, which leads to quite a big "surface area" that the user needs to know/care about, and I do wonder whether we can make that surface area smaller (to make it easier for a user to understand but also expose fewer implementation details that we become committed to supporting)

I don't expect anyone to use the low-level API except if one wants to write a proxy or something but I agree it sucks to expose the jsonrpsee settings but it's really no way around besides to providing some function with limited of the settings.

server/src/transport/http.rs
niklasad1: Settings are "static configuration" of the server.
Params contains "static configuration" + a few connection related things such as connection_id, conn_permit etc.

examples/examples/jsonrpsee_server_low_level_api.rs
niklasad1: Changed the code to avoid writing such weird comments

server/src/future.rs
niklasad1: It would be great to get rid of that but it's tricky because once the `tx handle` is dropped then the server is stopped.
So it would be hard to hide that without running to the server forever i.e, not be able to stop it.

Perhaps we can provide a helper function that avoid constructing the channel at least.

server/src/transport/http.rs
niklasad1: Removed now, I think it's easier to understand now

examples/examples/jsonrpsee_server_low_level_api.rs
niklasad1: renamed the middleware and added some docs that should be sufficient.

server/src/future.rs
jsdw: I like the function better yup, and it means we can change the internals without worry :)

server/src/future.rs
```diff
@@ -30,19 +30,27 @@ use jsonrpsee_core::Error;
...
-	pub(crate) fn new(rx: watch::Receiver<()>) -> Self {
+	/// Create a new stop handle.
+	pub fn new(rx: watch::Receiver<()>) -> Self {
```
jsdw, line 47: I wonder whether this can be `pub(crate)` again now or whethery ou'd prefer it was pub?

server/src/server.rs
```diff
@@ -620,54 +823,81 @@ impl<HttpMiddleware, RpcMiddleware> Builder<HttpMiddleware, RpcMiddleware> {
...
-			id_provider: self.id_provider,
 			http_middleware: self.http_middleware,
 		})
 	}
 }
 
 /// Data required by the server to handle requests.
 #[derive(Debug, Clone)]
-pub(crate) struct ServiceData {
+pub struct ServiceData {
```
jsdw, line 833: Is this still exposed publically or can the `pub`s now be removed?

Offhand it looks like it's only configured internally again now but I might be missing something!

server/src/transport/ws.rs
jsdw: How do you mean? I don't quite follow and am probably being dumb :)

server/src/server.rs
```diff
@@ -329,22 +465,85 @@ impl Builder<Identity, Identity> {
...
 }
 
+impl<RpcMiddleware: Clone, HttpMiddleware: Clone> TowerServiceBuilder<RpcMiddleware, HttpMiddleware> {
+	/// Build a tower service.
+	pub fn build(
+		&self,
+		methods: impl Into<Methods>,
+		stop_handle: StopHandle,
+	) -> TowerService<RpcMiddleware, HttpMiddleware> {
+		let conn_id = self.conn_id.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
```
niklasad1, line 475: this could have been a u32 but it's so annoying with mutable stuff in  FnMut closure's such as `hyper::service_fn`

server/src/server.rs
```diff
@@ -298,25 +334,125 @@ impl Default for Settings {
...
+}
+
+impl ServerConfigBuilder {
+	/// Create a new [`ServerConfigBuilder`].
+	pub fn new() -> Self {
+		Self::default()
+	}
+
+	/// See [`Builder::max_request_body_size`](method@Builder::max_request_body_size) for documentation.
+	pub fn max_request_body_size(mut self, size: u32) -> Self {
```
niklasad1, line 375: this is duplicated :(

we could add a trait for but it doesn't seems worth it for now

server/src/server.rs
```diff
@@ -618,56 +819,83 @@ impl<HttpMiddleware, RpcMiddleware> Builder<HttpMiddleware, RpcMiddleware> {
...
+/// jsonrpsee tower service.
+///
+/// This will not apply specific HTTP middleware that
+/// can be enabled by [`Builder::set_http_middleware`](method@Builder::set_http_middleware).
 ///
 /// # Note
 /// This is similar to [`hyper::service::service_fn`].
 #[derive(Debug, Clone)]
-pub struct TowerService<L> {
+pub struct TowerServiceNoHttp<L> {
```
niklasad1, line 893: bikeshed on naming,

I think this shouldn't be used by external users because the tower middleware from ServerBuilder isn't used then

server/src/transport/ws.rs
niklasad1: I was little scared about cancel-safety and it may be better to inject a channel in the background task and send a message/drop the channel to have clean shutdown.

However, I think it should be safe to drop because it's just to soketto stuff that isn't cancel-safe and that is wrapped in stream anyway.

server/src/server.rs
niklasad1: nice catch, it should be fixed now

server/src/server.rs
```diff
@@ -224,6 +224,23 @@ struct Settings {
 	message_buffer_capacity: u32,
 	/// Ping settings.
 	ping_config: PingConfig,
+	/// ID provider.
+	id_provider: Arc<dyn IdProvider>,
+}
+
+/// Service config.
```
niklasad1, line 231:
```suggestion
/// Builder for [`TowerService`].
```

server/src/server.rs
niklasad1: changed this to jsonrpsee internal thingy now

server/src/transport/http.rs
```diff
@@ -1,39 +1,117 @@
...
-pub(crate) fn is_json(content_type: Option<&hyper::header::HeaderValue>) -> bool {
+pub fn is_json(content_type: Option<&hyper::header::HeaderValue>) -> bool {
 	content_type.and_then(|val| val.to_str().ok()).map_or(false, |content| {
 		content.eq_ignore_ascii_case("application/json")
 			|| content.eq_ignore_ascii_case("application/json; charset=utf-8")
 			|| content.eq_ignore_ascii_case("application/json;charset=utf-8")
 	})
 }
 
-pub(crate) async fn reject_connection(socket: tokio::net::TcpStream) {
```
niklasad1, line 17: moved to the jsonrpsee::TowerService

tadeohepperle (approved):
I did read the code and give it a tick, but I do not feel able to give any feedback :)

examples/examples/jsonrpsee_as_service.rs
```diff
@@ -0,0 +1,208 @@
...
+// ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED
+// TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+// PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT
+// SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+// CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR
+// IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+// DEALINGS IN THE SOFTWARE.
+
+//! This examples shows how to use the `jsonrpsee::server`` as
```
lexnv, line 27: nit:
```suggestion
//! This examples shows how to use the `jsonrpsee::server` as
```

lexnv (approved):
LGTM! I believe the low-level API enables users to implement all use-cases, and handles most future requests for other features. Being low-level API seems worth it to me, especially when most users will not bother implementing it! 👍

examples/examples/jsonrpsee_as_service.rs
```diff
@@ -0,0 +1,209 @@
...
+// ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED
+// TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+// PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT
+// SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+// CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR
+// IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+// DEALINGS IN THE SOFTWARE.
+
+//! This examples shows how to use the `jsonrpsee::server` as
```
niklasad1, line 27:
```suggestion
//! This example shows how to use the `jsonrpsee::server` as
```

PR #1227, opened 2023-10-31
fix(host filtering): allow hosts with multiple ports

niklasad1:
The refactoring moved to another library where an `Authority` with the "same host name" but a different port was overwritten and this commit fixes that several ports with the same host name can be configured and work properly.


server/src/middleware/host_filter.rs
```diff
@@ -145,14 +161,14 @@ where
 impl WhitelistedHosts {
 	fn recognize(&self, other: &Authority) -> bool {
 		if let Ok(p) = self.0.recognize(&other.host) {
-			let p = p.handler();
+			let ports = p.handler();
 
-			match (p, &other.port) {
+			ports.iter().any(|p| match (p, &other.port) {
```
niklasad1, line 166: I expect this list to be small and linear search should ok here, we could change this to have HashSet or BTreeSet though

lexnv (approved):

server/src/middleware/host_filter.rs
jsdw: I suspect you're right!

jsdw (approved):

PR #1228, opened 2023-11-02
Adding gitspiegel-trigger workflow

mutantcornholio:
Using a workflow to trigger mirroring instead of a webhook allows us to reuse "Approving workflow runs from public forks" GitHub feature to somewhat protect us from malicious PRs


alvicsam (approved):

niklasad1:
I think we can remove the gitlab CI jobs instead, we are relying the GH CI and these were just added a test way back.

niklasad1 (approved):

PR #1230, opened 2023-11-06
ci: remove gitlab ci

niklasad1:


lexnv (approved):
LGTM!

IIUC we running the benchmarks in git CI and not gitlabCI. Could we rely on those results to make decisions wrts to performance? Or do we believe they are flaky enough that the data is only directionally correct, maybe even incorrect? 🤔

niklasad1:
The benchmarks are running on the gitlab CI/gitlab machine because the github machines were too flaky. After we migrated to gitlab the results have been much more reliable.

However, it still not needed to waste CPU cycles to run the CI tests twice and github actions works well enough for us. 
The windows job is a bit slow now but acceptable.

sergejparity (approved):

PR #1234, opened 2023-11-09
fix `RpcServiceBuilder::option_layer`

### LINKED ISSUE #1231: [middleware]: `RpcServiceBuilder::option_layer` doesn't work
`tower` has its own `Either type` which doesn't satisfy the RpcServiceT trait bound and jsonrpsee needs to mimic that type to work.

https://github.com/paritytech/jsonrpsee/blob/master/server/src/middleware/rpc/mod.rs#L92-#L94, either remove this API or provide a wrapper for that similar to how tower is doing it.

niklasad1:
Close https://github.com/paritytech/jsonrpsee/issues/1231

### DIFF
--- a/server/src/middleware/rpc/layers.rs
+++ b/server/src/middleware/rpc/layers.rs
@@ -24,12 +24,8 @@
 // IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 // DEALINGS IN THE SOFTWARE.
 
-use std::future::Future;
-use std::pin::Pin;
 use std::sync::Arc;
-use std::task::{Context, Poll};
 
-use futures_util::ready;
 use jsonrpsee_core::server::{
 	BoundedSubscriptions, MethodCallback, MethodResponse, MethodSink, Methods, SubscriptionState,
 };
@@ -37,8 +33,6 @@ use jsonrpsee_core::tracing::{rx_log_from_json, tx_log_from_str};
 use jsonrpsee_core::traits::IdProvider;
 use jsonrpsee_types::error::{reject_too_many_subscriptions, ErrorCode};
 use jsonrpsee_types::{ErrorObject, Request};
-use pin_project::pin_project;
-use tower::{BoxError, Layer};
 
 use super::{RpcServiceT, TransportProtocol};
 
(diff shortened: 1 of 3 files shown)


server/src/middleware/rpc/layers.rs
```diff
@@ -172,3 +179,60 @@ where
...
+	/// One type of backing [`RpcServiceT`].
+	A(#[pin] A),
+	/// The other type of backing [`RpcServiceT`].
+	B(#[pin] B),
+}
+
+impl<A, B, T, AE, BE> Future for Either<A, B>
+where
+	A: Future<Output = Result<T, AE>>,
+	AE: Into<Box<dyn StdError + Send + Sync>>,
```
lexnv, line 197: super nit: Might be worth creating a `pub type BoxError` type for this

server/src/middleware/rpc/layers.rs
```diff
@@ -172,3 +179,60 @@ where
 		rp
 	}
 }
+
+/// Similar to [`tower::util::Either`] but
+/// adjusted to satisfy the traits bound [`RpcServiceT].
```
niklasad1, line 184:
```suggestion
/// adjusted to satisfy the trait bound [`RpcServiceT].
```

server/src/middleware/rpc/layers.rs
```diff
@@ -172,3 +179,60 @@ where
...
+	}
+}
+
+#[async_trait::async_trait]
+impl<'a, A, B> RpcServiceT<'a> for Either<A, B>
+where
+	A: RpcServiceT<'a> + Send + Sync,
+	B: RpcServiceT<'a> + Send + Sync,
+{
+	async fn call(&self, request: Request<'a>, transport: TransportProtocol) -> MethodResponse {
```
lexnv, line 232: Dq unrelated to this PR : do we want to expose in the future extra information together with the `TransportPortocol` to the user? 🤔 Generally I'm thinking towards a `CallContext` that has `fn transport_protocol(&self) -> &TransportProtocol`

server/src/middleware/rpc/layers.rs
```diff
@@ -172,3 +179,60 @@ where
...
+#[pin_project(project = EitherProj)]
+#[derive(Clone, Debug)]
+pub enum Either<A, B> {
+	/// One type of backing [`RpcServiceT`].
+	A(#[pin] A),
+	/// The other type of backing [`RpcServiceT`].
+	B(#[pin] B),
+}
+
+impl<A, B, T, AE, BE> Future for Either<A, B>
```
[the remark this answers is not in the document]
jsdw: A bunch of this I think just aligns with https://docs.rs/tower/latest/src/tower/util/either.rs.html#59

lexnv (approved):
Nice one! I like that this also allows users to conditionally combine service layers, similar to the `fn option_layer` 👍

niklasad1:
> Maybe a dumb Q: can we not just implement the relevant traits on future::util::Either?

That's a good question but because tower::Service requires &mut self it's not possible to implement because AFAIU

For instance what comes to mind is:

```rust
#[async_trait::async_trait]
impl<'a, A, B> RpcServiceT<'a> for tower::util::Either<A, B>
where
	A: Service<Request<'a>, Response = MethodResponse> + Send + Sync,
	B: Service<Request<'a>, Response = MethodResponse> + Send + Sync,
	A::Future: Send,
	B::Future: Send,
	A::Error: std::fmt::Debug,
	B::Error: std::fmt::Debug,
{
	async fn call(&self, request: Request<'a>, _transport: TransportProtocol) -> MethodResponse {
		match self {
                        // requires &mut self
			tower::util::Either::A(service) => service.call(request).await.unwrap(),
                         // requires &mut self
			tower::util::Either::B(service) => service.call(request).await.unwrap(),
		}
	}
}




```

jsdw (approved):

jsdw:
But in your impl , `Either::A` and `Either::B` are `RpcServiceT'`s anyway, so I'd have thought you could basically impl this bit on tokios Either rather than your own, since the other bits are the same I think:

```rust
#[async_trait::async_trait]
impl<'a, A, B> RpcServiceT<'a> for tokio::utils::Either<A, B>
where
	A: RpcServiceT<'a> + Send + Sync,
	B: RpcServiceT<'a> + Send + Sync,
{
	async fn call(&self, request: Request<'a>, transport: TransportProtocol) -> MethodResponse {
		match self {
			Either::A(service) => service.call(request, transport).await,
			Either::B(service) => service.call(request, transport).await,
		}
	}
}
```

But maybe the other bits are different and I haven't looked closely enough :)

niklasad1:
> But in your impl , Either::A and Either::B are RpcServiceT's anyway, so I'd have thought you could basically impl this bit on tokios Either rather than your own, since the other bits are the same I think:

Nah, it's from tower but sure good idea to use tokio::Either or futures::Either

niklasad1:
I ran into issues with that as well because it's not possible to implement the Layer trait on a foreign type

```
impl<S, A, B> tower::Layer<S> for Either<A, B>
where
	A: tower::Layer<S>,
	B: tower::Layer<S>,
{
	type Service = Either<A::Service, B::Service>;

	fn layer(&self, inner: S) -> Self::Service {
		match self {
			Either::Left(layer) => Either::A(layer.layer(inner)),
			Either::Right(layer) => Either::B(layer.layer(inner)),
		}
	}
}
```

```bash
error[E0210]: type parameter `S` must be used as the type parameter for some local type (e.g., `MyStruct<S>`)
   --> server/src/middleware/rpc/mod.rs:142:6
    |
142 | impl<S, A, B> tower::Layer<S> for Either<A, B>
    |      ^ type parameter `S` must be used as the type parameter for some local type
    |
    = note: implementing a foreign trait is only possible if at least one of the types for which it is implemented is local
    = note: only traits defined in the current crate can be implemented for a type parameter

For more information about this error, try `rustc --explain E0210`.
warning: `jsonrpsee-server` (lib) generated 5 warnings
error: could not compile `jsonrpsee-server` (lib) due to previous error; 5 warnings emitted


```

PR #1242, opened 2023-11-20
refactor: RpcServiceT make it possible !BoxFuture

### LINKED ISSUE #1238: middleware: `RpcServiceT` make it possible to use !BoxedFuture
https://github.com/paritytech/jsonrpsee/pull/1215#discussion_r1363489019

After #1215 was merged a new RpcServiceT which requires the Future to be boxed and because it is used for every JSON-RPC call it can significantly increase the memory usage if many layers are used.

Let's introduce a trait that makes it possible to not box all futures:

```rust
pub trait RpcServiceT<'a> {
    /// The future response value.
    type Future<'cx>: Future<Output = MethodResponse>>
    where
        Self: 'cx,
        Request: 'cx;

    /// Process the request and return the response asynchronously.
    fn call(&self, req: Request<'a>) -> Self::Future<'_>;
}
```

niklasad1:
Close #1238


core/src/server/rpc_module.rs
```diff
@@ -59,9 +59,8 @@ pub type SyncMethod = Arc<dyn Send + Sync + Fn(Id, Params, MaxResponseSize) -> M
 pub type AsyncMethod<'a> =
 	Arc<dyn Send + Sync + Fn(Id<'a>, Params<'a>, ConnectionId, MaxResponseSize) -> BoxFuture<'a, MethodResponse>>;
 /// Method callback for subscriptions.
-pub type SubscriptionMethod<'a> = Arc<
-	dyn Send + Sync + Fn(Id, Params, MethodSink, SubscriptionState) -> BoxFuture<'a, Result<MethodResponse, Id<'a>>>,
->;
+pub type SubscriptionMethod<'a> =
```
niklasad1, line 62: Changed this this align better with async method calls

server/src/middleware/rpc/layer/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+use std::pin::Pin;
+use std::task::{Context, Poll};
+
+use futures_util::Future;
+use jsonrpsee_core::server::MethodResponse;
+use pin_project::pin_project;
+
+/// Response which may be ready or a future that needs to be
+/// polled.
+#[pin_project(project = ResponseStateProj)]
```
niklasad1, line 45: I reckon that this ResponseFuture may be useful but we could also make it private.........

server/src/middleware/rpc/mod.rs
```diff
@@ -26,24 +26,30 @@
...
-
 /// Similar to `tower::Service` but specific for jsonrpsee and
 /// doesn't requires `&mut self` for performance reasons.
-#[async_trait::async_trait]
 pub trait RpcServiceT<'a> {
+	/// The future response value.
+	//
+	// type Future<'cx>: Future<Output = MethodResponse> where Self: 'cx, 'a: 'cx;
+	// would be better but ran into https://github.com/rust-lang/rust/issues/100013
+	type Future: Future<Output = MethodResponse> + Send;
```
niklasad1, line 47: TODO: remove send bound

tests/tests/metrics.rs
```diff
@@ -60,30 +62,38 @@ pub struct CounterMiddleware<S> {
 #[async_trait]
 impl<'a, S> RpcServiceT<'a> for CounterMiddleware<S>
 where
-	S: RpcServiceT<'a> + Send + Sync,
+	S: RpcServiceT<'a> + Send + Sync + Clone + 'static,
```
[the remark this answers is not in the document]
niklasad1: //cc @jsdw do you know if it's possible to avoid cloning the service to move it to the async block below?

I'm probably need a lifetime on the future for that to work here and bummer that it worked with async_trait :thinking:

server/src/middleware/rpc/mod.rs
niklasad1: I changed my mind it's a mess to change that and every future in jsonrpsee requires Send

tests/tests/metrics.rs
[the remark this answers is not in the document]
jsdw: As it stands, with the service being taken as a reference through `&self`, I can't see how you'd avoid cloning it unless you put it into a `RefCell`/`Mutex` and moved it out into the `async` block and then put it back at the end (but of course, since `call` only takes `&self`, there's nothing that prevents it being called more than once at a time at compile time at least, which would lead to issues!)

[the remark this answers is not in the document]
niklasad1: It should be quite cheap to clone the underlying service RpcService anyway, it's cloning a couple of Arc's but if someone puts a bunch of heavy stuff in a wrapped service it could be expensive...

examples/examples/jsonrpsee_as_service.rs
```diff
@@ -58,30 +57,33 @@ struct Metrics {
...
 	headers: HeaderMap,
 	inner: S,
 	#[allow(unused)]
 	transport_label: &'static str,
 }
 
-#[async_trait]
 impl<'a, S> RpcServiceT<'a> for AuthorizationMiddleware<S>
 where
 	S: Send + Sync + RpcServiceT<'a>,
```
lexnv, line 70: Should we also add the `Clone` bound to `S`? 🤔

server/src/middleware/rpc/layer/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+
+impl<F: Future<Output = MethodResponse>> Future for ResponseFuture<F> {
+	type Output = MethodResponse;
+
+	fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+		let this = self.project();
+
+		match this {
+			ResponseStateProj::Poll(fut) => fut.poll(cx),
+			ResponseStateProj::Ready(rp) => Poll::Ready(rp.take().expect("Future not polled after Ready; qed")),
```
lexnv, line 83: dq: When would we create a `ResponseFuture` with `Ready(None)` / do we need `Ready(Option<>)` to impl default on it?

lexnv (approved):
LGTM!

server/src/middleware/rpc/layer/mod.rs
niklasad1: Never, the `Option` is just a hack to take out the result instead of cloning it as `poll` isn't called after `Poll::Ready` is called.

We could also just clone the response.... but I prefer not to do it :P

examples/examples/jsonrpsee_as_service.rs
niklasad1: fixed, ty

examples/examples/jsonrpsee_as_service.rs
```diff
@@ -58,30 +57,33 @@ struct Metrics {
...
 	inner: S,
 	#[allow(unused)]
 	transport_label: &'static str,
 }
 
-#[async_trait]
 impl<'a, S> RpcServiceT<'a> for AuthorizationMiddleware<S>
 where
-	S: Send + Sync + RpcServiceT<'a>,
+	S: Send + Clone + Sync + RpcServiceT<'a>,
```
jsdw, line 70: It didn't occur to me before but I guess, since you only need to take `&self`, that `S` could be wrapped in an `Arc` or something if you want to be able to cheaply clone it while avoiding the `Clone` bounds on `S`.

But `Clone` bounds on `S` might be better because it doesn't dictate how `S` is made cloneable and so they can opt to wrap it in  `Arc` or maybe just copy it or whatever if it's cheap etc.

server/src/middleware/rpc/layer/logger.rs
```diff
@@ -0,0 +1,111 @@
...
+/// A middleware that logs each RPC call and response.
+#[derive(Debug)]
+pub struct RpcLogger<S> {
+	max: u32,
+	service: S,
+}
+
+impl<'a, S> RpcServiceT<'a> for RpcLogger<S>
+where
+	S: RpcServiceT<'a> + Send + Sync + Clone + 'static,
```
jsdw, line 72: I geuss my comment on `Clone` should have been here and not on an example but whatever :)

server/src/middleware/rpc/layer/either.rs
```diff
@@ -0,0 +1,116 @@
...
+}
+
+/// Response future for the either layer.
+#[pin_project(project = Fut)]
+pub enum ResponseFuture<A, B> {
+	/// Left future
+	Left(#[pin] A),
+	/// Right future.
+	Right(#[pin] B),
+}
```
jsdw, line 91: Could this bit just use `futures::Either` (https://docs.rs/futures/latest/futures/future/enum.Either.html#impl-Future-for-Either%3CA,+B%3E) or `tokio_util::Either`? https://docs.rs/tokio-util/latest/tokio_util/either/enum.Either.html. Not a big deal but since we already import those :)

server/src/middleware/rpc/layer/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+
+/// Response which may be ready or a future that needs to be
+/// polled.
+#[pin_project(project = ResponseStateProj)]
+pub enum ResponseFuture<F> {
+	/// The response is ready.
+	Ready(Option<MethodResponse>),
+	/// The response has to be polled.
+	Poll(#[pin] F),
+}
```
jsdw, line 51: could this just be `tokio_util::Either<std::future::Ready<MethodResponse>, F>`?

server/src/middleware/rpc/layer/rpc_service.rs
```diff
@@ -0,0 +1,141 @@
...
+	CallsAndSubscriptions {
+		bounded_subscriptions: BoundedSubscriptions,
+		sink: MethodSink,
+		id_provider: Arc<dyn IdProvider>,
+		_pending_calls: tokio::sync::mpsc::Sender<()>,
+	},
+}
+
+impl RpcService {
+	/// Create a new service with doesn't support subscriptions.
```
jsdw, line 65: Is the comment wrong? Looks like it can support subscriptions if you pass the right `RpcServiceCfg` in?

tests/tests/metrics.rs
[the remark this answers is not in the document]
jsdw: At least we give them the choice; they could choose to put their service behind an `Arc` of could avoid it entirely if it's a cheap service that is trivially cloneable :)

jsdw (approved):
Looks great to me; lovely that Services can avoid boxing now!

examples/examples/jsonrpsee_as_service.rs
niklasad1: good idea, it's probably a good practice to wrap in an Arc instead cloning the entire service

server/src/middleware/rpc/layer/rpc_service.rs
niklasad1: yes

examples/examples/jsonrpsee_as_service.rs
jsdw: But I mean that as it is right now, it's up to the user whether they want to wrap the inner service in an Arc or whether they don't need to, so I do sortof like how it is now :)

niklasad1: yepp, I was referring to the examples that we should point to folks to wrap it an Arc instead of cloning the entire service because it can be expensive such if one put HTTP headers or something heavier in there

PR #1168, opened 2023-08-07
refactor: update and turn `WsTransportClientBuilder` generic

oleonardolima:
e7ecb9d5bc41ee0cb018a0d4a97c20ed8cb46966 refactor(jsonrpsee-client-transport): update and turn `WsTransportClientBuilder` generic 

- fix(docs): typo on `WsTransportClientBuilder` doc on `Receiver`
  reference.
- refactor: add initial fn signatures for `build_with_stream`,
  `try_connect_over_tcp`, and `try_connect`.
- refactor: expose `EitherStream` visibility to public.
- refactor: make `Sender` and `Receiver` generic over T, a data stream.
- refactor: make `TransportSenderT` and `TransportReceiverT`
  implementations over generic `Sender` and `Receiver`, bound to
`AsyncRead`, `AsyncRead`, `MaybeSend` and `'static`.
- refactor: turn old `try_connect` TCP steps into
  `try_connect_over_tcp`.
- feat: implement `build_with_stream` and `try_connect` to handle and
  handle the handshake for a generic data stream `T`.
- feat: add new `Redirected` error variant to `WsHandshakeError`, as it
  should be handled by the client when using a generic data stream `T`.

c7ae7dd79c781dc87e169838c0336c06b1f5e62f refactor(jsonrpsee-ws-client): add new fns to WsClientBuilder 

- feat: add new `WsClientBuilder::build_with_transport` that builds and
  returns a `WsClient` with the given `Sender` and `Receiver`.
- feat: add new `WsClientBuilder::build_with_stream` that uses the new
  `WsTransportClientBuilder::build_with_stream`, building and returning
  the `WsClient` with the given `data_stream` as transport layer.
- refactor: update the `WsClientBuilder::build` to use the new
  `build_with_transport`, it helps not having duplicated code.
  
10e78a045cc07d5b39f8609f13a7ace8284e741f refactor: re-export EitherStream & sort current list

53641d1ec73c0f4a17191cd7f32b52f646226895 test: add integration tests and helper fns

- add new helper fns to spawn a socks5 server, using `fast-socks5`
- add new helper enum for `DataStream` that acts as a wrapper to
  `Socks5Stream<T>`, similar to what a client would need to do.
- impl AsyncRead + AsyncWrite for the helper `DataStream` enum, to make
  it compatible between futures::io and tokio::io.
- add new tests that connects over a socks5 proxy, and use the new
  `WsClientBuilder::default()::build_with_stream(...) fn.

fixes #1162 #870

PR #1244, opened 2023-11-21
fix: unsubscription test

niklasad1:
Fixes https://github.com/paritytech/jsonrpsee/pull/1168#discussion_r1400415439

### DIFF
--- a/tests/tests/integration_tests.rs
+++ b/tests/tests/integration_tests.rs
@@ -116,7 +116,7 @@ async fn ws_unsubscription_works() {
 
 	sub.unsubscribe().await.unwrap();
 
-	let res = rx.next().with_default_timeout.await.expect("Test must complete with 1 min");
+	let res = rx.next().with_default_timeout().await.expect("Test must complete in 1 min");
 	// When the subscription is closed a message is sent out on this channel.
 	assert!(res.is_some());
 }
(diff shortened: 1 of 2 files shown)


jsdw (approved):
Nice!

tests/tests/integration_tests.rs
```diff
@@ -104,74 +106,40 @@ async fn ws_subscription_works_over_proxy_stream() {
...
-		if res.is_ok() {
-			success = true;
-			break;
-		}
-		tokio::time::sleep(std::time::Duration::from_millis(100)).await;
-	}
-
-	assert!(success);
+	// When the subscription is closed a message is sent out on this channel.
+	let res = rx.next().with_timeout(std::time::Duration::from_secs(10)).await.expect("Timeout 10 secs exceeded");
```
lexnv, line 120: Would there be a way to avoid relying on the sleep here? Might be possible that the CI will lag at some point and we'd exceed this?

lexnv (approved):

tests/tests/integration_tests.rs
niklasad1: I can remove it, it just a way to ensure that the test fails if the future never completes.

The underlying subscription will run ~5 hours and can only be cancelled if the connection is closed or unsubscribed. https://github.com/paritytech/jsonrpsee/blob/d3a34030f04c43359e380a9938430d515c06d794/tests/tests/helpers.rs#L176-#L188

PR #1246, opened 2023-11-27
feat(client): add `disconnect_reason` API

### LINKED ISSUE #1196: How to get disconnect reason?
I am using `on_disconnect` to get notified on disconnect and retry connection.

https://github.com/paritytech/jsonrpsee/blob/d4ea3464dc9866fac1ec284e65090b52015e10b6/core/src/client/async_client/mod.rs#L320

But how can I get disconnect reason so that I can implement advanced retry strategy?

niklasad1:
This commit adds an API to get the error reason why the backend was disconnected.

Close #1196

### DIFF
--- a/client/ws-client/src/tests.rs
+++ b/client/ws-client/src/tests.rs
@@ -25,9 +25,10 @@
 // DEALINGS IN THE SOFTWARE.
 
 #![cfg(test)]
-use crate::types::error::{ErrorCode, ErrorObject};
 
+use crate::types::error::{ErrorCode, ErrorObject};
 use crate::WsClientBuilder;
+
 use jsonrpsee_core::client::{BatchResponse, ClientT, SubscriptionClientT};
 use jsonrpsee_core::client::{IdKind, Subscription};
 use jsonrpsee_core::params::BatchRequestBuilder;
@@ -68,12 +69,8 @@ async fn method_call_with_wrong_id_kind() {
 	let client =
 		WsClientBuilder::default().id_format(IdKind::String).build(&uri).with_default_timeout().await.unwrap().unwrap();
 
-	match client.request::<String, _>("o", rpc_params![]).with_default_timeout().await.unwrap() {
-		Err(Error::RestartNeeded(e)) => {
-			assert!(matches!(*e, Error::InvalidRequestId(_)));
-		}
-		_ => panic!("Test failed"),
-	};
+	let err: Result<String, _> = client.request::<String, _>("o", rpc_params![]).with_default_timeout().await.unwrap();
+	assert!(matches!(err, Err(Error::RestartNeeded(e)) if matches!(*e, Error::InvalidRequestId(_))));
 }
 
 #[tokio::test]
--- a/client/ws-client/src/tests.rs
+++ b/client/ws-client/src/tests.rs
@@ -69,7 +69,7 @@ async fn method_call_with_wrong_id_kind() {
 	let client =
 		WsClientBuilder::default().id_format(IdKind::String).build(&uri).with_default_timeout().await.unwrap().unwrap();
 
-	let err: Result<String, _> = client.request::<String, _>("o", rpc_params![]).with_default_timeout().await.unwrap();
+	let err: Result<String, _> = client.request("o", rpc_params![]).with_default_timeout().await.unwrap();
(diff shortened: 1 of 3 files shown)


core/src/error.rs
```diff
@@ -57,7 +57,7 @@ pub enum Error {
 	InvalidResponse(Mismatch<String>),
 	/// The background task has been terminated.
 	#[error("The background task been terminated because: {0}; restart required")]
-	RestartNeeded(String),
+	RestartNeeded(Arc<Error>),
```
niklasad1, line 60: This is awkward to match on but definitely better than a String I reckon..

core/src/client/async_client/mod.rs
```diff
@@ -53,34 +53,57 @@ impl ThreadSafeRequestManager {
...
+					// This should never happen because the receiving end is still alive.
+					// Before shutting down the background task a error message should
+					// be emitted.
+					Err(_) => Error::Custom("Error reason could not be found. This is a bug. Please open an issue.".to_string()),
+				});
+				*write = ReadErrorOnce::Read(arc_err.clone());
+				arc_err
+			}
+			ReadErrorOnce::Read(arc_err) => {
+				*write = ReadErrorOnce::Read(arc_err.clone());
```
niklasad1, line 85: `state` is moved on the match and must be "initialized again here"

tadeohepperle (approved):
Looks good to me :)

client/ws-client/src/tests.rs
```diff
@@ -68,8 +68,12 @@ async fn method_call_with_wrong_id_kind() {
 	let client =
 		WsClientBuilder::default().id_format(IdKind::String).build(&uri).with_default_timeout().await.unwrap().unwrap();
 
-	let err: Result<String, Error> = client.request("o", rpc_params![]).with_default_timeout().await.unwrap();
-	assert!(matches!(err, Err(Error::RestartNeeded(e)) if e == "request ID=0 is not a pending call"));
+	match client.request::<String, _>("o", rpc_params![]).with_default_timeout().await.unwrap() {
```
lexnv, line 71: nit: Maybe `assert_matches!` cleans this up a bit, but since its part of testing I wouldn't worry too much about it

core/src/client/async_client/mod.rs
```diff
@@ -302,13 +325,17 @@ impl Client {
...
 
-	// Reads the error message from the backend thread.
-	async fn read_error_from_backend(&self) -> Error {
-		let mut err_lock = self.error.lock().await;
-		let from_back = std::mem::replace(&mut *err_lock, ErrorFromBack::Read(String::new()));
-		let (next_state, err) = from_back.read_error().await;
-		*err_lock = next_state;
-		err
+	/// This is similar to to [`Client::on_disconnect`] but it can be used to get
+	/// reason why the client was disconnected but it's not cancel-safe.
```
lexnv, line 329: nit: s/reason/the reason

lexnv (approved):
Solid work! I only had a few small nits that you can ignore if they don't make sense to you 👍

core/src/client/async_client/mod.rs
```diff
@@ -53,34 +53,57 @@ impl ThreadSafeRequestManager {
...
+		Self(AsyncRwLock::new(ReadErrorOnce::Unread(unread)))
+	}
+
+	async fn read_error(&self) -> Error {
+		if let ReadErrorOnce::Read(ref err) = *self.0.read().await {
+			return Error::RestartNeeded(err.clone());
+		};
+
+		let mut write = self.0.write().await;
+		let state = std::mem::replace(&mut *write, ReadErrorOnce::Unreachable);
```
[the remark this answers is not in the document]
niklasad1: yeah fair enough,

Ideally we would remove this AsyncRwLock with something like https://docs.rs/async-once-cell/0.5.3/async_once_cell/struct.Lazy.html# but it requires a bunch of unstable features but would be nice to get rid of for sure.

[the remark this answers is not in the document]
lexnv: Looks clean as it is, no need to rely on `unstable` features for now :D

client/ws-client/src/tests.rs
niklasad1: I wanted to show that working with Error::RestartNeeded(Arc<Error>) can be awkward :P

PR #1250, opened 2023-12-01
client(subscription): improve stream API

niklasad1:
It's was hard to know whether a subscription was closed or that the entire connection was closed without calling `Client::is_connected`.

This commit adds an explicit `Error::Lagged` to indicate that the subscription was dropped because it could not keep with the server.

Hopefully, the explicit error and added documentation will help with any confusion regarding that.


niklasad1:
I'm not happy with this yet, closing

PR #1252, opened 2023-12-01
Enable all http versions

venugopv:
Fixing HTTP2 client.  When creating the HTTPS connector with hyper_rustls::HttpsConnectorBuilder, the use of enable_http1() was preventing the initiation of an HTTP2 connection. Switching to enable_all_versions() successfully initiates HTTP2. 

I've successfully tested this with my local fork.

### DIFF
--- a/client/http-client/Cargo.toml
+++ b/client/http-client/Cargo.toml
@@ -16,7 +16,7 @@ publish = true
 [dependencies]
 async-trait = "0.1"
 hyper = { version = "0.14.10", features = ["client", "http1", "http2", "tcp"] }
-hyper-rustls = { version = "0.24", optional = true, default-features = false, features = ["http1", "tls12", "logging"] }
+hyper-rustls = { version = "0.24", optional = true, default-features = false, features = ["http1", "http2", "tls12", "logging"] }
 jsonrpsee-types = { workspace = true }
 jsonrpsee-core = { workspace = true, features = ["client", "http-helpers"] }
 serde = { version = "1.0", default-features = false, features = ["derive"] }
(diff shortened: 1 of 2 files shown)


niklasad1 (commented):
Looks good to me, how does it work with the `hyper::Request::post`?
With  both http2/http1 enabled which is used for the request? 

Thus, it's not possible to select whether to use explicitly use http1 or http2 and I think we should expose that in the HttpClientBuilder.....

venugopv:
Thank you, @niklasad1, for your quick review comments.

I have done tests on HTTPS hosts to see how _hyper::Request::post_ behaves, Here are my findings:

- For hosts that support both HTTP/1.1 and HTTP/2, the _hyper::Request::post_ method defaults to initiating an HTTP/2 connection.
- Conversely, for hosts that only support HTTP/1.1, the same method defaults to an HTTP/1.1 connection.

So it is working as expected in HTTP protocol negotiation.

I also felt having the option to explicitly choose between HTTP/1 and HTTP/2 in HttpClientBuilder() would be beneficial. However, Hyper does not provide an option for explicitly opting for HTTP/1.1 only. It does offer a http2_only: https://github.com/hyperium/hyper/blob/0.14.x/src/client/client.rs#L1194

So I am not sure how to configure http1_only for HTTP without a explicit hyper method.

For HTTPS connections, this could be managed within HttpClientBuilder() by specifying the preferred protocol when creating a hyper_rustls::HttpsConnector.

niklasad1 (approved):

jsdw (approved):
Offhand sounds/looks reasonable to me!

PR #1255, opened 2023-12-05
refactor(client subscription): replace old messages when the subscription can't keep up with the server

### LINKED ISSUE #1291: [client]: simplify subscription API
The "current" subscription API for the client doesn't expose how a subscription was closed.

The subscription can return :
    - None: The connection was closed
    - None: The subscription could not keep up with server a.k.a slow-receiver problem.
    - Err(e):  Failed to parse the subscription item as T

Thus, we should modify the API to do the following:
- None: The connection was closed
- Err(SubscriptionError::Lagged) - The subscription lagged too far behind i.e, could not keep up with the server.  The error itself indicates how many messages that were missed. 
- Err(Subscription::Parse) - Failed to decode item as T

#### Some notes about SubscriptionError::Lagged 

When this occurs it's up the user the decide whether that is okay to loose a few messages to continue or re-subscribe.

niklasad1:
The "current" subscription API for the client doesn't expose how a subscription was closed and drops the subscription when it can't keep up with the server

It can be closed by:

- The connection was closed
- The subscription could not keep up with server a.k.a slow-receiver problem.

This PR changes that the underlying channel to `async_broadcast` which exposes
`RecvError::Overflowed(n)` when the channel lagged behind and oldest message(s) are replaced with new messages such that users can decide themselves what to do if a subscription message was dropped/replaced.

Close https://github.com/paritytech/jsonrpsee/issues/1291


tests/tests/integration_tests.rs
```diff
@@ -296,7 +296,7 @@ async fn ws_subscription_several_clients_with_drop() {
 	let mut clients = Vec::with_capacity(10);
 	for _ in 0..10 {
 		let client = WsClientBuilder::default()
-			.max_buffer_capacity_per_subscription(u32::MAX as usize)
+			.max_buffer_capacity_per_subscription(u16::MAX as usize)
```
niklasad1, line 299: This failed locally because it tried to allocate lots of memory.

https://docs.rs/tokio/latest/src/tokio/sync/broadcast.rs.html#501-534 it reserves the capacity which is different from the mpsc channel

core/src/client/mod.rs
```diff
@@ -213,18 +218,39 @@ pub enum SubscriptionKind {
...
+/// 
+/// ## Lagging
+/// 
+/// As all sent messages must be kept in a buffer until the underlying
+/// stream is polled, it's possible that the server is producing
+/// subscription notifications faster than the client can handle.
+/// 
+/// When that occurs an error [`SubscriptionError::Lagged`] is emitted
+/// to indicate the n messages were lost. You can decide to continue using
+/// the subscription if that's accepted the oldest messaage in the buffer
```
lexnv, line 233: nit: `, the oldest message`

core/src/client/mod.rs
```diff
@@ -213,18 +218,39 @@ pub enum SubscriptionKind {
...
+/// will be returned on the next read operation.
+/// 
+/// To avoid `Lagging` from happening you may increase the buffer capacity
+/// or ensure that [`Subscription::next`] polled often enough such as
```
lexnv, line 237: nit: is polled

core/src/client/mod.rs
```diff
@@ -343,29 +371,52 @@ where
 	Notif: DeserializeOwned,
 {
 	/// Returns the next notification from the stream.
-	/// This may return `None` if the subscription has been terminated,
-	/// which may happen if the channel becomes full or is dropped.
+	/// 
+	/// The return values can be:
+	/// - `Some(Ok(val)`: The message was succesfully received and decoded.
+	/// - `Some(Err(SubscriptionError::Lagged(n)))`: The subscription buffer was full and n messages were lost.
+	/// - `Some(Err(SubscriptionError::Parse: the value failed to be decode as `Notif`
```
lexnv, line 378: nit: "Some(Err(SubscriptionError::Parse))" (closing brackets)

core/src/client/mod.rs
```diff
@@ -343,29 +371,52 @@ where
...
+	/// - `None`: The connection was closed.
```
lexnv, line 379: So, `None` is returned when: (a) connection is dropped; and (b) buffer is drained?

lexnv (approved):
Looks clean!

core/src/client/mod.rs
niklasad1: No, it should only return None when the stream is "closed" 

https://docs.rs/tokio-stream/latest/src/tokio_stream/wrappers/broadcast.rs.html#54-67

core/src/client/mod.rs
```diff
@@ -213,18 +218,42 @@ pub enum SubscriptionKind {
...
 
-/// Active subscription on the client.
-///
-/// It will try to `unsubscribe` in the drop implementation
-/// but it may fail if the underlying buffer is full.
-/// Thus, if you want to ensure it's actually unsubscribed then
-/// [`Subscription::unsubscribe`] is recommended to use.
+/// Represent a client-side subscription which is implemented on top of 
+/// a bounded channel where it's possible that the receiver may
+/// not keep it with the sender side a.k.a "slow receiver problem"
```
jsdw, line 223:
```suggestion
/// not keep up with the sender side a.k.a "slow receiver problem"
```

core/src/client/mod.rs
```diff
@@ -213,18 +218,42 @@ pub enum SubscriptionKind {
...
+/// in a separate tokio task.
+/// 
+/// ## Connection closed
+/// 
+/// When the connection is closed the underlying stream will eventually
+/// return `None` to indicate that.
+/// 
+/// Because the subscription is implemented on top of a bounded channel
+/// it will not instantly return `None` when the connection is closed 
+/// because all messages buffered must be read before it to return `None`.
```
jsdw, line 250:
```suggestion
/// because all messages buffered must be read before it returns `None`.
```

core/src/client/async_client/mod.rs
```diff
@@ -162,8 +162,15 @@ impl ClientBuilder {
 	///
 	/// # Panics
 	///
-	/// This function panics if `max` is 0.
+	/// This function panics if `max` is 0 or bigger than usize::MAX / 2.
```
jsdw, line 165: This now allocates the entire capacity up front, so I think we should be super clear that this is now allocating when before it wasn't.

I'd suggest changing the name entirely to eg `with_capacity_per_subscription` (partly so that anybody using this function is forced to find the new one) and make it clear in the docs that it's not a "max capacity" but an actual allocating "capacity" (like `Vec::with_capacity`)

core/src/client/mod.rs
```diff
@@ -213,18 +218,42 @@ pub enum SubscriptionKind {
...
-/// [`Subscription::unsubscribe`] is recommended to use.
+/// Represent a client-side subscription which is implemented on top of 
+/// a bounded channel where it's possible that the receiver may
+/// not keep it with the sender side a.k.a "slow receiver problem"
+/// 
+/// ## Lagging
+/// 
+/// As all sent messages must be kept in a buffer until the underlying
+/// stream is polled, it's possible that the server is producing
+/// subscription notifications faster than the client can handle.
```
jsdw, line 229: Just to check that I understand this lagging stuff; I think the idea is that the server is sending messages to the client, but if you don't poll the client enough then the buffer on the client will fill up, and then we'll start missing messages if we aren't polling the client subscription enough to drain the buffer right?

```suggestion
/// All messages from the server must be kept in a buffer in the client
/// until they are read by polling the [`Subscription`]. If you don't 
/// poll the client subscription quickly enough, the buffer may fill 
/// up, which will result in messages being lost.
```

jsdw:
Looks good to me; just a couple of bits to clarify and make sure people aren't surprised by the sudden up-front allocating. (I do wonder how much that up front allocating might impact the memory usage in a client, because now the max buffer size is just the actual buffer size :))

core/src/client/async_client/mod.rs
niklasad1: It's doesn't allocate when the client is instantiated but when each subscription is created.

Sure good idea.

client/ws-client/src/lib.rs
```diff
@@ -198,8 +198,11 @@ impl WsClientBuilder {
 	}
 
 	/// See documentation [`ClientBuilder::max_buffer_capacity_per_subscription`] (default is 1024).
-	pub fn max_buffer_capacity_per_subscription(mut self, max: usize) -> Self {
-		self.max_buffer_capacity_per_subscription = max;
+	pub fn max_buffer_capacity_per_subscription(mut self, capacity: usize) -> Self {
+		// https://docs.rs/tokio/latest/src/tokio/sync/broadcast.rs.html#501-506
```
niklasad1, line 202:
```suggestion
```

core/Cargo.toml
```diff
@@ -25,15 +25,17 @@ tracing = "0.1.34"
...
 async-lock = { version = "3.0", optional = true }
+# Latest version doesn't compile for WASM: https://github.com/smol-rs/async-broadcast/issues/50
+async-broadcast = { version = "0.5", optional = true }
 futures-util = { version = "0.3.14", default-features = false, optional = true }
 hyper = { version = "0.14.10", default-features = false, features = ["stream"], optional = true }
 rustc-hash = { version = "1", optional = true }
 rand = { version = "0.8", optional = true }
 parking_lot = { version = "0.12", optional = true }
 tokio = { version = "1.16", optional = true }
+tokio-stream = { version = "0.1", optional = true, features = ["sync"] }
```
niklasad1, line 36:
```suggestion
```

core/src/client/mod.rs
```diff
@@ -388,6 +474,27 @@ impl<Notif> Drop for Subscription<Notif> {
 	}
 }
 
+impl<Notif> Stream for Subscription<Notif>
```
niklasad1, line 477: technically we should be able to do this instead but seems to some waker issue which doesn't always occur


```rust
impl<Notif> Stream for Subscription<Notif>
where
	Notif: DeserializeOwned,
{
	type Item = Result<Notif, SubscriptionError>;

	fn poll_next(
		self: std::pin::Pin<&mut Self>,
		cx: &mut std::task::Context<'_>,
	) -> std::task::Poll<Option<Self::Item>> {
		let mut this = Pin::new(self);

		let mut fut = this.rx.inner.recv();

		match fut.poll_unpin(cx) {
			Poll::Ready(Ok(v)) => Poll::Ready(Some(serde_json::from_value(v).map_err(Into::into))),
			Poll::Ready(Err(RecvError::Closed)) => Poll::Ready(None),
			Poll::Ready(Err(RecvError::Overflowed(n))) => Poll::Ready(Some(Err(SubscriptionError::Lagged(n)))),
			Poll::Pending => Poll::Pending,
		}
	}
}

```

niklasad1: seems like some race-condition that doesn't occur with the default stream impl

niklasad1: yeah ok I see the recv future has a `listener` which needs dropped when poll_next returns which makes it only work if channel is non-empty and it may not be woken up again after Poll::Pending

niklasad1: aight, I see that I could add a field `pending` to the struct to be able to keep the state but tricky with borrowing....

niklasad1: just to prove my point this hack fixes the stream impl and nothing else is causing that:

```rust
impl Stream for SubscriptionRx {
	type Item = Result<serde_json::Value, SubscriptionError>;

	fn poll_next(
		self: std::pin::Pin<&mut Self>,
		cx: &mut std::task::Context<'_>,
	) -> std::task::Poll<Option<Self::Item>> {
		let mut this = Pin::new(self);

		loop {
			if let Some(f) = this.wake.as_mut() {
				futures_util::ready!(f.poll_unpin(cx));
			}

			// The timer has expired and mark it as done;
			this.wake = None;

			// Attempt to receive a message.
			match this.inner.try_recv() {
				Ok(msg) => return Poll::Ready(Some(Ok(msg))),
				Err(TryRecvError::Closed) => return Poll::Ready(None),
				Err(TryRecvError::Overflowed(n)) => {
					return Poll::Ready(Some(Err(SubscriptionError::Lagged(n))));
				}
				Err(TryRecvError::Empty) => {
					tracing::info!("sub empty");
				}
			}

			if this.wake.is_none() {
				tracing::info!("enable sleep 3 seconds");
				this.wake = Some(futures_timer::Delay::new(Duration::from_secs(3)));
			}
		}
	}
}
```

niklasad1: /cc @jsdw any ideas here? :)

jsdw: Mmm I see the issue! You can't store `this.inner.rx.recv()` because of the lifetime, but would need to otherwise dropping it throws the waker away and so nothing is woken up.

By the looks of it, `async_broadcast::Receiver` impls Stream, so you _could _ expose a way to call `self.inner.rx.inner.poll_next` directly instead?

niklasad1: yeah I'm already doing that in this PR but the issue with that is https://docs.rs/async-broadcast/latest/src/async_broadcast/lib.rs.html#1385, thus the "lagged/overflowed" is ignored and that's not propagated on the stream.

The entire point of this manual impl is to include "lagged/overflowed" error as well....

niklasad1: Thus, the behavior between `StreamExt::next` and `Subscription::recv` are different and for example in subxt we are relying on the stream impl and if the stream lags then we will replace old items and the user is not notified about it.

Not ideal and not the intention with this PR ^^

jsdw: Ooh I see; I didn't pay enough attention to notice that the vehaviours were different! That sucks. It doesn't look like you can recreate the logic in their `poll_next` and tweak it in the way you'd need offhand, so it feels like the only options left are:

- Open issue/PR with `async-broadcast` people to see if we can expose what we need there (perhaps by writing a Stream impl that returns all of the details exposed by `recv`, which needn't be the default or chaneg any of their APIs, or
- Write our own naive async broadcast channel so that we have total control over it, or
- Try and find something else out there that does what we want

PR #1248, opened 2023-11-29
server: change ws ping API 

niklasad1:
Close #1195

PR #1258, opened 2023-12-06
refactor(client): unify ws ping/pong API with the server

niklasad1:
This changes the clients to support a similar WebSocket ping/pong as the server which
was introduced in https://github.com/paritytech/jsonrpsee/pull/1248

The `PingConfig` type is duplicated in the server and client but I'm fine with that for now.

After introducing `Instant` in code I realized that the ping/pong interval would panic in wasm and I have removed ping/pongs completely because gloo-net doesn't support that anyway.

The other solution is to use some other crates where instants/interval stream are supported but it's not worth the effort IMHO and I think there was some reason in JS land where it's not possible to have client-side initiated pings.

### DIFF
--- a/core/src/client/async_client/mod.rs
+++ b/core/src/client/async_client/mod.rs
@@ -74,12 +74,14 @@ const LOG_TARGET: &str = "jsonrpsee-client";
 /// an inactive connection.
 ///
 /// jsonrpsee doesn't associate the ping/pong frames just that if
-/// pong frame isn't received within `inactive_limit` then it's regarded
+/// a pong frame isn't received within the `inactive_limit` then it's regarded
 /// as missed.
 ///
 /// Such that the `inactive_limit` should be configured to longer than a single
 /// WebSocket ping takes or it might be missed and may end up
 /// terminating the connection.
+///
+/// Default: ping_interval: 30 seconds, max failures: 1 and inactive limit: 40 seconds.
 #[derive(Debug, Copy, Clone)]
 pub struct PingConfig {
 	/// Interval that the pings are sent.
--- a/core/src/client/async_client/mod.rs
+++ b/core/src/client/async_client/mod.rs
@@ -16,12 +16,16 @@ use crate::JsonRawValue;
 use std::borrow::Cow as StdCow;
 
 use core::time::Duration;
+use std::num::NonZeroUsize;
+use std::time::Instant;
 use helpers::{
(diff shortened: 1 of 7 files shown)


core/src/client/async_client/mod.rs
```diff
@@ -970,3 +1078,5 @@ impl<Fut: Future> Stream for MaybePendingFutures<Fut> {
 		self.futs.poll_next_unpin(cx)
 	}
 }
+
+
```
niklasad1, line 1082:
```suggestion
```

core/src/client/async_client/mod.rs
```diff
@@ -831,53 +952,36 @@ where
...
 				}
-				_ = ping.tick() => {
+				_ = ping_interval.next() => {
 					if let Err(err) = sender.send_ping().await {
 						tracing::error!(target: LOG_TARGET, "Could not send ping frame: {err}");
 						break Err(Error::Custom("Could not send ping frame".into()));
 					}
 				}
 			}
-		}
```
jsdw, line 841: Nice to de-duplicate these bits :)

jsdw (approved):
Looks solid to me; good job!

core/src/client/async_client/mod.rs
```diff
@@ -256,14 +370,21 @@ impl ClientBuilder {
 		S: TransportSenderT,
 		R: TransportReceiverT,
 	{
+		use futures_util::stream::Pending;
+
+		pub(crate) type PendingIntervalStream = IntervalStream<Pending<()>>;
```
niklasad1, line 375:
```suggestion
		type PendingIntervalStream = IntervalStream<Pending<()>>;		
```

core/src/client/async_client/mod.rs
```diff
@@ -22,26 +50,92 @@ use helpers::{
...
+/// Configuration for WebSocket ping/pong mechanism and it may be used to disconnect
+/// an inactive connection.
+///
+/// jsonrpsee doesn't associate the ping/pong frames just that if
+/// pong frame isn't received within `inactive_limit` then it's regarded
+/// as missed.
+///
+/// Such that the `inactive_limit` should be configured to longer than a single
+/// WebSocket ping takes or it might be missed and may end up
+/// terminating the connection.
```
lexnv, line 82: nit: Might be worth stating the default values

lexnv (approved):
Nice one! 👍

PR #1263, opened 2023-12-11
refactor: set `tcp_nodelay == true` by default and explicit APIs

### LINKED ISSUE #1262: Consider adding explicit capability to set TCP_NODELAY on client
Currently in order to set `TCP_NODELAY` on a socket used by an `HttpClient`, one can use (abuse?) the ability to set middleware in the `HttpClientBuilder` like this:

```rust
let client_builder =
    HttpClientBuilder::default().set_middleware(tower::ServiceBuilder::new().layer_fn(|_| {
        let mut connector = HttpConnector::new();
        connector.set_nodelay(true);

        // TODO handle https? copy line 120-135 in client/http-client/src/transport.rs
        HttpBackend::Http(Client::builder().build(connector))
    }));

let client = client_builder.build(url).unwrap();
```

This feels kind of like a hack since I must override the `HttpBackend` and essentially render [this](https://github.com/paritytech/jsonrpsee/blob/073aa2815f0581d5669ab5910e8313f508e543b5/client/http-client/src/transport.rs#L117C2-L117C2) whole block of code useless. Also not shown in the example snippet above is the necessary logic to construct an `HttpBackend::Https` if the url I plan to connect to is https, which breaks the expectation that `client_builder.build(url)` works for either http or https since I now must know the url in advance.

All of this complexity is required to set an option that perhaps should be set by default in an RPC library. At least, there should be a more ergonomic way to set this, so users don't have to reimplement logic internal to the library while also breaking expected behavior.

Without `TCP_NODELAY`, I can observe latency on a request to `localhost` stepping up from `1ms` to `40ms` after adding one byte to the request body beyond a certain threshold. In latency-sensitive use-cases, this matters a lot and it is necessary for users to be able to set `TCP_NODELAY` on the client.

Is there a better way to set this option that I have missed? Is the way I have found to set it through overriding the `HttpBackend` service through middleware even something that was intended? If there was no intended way to set this, are you open to adding a more ergonomic way to do so? Do you have any strong opinions about setting `TCP_NODELAY` to true by default on the client, or perhaps disabling delayed ACKs by default on the server?

niklasad1:
Close #1262

### DIFF
--- a/client/http-client/src/transport.rs
+++ b/client/http-client/src/transport.rs
@@ -123,25 +123,27 @@ where
 			}
 			#[cfg(feature = "__tls")]
 			"https" => {
-				let builder = match cert_store {
+				let mut http_conn = HttpConnector::new();
+				http_conn.set_nodelay(tcp_no_delay);
+				http_conn.enforce_http(false);
+
+				let https_conn = match cert_store {
 					#[cfg(feature = "native-tls")]
 					CertificateStore::Native => hyper_rustls::HttpsConnectorBuilder::new()
 						.with_native_roots()
 						.https_or_http()
-						.enable_all_versions(),
+						.enable_all_versions()
+						.wrap_connector(http_conn),
 					#[cfg(feature = "webpki-tls")]
 					CertificateStore::WebPki => hyper_rustls::HttpsConnectorBuilder::new()
 						.with_webpki_roots()
 						.https_or_http()
-						.enable_all_versions(),
+						.enable_all_versions()
+						.wrap_connector(http_conn),
 					_ => return Err(Error::InvalidCertficateStore),
 				};
 
-				let mut connector = HttpConnector::new();
-				connector.set_nodelay(tcp_no_delay);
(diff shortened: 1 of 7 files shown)


lexnv (approved):
Nice one!

client/http-client/src/transport.rs
```diff
@@ -256,169 +373,70 @@ impl From<hyper::Error> for Error {
...
-			80,
-			"http://localhost:-43",
-			80,
-			CertificateStore::Native,
-			80,
-			HeaderMap::new(),
-			tower::ServiceBuilder::new(),
-		)
-		.unwrap_err();
+		let err = HttpTransportClientBuilder::new().build("ws://localhost:-43").unwrap_err();
```
niklasad1, line 399:
```suggestion
		let err = HttpTransportClientBuilder::new().build("http://localhost:-43").unwrap_err();
```

lexnv (approved):

PR #1277, opened 2024-01-18
fix: allow application/json-rpc http content type

dan-starkware:
Hi, consider allowing 'application/json-rpc' content type.

I came across [this](https://www.jsonrpc.org/historical/json-rpc-over-http.html) (deprecated?) JSON-RPC over HTTP specification file that indicates that content-type should be 'application/json-rpc'.
Also, several widely used libraries, such as jsonrpc4j [use](https://github.com/briandilley/jsonrpc4j/blob/755cccdf325f8996d89cbaaca82de0b4ae3acf85/src/main/java/com/googlecode/jsonrpc4j/JsonRpcBasicServer.java#L37) it.


niklasad1 (approved):
Fair enough, I see no problem by supporting these content types.

The assumption from our side is that "application/json" is mainly used for JSON-RPC v2 but it's not really specified there.

lexnv (approved):
Thanks for contributing!

Issue #1067, opened 2023-04-04
proc macros: IntoResponse for client API


niklasad1:
After #1057 is merged then it possible for the server implementation to add custom return types which isn't necessarily `Result<T, E>` i.e, it could a type that just implements `IntoResponse`. While this is a nice improvement for the server, it doesn't work for the clients as the responses could fail a bunch of reason such as I/O, decoding error and similar.

Thus, currently `Result<T, E>` only works when both the client and server attributes are enabled at an API.
For example the following won't compile:

```rust
#[derive(Serialize)]
struct Custom;

impl IntoResponse for Custom {
	type Output = Self;

	fn into_response(self) -> PartialResponse<Self::Output> {
		let error_object = jsonrpsee::types::ErrorObjectOwned::owned(0, "custom_error", None::<()>);
		PartialResponse::Error(error_object)
	}
}

#[rpc(server, namespace = "foo")]
pub trait Rpc {
	#[method(name = "method1")]
	async fn method1(&self) -> Custom;
}
```

For example we could try to parse the return type if it's not a result type and generate:

```rust
  async fn method1(&self) -> Result<Custom, Error>;
```

This leads to question whether group the server and client is such a good idea since there are basically to much magic happens in the proc macro. 

Thus, it might be better split them up but I still very much like the idea that an API can be shared by a server/client to avoid duplicated all things.

PR #1281, opened 2024-01-30
feat: async API when `Response` has been processed.

### LINKED ISSUE #1279: add low-level API for method implementations
It has come up with a special use-case that a method implementation would want to start sending out
notifications after the method call finished.

In jsonrpsee the implementation looks roughly:

```rust
	fn bar(&self, sink: SubscriptionSink) -> u32 {
                let id: u32 = rand();
                tokio::spawn(async {
                     let mut cnt = 0;
                     loop {
                           sink.send(format("{id};{cnt}").await;
                     }
                });
                id
	}
```

As you may detect we want to ensure that "id" is sent before stuff returned back before the notifications are sent
out with this id.

Ideally, we would want some kind of closure that is invoked once the fn bar returns.

niklasad1:
Resolves #1279


examples/examples/ws_notify_on_method_answered.rs
```diff
@@ -0,0 +1,84 @@
...
+impl RpcServer for RpcServerImpl {
+	fn storage_keys(&self) -> ResponsePayloadV2<'static, String> {
+		let (tx, rx) = response_channel();
+
+		tokio::spawn(async move {
+			rx.is_sent().await.unwrap();
+			println!("Method response to `state_getKeys` finished");
+		});
+
+		ResponsePayloadV2::result("ehheeheh".to_string()).notify_on_success(tx)
```
niklasad1, line 53: Basically this is the new API which enforces that `rx` doesn't return until the message has been sent to internal websocket buffer

core/src/server/rpc_module.rs
```diff
@@ -485,7 +484,71 @@ impl<Context> From<RpcModule<Context>> for Methods {
...
+	///        loop {
+	///           tokio::time::sleep(std::time::Duration::from_millis(100)).await;
+	///           println!("after method response");
+	///       }
+	///    });
+	///
+	///    MethodResponse::response(id, ResponsePayloadV2::result("foo"), max_response_size).notify_on_success(tx)
+	/// }).unwrap();
+	///```
+	pub fn register_raw_method<F>(
```
niklasad1, line 523: this doesn't bring much functionality rather than access to all data such as raw JSON params and ID, not sure if we need it

core/src/server/helpers.rs
```diff
@@ -358,25 +494,101 @@ impl BatchResponseBuilder {
...
 /// Create a JSON-RPC error response.
 pub fn batch_response_error(id: Id, err: impl Into<ErrorObject<'static>>) -> String {
 	let err = ResponsePayload::error_borrowed(err);
 	serde_json::to_string(&Response::new(err, id)).expect("JSON serialization infallible; qed")
 }
 
+/// Similar to [`ResponsePayload`] but possible to with an async-like
+/// API to detect when a method response has been sent.
+#[derive(Debug)]
+pub struct ResponsePayloadV2<'a, T>
```
niklasad1, line 521: TODO: rename this to `ResponsePayload` technically a breaking change because `ResponsePayload` was enum before but I may treat as non-breaking change to easily backport it v0.20

proc-macros/src/rpc_macro.rs
```diff
@@ -71,6 +73,17 @@ impl RpcMethod {
...
 		}
 
+		if raw && sig.asyncness.is_some() {
+			return Err(syn::Error::new(sig.span(), "Raw method must be synchronous"));
+		}
+
+		if raw && sig.inputs.len() != 4 {
+			return Err(syn::Error::new(
+				sig.span(),
+				"Raw methods must have arity 4, such as `fn foo(&self, id: Id, params: Params, max_response: usize)`",
```
lexnv, line 83: nit: 4 parameters?

core/src/server/helpers.rs
```diff
@@ -24,20 +24,67 @@
...
 
-use super::{DisconnectError, SendTimeoutError, SubscriptionMessage, TrySendError};
+/// Wrapper over [`tokio::mpsc::oneshot::channel`]..
+pub fn response_channel() -> (MethodResponseNotifyTx, MethodResponseNotifyRx) {
+	let (tx, rx) = tokio::sync::oneshot::channel();
+	(tx, MethodResponseNotifyRx(rx))
+}
+
+/// ..
+pub type MethodResponseNotifyTx = tokio::sync::oneshot::Sender<NotifyMsg>;
```
lexnv, line 49: nit: We could wrap this into a struct to avoid breaking changes in user-code; although they are not supposed to use it

lexnv:
Overall looks good to me, nice one 👍 ! And in combination with the [polkadot-sdk commit](https://github.com/paritytech/polkadot-sdk/commit/df5c0da7b1df970e87145e5c42cdf5e26bb8ed47#diff-85e700332cc343b498a3d97dd807aa72af6e0abfcad6046645d351d96010943bR343-R360) this should resolve any races from happening.

I find the `is_sent()` method name a bit confusing; in the sense that the message is sent to our internal buffers; but can't think of any other names. This should be fine since this is a niece low-level API

niklasad1:
> I find the is_sent() method name a bit confusing; in the sense that the message is sent to our internal buffers; but can't think of any other names. This should be fine since this is a niece low-level API

Yeah, agree. I don't like it either I thought it was nice wrapper but it makes things more confusing I could remove the wrapper completely...

core/src/server/method_response.rs
```diff
@@ -0,0 +1,537 @@
...
+	where
+		T: Serialize + Clone,
+	{
+		let mut rp = Self::response(id, result, max_response_size);
+		rp.kind = ResponseKind::Subscription;
+		rp
+	}
+
+	/// Optional callback that may be utilized to notify that the method response was
+	/// a succesfully JSON-RPC object and processed.
```
jsdw, line 191:
```suggestion
	/// Optional callback that may be utilized to notify that the method response was
	/// processed as a succesfulJSON-RPC object.
```

core/src/server/method_response.rs
```diff
@@ -0,0 +1,537 @@
...
+	pub fn notify_on_success(mut self, tx: MethodResponseNotifyTx) -> Self {
+		self.on_close = Some(NotifyKind::Success(tx));
+		self
+	}
+
+	/// Optional callback that may be utilized to notify that the method response was a
+	/// an JSON-RPC error object and processed.
```
jsdw, line 198:
```suggestion
	/// Optional callback that may be utilized to notify that the method response was
	/// processed as a JSON-RPC error object.
```

core/src/server/method_response.rs
```diff
@@ -0,0 +1,537 @@
...
+	pub fn notify_on_error(mut self, tx: MethodResponseNotifyTx) -> Self {
```
jsdw, line 199: Also; how this is set up at the moment means that if somebody does

```rust
response
    .notify_on_success(...)
    .notify_on_error(...)
```

They might expect one of those to fire in each case, but in fact only the last registered one has any chance of ever firing.

I guess I'd either have more than one callback that can be registered, or safer still, just have `notify_on_response` and tell that whether success or error happened.

core/src/server/method_response.rs
```diff
@@ -0,0 +1,537 @@
...
+		InnerResponsePayload::Result(StdCow::Owned(t)).into()
+	}
+
+	/// Create successful borrowed response payload.
+	pub fn result_borrowed(t: &'a T) -> Self {
+		InnerResponsePayload::Result(StdCow::Borrowed(t)).into()
+	}
+
+	/// Get a notification when the successful method response has been sent out.
+	pub fn notify_on_success(mut self, tx: MethodResponseNotifyTx) -> Self {
```
jsdw, line 389: Another small thought/question about this API; is there any benefit to the user passing in a `MethodResponseNotifyTx`?

You could have `response.notify_on_success(&mut self) -> MethodResponseNotifyRx` for instance, to save having to import the relevant type and create a notify channel pair.

Sample usage might be:

```rust
let res = ResponsePayload::result(..);

let on_complete = res.notify_on_completion();

tokio::spawn(async move {
    if let false = on_complete.await {
        return
    };

    // do things after message sent and successful.
});

res
```

core/src/server/rpc_module.rs
jsdw: I'd probably not bother adding it for now because the .notify_on_ API does what we need anyways

If we do keep it, then the comment should prob be tweaked, since "it's possible to get enforce order of things on the connection" isn't unique to this any more now anyways :)

examples/examples/ws_notify_on_method_answered.rs
```diff
@@ -0,0 +1,84 @@
...
+impl RpcServer for RpcServerImpl {
+	fn storage_keys(&self) -> ResponsePayload<'static, String> {
+		let (tx, rx) = response_channel();
+
+		tokio::spawn(async move {
+			rx.await.unwrap();
+			println!("Method response to `state_getKeys` finished");
+		});
+
+		ResponsePayload::result("ehheeheh".to_string()).notify_on_success(tx)
```
jsdw, line 53: I can see that a benefit of the more explicit `response_channel()` thing over my suggestion above is that you can keep the order nice here and just return a response payload immediately!

proc-macros/src/render_client.rs
```diff
@@ -109,10 +105,20 @@ impl RpcDescription {
 			let ret_ty = args.last_mut().unwrap();
 			let err_ty = self.jrps_client_item(quote! { core::client::Error });
 
+			quote! { core::result::Result<#ret_ty, #err_ty> }
+		} else if type_name.ident == "ResponsePayload" {
```
jsdw, line 109: I'm sad that we have to match on the string type names :D

I guess it wasn't possible to use a trait to convert these into results automatically or something; I'm sure it was something like that but can't remember the issue now! But ah well!

proc-macros/src/render_client.rs
```diff
@@ -72,19 +72,15 @@ impl RpcDescription {
 	/// Verify and rewrite the return type (for methods).
 	fn return_result_type(&self, mut ty: syn::Type) -> TokenStream2 {
 		// We expect a valid type path.
-		let syn::Type::Path(ref mut type_path) = ty else {
-			return quote_spanned!(ty.span() => compile_error!("Expecting something like 'Result<Foo, Err>' here. (1)"));
-		};
+		let syn::Type::Path(ref mut type_path) = ty else { return self.untyped_client() };
```
jsdw, line 75: I don't really understand the effects of this return self.untyped_client() change; it looks like it turns some compile errors into successful returns; is this to play it safe because if the type the user returned isn't valid, we'll hit a type error anyways at this point?

jsdw:
Looks good to me; the only actual thing I think should be addressed is making `res.notify_on_success(..).notify_on_error(..)` either impossible or do as you might expect (ie fire all of the provided channels).

core/src/server/method_response.rs
jsdw: Though I'd also probably want it so that `response.notify_on_completion(a).notify_on_completion(b)` also worked and fired both a and b; one could store a vec of things-to-be-notified perhaps?

core/src/server/method_response.rs
niklasad1: Good idea :)

proc-macros/src/render_client.rs
niklasad1: This is a really opinionated change, basically any type that isn't `Result<T, Error>`, `RpcResult<T>` or `ResponsePayload<T>` then I changed the client to generate `Result<serde_json::Value, Error>`.

To fix  https://github.com/paritytech/jsonrpsee/issues/1067

proc-macros/src/render_client.rs
niklasad1: Nah, it's that the client codegen is super opinionated and if the error type must be `jsonrpsee::core::client::Error` otherwise it won't compile. Yeah, it's ugly but just proc macro code thank god.

IntoResponse stuff doesn't work there because we have deal with I/O errors and bunch of things

proc-macros/src/render_client.rs
niklasad1: Revert this, to have backward compatible stuff but technically right now the IntoResponse is breaking because it returns `jsonrpsee::core::server::ResponsePayload` instead of ` `jsonrpsee::types::ResponsePayload`

proc-macros/tests/ui/correct/custom_ret_types.rs
```diff
@@ -4,8 +4,7 @@ use std::net::SocketAddr;
 
 use jsonrpsee::core::{async_trait, ClientError, Serialize};
 use jsonrpsee::proc_macros::rpc;
-use jsonrpsee::server::{IntoResponse, ServerBuilder};
-use jsonrpsee::types::ResponsePayload;
+use jsonrpsee::server::{IntoResponse, ResponsePayload, ServerBuilder};
```
niklasad1, line 7: This is a breaking change as well

jsdw: Aah I see; it moved! So will you do a proper release then instead of a patch one do you think? Or alias this type in `jsonrpsee::types` in the release to avoid this break?

proc-macros/tests/ui/correct/custom_ret_types.rs
```diff
@@ -31,7 +30,7 @@ impl IntoResponse for CustomError {
 		let data = data.map(|val| serde_json::value::to_raw_value(&val).unwrap());
 
 		let error_object = jsonrpsee::types::ErrorObjectOwned::owned(code, "custom_error", data);
-		ResponsePayload::Error(error_object)
+		ResponsePayload::error(error_object)
```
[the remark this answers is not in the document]
jsdw: Aaahhh I see! I was thinking of something else :)

core/src/server/method_response.rs
```diff
@@ -482,6 +483,11 @@ impl Future for MethodResponseNotifyRx {
 	}
 }
 
+/// An error that may occur if ResponsePayload::notify_on has been called more than once.
+#[derive(thiserror::Error, Clone, Copy, Debug)]
+#[error("ResponsePayload::notify_on can only be called once")]
+pub struct AlreadyNotifiedError;
```
jsdw, line 489: Is this actually used now?

jsdw (approved):
Looks good to me offhand!

It feels like we'll need to do a major release given the breaks unless you mangle a couple of things in a patch release to keep the old APIs/imports :)

proc-macros/tests/ui/correct/custom_ret_types.rs
niklasad1: I think we have to make a proper release because the MethodResponse is also a breaking change and I can't do proper alias in types because then the `feature flag core/server` needs to be enabled for ResponsePayload to be available.

core/src/server/method_response.rs
niklasad1: no :)

examples/examples/ws_notify_on_method_answered.rs
```diff
@@ -0,0 +1,84 @@
+// Copyright 2019-2021 Parity Technologies (UK) Ltd.
```
lexnv, line 1: nit: Maybe copyright years need to change

lexnv (approved):
Looks good to me!

proc-macros/src/render_server.rs
```diff
@@ -299,7 +299,7 @@ impl RpcDescription {
 		let params_fields = quote! { #(#params_fields_seq),* };
 		let tracing = self.jrps_server_item(quote! { tracing });
 		let sub_err = self.jrps_server_item(quote! { SubscriptionCloseResponse });
-		let response_payload = self.jrps_server_item(quote! { types::ResponsePayload });
+		let response_payload: TokenStream2 = self.jrps_server_item(quote! { ResponsePayload });
```
niklasad1, line 302:
```suggestion
		let response_payload = self.jrps_server_item(quote! { ResponsePayload });
```

core/src/server/method_response.rs
```diff
@@ -0,0 +1,571 @@
...
+		Self::error(err)
+	}
+}
+
+impl<'a> ResponsePayload<'a, ()> {
+	/// Similar to [`ResponsePayload::error`] but assigns `T = ()`.
+	///
+	/// This is useful when one only want to return the error and
+	/// `T` can't resolved by rustc to avoid type annonations.
+	pub fn unit_error(e: impl Into<ErrorObjectOwned>) -> Self {
```
niklasad1, line 448: /cc @jsdw @lexnv any thoughts on this one?

Basically, I would want a nice wrapper to avoid doing `ResponsePayload::<()>::error(err)` but maybe unit error is weird because it kind of reads like the error is `()`

examples/examples/ws_notify_on_method_answered.rs
niklasad1: let's take of it in another PR

core/src/server/method_response.rs
jsdw: When you first mentioned it I assumed it meant error is `()` which is what confused me :D

That said, I'm struggling to come up with a better name! I'm ok with just doing `ResponsePayload::<()>::error(err)` too I guess; if thename for any longer you'd not be saving any chars any more anyways!

server/src/transport/ws.rs
```diff
@@ -155,8 +154,20 @@ where
...
+
+					// The connection is closed, just quit.
+					if sink.send(serialized_rp).await.is_err() {
+						return;
+					}
+
+					// Notify that the message has been sent out to the internal
+					// WebSocket buffer.
+					if let Some(n) = on_close.take() {
+						_ = n.notify(is_success);
```
niklasad1, line 169:
```suggestion
						n.notify(is_success);
```

PR #1283, opened 2024-02-02
client(error): make display impl less verbose

### LINKED ISSUE #1280: core::ClientError: improve Display impl
The display implementations of [ClientError](https://github.com/paritytech/jsonrpsee/blob/master/core/src/client/error.rs#L35) should be improved for instance an I/O error when the background task got closed generates the following:

```bash
The background task been terminated because: Networking or low-level protocol error: WebSocket connection erro
r: i/o error: Transport endpoint is not connected (os error 107); restart required
```

Lots are redundant information and the I/O part should be sufficient.

niklasad1:
Resolves #1280

### DIFF
--- a/core/src/http_helpers.rs
+++ b/core/src/http_helpers.rs
@@ -67,7 +67,7 @@ where
 	let mut is_single = None;
 
 	while let Some(d) = body.data().await {
-		let data = d.map_err(|e| HttpError::Stream(e))?;
+		let data = d.map_err(HttpError::Stream)?;
 
 		// If it's the first chunk, trim the whitespaces to determine whether it's valid JSON-RPC call.
 		if received_data.is_empty() {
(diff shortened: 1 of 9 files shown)


jsdw (approved):
So basically this change restricts the body error type a bit and as such can avoid the generic boxed error (and a couple of display tweaks)?

Looks good to me!

lexnv (approved):

core/src/http_helpers.rs
```diff
@@ -61,7 +67,7 @@ where
 	let mut is_single = None;
 
 	while let Some(d) = body.data().await {
-		let data = d.map_err(|e| GenericTransportError::Inner(anyhow!(e.into())))?;
+		let data = d.map_err(|e| HttpError::Stream(e))?;
```
niklasad1, line 70:
```suggestion
		let data = d.map_err(HttpError::Stream)?;
```

PR #1284, opened 2024-02-05
feat(server): add `TowerService::on_session_close`

### LINKED ISSUE #1264: jsonrpsee TowerService: how to detect when a WebSocket connection is closed.
Previously with the RpcLogger it was easy to log when a connection was opened/closed but
it's not easy anymore because the `TowerService::call` returns once the HTTP request has 
been answered.

Thus, for instance I guess a common use-case is the emit prometheus metrics when a connection is opened and closed but the following won't work correctly:

```rust
    let make_service = make_service_fn(move |_conn: &AddrStream| {
        async move {
            Ok::<_, Infallible>(service_fn(move |req| {
                let metrics = metrics.clone();
                let svc_builder = svc_builder.clone();
                let is_websocket = ws::is_upgrade_request(&req);
                let transport_label = if is_websocket { "ws" } else { "http" };

                let metrics = metrics.map(|m| MetricsLayer::new(m, transport_label));
                let rpc_middleware = RpcServiceBuilder::new().option_layer(metrics.clone());
                let mut svc = svc_builder
                    .set_rpc_middleware(rpc_middleware)
                    .build(methods.clone(), stop_handle.clone());

                async move {
                    if is_websocket {
                        let now = std::time::Instant::now();
                        metrics.as_ref().map(|m| m.ws_connect());
                        // This will return once the HTTP upgrade request has been answered.
                        let rp = svc.call(req).await;
                        // Thus, this will not actually be when the connection was closed
                        // but once the WS upgrade handshake is done
                        metrics.as_ref().map(|m| m.ws_disconnect(now));
                        rp
                    } else {
                        svc.call(req).await
                    }
                }
                .boxed()
            }))
        }
    });
```

## Option 1

Introduce an alternative API to build the TowerService with future that resolves
once the connection is closed.

```rust
	/// Build a tower service and get a future to resolves when the connection
	/// has been terminated.
	pub fn build_with_close_handle(
		self,
		methods: impl Into<Methods>,
		stop_handle: StopHandle,
	) -> (TowerService<RpcMiddleware, HttpMiddleware>,  impl Future<Output = ()>) {
```

That is not great either because the TowerServiceBuilder can actually be reused/cloned to make another request....
Thus, the `future` will be fired once

## Option 2

Tell folks to use the low-level API but annoying for just the logging stuff.

niklasad1:
This PR adds another builder option to the `TowerServiceBuilder` get subscribe to when the connection/session has closed because when `TowerService::call` resolves it means that HTTP call has been finished but doesn't mean that WebSocket connection has closed. 

Thus, to know that we need another future sadly but worth the effort because otherwise the only way for folks it use the low-level API.

Resolves #1264

### DIFF
--- a/server/src/future.rs
+++ b/server/src/future.rs
@@ -182,11 +182,9 @@ impl Future for SessionClosedFuture {
 	fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
 		match self.0.poll_next_unpin(cx) {
 			Poll::Pending => Poll::Pending,
-			// A message is only sent when
-			Poll::Ready(x) => {
-				tracing::info!("{:?}", x);
-				Poll::Ready(())
-			}
+			// Only message is only sent and
+			// ignore can't keep up errors.
+			Poll::Ready(_) => Poll::Ready(()),
 		}
 	}
 }
--- a/server/src/future.rs
+++ b/server/src/future.rs
@@ -30,7 +30,7 @@ use std::pin::Pin;
 use std::sync::Arc;
 use std::task::{Context, Poll};
 
-use futures_util::{Stream, StreamExt};
+use futures_util::{Future, Stream, StreamExt};
 use pin_project::pin_project;
 use tokio::sync::{watch, OwnedSemaphorePermit, Semaphore, TryAcquireError};
 use tokio::time::Interval;
@@ -156,4 +156,28 @@ impl Stream for IntervalStream {
 			Poll::Pending
 		}
 	}
+}
+
+#[derive(Debug, Clone)]
+pub(crate) struct SessionCloseTx(tokio::sync::mpsc::Sender<()>);
+
+/// A future that resolves when the a connection
+/// has been closed.
+#[derive(Debug)]
+pub struct SessionCloseFuture(tokio::sync::mpsc::Receiver<()>);
+
+impl Future for SessionCloseFuture {
+	type Output = ();
+
+	fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+		match self.0.poll_recv(cx) {
+			Poll::Pending => Poll::Pending,
+			Poll::Ready(_) => Poll::Ready(()),
+		}
+	}
+}
+
+pub(crate) fn on_session_close() -> (SessionCloseTx, SessionCloseFuture) {
+	let (tx, rx) = tokio::sync::mpsc::channel(1);
+	(SessionCloseTx(tx), SessionCloseFuture(rx))
 }
--- a/server/src/future.rs
+++ b/server/src/future.rs
@@ -34,6 +34,7 @@ use futures_util::{Future, Stream, StreamExt};
 use pin_project::pin_project;
 use tokio::sync::{watch, OwnedSemaphorePermit, Semaphore, TryAcquireError};
 use tokio::time::Interval;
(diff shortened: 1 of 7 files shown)


jsdw (approved):
LGTM I think! Is there a way to test this?

server/src/server.rs
```diff
@@ -941,6 +942,25 @@ pub struct TowerService<RpcMiddleware, HttpMiddleware> {
...
+	/// a bounded buffer of 4 such that if one creates more than 4
+	/// SessionCloseFuture's. Then any of these 4 first futures
+	/// must be polled or dropped to make any progress.
+	pub fn on_session_closed(&mut self) -> SessionClosedFuture {
+		if let Some(n) = self.rpc_middleware.on_session_close.as_mut() {
+			// If it's called more then once another listener is created.
+			n.closed()
+		} else {
+			let (session_close, fut) = session_close();
+			self.rpc_middleware.on_session_close = Some(session_close);
```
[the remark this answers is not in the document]
niklasad1: Added some extra docs on it, it's possible.

Also it's a little bit weird that `TowerService::call` can be used several times, in practice I don't think that's the case but a footgun.

That `on_session_close` on works the subsequent `TowerService::call` as it does Option::take

server/src/future.rs
```diff
@@ -157,3 +158,42 @@ impl Stream for IntervalStream {
...
+#[derive(Debug)]
+pub struct SessionClosedFuture(BroadcastStream<()>);
+
+impl Future for SessionClosedFuture {
+	type Output = ();
+
+	fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+		match self.0.poll_next_unpin(cx) {
+			Poll::Pending => Poll::Pending,
+			// A message is only sent when
```
lexnv, line 185: nit: when the connection is closed

server/src/future.rs
```diff
@@ -157,3 +158,42 @@ impl Stream for IntervalStream {
 		}
 	}
 }
+
+#[derive(Debug, Clone)]
+pub(crate) struct SessionClose(tokio::sync::broadcast::Sender<()>);
+
+impl SessionClose {
+	pub(crate) fn close(self) {
+		let _ = self.0.send(());
```
lexnv, line 167: nit: Could this be implemented also on `Drop`? And if we already called `close()` we'd do nothing on drop, otherwise, we'll call `self.0.send()`?

lexnv (approved):

server/src/future.rs
niklasad1: I skipped it because I'm scared of that as the tower stuff requires Clone and I'm not sure whether something is dropped at some point.

lexnv: Makes sense, thanks!

server/src/future.rs
```diff
@@ -157,3 +158,42 @@ impl Stream for IntervalStream {
...
+				Poll::Ready(())
+			}
+		}
+	}
+}
+
+pub(crate) fn session_close() -> (SessionClose, SessionClosedFuture) {
+	// SessionClosedFuture is closed after one message has been recevied
+	// and max one message is handled then it's closed.
+	let (tx, rx) = tokio::sync::broadcast::channel(1);
```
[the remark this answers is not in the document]
niklasad1: Yepp, I think the implementation is quite complicated/clever as the Receiver is not clone.

I think it just clone the message(s) to other receivers which is probably quite nice to avoid having a separate state for each receiver

PR #1287, opened 2024-02-07
refactor(rpc_module): RpcModule::raw_json_request -> String

niklasad1:
This PR changes the RpcModule to support the new async API in the MethodResponse and RpcModule::raw_json_request now returns a String instead of MethodResponse as consequence of this change.


tadeohepperle (approved):
:+1:

lexnv (approved):

PR #1288, opened 2024-02-07
fix(server): return err on WS handshake err

niklasad1:
I noticed this when I was replying to an issue and `ws::connect` should only return when the WS handshake has been completed.


jsdw (approved):
I think this makes sense; only return when upgrade is done, whcih also allows us to return any error emitted by it!

PR #1293, opened 2024-02-12
fix(client): change to `AtomicU64` to `AtomicUsize`

### LINKED ISSUE #1290: AtomicU64/AtomicI64 is no longer supported by the *-espidf targets.
AtomicU64/AtomicI64 is no longer supported by the *-espidf targets. [Refer to the changelog for version 0.42.4 of esp-idf-hal](https://github.com/esp-rs/esp-idf-hal/blob/81675ccbf5d6868fedb3fa19a57d84ee5cfc06b0/CHANGELOG.md?plain=1#L28C1-L29C99)

Which is my case I am using a ESP32-C6 target = "riscv32imac-esp-espidf"

Error:

```text
error[E0432]: unresolved import `std::sync::atomic::AtomicU64`
  --> /home/ei02008/.cargo/registry/src/index.crates.io-6f17d22bba15001f/jsonrpsee-core-0.21.0/src/client/mod.rs:40:25
   |
40 | use std::sync::atomic::{AtomicU64, Ordering};
   |                         ^^^^^^^^^
   |                         |
   |                         no `AtomicU64` in `sync::atomic`
   |                         help: a similar name exists in the module: `AtomicU32`
```

Would be nice to have this fixed.

Cheers,

niklasad1:
Close #1290 

Some targets may not support AtomicU64.
This PR moves to `AtomicUsize` instead to support more targets.

### DIFF
--- a/core/src/client/mod.rs
+++ b/core/src/client/mod.rs
@@ -37,7 +37,7 @@ pub use error::Error;
 use std::fmt;
 use std::ops::Range;
 use std::pin::Pin;
-use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::Arc;
 use std::task;
 
@@ -396,7 +396,7 @@ pub struct RequestIdManager {
 	/// Max concurrent pending requests allowed.
 	max_concurrent_requests: usize,
 	/// Get the next request ID.
-	current_id: AtomicU64,
+	current_id: CurrentId,
 	/// Request ID type.
 	id_kind: IdKind,
 }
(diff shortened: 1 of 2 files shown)


core/src/client/mod.rs
```diff
@@ -487,6 +487,22 @@ impl IdKind {
...
+impl CurrentId {
+	fn new() -> Self {
+		CurrentId(AtomicUsize::new(0))
+	}
+
+	fn next(&self) -> u64 {
+		self.0
+			.fetch_add(1, Ordering::Relaxed)
+			.try_into()
+			.expect("usize -> u64 infallible, there are no 128-bit CPUs; qed")
```
niklasad1, line 502:
```suggestion
			.expect("usize -> u64 infallible, there are no CPUs > 64 bits; qed")
```

lexnv (approved):

jsdw (approved):

PR #1294, opened 2024-02-13
Fix part of proc-macro documentation not rendering correctly in IDE

Exidex:
Fix this 
image

I am not 100% sure if it is the IDE (RustRover) issue or the code. But this was an easy fix so here we go.


lexnv (approved):

niklasad1 (approved):

niklasad1:
Thanks :)

PR #1295, opened 2024-02-16
server: Extend methods with optional connection context

lexnv:


core/src/server/rpc_module.rs
```diff
@@ -500,7 +495,7 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 	/// use jsonrpsee_core::server::RpcModule;
 	///
 	/// let mut module = RpcModule::new(());
-	/// module.register_method("say_hello", |_params, _ctx| "lo").unwrap();
+	/// module.register_method("say_hello", |_params, _connection_ctx, _ctx| "lo").unwrap();
```
niklasad1, line 498: /cc @jsdw so basically this is the grumble

we expose now two context's here one `conn_ctx` which isn't exclusively connection stuff and then `ctx` which is completely controlled by the user.

however, the `ctx` is shared state per "rpc per server" and `conn_ctx` is per connection....

core/src/server/context.rs
```diff
@@ -0,0 +1,61 @@
...
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR
+// IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+// DEALINGS IN THE SOFTWARE.
+
+//! The context of a JSON-RPC server implementation.
+
+/// The context of a JSON-RPC server that is passed to methods and subscriptions
+/// that enabled the `with_context` attribute.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct ConnectionContext {
```
niklasad1, line 32: I would rather re-name this to `PerConnectionContext` or something to easily distinguish from the other context

proc-macros/src/render_server.rs
```diff
@@ -77,6 +85,15 @@ impl RpcDescription {
 			let mut sub_sig = sub.signature.clone();
 
 			sub_sig.sig.inputs.insert(1, subscription_sink);
+
+			if sub.with_context {
```
niklasad1, line 89: In the proc macros I'm not convinced that should expose ConnectionContext.

Just `connection_id: usize` should be sufficient and because everything internally has connection_id can't we just require everything to expose the connection ID?

proc-macros/src/rpc_macro.rs
```diff
@@ -48,12 +48,13 @@ pub struct RpcMethod {
...
 	pub aliases: Vec<String>,
+	pub with_context: bool,
 }
 
 impl RpcMethod {
 	pub fn from_item(attr: Attribute, mut method: syn::TraitItemMethod) -> syn::Result<Self> {
-		let [aliases, blocking, name, param_kind] =
-			AttributeMeta::parse(attr)?.retain(["aliases", "blocking", "name", "param_kind"])?;
+		let [aliases, blocking, name, param_kind, with_context] =
+			AttributeMeta::parse(attr)?.retain(["aliases", "blocking", "name", "param_kind", "with_context"])?;
```
niklasad1, line 57: so, `with_context` shouldn't be needed if we decide to expose connection_id  globally...

core/src/server/rpc_module.rs
```diff
@@ -56,22 +57,15 @@ use super::IntoResponse;
 /// implemented as a function pointer to a `Fn` function taking four arguments:
 /// the `id`, `params`, a channel the function uses to communicate the result (or error)
 /// back to `jsonrpsee`, and the connection ID (useful for the websocket transport).
-pub type SyncMethod = Arc<dyn Send + Sync + Fn(Id, Params, MaxResponseSize) -> MethodResponse>;
```
niklasad1, line 59: another option is to add new `MethodKind` here such as:

```rust
RawMethod = Arc<dyn Send + Sync + Fn(Id, Params, ConnectionContext) -> MethodResponse>
```

then let folks spawn futures and stuff them selves if they need and just provide a new attribute in the procs macros as you already did.

then register_blocking, async, etc are not needed.

niklasad1: then we don't have break existing APIs

core/src/server/rpc_module.rs
jsdw: Yeah ok I see!

My general grumble is that we should have some way to pass generic data from middleware or whatever into RPC calls. If we had that then we wouldn't need to expose eg the internal connection ID as a hardcoded thing, because the user could just create their own connection ID counter in middleware and pass that in.

Methods might want access to a bunch of other stuff from middleware, and I wouldn't want us to always have to work around it by trying to provide those things in a hrdcoded way :)

lexnv: For the first idea, I've been exploring with a custom type as field to the `PerConnectionContext`: `Arc<Mutex<anymap::Map<dyn anymap::any::Any + Send + Sync>>>`. This should allow users to insert in the map any type they want.

I might be missing something here, but the trick part with this comes from the fact that the `PerConnectionContext` is constructed as the last step in our `RpcServiceT` middleware. This means that users will only access it from methods, but cannot alter its state from another `RpcServiceT` middleware.
From my perspective, if the `PerConnectionContext` can only be accessed / altered from within methods, I think this brings the same usefulness as `connection: usize`.  Since `chainHead_*` methods could store internally a `Map<connection, Data>`.

https://github.com/paritytech/jsonrpsee/blob/736dd65259ba01bd31ffc8cae9d7b4d5b237d4d6/server/src/middleware/rpc/mod.rs#L42-L50

For this to be useful outside the method call; we'd probably need to extend the `call` function to include the (per connection) context : `fn call(&self, request: Request<'a>, context: PerConnectionContext) -> Self::Future;`. I do remember this was the original design for the middleware, but I think we had some difficulties implementing it?
And I think an extension of this is the second idea, however I don't know how much harder that is to implement.

core/src/server/rpc_module.rs
lexnv: That makes sense, I've implemented that in: https://github.com/paritytech/jsonrpsee/pull/1297/files#diff-dc533588250ea236f580d7fdaf87cdfce2ec80da031de9067b1088465a555998R64

One downside with doing a `RawMethod` like this is that we lose the ability to have:
- async methods (since this is a `SyncMethod` that receives an extra connection ID)
- blocking methods  (since the server-context is passed differently to a blocking method -- its need to be `Arc<Context>` instead of `&Context` so our callbacks are different)

This makes me wonder if I should also add the counter-part RawAsyncMethod and another `register_raw_blocking`. I think a single `RawMethod` would solve the connection ID exposure to complete the chainHead issue in substrate; but it might require us to extend it in the future.

At some point in the future , if we choose to implement `RawAsyncMethod`, it might be more beneficial to make a breaking change and add the `connection Id` to every method, for now I think we could avoid that

core/src/server/rpc_module.rs
niklasad1: Thus, it would be neat to have a have closure similar for the context/user defined data in the rpc API such that users can decide whether the state for rpc impl should be shared or per connection...

niklasad1:
Closing this replaced by https://github.com/paritytech/jsonrpsee/pull/1297

PR #1297, opened 2024-02-21
server: Register raw method with connection ID

lexnv:
This PR exposes the connection ID to raw methods.

To achieve this, a new `register_raw_method` is exposed on the RpcModule.
The `with-context` attribute is implemented in the proc-macros to use this method.
When the `with-context` attribute is used, all methods of a server are rendered with an extra connection Id parameter.
This is similar to the `PendingSubscriptionSink` that is injected into subscription methods.

### Examples

The example `server_with_context` implements a server with the context attribute enabled.
A function of the server returns the connection ID from its parameters.
Users on different connections will receive a different connection ID.

### Testing
- [x] Add proc-macro tests to capture the incompatibilities between `with-context`, sync methods and blocking methods
- [x] Add a unit-test to validate this behavior further
- [x] Build on the example to allow methods to be called from a single connection context

A simplified version of https://github.com/paritytech/jsonrpsee/pull/1295, that avoids a user-breaking change.


core/src/server/rpc_module.rs
```diff
@@ -598,6 +604,36 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
...
+	/// ```
+	pub fn register_raw_method<R, F>(
+		&mut self,
+		method_name: &'static str,
+		callback: F,
+	) -> Result<&mut MethodCallback, RegisterMethodError>
+	where
+		Context: Send + Sync + 'static,
+		R: IntoResponse + 'static,
+		F: Fn(Params, ConnectionId, &Context) -> R + Send + Sync + 'static,
```
niklasad1, line 625: let's change this to Arc<Context> such that it's Clone + Send + Sync it may be needed for tokio::spawn, tokio::spawn_blocking

```suggestion
		F: Fn(Params, ConnectionId, Arc<Context>) -> R + Send + Sync + 'static,
```

examples/examples/server_with_context.rs
```diff
@@ -0,0 +1,129 @@
...
+use jsonrpsee::proc_macros::rpc;
+use jsonrpsee::server::{PendingSubscriptionSink, Server, SubscriptionMessage};
+use jsonrpsee::types::ErrorObjectOwned;
+use jsonrpsee::ws_client::WsClientBuilder;
+use jsonrpsee::ConnectionId;
+
+type ExampleHash = [u8; 32];
+type ExampleStorageKey = Vec<u8>;
+
+#[rpc(server, client, with_context)]
```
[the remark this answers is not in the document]
niklasad1: For subscriptions we already got connect_id on the PendingSubscriptionSink and SubscriptionSink 

So should be sufficient I guess

niklasad1 (commented):
This should work and overall I like it but some small comments to address

examples/examples/server_with_context.rs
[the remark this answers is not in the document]
lexnv: I generally like this idea. We now have control over individual methods, instead of having an all-or-nothing server.
Can't remember why I switched the implementation from https://github.com/paritytech/jsonrpsee/pull/1295, I guess I was a bit lazy with this one :D

For subscriptions there is nothing to extend, as the connection Id is already provided yep :D

core/src/server/rpc_module.rs
lexnv: That makes sense. I modified this to receive an `Arc<Context>` and added the `Clone` trait bound. I believe we'll have everything we'd need to do a `spawn_blocking`, thanks!

niklasad1 (approved):
LGTM, nice work

I wonder if we should provide a type with connection_id() to avoid breaking it if we decide to add some other field but I think that's unlikely though.

tests/tests/helpers.rs
```diff
@@ -137,7 +137,7 @@ pub async fn server() -> SocketAddr {
 	let mut module = RpcModule::new(());
 	module.register_method("say_hello", |_, _| "hello").unwrap();
 
-	module.register_raw_method("raw_method", |_, connection_id, _| connection_id).unwrap();
+	module.register_raw_method("raw_method", |_, connection_id, _| async { connection_id }).unwrap();
```
niklasad1, line 140:
```suggestion
	module.register_raw_method("raw_method", |_, connection_id, _| async move { connection_id }).unwrap();
```

jsdw (approved):
Looks good to me; just a couple of thoughts:

So, basically, `register_raw_method` is the same as `register_async_method` but with a connection ID. I'd wonder about calling it something more similar to that, eg: `register_async_method_with_details` (especially if the below was done).

In general I guess being able to pass `Extensions` between middlewares and method would remove the need for this again? But happy to just merge this and cross that bridge when we get there anyways! 

> I wonder if we should provide a type with connection_id() to avoid breaking it if we decide to add some other field but I think that's unlikely though.

I wondered too about wrapping the ConnectionId in a type, just to give more flexibility eg:

```rust
pub struct ConnectionDetails {
    id: ConnectionId
}

impl ConnectionDetails {
    pub fn id(&self) -> &ConnectionId
}
```

core/src/server/rpc_module.rs
```diff
@@ -60,6 +60,9 @@ pub type SyncMethod = Arc<dyn Send + Sync + Fn(Id, Params, MaxResponseSize) -> M
 /// Similar to [`SyncMethod`], but represents an asynchronous handler.
 pub type AsyncMethod<'a> =
 	Arc<dyn Send + Sync + Fn(Id<'a>, Params<'a>, ConnectionId, MaxResponseSize) -> BoxFuture<'a, MethodResponse>>;
+/// Similar to [`SyncMethod`], but represents an asynchronous handler with connection details.
```
niklasad1, line 63:
```suggestion
/// Similar to [`AsyncMethod`], but represents an asynchronous handler with connection details.
```

core/src/server/rpc_module.rs
```diff
@@ -80,6 +83,45 @@ pub type MaxResponseSize = usize;
...
+
+	/// Get the connection ID.
+	pub fn id(&self) -> ConnectionId {
+		self.id
+	}
+}
+
+#[derive(Debug, Clone)]
+#[allow(missing_copy_implementations)]
+/// The connection details exposed to the server methods.
```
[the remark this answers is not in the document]
lexnv: The `RpcService` should be the last "layer" before the user-callback is called.

We construct the `ConnectionDetails` in this layer:

https://github.com/paritytech/jsonrpsee/blob/87585ad948dcdbd7a31773ba8305f38cc9c21fbe/server/src/middleware/rpc/layer/rpc_service.rs#L102-L104

I think we could just as easily provide a constructor for this: 

```rust
impl ConnectionDetails {
  pub fn new(id: ConnectionId) ..
}
```

The `new` method needs to be public, because the [ConnectionDetails](https://github.com/paritytech/jsonrpsee/blob/87585ad948dcdbd7a31773ba8305f38cc9c21fbe/core/src/server/rpc_module.rs#L89-L91) is declared in the `core` crate. And the `ConnectionDetails` needs to be instantiated in the `server` crate as well.

We'd have to eventually change this method to `new(id: ConnectionId, extensions: Extenstions)` which will be a breaking change (although we should be the only ones building this object).

TLDR; since we'll do a breaking change either way, I think a `new` method should be fine for now! 🙏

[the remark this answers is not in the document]
niklasad1: ok, I see makes sense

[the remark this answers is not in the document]
jsdw: Maybe have it be `#[doc(hidden)]` and comment that it's not part of the public APi and can change without notice?

core/src/server/rpc_module.rs
```diff
@@ -607,27 +655,27 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 	/// use jsonrpsee_core::server::RpcModule;
 	///
 	/// let mut module = RpcModule::new(());
-	/// module.register_raw_method("say_hello", |_params, _connection_id, _ctx| async { "lo" }).unwrap();
+	/// module.register_async_with_details("say_hello", |_params, _connection_details, _ctx| async { "lo" }).unwrap();
 	/// ```
-	pub fn register_raw_method<R, Fun, Fut>(
+	pub fn register_async_with_details<R, Fun, Fut>(
```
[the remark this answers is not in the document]
jsdw: How about `register_async_method_eith_details`? since we are provigin `ConnectionDetails` to it, so the word details lines up well imo :)

niklasad1:
We may mark this feature/API as `#[doc(hidden)]` as we will break anyway in the next release?

What do you reckon?

jsdw:
> We may mark this entire feature/API as #[doc(hidden)] as we will break anyway in the next release?

Works for me too, since I geuss we only need it internally right now and we'll break it again soon! Could also be feature flagged to hdie it at compile time and then only enable in substrate?

core/src/server/rpc_module.rs
```diff
@@ -131,6 +159,8 @@ pub enum MethodCallback {
 	Sync(SyncMethod),
 	/// Asynchronous method handler.
 	Async(AsyncMethod<'static>),
+	/// Asynchronous method handler with details.
+	AsyncWithDetails(AsyncMethodWithDetails<'static>),
```
niklasad1, line 163: does it work here?

```suggestion
    #[doc(hidden)]
	AsyncWithDetails(AsyncMethodWithDetails<'static>),
```

examples/examples/server_with_raw_method.rs
niklasad1, line 1: rename perhaps?

lexnv: Have changed it to `server_wtih_connection_details`, thanks!

core/src/server/rpc_module.rs
```diff
@@ -60,6 +60,12 @@ pub type SyncMethod = Arc<dyn Send + Sync + Fn(Id, Params, MaxResponseSize) -> M
 /// Similar to [`SyncMethod`], but represents an asynchronous handler.
 pub type AsyncMethod<'a> =
 	Arc<dyn Send + Sync + Fn(Id<'a>, Params<'a>, ConnectionId, MaxResponseSize) -> BoxFuture<'a, MethodResponse>>;
+
+/// Similar to [`AsyncMethod`], but represents an asynchronous handler with connection details.
+#[doc(hidden)]
+pub type AsyncMethodWithDetails<'a> =
+	Arc<dyn Send + Sync + Fn(Id<'a>, Params<'a>, ConnectionDetails, MaxResponseSize) -> BoxFuture<'a, MethodResponse>>;
+/// Similar to [`SyncMethod`], but represents a raw handler that has access to the connection Id.
```
niklasad1, line 68:
```suggestion
```

PR #1306, opened 2024-02-29
feat: add `Extensions` to Request/MethodResponse

### LINKED ISSUE #1305: add `type map http::Extensions` to Request and MethodResponse types
After trying to "share state" between different layers of the middleware in jsonrpsee, which is hard to
do properly.

One possibility is to inject a shared state such as `Arc<Mutex<HashSet<(RequestId, MethodName, Params)>>>` but it's not a good solution because the RequestId isn't necessarily unique, and it's possible that an identical RPC call may overwrite a pending call or something similar.

For instance, my own use case is that I would want to know whether a call was rate-limited or not, which was controlled by another middleware layer.

A good solution is to copy what hyper/http has for the Request/Response in the [http::Extensions API](https://docs.rs/http/latest/http/struct.Extensions.html) so that one could just inject custom data in there, as in the following example:

```rust
impl<'a, S> RpcServiceT<'a> for Metrics<S>
where
    S: Send + RpcServiceT<'a>,
{
    type Future = BoxFuture<'a, MethodResponse>;

    fn call(&self, req: Request<'a>) -> Self::Future {
        if self.is_rate_limited() {
            req.extensions_mut().insert("rate_limited", true);
        }

        let rp = self.service.call(req);

        if rp.extensions.get("rate_limited").is_some() {
            update_metrics(true);
        } else {
            update_metrics(false);
        }
    }
}

impl<'a, S> RpcServiceT<'a> for RateLimit<S>
where
    S: Send + RpcServiceT<'a>,
{
    type Future = BoxFuture<'a, MethodResponse>;

    fn call(&self, req: Request<'a>) -> Self::Future {
        if self.guard.check() {
            req.extensions_mut().insert("rate_limited", true);
            wait_until_ready();
        }

        self.service.call(req)
    }
}
```

niklasad1:
Fixes #1305 

In summary this PR adds `http::Extensions` to Request and Responses this may be used to share state between middlewares and similar.

Most of the changes are chore but I had to add this to all RPC handlers and the proc macro API.
Thus, the most important things to review is the changes to Request and MethodResponse.

### DIFF
--- a/tests/tests/helpers.rs
+++ b/tests/tests/helpers.rs
@@ -28,22 +28,29 @@
 #![allow(dead_code)]
 
 use std::net::SocketAddr;
+use std::sync::atomic::AtomicU32;
+use std::sync::Arc;
 use std::time::Duration;
 
 use fast_socks5::client::Socks5Stream;
 use fast_socks5::server;
+use futures::future::{self, Either};
 use futures::{SinkExt, Stream, StreamExt};
+use hyper_util::rt::{TokioExecutor, TokioIo};
 use jsonrpsee::server::middleware::http::ProxyGetRequestLayer;
 
+use jsonrpsee::server::middleware::rpc::RpcServiceT;
 use jsonrpsee::server::{
-	PendingSubscriptionSink, RpcModule, Server, ServerBuilder, ServerHandle, SubscriptionMessage, TrySendError,
+	stop_channel, PendingSubscriptionSink, RpcModule, RpcServiceBuilder, Server, ServerBuilder, ServerHandle,
+	SubscriptionMessage, TrySendError,
 };
 use jsonrpsee::types::{ErrorObject, ErrorObjectOwned};
-use jsonrpsee::SubscriptionCloseResponse;
+use jsonrpsee::{Methods, SubscriptionCloseResponse};
 use serde::Serialize;
 use tokio::net::TcpStream;
 use tokio::time::interval;
 use tokio_stream::wrappers::IntervalStream;
+use tower::Service;
 use tower_http::cors::CorsLayer;
 
 pub async fn server_with_subscription_and_handle() -> (SocketAddr, ServerHandle) {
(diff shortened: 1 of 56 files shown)


types/src/request.rs
```diff
@@ -34,11 +34,12 @@ use crate::{
 	Params,
 };
 use beef::Cow;
+use http::Extensions;
 use serde::{Deserialize, Serialize};
 use serde_json::value::RawValue;
 
 /// JSON-RPC request object as defined in the [spec](https://www.jsonrpc.org/specification#request-object).
-#[derive(Serialize, Deserialize, Debug, Clone)]
+#[derive(Serialize, Deserialize, Debug)]
```
niklasad1, line 42: breaking change, Extensions is !Clone in http v0.2

lexnv (approved):
Looks good to me!

Once we merge this, I think we could replace the `connection_id` we expose to raw methods to `Params::Extensions`. And then guarantee by our middleware that we add a unique type for connection_id. 

And we'll have a way to easily pass data not only between middlewares, but also middleware -> methods

But that could happen later :D

jsdw:
Do we have an example of ths Extensions stuff being used? Would be great to see it in practice :)

tests/tests/helpers.rs
```diff
@@ -133,19 +146,30 @@ pub async fn server_with_subscription() -> SocketAddr {
...
+
+		fn call(&self, mut request: jsonrpsee::types::Request<'a>) -> Self::Future {
+			request.extensions_mut().insert(self.connection_id);
+			self.inner.call(request)
+		}
+	}
+
+	let mut module = RpcModule::new(());
+	module.register_method("say_hello", |_, _, _| "hello").unwrap();
+	module.register_method("raw_method", |_, _, ext| *ext.get::<u32>().unwrap()).unwrap();
```
jsdw, line 169: Nit: I wonder whether we should rename this to eg "get_connection_id" to make it clearer what it's doing/testing?

tests/tests/helpers.rs
```diff
@@ -159,14 +183,83 @@ pub async fn server() -> SocketAddr {
...
 
-	let addr = server.local_addr().unwrap();
+	let listener = tokio::net::TcpListener::bind("127.0.0.1:0").await.unwrap();
+	let addr = listener.local_addr().unwrap();
+	let (stop_hdl, server_hdl) = stop_channel();
+	let methods: Methods = module.into();
 
-	let server_handle = server.start(module);
+	let methods2 = methods.clone();
+	tokio::spawn(async move {
```
jsdw, line 194: Ah I geuss this is all added so that we can inject connection details!

I wonder whether there's something wecan do in the future to simplify this "low level" initialising

jsdw (approved):
This looks great! IMO it would benefit from an example showing extensions being used to eg share details between middlewares to the request method or whatever :)

Getting details in I guess requires the "low level" interface as in the test stuff, and I wonder if we can simplify injecting data on that side of things!

niklasad1:
> Do we have an example of ths Extensions stuff being used? Would be great to see it in practice and offhand I can't see one :)

Yes, https://github.com/paritytech/jsonrpsee/pull/1306/files#diff-aa58ad08343facdd5487fab413ac599701293922cbcd160d00c7022a7577fc4cR50-R60

tests/tests/helpers.rs
niklasad1: We could probably inject the connection id by default but it's quite weird to explain/document it so I didn't do it.

What does @lexnv reckon? :)

jsdw:
> > Do we have an example of ths Extensions stuff being used? Would be great to see it in practice and offhand I can't see one :)
> 
> Yes, https://github.com/paritytech/jsonrpsee/pull/1306/files#diff-aa58ad08343facdd5487fab413ac599701293922cbcd160d00c7022a7577fc4cR50-R60

Ah awesome; I missed that!

tests/tests/helpers.rs
lexnv: I think this is fine, substrate could then inject the connection ID in a similar way

lexnv (approved):
LGTM!

PR #1307, opened 2024-02-29
Do not drop a subscription when buffer is full: provide the subscriber an option to drain the buffer.

RGafiyatullin:


niklasad1:
hey @RGafiyatullin thanks for your PR it looks good but I have already a PR that does something similar already up but it actually replaces the oldest message if the stream couldn't keep up with the server and provides a slighter nicer API, https://github.com/paritytech/jsonrpsee/pull/1255.

So please hold off on this for a bit

RGafiyatullin:
Oh, indeed you have. Sorry :)

Issue #1308, opened 2024-02-29
[client]: individual configaration per subscription instead of globally


niklasad1:
Currently, the jsonrpsee client provides the same max buffer size for every subscription but once we have moved `async-broadcast/tokio-broadcast` it would be nice to make configuration per subscription basis.

For instance `subscriptionA` may be okay to drop the oldest message but `subscriptionB` is not allowed to drop messages.

Thus, I propose a API like:

```rust
enum StrategyWhenSubscriptionSlow {
   DropOldest,
   DropLatest,
   Shutdown,
   Drain
}


pub struct SubscriptionConfig {
   max_buf_len: usize,
   strategy: StrategyWhenSubscriptionSlow
}

async fn subscribe(sub_method: &str, params: Params,  unsub_method: &str, cfg: SubscriptionConfig) -> Subscription
```

niklasad1:
Closing this because we decided to take another solution here.
Because the Subscription implements stream it should be quite easy to extend such as the following example:

```rust
fn drop_oldest_when_lagging<T: Clone + DeserializeOwned + Send + Sync + 'static>(
	mut sub: Subscription<T>,
	buffer_size: usize,
) -> impl Stream<Item = T> {
	let (mut tx, rx) = async_broadcast::broadcast(buffer_size);

	tx.set_overflow(true);

	tokio::spawn(async move {
		// poll sub, sending msgs to our chan which throws stuff away
		while let Some(n) = sub.next().await {
			let msg = match n {
				Ok(msg) => msg,
				Err(e) => {
					tracing::error!("Failed to recv subscription message: {e}");
					continue;
				}
			};

			tx.try_broadcast(msg).expect("Infallible in overflowing mode; qed");
		}
	});
	rx
}
```

PR #1310, opened 2024-03-01
WIP: refactor(client): individual subscription and different strategies

niklasad1:
Follow-up/another solution to #1255 
Essentially the suggestion is to make each subscription decide how to deal with "slow receiver":

```rust
enum StrategyWhenSubscriptionSlow {
   DropOldest,
   DropLatest,
   Close,
   Drain
}

pub struct SubscriptionConfig {
   buf_capacity: usize,
   strategy: StrategyWhenSubscriptionSlow
}

async fn subscribe(sub_method: &str, params: Params,  unsub_method: &str, cfg: SubscriptionConfig) -> Subscription
```

As I tried to explain in https://github.com/paritytech/jsonrpsee/issues/1308


jsdw:
Overall this looks good!

My maybe dumb thought is; is there a sensible way to add this same behaviour outside of jsonrpsee itself (or is it even worth doing)? eg:

```rust
let sub: Subscription<i32> =
    client1.subscribe("subscribe_hello", rpc_params![], "unsubscribe_hello").await?;

// drop oldest messages from subscription:
let sub = drop_oldest(sub, buffer_size);

// maybe this is defined something like:
fn drop_oldest<T>(mut sub: Stream<T>, buffer_size: usize) -> impl Stream<T> {
    let buf = Vec::with_capacity(usize);
    let (tx, rx) = async_broadcast::channel(); 
   tokio::spawn(async move {
        while let Some(msg) = sub.next().await { tx.send(msg) }
    })
    rx
}
```

The thinking is that, if subscriptions coming back from jsonrpsee client honour backpressure then we can bolt things in front of them rather than add logic inside them and then we/users can do whatever arbitrary behaviour they like :)

I realise this might all be impossible though at present because maybe we do a tokio::spawn internall in the client and so subscriptions don't get backpressure or whatever!

core/src/client/mod.rs
```diff
@@ -592,9 +682,233 @@ impl<'a, R> IntoIterator for BatchResponse<'a, R> {
...
+	/// The subscription may suffer from the slow-receiver problem
+	pub fn set_lagging_strategy(mut self, strategy: LaggingStrategy) -> Self {
+		self.strategy = strategy;
+		self
+	}
+}
+
+/// Strategy to use when a message could not be handled
+#[derive(Debug, Copy, Clone)]
+pub enum LaggingStrategy {
```
lexnv, line 866: dq: How easy would it be for a user to define their own custom strategy?

niklasad1: it's not possible because we control stream/recv impl in that sense what James suggested is probably nicer and more generic...

but this should cover 99% of the use cases I reckon

niklasad1:
> The thinking is that, if subscriptions coming back from jsonrpsee client honour backpressure then we can bolt things in front of them rather than add logic inside them and then we/users can do whatever arbitrary behaviour they like :)

I realise this might all be impossible though at present because maybe we do a tokio::spawn internall in the client and so subscriptions don't get backpressure or whatever!

So indeed we could change that to "await" [here](https://github.com/paritytech/jsonrpsee/blob/master/core/src/client/async_client/helpers.rs#L109-#L121) but it may cause the client to handle other messages slower but should be fine I guess

niklasad1:
Superseeded by #1316

PR #1316, opened 2024-03-06
refactor(subscription): delegate handling to user and do not close when lagging.

niklasad1:
This PR changes that subscription is not closed if couldn't keep up with the server. 

Instead entire connection is "backpressured" and until the subscription is polled by the user no other messages are processed until doing.

The benefit is that jsonrpsee doesn't have to provide that logic and is less opinionated. Then it's easy to extend to proper behavior, and the internal buffer for subscription is removed or replaced with a capacity of 1.

The downside is that if something is polled in the wrong order it may "stall the connection", thus if one has to several subscriptions and poll them in the wrong order the connection will never make progress.

Example to drop the oldest message.

```rust
let sub: Subscription<i32> =
    client1.subscribe("subscribe_hello", rpc_params![], "unsubscribe_hello").await?;

// drop oldest messages from subscription:
let sub = drop_oldest(sub, buffer_size);

fn drop_oldest<T>(mut sub: Stream<T>, buffer_size: usize) -> impl Stream<T> {
    let buf = Vec::with_capacity(usize);
    let (tx, rx) = async_broadcast::channel(); 
    tokio::spawn(async move {
        while let Some(msg) = sub.next().await { tx.send(msg) }
    })
    rx
}
```

### DIFF
--- a/client/wasm-client/src/lib.rs
+++ b/client/wasm-client/src/lib.rs
@@ -62,7 +62,6 @@ use jsonrpsee_core::client::{ClientBuilder, Error, IdKind};
 pub struct WasmClientBuilder {
 	id_kind: IdKind,
 	max_concurrent_requests: usize,
-	max_buffer_capacity_per_subscription: usize,
 	max_log_length: u32,
 	request_timeout: Duration,
 }
@@ -73,7 +72,6 @@ impl Default for WasmClientBuilder {
 			id_kind: IdKind::Number,
 			max_log_length: 4096,
 			max_concurrent_requests: 256,
-			max_buffer_capacity_per_subscription: 1024,
 			request_timeout: Duration::from_secs(60),
 		}
 	}
@@ -97,12 +95,6 @@ impl WasmClientBuilder {
 		self
 	}
 
-	/// See documentation [`ClientBuilder::max_buffer_capacity_per_subscription`] (default is 1024).
-	pub fn max_buffer_capacity_per_subscription(mut self, max: usize) -> Self {
-		self.max_buffer_capacity_per_subscription = max;
-		self
-	}
-
 	/// See documentation for [`ClientBuilder::id_format`] (default is Number).
 	pub fn id_format(mut self, kind: IdKind) -> Self {
 		self.id_kind = kind;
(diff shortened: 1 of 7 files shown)


jsdw:
I love the idea of course; I thought it would be hard actually getting backpressuring all the way to the client side so wasn't expecting it to be possible :D 

Having not properly looked at the code yet, my question is: if one subscription doesn't poll fast enough to keep up, does that impact other subscriptions too (ie we end up blocking the entire connection) or they continue working normally?

niklasad1:
> Having not properly looked at the code yet, my question is: if one subscription doesn't poll fast enough to keep up, does that impact other subscriptions too (ie we end up blocking the entire connection) or they continue working normally?

We will block/backpressure the entire connection, I tried to write about in the PR description.
Thus, it could be a footgun.

I noticed this myself in our tests that we had two concurrent subscriptions and then one needs to use select in these overlaps such if one awaits on the "wrong" subscription it will be blocked/deadlocked :(

jsdw:
Ok so I was pondering this a little and one realisation I had is that backpressure on individual subscriptions isn't necessary to give the user control over how to handle lagging or whatever. ie to explain in the form of comments added to the above:

```rust
// Imagine this subscription bufer was being filled up even if it wasn't being polled 
// (ie because the background task is pushing to its buffer): 
let sub: Subscription<i32> =
    client1.subscribe("subscribe_hello", rpc_params![], "unsubscribe_hello").await?;

// The implementation of `drop_oldest` doesn't care; it too is just running a task that 
// takes messages off this internal subscription buffer as fast as it can, and drops old
// ones if its own buffer isn't filled in time.
let sub = drop_oldest(sub, buffer_size);
``` 

So even without backpressure on each subscription, a user can write some code in front of a subscription _today_ to drop messages or close the subscription if it isn't polled quickly enough etc.

Backpressure is useful because it tells the backend that we aren't keeping up though, and lets the server aggregate messages or close subs or whatever to try and reduce the amount of load it puts on the client.

With those things in mind I can think of two approaches offhand with different tradeoffs:

1. Leave things roughly as they are today, but consider optionally exposing the "Driver" task so that, while the default would still be to run it in a `tokio::spawn`, the user can take it and poll it themselves if they want. They can then apply backpressure by polling it more slowly. For example they could do this to run things "sequentially":

```rust
let (client, driver) = make_jsonrpsee_client_exposing_driver();

let sub1 = client.subscribe("foo", rpc_params![], "unsubscribe_foo").await?;
let sub2 = client.subscribe("bar", rpc_params![], "unsubscribe_bar").await?;
let sub3 = drop_oldest(client.subscribe("wibble", rpc_params![], "unsubscribe_wibble").await?);

select! {
    msg1 = sub1.next() => { /* handle msg from sub1 */ },
    msg2 = sub2.next() => { /* handle msg from sub2 */ },
    msg3 = sub3.next() => { /* handle msg from sub3 */ },
   () = driver => { /* poll driver. If handling sub messages is slow then this is polled less, applying backpressure */ }
}
```

2. Do something like this PR: every subscription etc needs polling to allow progress to be made, but then if any of them aren't polled quickly enough then the slight foot gun that everything freezes (and then as you said, maybe a timeout on each thing is a good idea to work around this). 
    - No explicit driver thing needed but comes with this footgun.
    - User can still add eg `drop_oldest` or whatever in front, which would likely spawn a task and ensure the sub is polled and then whatever strategy is applies.

I think my conclusion is that whatever we do, we can add things like `drop_oldest` in front of subscriptions rather than bake any logc into the core of `jsonrpsee`, which is nice. Then it's just a case of, offer an explicit driver to allow the user to control backpressure, or try to control backpressure more implicitly by how fast all of the subscriptions are polled.

I'm leaning towards the first one I guess now, because of the slightly scary footgun :D What do you reckon?!

niklasad1:
yeah for now it's probably better to go with 1), if we take similar approach as https://github.com/paritytech/jsonrpsee/pull/1307 then I'm happy with it.

i.e, a way to distinguish between slow and closed.

PR #1320, opened 2024-03-12
client: add `Subscription::close_reason`

### LINKED ISSUE #1291: [client]: simplify subscription API

niklasad1:
This PR changes adds a new API to the subscription know how the subscription was closed.

It still drops the subscription if it's lagging but the subscription should easily-extendable to
implement specific handling for lagging, see the example below

### Example of custom subscription logic

```rust
fn drop_oldest_when_lagging<T: Clone + DeserializeOwned + Send + Sync + 'static>(
	mut sub: Subscription<T>,
	buffer_size: usize,
) -> impl Stream<Item = T> {
	let (mut tx, rx) = async_broadcast::broadcast(buffer_size);

	tx.set_overflow(true);

	tokio::spawn(async move {
		while let Some(n) = sub.next().await {
			let msg = match n {
				Ok(msg) => msg,
				Err(e) => {
					tracing::error!("Failed to recv subscription message: {e}");
					continue;
				}
			};

			tx.try_broadcast(msg).expect("Infallible in overflowing mode; qed");
		}
	});
	rx
}
```

Resolves https://github.com/paritytech/jsonrpsee/issues/1291

### DIFF
--- a/core/src/client/mod.rs
+++ b/core/src/client/mod.rs
@@ -247,7 +247,7 @@ pub enum SubscriptionError {
 	/// The subscription lagged too far behind i.e, could not keep up with the server.
 	#[error("The subscription was too slow and at least one message was lost")]
 	TooSlow,
-	/// The subscription was closed because the connection was closed.
+	/// The subscription was closed.
 	#[error("The subscription was closed")]
 	Closed,
 	/// The subscription notification parsing failed.
--- a/core/src/client/mod.rs
+++ b/core/src/client/mod.rs
@@ -33,12 +33,13 @@ cfg_async_client! {
 
 pub mod error;
 pub use error::Error;
+use tokio::sync::mpsc::error::TrySendError;
 
 use std::fmt;
 use std::ops::Range;
 use std::pin::Pin;
 use std::sync::atomic::{AtomicUsize, Ordering};
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 use std::task::{self, Poll};
 
 use crate::params::BatchRequestBuilder;
@@ -50,6 +51,31 @@ use jsonrpsee_types::{ErrorObject, Id, SubscriptionId};
 use serde::de::DeserializeOwned;
 use serde_json::Value as JsonValue;
 use tokio::sync::{mpsc, oneshot};
+
+/// Shared state whether a subscription has lagged or not.
+#[derive(Debug, Clone)]
+pub(crate) struct SubscriptionLagged(Arc<Mutex<bool>>);
+
+impl SubscriptionLagged {
+	/// Create a new [`SubscriptionLagged`].
+	pub(crate) fn new() -> Self {
+		Self(Arc::new(Mutex::new(false)))
+	}
+
+	/// A message has been missed.
+	pub(crate) fn set_missed(&self) {
+		*self.0.lock().unwrap() = true;
+	}
+
+	/// Check whether the subscription has missed a message.
+	pub(crate) fn has_missed_message(&self) -> bool {
+		let mut lock = self.0.lock().unwrap();
+
+		let res = *lock;
+		*lock = false;
+		res
+	}
+}
 
 // Re-exports for the `rpc_params` macro.
 #[doc(hidden)]
@@ -265,9 +291,7 @@ pub struct Subscription<Notif> {
 	/// Channel to send requests to the background task.
 	to_back: mpsc::Sender<FrontToBack>,
 	/// Channel from which we receive notifications from the server, as encoded `JsonValue`s.
-	rx: mpsc::Receiver<JsonValue>,
-	/// Whether the `rx` lagged behind i.e, could not keep up with server.
-	rx_lagged: mpsc::Sender<JsonValue>,
+	rx: SubscriptionReceiver,
 	/// Callback kind.
 	kind: Option<SubscriptionKind>,
 	/// Marker in order to pin the `Notif` parameter.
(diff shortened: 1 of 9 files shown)


examples/examples/client_subscription_drop_oldest_item.rs
```diff
@@ -0,0 +1,118 @@
...
+	let sub: Subscription<i32> = client.subscribe("subscribe_hello", rpc_params![], "unsubscribe_hello").await?;
+
+	// drop oldest messages from subscription:
+	let mut sub = drop_oldest_when_lagging(sub, 10);
+
+	// Simulate that polling takes a long time.
+	tokio::time::sleep(Duration::from_secs(1)).await;
+
+	// The subscription starts from zero but you can
+	// noticed that many items have been replaced
```
niklasad1, line 58:
```suggestion
	// notice that many items have been replaced
```

core/src/client/mod.rs
```diff
@@ -51,6 +52,31 @@ use serde::de::DeserializeOwned;
 use serde_json::Value as JsonValue;
 use tokio::sync::{mpsc, oneshot};
 
+/// Shared state whether a subscription has lagged or not.
+#[derive(Debug, Clone)]
+pub(crate) struct SubscriptionLagged(Arc<Mutex<bool>>);
```
jsdw, line 57: Any reason not to use `Arc<AtomicBool>` instead? :)

jsdw: Actually I endedu pthinking more about this and the mutex does offer stronger guarantees I think re ordering! It's probably the safe option regardless without thinking very carefully about it!

core/src/client/mod.rs
```diff
@@ -592,6 +655,61 @@ impl<'a, R> IntoIterator for BatchResponse<'a, R> {
...
+}
+
+pub(crate) struct TooSlow;
+
+impl Stream for SubscriptionReceiver {
+	type Item = Result<JsonValue, TooSlow>;
+
+	fn poll_next(mut self: Pin<&mut Self>, cx: &mut task::Context<'_>) -> task::Poll<Option<Self::Item>> {
+		if self.lagged.has_missed_message() {
+			return Poll::Ready(Some(Err(TooSlow)));
```
jsdw, line 698: I wonder whether a potential issue here is that this "missed messages" thing being set to true doesn't wake up the task. It might be irrelevant though since, if messages are being missed, it is because this thing isn't being polled quickly enough already, so I'm inclined to think it's not an issue.

core/src/client/mod.rs
```diff
@@ -592,6 +655,61 @@ impl<'a, R> IntoIterator for BatchResponse<'a, R> {
...
+				Err(TrySubscriptionSendError::TooSlow(m))
+			}
+		}
+	}
+}
+
+#[derive(Debug)]
+pub(crate) struct SubscriptionReceiver {
+	inner: mpsc::Receiver<JsonValue>,
+	lagged: SubscriptionLagged,
```
jsdw, line 688: It's a shame we need a mutex here, because that's potentially polled a lot, so I wonder what perf impact it has.

Another way to implement this could be to have a channel with CAPACITY + 1, and then if we get up to CAPACITY then push the "Err(MissedMessages)" error to the final vacant slot. Then all subsequent messages are thrown away until the channel starts draining and every time it gets up to CAPACITY again we push another err on. But this could lead to silly things like a whole channel full of such errs if the client is too slow.

I still wonder whether we shouldn't just disconnect them when they are too slow. If they want to avoid it then they use something like in your example here https://github.com/paritytech/jsonrpsee/pull/1320/files#diff-0255f26eec4d19e28d74dcf6bd4ace6bfa5e4e45a1b1e9a63594548150227b60R68. That keeps the impl simple and avoids a mutex at least :)

core/src/client/mod.rs
niklasad1: `has_missed_message` will flip the flag and two reads will end up reading the underlying stream but it's possible that the stream will never be read if the sender/receiver at interleaved:

1. send (full)
2. read (too slow)
3. send (full)
4. read (too slow)
...

and so on but I find it unlikely to be an issue because it should be very unlikely...

tests/tests/integration_tests.rs
```diff
@@ -112,14 +113,15 @@ async fn ws_unsubscription_works() {
...
-	sub.unsubscribe().await.unwrap();
+	sub.unsubscribe().with_default_timeout().await.unwrap().unwrap();
 
 	let res = rx.next().with_default_timeout().await.expect("Test must complete in 1 min");
 	// When the subscription is closed a message is sent out on this channel.
 	assert!(res.is_some());
 }
 
 #[tokio::test]
+#[ignore]
```
niklasad1, line 124:
```suggestion
```

examples/examples/client_subscription_drop_oldest_item.rs
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	Ok(())
+}
+
+fn drop_oldest_when_lagging<T: Clone + DeserializeOwned + Send + Sync + 'static>(
+	mut sub: Subscription<T>,
+	buffer_size: usize,
+) -> impl Stream<Item = Result<T, BroadcastStreamRecvError>> {
+	let (tx, rx) = tokio::sync::broadcast::channel(buffer_size);
```
jsdw, line 80: Ooh I didn't realise that this channel handled lagging messages, TIL :)

jsdw (approved):
Cool! Looks good to me :)

core/src/client/mod.rs
```diff
@@ -262,64 +302,85 @@ impl<Notif> Subscription<Notif> {
...
 	}
+
+	/// The reason why the subscription was closed.
+	///
+	/// Returns Some(reason) is the subscription was closed otherwise
+	/// None is returned.
+	pub fn close_reason(&self) -> Option<SubscriptionCloseReason> {
+		let lagged = self.rx.lagged.has_lagged();
+
+		// The `is_closed` is only set if the subscription has been polled
```
niklasad1, line 317:
```suggestion
		// `is_closed` is only set if the subscription has been polled
```

lexnv (approved):
LGTM! Nice one 👍 

With this PR we should have a bit more info about what happens when the subscription stream ends, which makes our life easier when debugging connectivity issues 🙏

PR #1324, opened 2024-03-15
feat: RpcModule::from_arc

aatifsyed:
I want to share state between my `RpcModule` and another service.
Because `RpcModule` only `Deref`s to `Context`, I have to have a _second_ `Arc`, and my handlers have to accept `Arc<Arc<Context>>`.

This would allow me to clone the inner context, obviating the need for the double indirection.

Alternative approaches:
- impl `Deref<Target = Arc<Context>>`. I think this is backwards-compatible.
- `RpcModule::new(_: impl Into<Arc<Context>>)`, which is backwards-compatible.
  I could clone the arc before passing it to the ctor, where it would stay without a new `Arc`

### DIFF
--- a/core/src/server/rpc_module.rs
+++ b/core/src/server/rpc_module.rs
@@ -507,12 +507,14 @@ pub struct RpcModule<Context> {
 impl<Context> RpcModule<Context> {
 	/// Create a new module with a given shared `Context`.
 	pub fn new(ctx: Context) -> Self {
-		Self { ctx: Arc::new(ctx), methods: Default::default() }
+		Self::from_arc(Arc::new(ctx))
 	}
 
-	/// Get a reference to the shared `Context`.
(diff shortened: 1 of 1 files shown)


niklasad1 (commented):
hmm I may be slow but I suppose you want to something like:

```rust
let shared_state = shared_state();
let rpc1 = RpcModule::new(shared_state);
let shared_state = rpc1.context().clone();
let rpc2 = RpcModule::new(shared_state);
```

If that's the case I would prefer to add another API `RpcModule::from_arc` instead

```rust
let shared_state = Arc::new(shared_state());
let rpc1 = RpcModule::from_arc(shared_state.clone());
let rpc2 = RpcModule::from_arc(shared_state);
```

aatifsyed:
I've implemented your suggestion :)

niklasad1 (approved):
Thanks, looks good to me

PR #1326, opened 2024-03-20
Revert "fix(server): return err on WS handshake err (#1288)"

niklasad1:
https://github.com/paritytech/jsonrpsee/pull/1288 never worked and let's revert it.

### DIFF
--- a/server/src/transport/ws.rs
+++ b/server/src/transport/ws.rs
@@ -472,7 +472,7 @@ where
 		}
 		Err(e) => {
 			tracing::debug!(target: LOG_TARGET, "WS upgrade handshake failed: {}", e);
-			Err(hyper::Response::new(hyper::Body::from(format!("Could not upgrade connection: {e}"))))
+			Err(hyper::Response::new(hyper::Body::from(format!("WS upgrade handshake failed: {e}"))))
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


server/src/transport/ws.rs
```diff
@@ -472,7 +472,7 @@ where
 		}
 		Err(e) => {
 			tracing::debug!(target: LOG_TARGET, "WS upgrade handshake failed: {}", e);
-			Err(hyper::Response::new(hyper::Body::from(format!("WS upgrade handshake failed: {e}"))))
+			Err(hyper::Response::new(hyper::Body::from(format!("Could not upgrade connection: {e}"))))
```
niklasad1, line 475:
```suggestion
			Err(hyper::Response::new(hyper::Body::from(format!("WS upgrade handshake failed: {e}"))))
```

server/src/transport/ws.rs
```diff
@@ -448,6 +440,14 @@ where
 			let rpc_service = rpc_middleware.service(rpc_service);
 
 			let fut = async move {
+				let upgraded = match hyper::upgrade::on(req).await {
```
jsdw, line 443: Yup ok, I have no idea why moving this up would break anything offhand!

niklasad1: I forgot about https://github.com/hyperium/hyper/blob/master/examples/upgrades.rs#L56-#L72 so then it adds up :)

PR #1327, opened 2024-03-21
fix(client): support batched notifications

lavigneer:
As per the examples in the [jsonrpc specification](https://www.jsonrpc.org/specification#examples), it is valid to include only notifications in a batch. 

rpc call Batch (all notifications):
```
--> [
        {"jsonrpc": "2.0", "method": "notify_sum", "params": [1,2,4]},
        {"jsonrpc": "2.0", "method": "notify_hello", "params": [7]}
    ]
<-- //Nothing is returned for all notification batches
```

This change adds support for handling this use case.

I ran into this while trying to leverage this library for a proof of conept I was working on a little while back,. My particular use case was using the client crate of this library to communicate with an existing service. I realized I never upstreamed this change and figured it might be useful to others.

I didn't see any contribution guidelines for this project, so please let me know if I'm missing any information. If you're not accepting external contributions I understand and can close this out. Thanks for your time.

### DIFF
--- a/test-utils/src/helpers.rs
+++ b/test-utils/src/helpers.rs
@@ -186,6 +186,11 @@ pub fn server_notification(method: &str, params: Value) -> String {
 	format!(r#"{{"jsonrpc":"2.0","method":"{}", "params":{} }}"#, method, serde_json::to_string(&params).unwrap())
 }
 
+/// Batched server originated notification
+pub fn server_batched_notification(method: &str, params: Value) -> String {
+	format!(r#"[{{"jsonrpc":"2.0","method":"{}", "params":{} }}]"#, method, serde_json::to_string(&params).unwrap())
+}
+
 pub async fn http_request(body: Body, uri: Uri) -> Result<HttpResponse, String> {
 	let client = hyper::Client::new();
 	http_post(client, body, uri).await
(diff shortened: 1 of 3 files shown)


test-utils/src/helpers.rs
```diff
@@ -186,6 +186,12 @@ pub fn server_notification(method: &str, params: Value) -> String {
 	format!(r#"{{"jsonrpc":"2.0","method":"{}", "params":{} }}"#, method, serde_json::to_string(&params).unwrap())
 }
 
+/// Batched server originated notification
+pub fn server_batched_notification(method: &str, params: Value) -> String {
+	format!(r#"[{{"jsonrpc":"2.0","method":"{}", "params":{} }}]"#, method, serde_json::to_string(&params).unwrap())
+}
+
+
```
niklasad1, line 194:
```suggestion
```

niklasad1 (approved):
Thanks, looks good

EDIT: just run `cargo +stable fmt --all` on your code to make the CI happy

lavigneer:
> Thanks, looks good
> 
> EDIT: just run `cargo +stable fmt --all` on your code to make the CI happy

Done and pushed. Thanks

PR #1332, opened 2024-03-25
client: support batched subscription notifs

niklasad1:
Follow-up on https://github.com/paritytech/jsonrpsee/pull/1327

### DIFF
--- a/client/ws-client/src/tests.rs
+++ b/client/ws-client/src/tests.rs
@@ -212,6 +212,29 @@ async fn batched_notification_handler_works() {
 }
 
 #[tokio::test]
+async fn batched_subscription_notif_works() {
+	init_logger();
+
+	let server = WebSocketTestServer::with_hardcoded_subscription(
+		"127.0.0.1:0".parse().unwrap(),
+		server_subscription_id_response(Id::Num(0)),
+		server_batched_subscription("sub", "batched_notif".into()),
+	)
+	.with_default_timeout()
+	.await
+	.unwrap();
+
+	let uri = to_ws_uri_string(server.local_addr());
+	let client = WsClientBuilder::default().build(&uri).with_default_timeout().await.unwrap().unwrap();
+	{
+		let mut nh: Subscription<String> =
+			client.subscribe("sub", rpc_params![], "unsub").with_default_timeout().await.unwrap().unwrap();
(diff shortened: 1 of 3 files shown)


client/ws-client/src/tests.rs
```diff
@@ -211,6 +211,29 @@ async fn batched_notification_handler_works() {
...
+	.with_default_timeout()
+	.await
+	.unwrap();
+
+	let uri = to_ws_uri_string(server.local_addr());
+	let client = WsClientBuilder::default().build(&uri).with_default_timeout().await.unwrap().unwrap();
+	{
+		let mut nh: Subscription<String> =
+			client.subscribe("sub", rpc_params![], "unsub").with_default_timeout().await.unwrap().unwrap();
+		let response: String = nh.next().with_default_timeout().await.unwrap().unwrap().unwrap();
```
lexnv, line 232: nit: It would be interesting to extend this for 2 subscriptions and 2 (or more) batched notifications and ensure that responses to go appropriate handlers:

```rust
	let server = WebSocketTestServer::with_hardcoded_subscription(
		"127.0.0.1:0".parse().unwrap(),
		server_subscription_id_response(Id::Num(0)),
		server_batched_subscription(&[("sub1", "batched_notif1"), ("sub2", "resp2")]),
	)

    let mut sub1 = client.subscribe("sub1", rpc_params![], "unsub");
    let mut sub2 = client.subscribe("sub2", rpc_params![], "unsub");

    assert_eq!(sub1.next(), "batched_notif1");
    assert_eq!(sub2.next(), "resp2");
```

lexnv (approved):
LGTM!

client/ws-client/src/tests.rs
jsdw: This would be a nice extra check yup; right now we can see that the subscription response comes back ok but would be nice to know that eg a few batched subscription responses all end up in the right place :)

jsdw (approved):
LGTM :)

client/ws-client/src/tests.rs
niklasad1: If you have a look at the updated test we now emit `[notif1, notif2, sub]` and check that it goes to the correct handler (notif=sub, sub_notif=sub)

Sure we could spawn a few more subscriptions to ensure the responses are propagated properly in the Vec

PR #1347, opened 2024-04-08
fix(client): `disconnect_reason/read_error` is cancel-safe

niklasad1:
If/when the connection is closed the cause is fetched by `read_error` from the background task. It was not cancel-safe which could have side-affects and mutate state such the internal Option could be `None` and end-up in a panic.

### DIFF
--- a/core/src/client/async_client/mod.rs
+++ b/core/src/client/async_client/mod.rs
@@ -363,7 +363,7 @@ impl ClientBuilder {
 		type PendingIntervalStream = IntervalStream<Pending<()>>;
 
 		let (to_back, from_front) = mpsc::channel(self.max_concurrent_requests);
-		let (err_to_front, err_from_back) = oneshot::channel::<Error>();
+		let disconnect_reason = SharedDisconnectReason::default();
 		let max_buffer_capacity_per_subscription = self.max_buffer_capacity_per_subscription;
 		let (client_dropped_tx, client_dropped_rx) = oneshot::channel();
 		let (send_receive_task_sync_tx, send_receive_task_sync_rx) = mpsc::channel(1);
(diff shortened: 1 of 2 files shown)


lexnv (approved):

core/src/client/async_client/mod.rs
```diff
@@ -145,64 +145,36 @@ impl ThreadSafeRequestManager {
 		self.0.lock().expect("Not poisoned; qed")
 	}
 }
+
+pub(crate) type SharedDisconnectReason = Arc<AsyncRwLock<Option<Arc<Error>>>>;
```
jsdw, line 149: Do we need the inner arc in the `Arc<Error>`?

core/src/client/async_client/mod.rs
```diff
@@ -145,64 +145,36 @@ impl ThreadSafeRequestManager {
...
-				arc_err
-			}
-			ReadErrorOnce::Read(arc_err) => {
-				*write = Some(ReadErrorOnce::Read(arc_err.clone()));
-				arc_err
-			}
-		};
-
-		Error::RestartNeeded(err)
+		if let Some(err) = self.disconnect_reason.read().await.as_ref() {
```
jsdw, line 170: Now we're not holding this lock over any await points (at least, I think we aren't?), should we make it a sync RwLock?

jsdw (approved):
Couple of small comments but looks ok!

Given that the method was actually listed as not being cancel safe, I do think it was prob their fault anyways if that was what led to the issue, but always nice to remove potential footguns :)

niklasad1:
>  I do think it was prob their fault anyways if that was what led to the issue, but always nice to remove potential footguns :)

Actually, we have https://docs.rs/jsonrpsee-ws-client/latest/jsonrpsee_ws_client/struct.WsClientBuilder.html#method.request_timeout where the "future/method call" is dropped and it may suffer from this cancel-safety issue and "nothing caused by the user". Thus, an internal jsonrpsee issue as well which I forgot about...

But it could be caused `disconnect_reason` which is on the user but better to keep everything cancel-safe if possible :)

core/src/client/async_client/mod.rs
niklasad1: The error is not clone so this a just hack around that, any better ideas?

jsdw: Ah ok, in that case it's all gravy :)

core/src/client/async_client/mod.rs
niklasad1: done nice catch, one dependency removed :)

maan2003:
I had a fix in different way, it uses `&mut Fut: Future` to achieve cancel safety. 
See c5ed2553fbc1

niklasad1:
> I had a fix in different way, it uses &mut Fut: Future to achieve cancel safety.
See https://github.com/paritytech/jsonrpsee/commit/c5ed2553fbc11ba47fd94c8aad9b615132a54363

Yepp, that works as well but I didn't like the old code and decided to refactor it. The RwLock is basically only "write_locked once" and should be quite efficient anyway because once the connection is closed it should never block (only read locks can be held)

I guess we could change it to a OnceLock instead but I'm quite happy with this for now

PR #1360, opened 2024-04-29
fix(proc-macros): make feature `server-core` compile

niklasad1:
Similar to #1359 and kudos to @koushiro for detecting this but I prefer not re-export tokio

PR #1361, opened 2024-04-29
fix(proc-macros): re-export tokio for params parsing

niklasad1:
Follow-up on #1360 and I forgot that we use tokio in param parsing.

I also added a test crate to ensure that it compiles

### DIFF
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -76,6 +76,7 @@ pub mod __reexports {
 	pub use serde;
 	pub use serde_json;
 
+	// Needed for the params parsing in the proc macro API.
 	cfg_client_or_server! {
 		pub use tokio;
 	}
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -75,6 +75,10 @@ pub mod __reexports {
 	pub use async_trait::async_trait;
 	pub use serde;
 	pub use serde_json;
+
+	cfg_client_or_server! {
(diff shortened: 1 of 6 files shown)


tests/proc-macro-core/src/lib.rs
```diff
@@ -0,0 +1,56 @@
+//! Test module for the proc-macro API to make sure that is compiles with core features.
```
niklasad1, line 1:
```suggestion
//! Test module for the proc-macro API to make sure that it compiles with core features.
```

niklasad1:
@CodiumAI-Agent /review

tests/proc-macro-core/src/lib.rs
```diff
@@ -0,0 +1,56 @@
+//! Test module for the proc-macro API to make sure that it compiles with core features.
```
niklasad1, line 1:
```suggestion
//! Test module for the proc-macro API to make sure that it compiles with the core features.
```

core/src/lib.rs
```diff
@@ -75,6 +75,10 @@ pub mod __reexports {
 	pub use async_trait::async_trait;
 	pub use serde;
 	pub use serde_json;
+
+	cfg_client_or_server! {
```
niklasad1, line 79:
```suggestion
	// Needed for the params parsing in the proc macro API.
	cfg_client_or_server! {
```

PR #1363, opened 2024-04-29
fix(proc macros): don't use `params` as name

### LINKED ISSUE #1362: Can't use `params` as parameter name when using `subscription` proc-macro attribute
Rust toolchain

```
stable-aarch64-apple-darwin (default)
rustc 1.77.2 (25ef9e3d8 2024-04-09)
```

Example:

```toml
[package]
name = "rpc-demo"
version = "0.1.0"
edition = "2021"

[dependencies]
jsonrpsee = { git = "https://github.com/paritytech/jsonrpsee", branch = "master", features = ["client-core", "server-core", "macros"] }
serde = { version = "1.0", features = ["derive"] }
```

```rust
use jsonrpsee::{core::SubscriptionResult, proc_macros::rpc};

#[derive(Debug, Serialize, Deserialize)]
pub enum PubSubKind {
    A,
    B,
}

#[derive(Debug, Serialize, Deserialize)]
pub struct PubSubParams {
    params: String,
}

#[derive(Debug, Serialize, Deserialize)]
pub struct PubSubItem {
    result: String,
}

#[async_trait]
pub trait FooPubSubApi {
    // Error
    // #[subscription(name = "subscribe" => "subscription", unsubscribe = "unsubscribe", item = PubSubItem)]
    // async fn sub(&self, kind: PubSubKind, params: Option<PubSubParams>) -> SubscriptionResult;
    
    // OK
    #[subscription(name = "subscribe" => "subscription", unsubscribe = "unsubscribe", item = PubSubItem)]
    async fn sub(&self, kind: PubSubKind, pubsub_params: Option<PubSubParams>) -> SubscriptionResult;
}
```

Error:
```
error[E0277]: the trait bound `ArrayParams: Serialize` is not satisfied
   --> src/lib.rs:24:40
    |
20  | #[rpc(client, server, namespace = "foo")]
    | ----------------------------------------- required by a bound introduced by this call
...
24  |     async fn sub(&self, kind: PubSubKind, params: Option<PubSubParams>) -> Subscript...
    |                                           ^^^^^^ the trait `Serialize` is not implemented for `ArrayParams`
    |
    = help: the following other types implement trait `Serialize`:
              bool
              char
              isize
              i8
              i16
              i32
              i64
              i128
            and 148 others
note: required by a bound in `ArrayParams::insert`
   --> /Users/qinxuan/.cargo/git/checkouts/jsonrpsee-5ae67701a3cedc92/c908eeb/core/src/params.rs:200:19
    |
200 |     pub fn insert<P: Serialize>(&mut self, value: P) -> Result<(), serde_json::Error> {
    |                      ^^^^^^^^^ required by this bound in `ArrayParams::insert`

For more information about this error, try `rustc --explain E0277`.
error: could not compile `rpc-demo` (lib) due to 1 previous error
```

niklasad1:
Because jsonrpsee's internal proc macro code was using `params` as internal variable name it could conflict and with `params` in the user's API in such case it may try to overwrite the "wrong" variable.

In this PR I have renamed this internal variable to something much weirder which should be quite unlikely that anyone will use and panic if those collide to make it easier to debug.

Close #1362

### DIFF
--- a/proc-macros/src/render_client.rs
+++ b/proc-macros/src/render_client.rs
@@ -243,7 +243,7 @@ impl RpcDescription {
 				}
 
 				quote!({
-					let mut #p #jsonrpsee::core::params::ObjectParams::new();
+					let mut #p = #jsonrpsee::core::params::ObjectParams::new();
 					#(
 						if let Err(err) = #p.insert( #params_insert ) {
 							panic!("Parameter `{}` cannot be serialized: {:?}", stringify!( #params_insert ), err);
--- a/proc-macros/src/render_client.rs
+++ b/proc-macros/src/render_client.rs
@@ -226,7 +226,9 @@ impl RpcDescription {
 		}
 
 		if params.iter().any(|(param, _)| param.ident == p) {
-			panic!("Cannot use `{}` as a parameter name", ILLEGAL_PARAM_NAME);
+			panic!(
+				"Cannot use `{}` as a parameter name because it's overlapping with an internal variable in the generated code. Change it something else to make it work", ILLEGAL_PARAM_NAME
+			);
 		}
 
 		match param_kind {
--- a/proc-macros/src/render_client.rs
+++ b/proc-macros/src/render_client.rs
@@ -29,7 +29,7 @@ use crate::rpc_macro::{RpcDescription, RpcMethod, RpcSubscription};
 use proc_macro2::TokenStream as TokenStream2;
 use quote::{quote, quote_spanned};
 use syn::spanned::Spanned;
-use syn::{AngleBracketedGenericArguments, FnArg, Pat, PatIdent, PatType, PathArguments, TypeParam};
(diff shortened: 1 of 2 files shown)


niklasad1:
@CodiumAI-Agent /review

proc-macros/src/render_client.rs
```diff
@@ -210,6 +210,8 @@ impl RpcDescription {
 		param_kind: &ParamKind,
 		signature: &syn::TraitItemFn,
 	) -> TokenStream2 {
+		const ILLEGAL_PARAM_NAME: &str = "__________________params";
```
jsdw, line 213: The only wee thing is that this might provoke some sort of clippy warning about _ being used in front of a variable that's in use, but perhaps not an issue (and you could call it eg jsonrpsee__params or whatever if it was) :)

jsdw (approved):

PR #1365, opened 2024-05-06
proc-macros: rename parameter names

deflator-hacker:
Decoupling rust parameter names from (de)serialized json is essential for me and useful in general.

Some usecases include:

- supporting parameter names that are reserved keywords in rust, such as `type` (currently impossible)
- talking to case sensitive RPC servers only accepting non-snake-case parameters (currently impossible without clippy complaining)

Those are both usecases I personally have.

The PR implements this by introducing a new attribute for the `rpc` macro called `argument`.

- documentation has been updated
- tests have been written

I previously responded to an outdated similar PR [here](https://github.com/paritytech/jsonrpsee/pull/1015#issuecomment-2094789210).

Minimal working example:

```rust
#[rpc(client, server)]
pub trait MyRpc {
	#[method(name = "foo")]
	async fn foo(
		&self,
		param_a: u8,
		#[argument(rename = "type")]
		r#type: String
	) -> RpcResult<u16>;
}
```

I divided the changes into multiple commits because they are a bit lengthy.
This is my first open source contribution, I am eager to learn, let me know what can be improved.


niklasad1:
Looks overall good, nice job

I'm not following why you need to rename parameters when those passed as Array (these are omitted by the client)? In such scenarios then variable doesn't correspond to the actual param name, so why is it needed?

The contrary applies to JSON-Object where the key is the same name passed in the call with some glue to convert snake_case to camaleCase. Then I follow why these need to be renamed to support reserved rust keywords etc....

I'll take a closer look tomorrow.

deflator-hacker:
> I'm not following why you need to rename parameters when those passed as Array (these are omitted by the client)? In such scenarios then variable doesn't correspond to the actual param name, so why is it needed?

You are absolutely right, there is no point (actually no way) to rename parameters in arrays.
I included a check for array params in [proc_macros/tests/ui/correct/rename_fn_params.rs](https://github.com/deflator-hacker/jsonrpsee/blob/master/proc-macros/tests/ui/correct/rename_fn_params.rs#L17) just because I wanted to assert that I did not break anything deserialization-related by accident.

proc-macros/src/render_server.rs
```diff
@@ -147,25 +150,23 @@ impl RpcDescription {
...
 						})
 					})
-				} else {
-					if method.signature.sig.asyncness.is_some() {
-						handle_register_result(quote! {
-							rpc.register_async_method(#rpc_method_name, |params, context| async move {
-								#parsing
-								#into_response::into_response(context.as_ref().#rust_method_name(#params_seq).await)
-							})
+				} else if method.signature.sig.asyncness.is_some() {
```
niklasad1, line 153: I suppose this is unrelated and just fixes a clippy warning?

proc-macros/src/rpc_macro.rs
```diff
@@ -37,13 +37,57 @@ use quote::quote;
 use syn::spanned::Spanned;
 use syn::{punctuated::Punctuated, Attribute, Token};
 
+/// argument to rpc function
```
niklasad1, line 40:
```suggestion
/// Represents a single argument in a RPC call.
```

proc-macros/src/rpc_macro.rs
```diff
@@ -37,13 +37,57 @@ use quote::quote;
...
+///
+/// stores modifications based on attributes
+#[derive(Debug, Clone)]
+pub struct RpcFnArg {
+	pub arg_pat: syn::PatIdent,
```
niklasad1, line 45: can you make these private because you have added getters fns below?

proc-macros/tests/ui/correct/basic.rs
```diff
@@ -111,6 +122,8 @@ async fn main() {
 	let client = WsClientBuilder::default().build(&server_url).await.unwrap();
 
 	assert_eq!(client.async_method(10, "a".into()).await.unwrap(), 42);
+	assert_eq!(client.rename_params(256, true).await.unwrap(), 128);
+	assert_eq!(client.rename_params(256, false).await.unwrap(), 256);
```
niklasad1, line 126: you can just add two additional tests here with object params.

```rust
	let mut params = ObjectParams::new();
	params.insert("type", 256).unwrap();
	params.insert("halfType", false).unwrap();

	assert_eq(client.request::<u16, ObjectParams>("rename_params", params).await.unwrap(), 128);
```

```rust
	let mut params = ObjectParams::new();
	params.insert("type", 256).unwrap();
	params.insert("halfType", true).unwrap();

	assert_eq(client.request::<u16, ObjectParams>("rename_params", params).await.unwrap(), 256);
```

Then remove the extra test file that you added below "proc-macros/tests/ui/correct/rename_fn_params.rs"

proc-macros/src/rpc_macro.rs
niklasad1: nvm, it's used to extract the ty in a couple of places

proc-macros/src/rpc_macro.rs
```diff
@@ -37,13 +37,57 @@ use quote::quote;
...
+
+		Ok(Self { arg_pat, rename_to, ty })
+	}
+
+	/// Return the pattern identifier of the argument.
+	pub fn arg_pat(&self) -> &syn::PatIdent {
+		&self.arg_pat
+	}
+	/// Return the string representation of this argument when (de)serializing.
+	pub fn rename_to(&self) -> &Option<String> {
```
niklasad1, line 75: is it possible to refactor this method then internally fetch the name from the rename if it exist otherwise fall back to self.arg_pat?

It would be easier to just call `fn_arg.name()`

```rust
pub fn name(&self) -> String {
	self.rename_to.clone().unwrap_or_else(|| self.arg_pat.ident.to_string())
}
```

proc-macros/src/render_server.rs
```diff
@@ -383,18 +384,30 @@ impl RpcDescription {
...
 				alias_vals.push(',');
 				alias_vals.push_str(&format!(
 					r#"alias = "{}""#,
-					heck::ToLowerCamelCase::to_lower_camel_case(name.ident.to_string().as_str())
+					heck::ToLowerCamelCase::to_lower_camel_case(
+						rename_to.as_deref().unwrap_or(arg_pat.ident.to_string().as_str())
+					)
 				));
 
+				let serde_rename = rename_to
```
niklasad1, line 406: we shouldn't need this serde rename would be better to rename the name in the struct instead, if you agree with my proposed API with a function name we could do this instead:

```rust
			let fields = params.iter().zip(generics.clone()).map(|(fn_arg, ty)| {
				// this could be a renamed name or not... but shouldn't matter at this point
                let name = fn_arg.name();

				let mut alias_vals = String::new();
				alias_vals.push_str(&format!(r#"alias = "{}""#, heck::ToSnakeCase::to_snake_case(name.as_str())));
				alias_vals.push(',');
				alias_vals
					.push_str(&format!(r#"alias = "{}""#, heck::ToLowerCamelCase::to_lower_camel_case(name.as_str())));

				let alias = TokenStream2::from_str(alias_vals.as_str()).unwrap();
				let serde_alias: Attribute = syn::parse_quote! {
					#[serde(#alias)]
				};
				let unquoted_name = TokenStream2::from_str(name.as_str()).unwrap();

				quote! {
					#serde_alias
					#unquoted_name: #ty,
				}
			});
```

proc-macros/src/render_server.rs
deflator-hacker: Yes, I can revert if this is not the right place to do this but CI tests linted this so I fixed it.

proc-macros/src/render_server.rs
deflator-hacker: I adopted the `RpcFnArg::name()` API.

I tested this change, it does not work because `RpcFnArg::name()` can be _any_ string, such as `"type"`, `"in"`, `"for"` etc.
If it was `"type"` the struct would be invalid because `"type"` is a reserved keyword.
Leveraging serde here gives us all possible strings that can be used with serde.

proc-macros/src/rpc_macro.rs
deflator-hacker: done

proc-macros/tests/ui/correct/basic.rs
deflator-hacker: done

proc-macros/src/rpc_macro.rs
deflator-hacker: done

niklasad1:
Looks great, I'll can push a fix to unblock the UI tests tomorrow it's probably the latest rustc release that broke them.

proc-macros/tests/ui/correct/basic.rs
```diff
@@ -24,6 +24,13 @@ pub trait Rpc {
 	#[method(name = "array_params")]
 	async fn array_params(&self, items: Vec<u64>) -> RpcResult<u64>;
 
+	#[method(name = "rename_params", param_kind = map)]
```
niklasad1, line 27: you shouldn't need "param_kind = map" because it's only used when you are using the proc macro API anyway it doesn't matter.

proc-macros/tests/ui/correct/basic.rs
```diff
@@ -111,6 +122,17 @@ async fn main() {
 	let client = WsClientBuilder::default().build(&server_url).await.unwrap();
 
 	assert_eq!(client.async_method(10, "a".into()).await.unwrap(), 42);
+
+	let mut params = ObjectParams::new();
+	params.insert("type", 256).unwrap();
+	params.insert("halfType", true).unwrap();
+	assert_eq!(client.request::<u16, ObjectParams>("foo_rename_params", params).await.unwrap(), 128);
```
niklasad1, line 129: can you add another call here with ArrayParams to ensure that the rename stuff doesn't break it?

proc-macros/tests/ui/correct/basic.rs
deflator-hacker: thanks, removed it

proc-macros/tests/ui/correct/basic.rs
deflator-hacker: added

proc-macros/src/render_server.rs
```diff
@@ -383,17 +384,17 @@ impl RpcDescription {
...
+				alias_vals.push_str(&format!(r#"alias = "{}""#, heck::ToSnakeCase::to_snake_case(name.as_str())));
 				alias_vals.push(',');
-				alias_vals.push_str(&format!(
-					r#"alias = "{}""#,
-					heck::ToLowerCamelCase::to_lower_camel_case(name.ident.to_string().as_str())
-				));
+				alias_vals
+					.push_str(&format!(r#"alias = "{}""#, heck::ToLowerCamelCase::to_lower_camel_case(name.as_str())));
+
+				let serde_rename = quote!(#[serde(rename = #name)]);
```
niklasad1, line 397: This is needed because arg_pat below can't be reserved rust names such as type

proc-macros/src/rpc_macro.rs
```diff
@@ -37,13 +37,57 @@ use quote::quote;
 use syn::spanned::Spanned;
 use syn::{punctuated::Punctuated, Attribute, Token};
 
+/// Represents a single argument in a RPC call.
+///
+/// stores modifications based on attributes
+#[derive(Debug, Clone)]
+pub struct RpcFnArg {
+	pub arg_pat: syn::PatIdent,
```
niklasad1, line 45:
```suggestion
	pub(crate) arg_pat: syn::PatIdent,
```

proc-macros/src/rpc_macro.rs
```diff
@@ -37,13 +37,57 @@ use quote::quote;
...
 use syn::{punctuated::Punctuated, Attribute, Token};
 
+/// Represents a single argument in a RPC call.
+///
+/// stores modifications based on attributes
+#[derive(Debug, Clone)]
+pub struct RpcFnArg {
+	pub(crate) arg_pat: syn::PatIdent,
+	rename_to: Option<String>,
+	pub ty: syn::Type,
```
niklasad1, line 47:
```suggestion
	pub(crate) ty: syn::Type,
```

niklasad1 (approved):
Nice work looks good

proc-macros/src/lib.rs
```diff
@@ -206,6 +206,16 @@ pub(crate) mod visitor;
...
 ///
+/// ### `argument` attribute
+///
+/// `argument` attribute is used to modify a function argument.
+///
+/// **Arguments:**
+///
+/// - `rename`: rename the generated JSON key.
+///
+/// **Method requirements:**
```
jsdw, line 217: This line isn't followed by anything? :)

```suggestion
```

proc-macros/tests/ui/correct/basic.rs
```diff
@@ -24,6 +24,13 @@ pub trait Rpc {
 	#[method(name = "array_params")]
 	async fn array_params(&self, items: Vec<u64>) -> RpcResult<u64>;
 
+	#[method(name = "rename_params")]
+	async fn rename_params(
+		&self,
+		#[argument(rename = "type")] r#type: u16,
```
jsdw, line 30: DQ: without this rename, would it not still be valid to pass a param with name "type"?

proc-macros/tests/ui/correct/basic.rs
```diff
@@ -24,6 +24,13 @@ pub trait Rpc {
...
+		#[argument(rename = "halfType")] half_type: bool,
```
jsdw, line 31: I don't recall here either; would `halfType` not already be accepted or do we not to any camel cas transform?

I'd feel more confident that the test was working if the rename was to something more different than the original :)

jsdw (approved):
LGTM, just a couple of nits!

proc-macros/tests/ui/correct/basic.rs
deflator-hacker: I am unsure what you are asking.
I think that it would not be valid because the server would expect params with name "r#type" instead of "type".

Actually, the original usecase that got me to writing this PR was that I had a non-jsonrpsee server accepting a param "in" which could not be done through the proc-macros API because "in" is a reserved keyword (as "type" is, too) and thus all params would be serialized to "r#in", preventing the server from recognizing the param.

proc-macros/tests/ui/correct/basic.rs
deflator-hacker: You are right for servers generated by jsonrpsee, they accept mutliple cases ([see here](https://github.com/paritytech/jsonrpsee/blob/master/proc-macros/src/render_server.rs#L390)).

When talking to other JSON-RPC servers this might not be the case, hence my PR.

proc-macros/tests/ui/correct/basic.rs
niklasad1: So the rpc definition looks like this:

```rust
    #[method(name = "foo")]
    fn bar(f: usize, r#type: usize) -> usize {
        f + r#type
    }
```

Then in codegen we generated a struct as JSON Object with param names as the idents from the params binding such that the following would be generated:

```rust
    #[derive(serde::Serialize, serde::Deserialize)
    struct ParamsObject<G0, G1> {
        #[serde(alias = "storage_key", alias = "storageKey")]
        f: G0,
        #[serde(alias = "r_type", alias = "rType")]
        r#type: G1,
    }
```

Thus, it was not possible to make call to the server with  `type` as parameter name because it's keyword in rust prior to this PR. All good

jsdw: Aah thanks for the explanations! I thought that `r#type` might just resolve to the ident `type` anyway (ignoring the raw prefix bit that allows you to actually type it at all), but I was wrong actually about that :)

PR #1368, opened 2024-05-10
upgrade to hyper v1.0

### LINKED ISSUE #1257: [deps]: update hyper to v1.0
Hyper has released a new version v1.0 which has plenty of changes that impacts jsonrpsee.

We want to update when the rest of the ecosystem has updated.

Me and Tadeo have chatted about this and the roadmap is the following:

1. Update the server to hyper v1.0 and the service_fn has be replaced by `loop { let stream = socket.accept(), ... }`
    - http has be bumped to v1.0
    - our shared helpers for reading bytes may need to be changed 
    - some other stuff
    - maybe tower has been bumped as well not sure
2. Bump the client and best would be the use the hyper util for the high level client API
3. Bump hyper-rustls, tower etc to hyper v1.0

niklasad1:
Close #1257 

Things that could be improved in this PR or in follow-up:
- add helper function serve/serve_with_graceful_shutdown such that users of the low-level API doesn't have to implement their graceful shutdown stuff (I have repeated that a few times already now)
- get rid of BoxBody if possible
- add additional feature flags to support `aws-lc-rs` TLS crypto backend


client/http-client/Cargo.toml
```diff
@@ -15,13 +15,16 @@ publish = true
 
 [dependencies]
 async-trait = "0.1"
-hyper = { version = "0.14.10", features = ["client", "http1", "http2", "tcp"] }
-hyper-rustls = { version = "0.24", optional = true, default-features = false, features = ["http1", "http2", "tls12", "logging"] }
+hyper = { version = "1.3", features = ["client", "http1", "http2"] }
+hyper-rustls = { version = "0.27.1", optional = true, default-features = false, features = ["http1", "http2", "tls12", "logging"] }
+hyper-util = { version = "*", features = ["client", "client-legacy"] }
```
niklasad1, line 20:
```suggestion
hyper-util = { version = "0.1.1", features = ["client", "client-legacy"] }
```

client/http-client/Cargo.toml
```diff
@@ -34,8 +37,8 @@ tokio = { version = "1.16", features = ["net", "rt-multi-thread", "macros"] }
 
 [features]
 default = ["native-tls"]
-native-tls = ["hyper-rustls/native-tokio", "__tls"]
-webpki-tls = ["hyper-rustls/webpki-tokio", "__tls"]
+native-tls = ["hyper-rustls/native-tokio", "hyper-rustls/ring", "__tls"]
```
niklasad1, line 40: let's discuss whether we should support both crypto backends.

Rustls now supports both ring and aws-lc-rs but migrating to aws-lc-rs requires more dependencies such as cmake, NASM etc.

client/ws-client/src/tests.rs
```diff
@@ -471,10 +473,12 @@ async fn redirections() {
 	.unwrap();
 
 	let server_url = format!("ws://{}", server.local_addr());
-	let redirect_url = jsonrpsee_test_utils::mocks::ws_server_with_redirect(server_url);
+	let redirect_url =
+		jsonrpsee_test_utils::mocks::ws_server_with_redirect(server_url).with_default_timeout().await.unwrap();
 
 	// The client will first connect to a server that only performs re-directions and finally
 	// redirect to another server to complete the handshake.
+	println!("Connecting to {redirect_url}");
```
niklasad1, line 481:
```suggestion
```

client/http-client/src/transport.rs
```diff
@@ -281,20 +284,22 @@ pub struct HttpTransportClient<S> {
...
 		}
 
-		let mut req = hyper::Request::post(&self.target);
+		let mut req = HttpRequest::post(&self.target);
 		if let Some(headers) = req.headers_mut() {
 			*headers = self.headers.clone();
 		}
-		let req = req.body(From::from(body)).expect("URI and request headers are valid; qed");
+
+		let req = req.body(body.into()).expect("Failed to create request");
```
niklasad1, line 302:
```suggestion
		let req = req.body(body.into()).expect("URI and request headers are valid; qed");
```

server/src/middleware/http/proxy_get_request.rs
```diff
@@ -26,13 +26,17 @@
 
 //! Middleware that proxies requests at a specified URI to internal
 //! RPC method calls.
-
+//!
```
niklasad1, line 29:
```suggestion
```

server/src/transport/ws.rs
```diff
@@ -468,11 +470,11 @@ where
 				background_task(params).await;
 			};
 
-			Ok((response.map(|()| hyper::Body::empty()), fut))
+			Ok((response.map(|()| HttpResponseBody::default()), fut)) // empty body here
```
niklasad1, line 473:
```suggestion
			Ok((response.map(|()| HttpResponseBody::default()), fut))
```

client/http-client/Cargo.toml
jsdw: I'm happy with not exposing it initially, and then as a non-breaking change we can always expose it later or whatever if there's demand :)

client/http-client/src/lib.rs
```diff
@@ -46,3 +46,10 @@ mod tests;
 pub use client::{HttpClient, HttpClientBuilder};
 pub use hyper::http::{HeaderMap, HeaderValue};
 pub use jsonrpsee_types as types;
+
+/// Default HTTP body for the client.
+pub type HttpBody = http_body_util::Full<hyper::body::Bytes>;
```
jsdw, line 51: Could this also be `jsonrpsee_core::http_helpers::ResponseBody`, since the below two things reference that?

(maybe rename `RepsonseBody` to `HttpBody` though?)

examples/examples/http_proxy_middleware.rs
```diff
@@ -63,17 +66,17 @@ async fn main() -> anyhow::Result<()> {
 	println!("[main]: response: {:?}", response);
 
 	// Use hyper client to manually submit a `GET /health` request.
-	let http_client = Client::new();
+	let http_client = Client::builder(TokioExecutor::new()).build_http();
```
jsdw, line 69: Interesting way to be agnostic over runtime; I could imagine we might adopt a similar thing in subxt to hide the spawning stuff behind the scenes unless the user wants to override the executor or manually drive stuff

server/src/transport/ws.rs
```diff
@@ -415,6 +416,15 @@ where
 
 	match server.receive_request(&req) {
 		Ok(response) => {
+			let upgraded = match hyper::upgrade::on(req).await {
+				Ok(u) => u,
+				Err(e) => {
+					tracing::debug!(target: LOG_TARGET, "WS upgrade handshake failed: {}", e);
+					return Err(HttpResponse::new(HttpResponseBody::from(format!("WS upgrade handshake failed {e}"))));
+				}
+			};
```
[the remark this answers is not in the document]
jsdw: Like, previously it would only be executed when the returned future is polled, but now it's executed in the `connect(..).await` and has to finish before the fut is returned

server/src/server.rs
```diff
@@ -787,54 +781,95 @@ impl<HttpMiddleware, RpcMiddleware> Builder<HttpMiddleware, RpcMiddleware> {
...
 	/// use jsonrpsee_server::{Methods, ServerHandle, ws, stop_channel};
 	/// use tower::Service;
 	/// use std::{error::Error as StdError, net::SocketAddr};
+	/// use futures_util::future::{self, Either};
+	/// use hyper_util::rt::{TokioIo, TokioExecutor};
 	///
-	/// fn run_server() -> ServerHandle {
-	///     let addr = SocketAddr::from(([127, 0, 0, 1], 0));
+	/// async fn run_server() -> ServerHandle {
+	///     let listener = tokio::net::TcpListener::bind(SocketAddr::from(([127, 0, 0, 1], 0))).await.unwrap();
```
jsdw, line 791: I guess this could move into the `tokio::spawn` and then `run_server` could be sync again (iirc this is in an example or two as well)?

jsdw (approved):
LGTM, nice one! Just a couple of small comments!

server/src/transport/ws.rs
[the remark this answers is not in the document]
niklasad1: https://github.com/hyperium/hyper/blob/master/examples/upgrades.rs#L56-#L72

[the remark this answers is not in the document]
niklasad1: Thus, it will be stuck until the response is returned but I can add a comment in the code because it confuses me as well

niklasad1:
@CodiumAI-Agent /review

server/src/server.rs
```diff
@@ -960,32 +995,28 @@ impl<RpcMiddleware, HttpMiddleware> TowerService<RpcMiddleware, HttpMiddleware>
 	}
 }
 
-impl<RpcMiddleware, HttpMiddleware> hyper::service::Service<hyper::Request<hyper::Body>>
-	for TowerService<RpcMiddleware, HttpMiddleware>
+impl<B, RpcMiddleware, HttpMiddleware> Service<HttpRequest<B>> for TowerService<RpcMiddleware, HttpMiddleware>
```
lexnv, line 998: nit: Maybe instead of `B` we can call this `Body`? To keep consistency with `RpcMiddleware` and `HttpMiddleware`?

client/ws-client/src/tests.rs
```diff
@@ -461,6 +461,8 @@ fn assert_error_response(err: Error, exp: ErrorObjectOwned) {
 
 #[tokio::test]
 async fn redirections() {
+	init_logger();
```
lexnv, line 464: nit: is this indended?

lexnv (approved):
LGTM!

PR #1373, opened 2024-05-21
refactor: use rustls-platform-verifier cert store

### LINKED ISSUE #1340: Replace rust-native-certs and webpki-roots with rust-platform-verifier
The current build doesn't support iOS/Android builds if you don't split off to use the webpki-tls on android and ios. While researching this, I came across https://github.com/rustls/rustls-platform-verifier/, which is the rusttls replacement for native-certs that optimizes cert verification for each platform.

I think switching to rust-platform-verifier should give a performance boost, reduce the choice of "Which tls to activate" and support more platforms out of the box.

niklasad1:
Close #1340

This PR replaces the current certificate stores `rust-native-cert` and `webpki` with `rustls-platform-verifier` and additionally it adds an API to use a custom certificate store for special use-cases e.g. such as disabling cert verification or use some custom certificate store on some niche platform.

### DIFF
--- a/client/http-client/Cargo.toml
+++ b/client/http-client/Cargo.toml
@@ -39,7 +39,6 @@ tokio = { version = "1.16", features = ["net", "rt-multi-thread", "macros"] }
 [features]
 default = ["tls"]
 
-# Internal feature to indicate whether TLS is enabled.
 # Does nothing on its own.
 tls = ["hyper-rustls", "rustls", "rustls-platform-verifier"]
 
--- a/client/http-client/Cargo.toml
+++ b/client/http-client/Cargo.toml
@@ -17,8 +17,7 @@ publish = true
 async-trait = "0.1"
 hyper = { version = "1.3", features = ["client", "http1", "http2"] }
 hyper-rustls = { version = "0.27.1", optional = true, default-features = false, features = ["http1", "http2", "tls12", "logging", "rustls-platform-verifier", "ring"] }
-rustls-platform-verifier = { version = "0.3", optional = true }
-hyper-util = { version = "0.1.1", features = ["client", "client-legacy"] }
+hyper-util = { version = "0.1.1", features = ["client", "client-legacy", "tokio", "http1", "http2"] }
 http-body = "1"
 jsonrpsee-types = { workspace = true }
 jsonrpsee-core = { workspace = true, features = ["client", "http-helpers"] }
@@ -41,7 +40,7 @@ default = ["tls"]
 
 # Internal feature to indicate whether TLS is enabled.
 # Does nothing on its own.
-tls = ["hyper-rustls", "rustls", "rustls-platform-verifier"]
+tls = ["hyper-rustls", "rustls"]
 
 [package.metadata.docs.rs]
 all-features = true
--- a/client/http-client/Cargo.toml
+++ b/client/http-client/Cargo.toml
@@ -16,7 +16,7 @@ publish = true
 [dependencies]
 async-trait = "0.1"
 hyper = { version = "1.3", features = ["client", "http1", "http2"] }
-hyper-rustls = { version = "0.27.1", default-features = false, features = ["http1", "http2", "tls12", "logging", "ring", "rustls-platform-verifier"], optional = true }
+hyper-rustls = { version = "0.27.1", default-features = false, features = ["http1", "http2", "tls12", "logging", "ring"], optional = true }
 hyper-util = { version = "0.1.1", features = ["client", "client-legacy", "tokio", "http1", "http2"] }
 http-body = "1"
 jsonrpsee-types = { workspace = true }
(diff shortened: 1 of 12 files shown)


core/src/client/mod.rs
```diff
@@ -522,16 +522,6 @@ impl<T: Clone> RequestIdGuard<T> {
 	}
 }
 
-/// What certificate store to use
```
niklasad1, line 525: remove this from here because I don't want to expose another feature flag for TLS stuff

client/http-client/Cargo.toml
```diff
@@ -35,13 +37,11 @@ jsonrpsee-test-utils = { path = "../../test-utils" }
 tokio = { version = "1.16", features = ["net", "rt-multi-thread", "macros"] }
 
 [features]
-default = ["native-tls"]
-native-tls = ["hyper-rustls/native-tokio", "hyper-rustls/ring", "__tls"]
-webpki-tls = ["hyper-rustls/webpki-tokio", "hyper-rustls/ring", "__tls"]
+default = ["tls"]
 
 # Internal feature to indicate whether TLS is enabled.
```
jsdw, line 42: Not internal any more then? :)

client/http-client/src/client.rs
```diff
@@ -110,31 +113,84 @@ impl<L> HttpClientBuilder<L> {
...
+	/// This is enabled with the default settings and features.
 	///
 	/// # Optional
 	///
-	/// This requires the optional `native-tls` feature.
-	#[cfg(feature = "native-tls")]
-	pub fn use_native_rustls(mut self) -> Self {
+	/// This requires the optional `tls` feature.
+	#[cfg(feature = "tls")]
+	pub fn with_rustls_platform_verifier(mut self) -> Self {
```
jsdw, line 124: Since this is already the default if you do nothing; is there a point in having this function?

client/http-client/src/transport.rs
```diff
@@ -123,9 +127,17 @@ impl HttpTransportClientBuilder<Identity> {
 }
 
 impl<L> HttpTransportClientBuilder<L> {
-	/// Set the certificate store.
-	pub fn set_certification_store(mut self, cert_store: CertificateStore) -> Self {
-		self.certificate_store = cert_store;
+	/// See docs [`crate::HttpClientBuilder::with_rustls_platform_verifier`] for more information.
+	#[cfg(feature = "tls")]
+	pub fn with_rustls_platform_verifier(mut self) -> Self {
```
jsdw, line 132: Same; do we really need this?

client/transport/src/ws/mod.rs
```diff
@@ -97,31 +113,28 @@ impl Default for WsTransportClientBuilder {
...
+	/// This is enabled with the default settings and features.
 	///
 	/// # Optional
 	///
-	/// This requires the optional `native-tls` feature.
-	#[cfg(feature = "native-tls")]
-	pub fn use_native_rustls(mut self) -> Self {
+	/// This requires the optional `tls` feature.
+	#[cfg(feature = "tls")]
+	pub fn with_rustls_platform_verifier(mut self) -> Self {
```
jsdw, line 125: Do we need this?

jsdw (approved):
Little comment to remove on the tls feature and wondering if we need some builder fns but looks good to me!

client/http-client/src/client.rs
niklasad1: Ah, good point let's remove it :)

client/http-client/Cargo.toml
niklasad1: We had two feature flags for tls (native-tls, webpki-tls) before which was public but we don't need them anymore because we only support rustls-platform-verifier and the internal feature flag is not needed anymore

niklasad1, line 42:
```suggestion
```

client/http-client/Cargo.toml
```diff
@@ -35,13 +37,10 @@ jsonrpsee-test-utils = { path = "../../test-utils" }
 tokio = { version = "1.16", features = ["net", "rt-multi-thread", "macros"] }
 
 [features]
-default = ["native-tls"]
-native-tls = ["hyper-rustls/native-tokio", "hyper-rustls/ring", "__tls"]
-webpki-tls = ["hyper-rustls/webpki-tokio", "hyper-rustls/ring", "__tls"]
+default = ["tls"]
 
-# Internal feature to indicate whether TLS is enabled.
 # Does nothing on its own.
```
niklasad1, line 42:
```suggestion
```

client/http-client/Cargo.toml
niklasad1: ah ok, I guess you were referring to the outdated comment, removed now

lexnv (approved):

Issue #1292, opened 2024-02-12
tracking: release jsonrpsee 1.0


niklasad1:
Tracking for releasing jsonrpsee 1.0,

jsonrpsee has been quite stable now and it's time to release 1.0 to avoid breaking the APIs anymore

- [ ] https://github.com/paritytech/jsonrpsee/issues/1289
- [ ] https://github.com/paritytech/jsonrpsee/issues/1257
- [ ] https://github.com/paritytech/jsonrpsee/issues/1291
- [ ] https://github.com/paritytech/jsonrpsee/issues/1229
- [ ] https://github.com/paritytech/jsonrpsee/issues/1054

PR #1377, opened 2024-05-23
refactor(client): remove MaxSlots limit

### LINKED ISSUE #1374: Use async backpressure instead of MaxSlotsExceeded error
Currently library has an annoying behavior where exceeding max number of requests returns an error immediately.

I would have expected for it to be delayed until request can be made and user can cancel request before it goes out. Current behavior would be something like `try_request()` instead.

This can be achieved with a semaphore internally before trying to do anything.

Without this capability we are forced to add semaphore on application level even though we have already configured number of concurrent requests on the client level.

niklasad1:
Close #1374 

The rationale behind this is that the "ws-client/async-client" already has a bounded channel which provides backpressure and it doesn't make sense to use slot limit to throw an error if it's exceeded.

For the http client it's another story where it's handled by the hyper threadpool and we have no channel that provides backpressure.

However, it's better that users themselves decide how many concurrent calls that they want...

### DIFF
--- a/client/http-client/src/client.rs
+++ b/client/http-client/src/client.rs
@@ -104,12 +104,6 @@ impl<L> HttpClientBuilder<L> {
 		self
 	}
 
-	/// Set max concurrent requests.
-	pub fn max_concurrent_requests(mut self, max: usize) -> Self {
-		self.max_concurrent_requests = max;
-		self
-	}
-
 	/// Force to use the rustls native certificate store.
 	///
 	/// Since multiple certificate stores can be optionally enabled, this option will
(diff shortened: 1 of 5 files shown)


nazar-pc (approved):
This was quick, thanks!
Changes make sense to me, channel backpressure is exactly what I relied on and was surprised to get max slots error.

jsdw (approved):
LGTM!

lexnv (approved):

niklasad1:
Hey, we are [planning a new breaking release 0.23/1.0 soon hopefully next week](https://github.com/paritytech/jsonrpsee/issues/1292) but it has some big changes such as moving to hyper v1.0 and I need to do some manually testing on it....

PR #1380, opened 2024-05-28
feat: add macro attr `with_extensions`

niklasad1:
Follow-up on #1306 

Let's add on option to use `Extensions` in the API or not because it may be annoying to just mark it as used and would be nice not to break the previous API for existing users.

```rust
#[rpc(client, server)]
pub trait Rpc {
        // legacy
	#[method(name = "foo"])
	async fn async_method(&self, param_a: u8, param_b: String) -> u16>;

        // with extensions
	#[method(name = "with_ext", with_extensions)]
	async fn f(&self) -> bool;
}

impl RpcServer for () {
	async fn async_method(&self, param_a: u8, param_b: String) -> u16 {
            12
        }

	// NOTE: ext is injected here in the API
        async fn f(&self, ext: &Extensions: b: String) -> {
             ext.get::<u32>().is_ok()
        }
}
```

### DIFF
--- a/proc-macros/src/render_server.rs
+++ b/proc-macros/src/render_server.rs
@@ -66,10 +66,12 @@ impl RpcDescription {
 			let docs = &method.docs;
 			let mut method_sig = method.signature.clone();
 
-			let ext_ty = self.jrps_server_item(quote! { Extensions });
-			// Add `Extension` as the second parameter to the signature.
-			let ext: syn::FnArg = syn::parse_quote!(ext: &#ext_ty);
-			method_sig.sig.inputs.insert(1, ext);
+			if method.with_extensions {
+				let ext_ty = self.jrps_server_item(quote! { Extensions });
+				// Add `Extension` as the second parameter to the signature.
+				let ext: syn::FnArg = syn::parse_quote!(ext: &#ext_ty);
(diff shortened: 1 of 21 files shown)


jsdw (approved):
Looks great to me; the attribute works well :)

proc-macros/src/render_server.rs
```diff
@@ -200,20 +222,39 @@ impl RpcDescription {
...
+							})
+						})
+					}
 				} else {
-					handle_register_result(quote! {
-						rpc.register_subscription_raw(#rpc_sub_name, #rpc_notif_name, #rpc_unsub_name, |params, mut pending, context, ext| {
-							#parsing
-							let _ = context.as_ref().#rust_method_name(pending, &ext, #params_seq);
-							#sub_err::None
+					if sub.with_extensions {
```
pkhry, line 241: can you collapse the `if` into `else if` following the recommendation from clippy, otherwise LGTM

pkhry (approved):

mudlee:
Nice work! May I ask when it'll approximately be released? :)

niklasad1:
> Nice work! May I ask when it'll approximately be released? :)

We are "hoping" to get a new release tomorrow

PR #1381, opened 2024-05-29
feat: inject connection id in extensions

niklasad1:
Fixes https://github.com/paritytech/jsonrpsee/pull/1306#discussion_r1613301472

I introduced a wrapper type ConnectionId to avoid collisions in the extensions because those the builtin rustc types may used by users and they might unintentionally overwrite usize for some else....


lexnv (approved):
LGTM!

jsdw (approved):

PR #1382, opened 2024-05-29
server: add `serve` and `serve_with_graceful_shutdown` helpers

niklasad1:


pkhry (approved):

server/src/utils.rs
```diff
@@ -79,6 +83,60 @@ where
 	}
 }
 
+/// Serve a service over a TCP connection.
+pub async fn serve<S, B, I>(io: I, service: S) -> Result<(), BoxError>
```
lexnv, line 87: nit: I would mention here that graceful shutdowns are not handled, and users should look at `serve_with_graceful_shutdown` in those cases or similar

lexnv (approved):
LGMT!

jsdw (approved):
LGTM, nice to see some examples get simpler and less overall code!

PR #1384, opened 2024-05-30
client: set `authorization header` from the URL

### LINKED ISSUE #1371: Username and password in the url ignored
From my experiment and reading the code:

https://github.com/paritytech/jsonrpsee/blob/a4533966b997e83632509ad97eea010fc7c3efc0/client/transport/src/ws/mod.rs#L327

it seems that the username and passwords are simply ignored.

As a workaround, I am able to manually set `Authorization` header with `set_headers` and that works.

Thought it seems that wasm client does not support `set_headers`. Instead it seems to pass the url unmodified, so I suspect (haven't tested yet), the browser will handle `user:pass@` just fine, making it all rather inconsistent.

niklasad1:
Close #1371 

The rationale is to unify the behavior how cuRL, browsers and similar are handling username and passwords from URLs.

### curl
```
$ curl -sSG "http://user:pwd@localhost:9944"
2024-05-31 09:25:59.366 TRACE tokio-runtime-worker jsonrpsee-server: Request { method: GET, uri: /, version: HTTP/1.1, headers: {"host": "localhost:9944", "authorization": "Basic dXNlcjpwd2Q=", "user-agent": "curl/8.2.1", "accept": "*/*"}, body: Body(UnsyncBoxBody) }
```

### jsonrpsee ws client
```
$ jsonrpsee-ws
2024-05-31 09:28:32.016 TRACE tokio-runtime-worker jsonrpsee-server: Request { method: GET, uri: /, version: HTTP/1.1, headers: {"host": "localhost:9944", "upgrade": "websocket", "connection": "Upgrade", "sec-websocket-key": "amixRuJIg8sPqd0K2tNSTw==", "authorization": "Basic dXNlcjpwd2Q=", "sec-websocket-version": "13"}, body: Body(UnsyncBoxBody) }
```

### jsonrpsee http client
```
$ jsonrpsee-http 
2024-05-31 09:39:21.580 TRACE tokio-runtime-worker jsonrpsee-server: Request { method: POST, uri: /, version: HTTP/1.1, headers: {"content-type": "application/json", "accept": "application/json", "authorization": "Basic dXNlcjpwd2Q=", "host": "localhost:9944", "content-length": "62"}, body: Body(UnsyncBoxBody) }
```

### DIFF
--- a/client/http-client/src/transport.rs
+++ b/client/http-client/src/transport.rs
@@ -260,19 +260,16 @@ impl<L> HttpTransportClientBuilder<L> {
 			}
 		}
 
-		match url.password() {
-			Some(pwd) if !url.username().is_empty() => {
-				if !cached_headers.contains_key(hyper::header::AUTHORIZATION) {
-					let digest = base64::engine::general_purpose::STANDARD.encode(format!("{}:{pwd}", url.username()));
-					cached_headers.append(
-						hyper::header::AUTHORIZATION,
-						HeaderValue::from_str(&format!("Basic {digest}"))
-							.map_err(|_| Error::Url("Header value `authorization basic user:pwd` invalid".into()))?,
-					);
-				}
+		if let Some(pwd) = url.password() {
+			if !cached_headers.contains_key(hyper::header::AUTHORIZATION) {
+				let digest = base64::engine::general_purpose::STANDARD.encode(format!("{}:{pwd}", url.username()));
+				cached_headers.append(
+					hyper::header::AUTHORIZATION,
+					HeaderValue::from_str(&format!("Basic {digest}"))
+						.map_err(|_| Error::Url("Header value `authorization basic user:pwd` invalid".into()))?,
+				);
 			}
-			_ => (),
-		};
+		}
 
 		Ok(HttpTransportClient {
 			target: url.as_str().to_owned(),
--- a/client/http-client/src/transport.rs
+++ b/client/http-client/src/transport.rs
@@ -6,6 +6,7 @@
 // that we need to be guaranteed that hyper doesn't re-use an existing connection if we ever reset
 // the JSON-RPC request id to a value that might have already been used.
 
+use base64::Engine;
 use hyper::body::Bytes;
 use hyper::http::{HeaderMap, HeaderValue};
 use hyper_util::client::legacy::connect::HttpConnector;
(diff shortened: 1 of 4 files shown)


client/http-client/src/transport.rs
```diff
@@ -258,6 +260,20 @@ impl<L> HttpTransportClientBuilder<L> {
 			}
 		}
 
+		match url.password() {
+			Some(pwd) if !url.username().is_empty() => {
```
jsdw, line 264: Does CURL also not set anything if no username is given?

client/transport/src/ws/mod.rs
```diff
@@ -438,8 +439,23 @@ impl WsTransportClientBuilder {
...
+					value: basic_auth.as_bytes(),
+				});
+
+				it1.chain(it2).collect()
+			} else {
+				self.headers.iter().map(|(key, value)| Header { name: key.as_str(), value: value.as_bytes() }).collect()
+			}
+		} else {
+			self.headers.iter().map(|(key, value)| Header { name: key.as_str(), value: value.as_bytes() }).collect()
+		};
```
jsdw, line 457: Maybe could remove some dupe code like this:

```suggestion
let headers = match &target.basic_auth {
    Some(basic_auth) if if !self.headers.contains_key(http::header::AUTHORIZATION) => {
        // auth header stuff
    },
    _ => self.headers.iter().map(|(key, value)| Header { name: key.as_str(), value: value.as_bytes() }).collect()
};
```

jsdw (approved):

client/http-client/src/transport.rs
niklasad1: Tested right now, curl allow empty username let's do the same then

niklasad1: remove this check

client/http-client/src/transport.rs
```diff
@@ -258,6 +260,17 @@ impl<L> HttpTransportClientBuilder<L> {
 			}
 		}
 
+		if let Some(pwd) = url.password() {
+			if !cached_headers.contains_key(hyper::header::AUTHORIZATION) {
+				let digest = base64::engine::general_purpose::STANDARD.encode(format!("{}:{pwd}", url.username()));
+				cached_headers.append(
```
niklasad1, line 266:
```suggestion
				cached_headers.insert(
```

PR #1385, opened 2024-05-31
types: make jsonrpc field optional

### LINKED ISSUE #1142: Subscriptions just returns None (no error) - manual tokio_tungstenite works fine
I am trying to connect to a jsonrpc websocket with your library but it doesn't work.
The strange thing is that every other library including manual implementation with tokio-tokio_tungstenite works absolutely fine.
Even a simple wscat via terminal works without any problems.

from ```stream.next().await``` i always get just a ```None``` - there is no error.
Seems like the connection gets successfully established but then there is just nothing to receive. Even if I loop the stream.next().await for 1 minute i never receive anything.
Super confused cause its just with jsonrpsee which would totally fit for my use case.

Is anyone familiar with that problem or has an idea how to solve that?

```rust
    #[tokio::test]
    async fn test_jsonrpsee() {
        let mut headers = HeaderMap::new();
        let bx_auth_header = "<bloxroute-auth-token>";
        headers.insert("Authorization", bx_auth_header.parse().unwrap());

        let ws_client = WsClientBuilder::default().set_headers(headers);

        let client = ws_client
            .build("wss://api.blxrbdn.com:443/ws")
            .await
            .unwrap();
        dbg!(&client);

        // let bx_params = BxParams::default();
        // dbg!(bx_params.clone().to_rpc_params());
        let params = rpc_params!["newTxs", doc! {"include": vec!["tx_hash"]}];

        let mut stream: Subscription<String> = client
            .subscribe("subscribe", params, "unsubscribe")
            .await
            .unwrap();

        tokio::time::sleep(Duration::from_secs(1)).await;

        dbg!(&stream);
        dbg!(stream.next().await);
    }

    #[tokio::test]
    async fn test_manual_impl() {
        let request_text = r#"{
            "jsonrpc": "2.0",
            "id": 0,
            "method": "subscribe",
            "params": ["newTxs", {"include": ["tx_hash"]}]
        }"#;

        // Generate a random Sec-WebSocket-Key
        let key = base64::encode("secret-key");

        let request = Request::get("wss://api.blxrbdn.com:443/ws")
            .header("Host", "api.blxrbdn.com:443")
            .header("Authorization", "<bloxroute-auth-token>")
            .header("Sec-WebSocket-Key", key)
            .header("Connection", "Upgrade")
            .header("Upgrade", "websocket")
            .header("Sec-WebSocket-Version", "13")
            .body(())
            .unwrap();

        let (mut ws_stream, _) = connect_async(request).await.expect("Failed to connect");

        ws_stream
            .send(Message::Text(request_text.to_string()))
            .await
            .expect("Failed to send a message");

        dbg!(ws_stream.next().await);
    }
```

niklasad1:
Close #1142 

Previously, we made the interpretation that jsonrpc field must exist but recently https://github.com/paritytech/jsonrpsee/issues/1045 explained to me that it could be interpreted as optional.

So, let's change it to optional all over jsonrpsee.

I'm not 100% convinced by the interpretation of request object but it doesn't bother me to support it for stuff without the jsonrpc header...

### DIFF
--- a/types/src/request.rs
+++ b/types/src/request.rs
@@ -300,7 +300,7 @@ mod test {
 			// Without params field
 			(r#"{"jsonrpc":"2.0","id":1,"method":"subtract"}"#, Some(id), None, method, Some(TwoPointZero)),
 			// Without params and ID.
-			(r#"{"jsonrpc":"2.0","id":null,"method":"subtract"}"#, None, None, method, Some(TwoPointZero)),
+			(r#"{"id":null,"method":"subtract"}"#, None, None, method, None),
 		];
 
 		for (ser, id, params, method, version) in test_vector.iter().cloned() {
--- a/types/src/request.rs
+++ b/types/src/request.rs
@@ -42,7 +42,7 @@ use serde_json::value::RawValue;
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub struct Request<'a> {
 	/// JSON-RPC version.
-	pub jsonrpc: TwoPointZero,
+	pub jsonrpc: Option<TwoPointZero>,
 	/// Request ID
 	#[serde(borrow)]
 	pub id: Id<'a>,
@@ -60,7 +60,13 @@ pub struct Request<'a> {
 impl<'a> Request<'a> {
 	/// Create a new [`Request`].
 	pub fn new(method: Cow<'a, str>, params: Option<&'a RawValue>, id: Id<'a>) -> Self {
-		Self { jsonrpc: TwoPointZero, id, method, params: params.map(StdCow::Borrowed), extensions: Extensions::new() }
+		Self {
+			jsonrpc: Some(TwoPointZero),
+			id,
+			method,
+			params: params.map(StdCow::Borrowed),
+			extensions: Extensions::new(),
+		}
 	}
 
 	/// Get the ID of the request.
@@ -102,7 +108,7 @@ pub struct InvalidRequest<'a> {
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub struct Notification<'a, T> {
 	/// JSON-RPC version.
-	pub jsonrpc: TwoPointZero,
+	pub jsonrpc: Option<TwoPointZero>,
 	/// Name of the method to be invoked.
 	#[serde(borrow)]
 	pub method: Cow<'a, str>,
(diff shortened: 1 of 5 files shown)


types/src/request.rs
```diff
@@ -255,27 +278,34 @@ mod test {
...
+				params,
+				method,
+				Some(TwoPointZero),
+			),
 			// Without params field
-			(r#"{"jsonrpc":"2.0","id":1,"method":"subtract"}"#, Some(id), None, method),
+			(r#"{"jsonrpc":"2.0","id":1,"method":"subtract"}"#, Some(id), None, method, Some(TwoPointZero)),
 			// Without params and ID.
-			(r#"{"jsonrpc":"2.0","id":null,"method":"subtract"}"#, None, None, method),
+			(r#"{"jsonrpc":"2.0","id":null,"method":"subtract"}"#, None, None, method, Some(TwoPointZero)),
```
jsdw, line 303: Was the idea of giving the version too so that you'd add a test which didn't have the `"jsonrpc"` field?

niklasad1, line 303:
```suggestion
			(r#"{"id":null,"method":"subtract"}"#, None, None, method, None),
```

jsdw:
> I'm not 100% convinced by the interpretation of request object but it doesn't bother me to support it for stuff without the jsonrpc header...

Yeah; I wasn't 100% convinced (I mean, without the version, how do you know how to interpret a given request/response?).

The spec does also say this:

> JSON-RPC 2.0 Request objects and Response objects may not work with existing JSON-RPC 1.0 clients or servers. However, it is easy to distinguish between the two versions as 2.0 **always** has a member named "jsonrpc" with a String value of "2.0" whereas 1.0 does not. Most 2.0 implementations should consider trying to handle 1.0 objects, even if not the peer-to-peer and class hinting aspects of 1.0.

Emphasis mine

niklasad1:
> 5 Response object

> When a rpc call is made, the Server MUST reply with a Response, except for in the case of Notifications. The Response is > expressed as a single JSON Object, with the following members:

> jsonrpc
 >   A String specifying the version of the JSON-RPC protocol. MUST be exactly "2.0".

> result
 >   This member is REQUIRED on success.
>    This member MUST NOT exist if there was an error invoking the method.
>    The value of this member is determined by the method invoked on the Server.

> error
>    This member is REQUIRED on error.
>    This member MUST NOT exist if there was no error triggered during invocation.
>    The value for this member MUST be an Object as defined in section 5.1.

> id
>    This member is REQUIRED.
>    It MUST be the same as the value of the id member in the Request Object.
>    If there was an error in detecting the id in the Request object (e.g. Parse error/Invalid Request), it MUST be Null. 


The only way around that is that is doesn't specify "REQUIRED" for jsonrpc stuff but yeah I agree

jsdw:
> The only way around that is that is doesn't specify "REQUIRED" for jsonrpc stuff but yeah I agree

Yeah, I was initially agreeing with this PR on account of that,  but maybe the spec just isn't quite being clear, or it's allowing it to be optional for implementations to handle 1.0 too (which doesn't have that field at all)?

PR #1387, opened 2024-05-31
feat(core): replace callbacks in `RpcModule` and `Methods`

### LINKED ISSUE #1053: Support overriding existing Methods Callbacks

prestwich:
Allows users to remove or replace methods in the `Methods` and `RpcModule` collections

closes #1053 

this is effectively a successor to #1058, that follows the `HashMap` idiom of returning items when they are replaced. This addresses the [earlier PR review](https://github.com/paritytech/jsonrpsee/pull/1058/files#r1151498070) that it is possible to replace without realizing it, by making all replacements explicit. The caller can then decide whether to drop the replaced functions or not

Also includes `insert_or_replace_if` and `merge_replacing_if` as proposed in #1058 pr review

### DIFF
--- a/core/src/server/rpc_module.rs
+++ b/core/src/server/rpc_module.rs
@@ -1159,9 +1159,6 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 		R: IntoSubscriptionCloseResponse + Send,
 	{
 		let prev = self.methods.remove(subscribe_method_name);
-		// This method is auto-generated within the `register_subscription`
-		// method. As such we can safely remove it without returning it.
-		self.methods.remove(unsubscribe_method_name);
 
 		// Errors here can be ignored, as we know the method is not already
 		// registered (we just removed it).
--- a/core/src/server/rpc_module.rs
+++ b/core/src/server/rpc_module.rs
@@ -261,6 +261,9 @@ impl Methods {
 	/// to determine whether to replace the existing method.
 	///
 	/// Returns the previous method callback if it was replaced.
+	///
+	/// Note: If a conflict exists, and `cond` evaluates to false, the method
+	/// will not be replaced and the new method will be dropped.
 	pub fn insert_or_replace_if(
 		&mut self,
 		name: &'static str,
@@ -325,7 +328,14 @@ impl Methods {
 	/// Merge two [`Methods`]'s by adding all [`MethodCallback`]s from `other`
 	/// into `self`. If a method with the same name already exists, evaluates
 	/// the provided `cond` with the method name to determine whether to
-	/// replace the existing method. Returns a list of removed methods.
+	/// replace the existing method. Uses [`Self::insert_or_replace_if`]
+	/// internally.
+	///
+	/// Returns a list of removed methods.
+	///
+	/// Note: If a conflict exists, and `cond` evaluates to false, the method
+	/// will not be replaced and the new method will be dropped. Methods
+	/// dropped this way will **not** be included in the returned list.
 	pub fn merge_replacing_if(
 		&mut self,
 		other: impl Into<Methods>,
--- a/core/src/server/rpc_module.rs
+++ b/core/src/server/rpc_module.rs
@@ -980,6 +980,8 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 		R: IntoSubscriptionCloseResponse + Send,
 	{
 		let prev = self.methods.remove(subscribe_method_name);
+		// This method is auto-generated within the `register_subscription`
+		// method. As such we can safely remove it without returning it.
 		self.methods.remove(unsubscribe_method_name);
 
 		// Errors here can be ignored, as we know the method is not already
--- a/core/src/server/rpc_module.rs
+++ b/core/src/server/rpc_module.rs
@@ -250,10 +250,8 @@ impl Methods {
 
 	/// Inserts the method callback for a given name, replacing any existing
 	/// method with the same name.
-	pub fn insert_replacing(&mut self, name: &'static str, callback: MethodCallback) -> Option<MethodCallback> {
-		let prev = self.remove(name);
-		let _ = self.verify_and_insert(name, callback);
-		prev
+	pub fn insert(&mut self, name: &'static str, callback: MethodCallback) -> Option<MethodCallback> {
+		self.mut_callbacks().insert(name, callback)
 	}
 
 	/// Inserts the method callback for a given name. If a method with the same
@@ -264,7 +262,7 @@ impl Methods {
 	///
 	/// Note: If a conflict exists, and `cond` evaluates to false, the method
 	/// will not be replaced and the new method will be dropped.
-	pub fn insert_or_replace_if(
+	pub fn replace_if(
 		&mut self,
 		name: &'static str,
 		callback: MethodCallback,
@@ -328,7 +326,7 @@ impl Methods {
 	/// Merge two [`Methods`]'s by adding all [`MethodCallback`]s from `other`
 	/// into `self`. If a method with the same name already exists, evaluates
 	/// the provided `cond` with the method name to determine whether to
-	/// replace the existing method. Uses [`Self::insert_or_replace_if`]
+	/// replace the existing method. Uses [`Self::replace_if`]
 	/// internally.
 	///
 	/// Returns a list of removed methods.
@@ -345,7 +343,7 @@ impl Methods {
 
 		let mut removed = Vec::with_capacity(other.callbacks.len());
 		for (name, callback) in other.mut_callbacks().drain() {
-			if let Some(prev) = self.insert_or_replace_if(name, callback, &cond) {
+			if let Some(prev) = self.replace_if(name, callback, &cond) {
 				removed.push((name, prev));
 			}
 		}
--- a/core/src/server/rpc_module.rs
+++ b/core/src/server/rpc_module.rs
@@ -311,7 +311,7 @@ impl Methods {
 	/// Merge two [`Methods`]'s by adding all [`MethodCallback`]s from `other`
 	/// into `self`, removing and returning any existing methods with the same
 	/// name.
-	pub fn merge_replacing(&mut self, other: impl Into<Methods>) -> Vec<(&'static str, MethodCallback)> {
+	pub fn merge_replace(&mut self, other: impl Into<Methods>) -> Vec<(&'static str, MethodCallback)> {
 		let mut other = other.into();
 		let callbacks = self.mut_callbacks();
 
@@ -336,7 +336,7 @@ impl Methods {
 	/// Note: If a conflict exists, and `cond` evaluates to false, the method
 	/// will not be replaced and the new method will be dropped. Methods
 	/// dropped this way will **not** be included in the returned list.
-	pub fn merge_replacing_if(
+	pub fn merge_replace_if(
 		&mut self,
 		other: impl Into<Methods>,
 		cond: impl Fn(&'static str) -> bool,
--- a/core/src/server/rpc_module.rs
+++ b/core/src/server/rpc_module.rs
@@ -257,8 +257,8 @@ impl Methods {
 	}
 
 	/// Inserts the method callback for a given name. If a method with the same
-	/// name already exists, evaluates the provided closure to determine whether
-	/// to replace the existing method.
+	/// name already exists, evaluates the provided `cond` with the method name
+	/// to determine whether to replace the existing method.
 	///
 	/// Returns the previous method callback if it was replaced.
 	pub fn insert_or_replace_if(
@@ -324,8 +324,8 @@ impl Methods {
 
 	/// Merge two [`Methods`]'s by adding all [`MethodCallback`]s from `other`
 	/// into `self`. If a method with the same name already exists, evaluates
-	/// the provided `cond` to determine whether to replace the existing
-	/// method. Returns a list of removed methods.
+	/// the provided `cond` with the method name to determine whether to
+	/// replace the existing method. Returns a list of removed methods.
 	pub fn merge_replacing_if(
 		&mut self,
 		other: impl Into<Methods>,
--- a/core/src/server/rpc_module.rs
+++ b/core/src/server/rpc_module.rs
@@ -551,6 +551,16 @@ impl<Context> From<RpcModule<Context>> for Methods {
 }
 
 impl<Context: Send + Sync + 'static> RpcModule<Context> {
+	/// Return a mutable reference to the currently registered methods.
+	pub fn methods_mut(&mut self) -> &mut Methods {
+		&mut self.methods
+	}
+
+	/// Remove a method by name.
+	pub fn remove_method(&mut self, method_name: &'static str) -> Option<MethodCallback> {
+		self.methods.remove(method_name)
+	}
+
 	/// Register a new synchronous RPC method, which computes the response with the given callback.
 	///
 	/// ## Examples
@@ -588,7 +598,9 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 		R: IntoResponse + 'static,
 		F: Fn(Params, &Context, &Extensions) -> R + Send + Sync + 'static,
 	{
-		let prev = self.methods.remove(method_name);
+		let prev = self.remove_method(method_name);
+		// Errors here can be ignored, as we know the method is not already
+		// registered (we just removed it).
 		let _ = self.register_method(method_name, callback);
 		prev
 	}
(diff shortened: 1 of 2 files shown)


core/src/server/rpc_module.rs
```diff
@@ -266,6 +279,23 @@ impl Methods {
 		Ok(())
 	}
 
+	/// Merge two [`Methods`]'s by adding all [`MethodCallback`]s from `other`
+	/// into `self`, removing and returning any existing methods with the same
+	/// name.
+	pub fn merge_replacing(&mut self, other: impl Into<Methods>) -> Vec<(&'static str, MethodCallback)> {
```
niklasad1, line 314: As discussed in the predecessor PR, this API isn't great because it's hard to know which callback that was replaced.

For instance if one has several RpcModule's and one would want to merge it into a single RpcModule that actual merge order matters and you may end up with unexpected callback is overwritten.

```rust
fn default_setup () -> RpcModule {
   let apis = Vec<RpcModule>;
   let mut api = RpcModule::new(());
   for a in apis {
      api.merge_replace(a);
   }
}

```

I would personally prefer merge_if but I guess we could have both

core/src/server/rpc_module.rs
```diff
@@ -266,6 +308,50 @@ impl Methods {
...
```
niklasad1, line 314: rename `merge_replace` is sufficient

core/src/server/rpc_module.rs
prestwich: unlike the previous PR, this returns an explicit list of methods overridden, which is the rust idiom for maps like this

merge_replace_if is also included now :)

core/src/server/rpc_module.rs
```diff
@@ -243,6 +243,48 @@ impl Methods {
...
 	}
 
+	/// Remove a method by name.
+	pub fn remove(&mut self, name: &'static str) -> Option<MethodCallback> {
+		self.mut_callbacks().remove(name)
+	}
+
+	/// Inserts the method callback for a given name, replacing any existing
+	/// method with the same name.
+	pub fn insert_replacing(&mut self, name: &'static str, callback: MethodCallback) -> Option<MethodCallback> {
```
niklasad1, line 253: rename `insert` is sufficient IMO :)

core/src/server/rpc_module.rs
```diff
@@ -243,6 +243,48 @@ impl Methods {
...
+
+	/// Inserts the method callback for a given name. If a method with the same
+	/// name already exists, evaluates the provided `cond` with the method name
+	/// to determine whether to replace the existing method.
+	///
+	/// Returns the previous method callback if it was replaced.
+	///
+	/// Note: If a conflict exists, and `cond` evaluates to false, the method
+	/// will not be replaced and the new method will be dropped.
+	pub fn insert_or_replace_if(
```
niklasad1, line 267:
```suggestion
	pub replace_if(
```

core/src/server/rpc_module.rs
niklasad1, line 253:
```suggestion
	pub fn insert(&mut self, name: &'static str, callback: MethodCallback) -> Option<MethodCallback> {
```

prestwich:
pushing renamings as you suggest :)

core/src/server/rpc_module.rs
niklasad1: cool, sorry missed that

prestwich: no worries, i think i pushed it after you started the review

core/src/server/rpc_module.rs
```diff
@@ -820,6 +960,34 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
...
+		F: (Fn(Params<'static>, PendingSubscriptionSink, Arc<Context>, Extensions) -> Fut)
+			+ Send
+			+ Sync
+			+ Clone
+			+ 'static,
+		Fut: Future<Output = R> + Send + 'static,
+		R: IntoSubscriptionCloseResponse + Send,
+	{
+		let prev = self.methods.remove(subscribe_method_name);
+		self.methods.remove(unsubscribe_method_name);
```
niklasad1, line 983: You can add a comment that this callback is generated jsonrpsee and that's is why it's just removed and not returned here

niklasad1 (commented):
Looks clean.

We just need something better for `replace_method/remove` and similar APIs because it would potentially overwrite a subscription callback with arbitrary method callback and the unsubscription would be still be in the HashMap which isn't good.

Thus, we need to check the kind of the method when it's removed as well to do that correctly.

prestwich:
quick idea:
- Add `unsubscribe_method_name`  as a prop of `MethodCallback::Subscription` so that we can always identify the unsub method

prestwich:
on reflection, i think that this would have to be a more significant refactor in a separate PR. My proposed solution for now would be to note this possibility in the documentation of every merge/remove/replace/replace_if method

niklasad1:
> on reflection, i think that this would have to be a more significant refactor in a separate PR. My proposed solution for now would be to note this possibility in the documentation of every merge/remove/replace/replace_if method

I can't accept such code but lemme a have look how difficult it's to fix that.

prestwich:
Is the main blocker that it is possible to construct modules with dangling handlers _at all_? or that users may do that accidentally?

If the latter, how would you feel about `#[doc(hidden)] pub trait HazardousModuleMutations` to encapsulate dangerous behavior behind an explicit user opt-in?

prestwich:
> quick idea:
> 
> * Add `unsubscribe_method_name`  as a prop of `MethodCallback::Subscription` so that we can always identify the unsub method

I started on the above and realized that it would be broken by the aliasing system, as you could always alias an unsubscribe method, and leave it dangling regardless of any care in the remove method. This means that we have to track aliases separately. So to that end, I have done the more significant refactor I didn't want to do 🫡 

New behavior:
- `Subscribe` and `Unsubscribes` now store the name of their corresponding method
- when `remove()` removes a subscribe or an unsubscribe, it removes the corresponding method as well.
- `remove()` called on an unsubscribe returns the subscribe
- Aliases are now first-class citizens instead of cloned Arcs. This is necessary to ensure that aliases never allow dangling subscribe/unsubscribe entires

Aliases now work as follows
- `MethodCallback` can be downgraded to `WeakMethodCallback`
- the `MethodCallback::Alias` variant stores a `WeakMethodCallback`
- the `WeakMethodCallback` can be upgraded to a `MethodCallback`, provided the original method still exists (presumably owned by the map) 
- `method` and  `method_with_name` now resolve aliases by checking for `MethodCallback::Alias` and upgrading the result.
- Aliases with dangling targets are treated like empty map entries for most purposes

However, there are small breaking interface changes
- Changing contents of `MethodCallback::Subscription` and `MethodCallback::Unsubscription`
- Adding the `Alias` variant to the `MethodCallback` enum
- `method` and `method_with_name` now return `MethodCallback` instead of `&MethodCallback)

Followup questions:
- Should we opportunistically prune dangling aliases? Reaping them would only save a fat pointer and an indirection, so it seems fine to leave them in the map
- should aliases with dangling targets be replacable by `insert` and `register` functions? (i.e. should `verify_method_name` check if the map contains a dangling alias at that method name, and allow replacement if the alias is dangling)

niklasad1:
It's not only a dangling unsubscribe handler that is an issue but it's possible to i) overwrite a "unsubscribe callback" with "method callback" ii) overwrite a "method callback" with a "subscription without a unsubscribe callback"..... There may other edge-cases as well....

A subscription without a unsubscribe callback is most concerning to me because it may not be possible to unsubscribe then without terminating the connection.

So the type of "method callback" must be checked to do this correctly. 

```rust
	pub fn merge_replace(&mut self, other: impl Into<Methods>) -> Vec<(&'static str, MethodCallback)> {
		let mut other = other.into();
		let callbacks = self.mut_callbacks();

		let mut removed = Vec::with_capacity(other.callbacks.len());
		for (name, callback) in other.mut_callbacks().drain() {
			let new = callback.kind();

			if let Some(prev) = callbacks.insert(name, callback) {
				let old = prev.kind();

				match (old, new) {
					(MethodKind::Subscription, MethodKind::Subscription)
					| (MethodKind::MethodCall, MethodKind::MethodCall)
					| (MethodKind::Unsubscription, MethodKind::Unsubscription) => {
						removed.push((name, prev));
					}
					(MethodKind::Subscription, other) if other.is_method_call() => {
						todo!("remove unsubscribe handler")
					}
					(MethodKind::Unsubscription, other) if other.is_method_call() =>   {
						todo!("remove subscription handler")
					}
					(_other, MethodKind::Subscription) => {
						todo!("can't replace method call with subscription handler without unsubscribe handler");
					}
					_ => {
						panic!("Unsupported")
					}
				};
			};
		}
		removed
	}
```


However, I think it's not possible to handle it when a "method call" is replaced by a "subscription call" rather than throwing an error or panic:ing but that defeats the purpose of merge_replace stuff.

/cc @jsdw @lexnv Thoughts? I'm not sure about this API anymore and may be better to not support it anyway because it ended up to something way more complicated that I had in mind....

prestwich:
> It's not only a dangling unsubscribe handler that is an issue but it's possible to i) overwrite a "unsubscribe callback" with "method callback" ii) overwrite a "method callback" with a "subscription without a unsubscribe callback"..... There may other edge-cases as well....

> So the type of "method callback" must be checked to do this correctly.

Yes exactly. This is the problem the latest commit resolves. Please see how those are handled by the [remove function](https://github.com/paritytech/jsonrpsee/pull/1387/files#diff-dc533588250ea236f580d7fdaf87cdfce2ec80da031de9067b1088465a555998R398-R410). By ensuring that the `Subscription` and `Unsubscription` are removed only as a unit we ensure that there can be no dangling handlers. The danger of using `mut_callbacks` is called out in the [documentation on that function](https://github.com/paritytech/jsonrpsee/pull/1387/files#diff-dc533588250ea236f580d7fdaf87cdfce2ec80da031de9067b1088465a555998R327-R350), and all other functions have been rewritten to use `remove` (see logic of [`replace`](https://github.com/paritytech/jsonrpsee/pull/1387/files#diff-dc533588250ea236f580d7fdaf87cdfce2ec80da031de9067b1088465a555998R415), [`replace_if`](https://github.com/paritytech/jsonrpsee/pull/1387/files#diff-dc533588250ea236f580d7fdaf87cdfce2ec80da031de9067b1088465a555998R460), [`merge_replace`](https://github.com/paritytech/jsonrpsee/pull/1387/files#diff-dc533588250ea236f580d7fdaf87cdfce2ec80da031de9067b1088465a555998R493), etc)

core/src/server/rpc_module.rs
```diff
@@ -243,9 +439,26 @@ impl Methods {
...
+		name: &'static str,
+		callback: MethodCallback,
+		cond: impl Fn(&'static str) -> bool,
+	) -> Option<MethodCallback> {
+		if self.callbacks.contains_key(name) && !cond(name) {
+			return None;
+		}
+
+		self.remove(name);
+		self.replace(name, callback)
```
prestwich, line 461: the `remove` is redundant, as `remove` is called in replace

jsdw:
Niklas is away for a couple more weeks still, and so just as a heads up @prestwich, it's unlikely that much progress will happen on this PR until he's returned! When back and caught up we'll be able to give better feedback and progress this :)

prestwich:
cool, any updates on when he'll be back?

niklasad1:
Hey, I'm back but I'll take a look at this again next week.

Sorry for slow response here

niklasad1:
Hey again,

I haven't changed my mind regarding the "dangling handler" this because I regard the `remove API` explicit which applies per RpcModule and states that it's the user's responsibility to remove handlers if it's a subscription but yeah it's a maybe a gray area and perhaps my interpretation is wrong.  

Anyway, this PR will allow merge any number RpcModule's "to be merged" which I believe is a footgun and much harder to know whether something was leaked. As we have been discussing earlier.

This was not really obvious when making this PR and I think the overall good quality of your PR was good.
I'll take some time review this thoroughly, thanks for your patience

prestwich:
> Anyway, this PR will allow merge any number RpcModule's "to be merged" which I believe is a footgun and much harder to know whether something was leaked as we have been discussing earlier.

as noted above, in the current PR it is impossible to "leak" anything at any point, as the data structure no longer permits dangling handlers at all

core/src/server/rpc_module.rs
```diff
@@ -521,6 +783,16 @@ impl<Context> From<RpcModule<Context>> for Methods {
 }
 
 impl<Context: Send + Sync + 'static> RpcModule<Context> {
+	/// Return a mutable reference to the currently registered methods.
+	pub fn methods_mut(&mut self) -> &mut Methods {
```
niklasad1, line 787: We don't want to expose this, can you remove it or make it private?

It's doesn't seem to used.

core/src/server/rpc_module.rs
```diff
@@ -221,6 +322,101 @@ impl Methods {
...
+	}
+
+	/// Register a new alias for a target method, or returns an error if the
+	/// alias name was already taken. If the target method does not exist, the
+	/// alias will not be registered.
+	pub fn register_alias(
+		&mut self,
+		alias: &'static str,
+		target_name: &'static str,
+	) -> Result<(), RegisterMethodError> {
```
niklasad1, line 387: We returned an error if the target_name wasn't not registered before and I would like to keep it that way.

Why did you change that (maybe I'm missing something)?

core/src/server/rpc_module.rs
```diff
@@ -142,6 +157,52 @@ impl CallOrSubscription {
...
 }
 
+/// The [`Weak`] equivalent of [`MethodCallback`]. This allows the [`Methods`]
+/// map to store aliases to the method callbacks and ensures that any dangling
+/// alias references are safely handled.
+///
+/// This type is an implementation detail of the method aliasing system and
+/// should not be used directly.
+#[derive(Clone, Debug)]
+pub enum WeakMethodCallback<'a> {
```
niklasad1, line 167: dq: do we really need this Weak stuff if remove/merge actually takes of the subscription handlers?

core/src/server/rpc_module.rs
```diff
@@ -150,9 +211,48 @@ pub enum MethodCallback {
...
-	Unsubscription(UnsubscriptionMethod),
+	Unsubscription {
+		/// The unsubscription method handler, usually auto-generated on
+		/// subscription insertion.
+		method: UnsubscriptionMethod,
+		/// The name of the associated subscription method.
+		subscribe_method_name: &'static str,
+	},
+	/// The method is an alias for another method.
+	Alias(WeakMethodCallback<'static>),
```
niklasad1, line 229: How is this intended to work?
If one enables an alias for method a and method a is replaced by another callback?
Is the alias just disabled i.e, dangling/weak or is supposed to point the new callback?

core/src/server/rpc_module.rs
```diff
@@ -221,6 +322,101 @@ impl Methods {
...
+	pub fn method_names(&self) -> impl Iterator<Item = &'static str> + '_ {
+		self.callbacks.keys().copied()
+	}
+
+	/// Returns the method callback along with its name. The returned name is same as the
+	/// `method_name`, but its lifetime bound is `'static`.
+	pub fn method_with_name(&self, method_name: &str) -> Option<(&'static str, MethodCallback)> {
+		self.callbacks.get_key_value(method_name).and_then(|(k, v)| {
+			let v = match v {
+				MethodCallback::Alias(alias) => alias.upgrade()?,
```
niklasad1, line 373: Ok, it's just using the inner callback as an alias then.

I think it's better to remove MethodCallback::Alias then it just adds a bunch of confusion when matching on it with unreachable.

niklasad1 (commented):
Looks overall reasonable and I left some comments in the code.

In order for me to happy with this I would like to have some tests for this such as such as merge two rpc modules with conflicting subscription and methods works as intended.

Then it looks to me like aliases won't work when after a handler with aliases is replaced, so I want additional info how it's intended to work.

core/src/server/rpc_module.rs
prestwich: > If one enables an alias for method a and method a is replaced by another callback?

like the current aliasing system, the alias points to the underlying callback, not to the aliased name. However, unlike the current aliasing system, the alias no longer keeps the callback "alive" like the canonical name does. Removing the canonical name will remove the callback from the map and the alias's will no longer work

Using a weak is important to preventing dangling handlers. If we allow the alias to keep the callback alive, then we can create dangling handlers by aliasing subscribe functions before removing them

PR #1389, opened 2024-06-04
server: pass `extensions` from http layer

### LINKED ISSUE #1388: Inheriting HTTP Header Option to RPC middlewares
With the recent merge of [PR #1305](https://github.com/paritytech/jsonrpsee/issues/1305), different layers of rpc middleware can now share data using Extensions.

Leveraging this functionality, we plan to set HTTP header data within the Request extensions of an RPC middleware call, enabling subsequent middleware layers to process the header information effectively.

In our use case, we aim to log each RPC method call, associating specific HTTP header values with each entry. This includes logging individual lines for each call and multiple lines for batch calls. Additionally, this approach facilitates other scenarios, such as HTTP header validation on a per-RPC method basis.

This can be achieved by passing header information to this [function](https://github.com/paritytech/jsonrpsee/blob/a3307b8c877ae3cf285842cfea10e6bbff2d1d2a/server/src/server.rs#L1285) and setting them to Extensions only when the option is enabled.

---

Using HTTP middleware is one possible approach. However, for websocket requests, the RPC middleware needs to propagate HTTP header as its context data to achieve similar functionality.

niklasad1:
Close #1388 

This PR changes that we pass the `Extensions` from the origin HTTP request which means that is seen every RpcMiddleware layer. This causes us to inject connectionID in the origin HTTP request to be seen by all layers.

Recall, that our RpcService is the lowest layer in the RpcMiddleware call tree and it doesn't work to inject stuff because it will not be seen by other middleware layers.

An annoyance with this is that we to clone the Extensions for every JSON-RPC call which includes `ConnectionId` and `hyper::upgrade::OnUpgrade` which roughly is 16 bytes of overhead but I'm okay with that.

### DIFF
--- a/core/src/server/method_response.rs
+++ b/core/src/server/method_response.rs
@@ -147,19 +147,6 @@ impl MethodResponse {
 	where
 		T: Serialize + Clone,
 	{
-		Self::response_with_extensions(id, rp, max_response_size, Extensions::new())
-	}
-
-	/// Similar to [`MethodResponse::response`] but with extensions.
-	pub fn response_with_extensions<T>(
-		id: Id,
-		rp: ResponsePayload<T>,
-		max_response_size: usize,
-		extensions: Extensions,
-	) -> Self
-	where
-		T: Serialize + Clone,
-	{
 		let mut writer = BoundedWriter::new(max_response_size);
 
 		let success_or_error = if let InnerResponsePayload::Error(ref e) = rp.inner {
@@ -209,7 +196,7 @@ impl MethodResponse {
 						success_or_error: MethodResponseResult::Failed(err.code()),
 						kind,
 						on_close: rp.on_exit,
-						extensions,
+						extensions: Extensions::new(),
 					}
 				}
 			}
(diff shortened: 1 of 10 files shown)


lexnv (approved):
LGTM!

We have a few clippy warnings here and there. Would be good to have a look over them and see if they aren't too opinionated so we could fix them :D

jsdw (approved):
I wonder what the cost to clone is but otherwise lgtm :)

niklasad1:
> I wonder what the cost to clone is but otherwise lgtm :)

It depends really what one does it the middleware implementation but we have two extensions by default `OnUpgrade` and `ConnectionId` in the server which is cheap however if one writes some middleware that injects something big like an Vec or something it could be expensive.

Actually, I tested by wrapping the Extension in an Arc but ended to have to `Arc::make_mut` when the Extension needs be modified but it ends up using similar memory because make_mut will clone it because strong_count > 1. Lemme check if I could make it a weak ref instead.

niklasad1:
@CodiumAI-Agent /review

PR #1392, opened 2024-06-06
server: use `ConnectionId` in subscription APIs

niklasad1:
The reason is to unify the APIs such that we were using `usize` as connectionID in the subscriptions APIs and `ConnectionId` in the extensions which was awkward. 

Let's unify them and break the API compared to v0.22


lexnv (approved):
Nice!

PR #1396, opened 2024-06-06
chore: release v0.23

niklasad1:
## [v0.23.0] - 2024-06-07

This is a new breaking release, and let's go through the changes.

### hyper v1.0

jsonrpsee has been upgraded to use hyper v1.0 and this mainly impacts users that are using
the low-level API and rely on the `hyper::service::make_service_fn`
which has been removed, and from now on you need to manage the socket yourself.

The `hyper::service::make_service_fn` can be replaced by the following example template:

```rust
async fn start_server() {
  let listener = tokio::net::TcpListener::bind("127.0.0.1:0").await.unwrap();

  loop {
    let sock = tokio::select! {
    res = listener.accept() => {
        match res {
          Ok((stream, _remote_addr)) => stream,
          Err(e) => {
            tracing::error!("failed to accept v4 connection: {:?}", e);
            continue;
          }
        }
      }
      _ = per_conn.stop_handle.clone().shutdown() => break,
    };

    let svc = tower::service_fn(move |req: hyper::Request<hyper::body::Incoming>| {
      let mut jsonrpsee_svc = svc_builder
        .set_rpc_middleware(rpc_middleware)
        .build(methods, stop_handle);
      // https://github.com/rust-lang/rust/issues/102211 the error type can't be inferred
      // to be `Box<dyn std::error::Error + Send + Sync>` so we need to convert it to a concrete type
      // as workaround.
      jsonrpsee_svc
        .call(req)
        .await
        .map_err(|e| anyhow::anyhow!("{:?}", e))
    });

    tokio::spawn(jsonrpsee::server::serve_with_graceful_shutdown(
      sock,
      svc,
      stop_handle.clone().shutdown(),
    ));
  }
}
```

Also, be aware that `tower::service_fn` and `hyper::service::service_fn` are different and it's recommended to use `tower::service_fn` from now.

### Extensions

Because it was not possible/easy to share state between RPC middleware layers
jsonrpsee has added `Extensions` to the Request and Response.
To allow users to inject arbitrary data that can be accessed in the RPC middleware
and RPC handlers.

Please be careful when injecting large amounts of data into the extensions because
It's cloned for each RPC call, which can increase memory usage significantly.

The connection ID from the jsonrpsee-server is injected in the extensions by default.
and it is possible to fetch it as follows:

```rust
struct LogConnectionId<S>(S);

impl<'a, S: RpcServiceT<'a>> RpcServiceT<'a> for LogConnectionId<S> {
	type Future = S::Future;

	fn call(&self, request: jsonrpsee::types::Request<'a>) -> Self::Future {
		let conn_id = request.extensions().get::<ConnectionId>().unwrap();
		tracing::info!("Connection ID {}", conn_id.0);

		self.0.call(request)
	}
}
```

In addition the `Extensions` is not added in the proc-macro API by default and
one has to enable `with_extensions` attr for that to be available:

```rust
#[rpc(client, server)]
pub trait Rpc {
	// legacy
	#[method(name = "foo"])
	async fn async_method(&self) -> u16>;

	// with extensions
	#[method(name = "with_ext", with_extensions)]
	async fn f(&self) -> bool;
}

impl RpcServer for () {
	async fn async_method(&self) -> u16 {
		12
	}

	// NOTE: ext is injected just after self in the API
	async fn f(&self, ext: &Extensions: b: String) -> {
		ext.get::<u32>().is_ok()
	}
}
```

### client - TLS certificate store changed

The default TLS certificate store has been changed to
`rustls-platform-verifier` to decide the best certificate
store for each platform.

In addition it's now possible to inject a custom certificate store
if one wants need some special certificate store.

### client - Subscription API modified

The subscription API has been modified:
- The error type has been changed to `serde_json::Error`
  to indicate that error can only occur if the decoding of T fails.
- It has been some confusion when the subscription is closed which can occur if the client "lagged" or the connection is closed.
  Now it's possible to call `Subscription::close_reason` after the subscription closed (i.e. has return None) to know why.

If one wants to replace old messages in case of lagging it is recommended to write your own adaptor on top of the subscription:

```rust
fn drop_oldest_when_lagging<T: Clone + DeserializeOwned + Send + Sync + 'static>(
    mut sub: Subscription<T>,
    buffer_size: usize,
) -> impl Stream<Item = Result<T, BroadcastStreamRecvError>> {
    let (tx, rx) = tokio::sync::broadcast::channel(buffer_size);

    tokio::spawn(async move {
        // Poll the subscription which ignores errors
        while let Some(n) = sub.next().await {
            let msg = match n {
                Ok(msg) => msg,
                Err(e) => {
                    tracing::error!("Failed to decode the subscription message: {e}");
                    continue;
                }
            };

            // Only fails if the receiver has been dropped
            if tx.send(msg).is_err() {
                return;
            }
        }
    });

    BroadcastStream::new(rx)
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -106,7 +106,7 @@ impl RpcServer for () {
 		12
 	}
 
-	// NOTE: ext is injected here in the API
+	// NOTE: ext is injected just after self in the API
 	async fn f(&self, ext: &Extensions: b: String) -> {
 		ext.get::<u32>().is_ok()
 	}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,7 +56,7 @@ async fn start_server() {
 }
 ```
 
-Also, beware that `tower::service_fn` and `hyper::service::service_fn` are different and it's recommended to use `tower::service_fn` from now.
+Also, be aware that `tower::service_fn` and `hyper::service::service_fn` are different and it's recommended to use `tower::service_fn` from now.
 
 ### Extensions
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -128,7 +128,38 @@ The subscription API has been modified:
 - The error type has been changed to `serde_json::Error`
   to indicate that error can only occur if the decoding of T fails.
 - It has been some confusion when the subscription is closed which can occur if the client "lagged" or the connection is closed.
-  Now it's possible to call `Subscription::close_reason` after the subscription closed to know why.
+  Now it's possible to call `Subscription::close_reason` after the subscription closed (i.e. has return None) to know why.
+
+If one wants to replace old messages in case of lagging it is recommended to write your own adaptor on top of the subscription:
+
+```rust
+fn drop_oldest_when_lagging<T: Clone + DeserializeOwned + Send + Sync + 'static>(
+    mut sub: Subscription<T>,
+    buffer_size: usize,
+) -> impl Stream<Item = Result<T, BroadcastStreamRecvError>> {
+    let (tx, rx) = tokio::sync::broadcast::channel(buffer_size);
+
+    tokio::spawn(async move {
+        // Poll the subscription which ignores errors
+        while let Some(n) = sub.next().await {
+            let msg = match n {
+                Ok(msg) => msg,
+                Err(e) => {
+                    tracing::error!("Failed to decode the subscription message: {e}");
+                    continue;
+                }
+            };
+
+            // Only fails if the receiver has been dropped
+            if tx.send(msg).is_err() {
+                return;
+            }
+        }
+    });
+
+    BroadcastStream::new(rx)
+}
+```
 
 ### [Added]
 - server: add `serve` and `serve_with_graceful_shutdown` helpers ([#1382](https://github.com/paritytech/jsonrpsee/pull/1382))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ This is a new breaking release, and let's go through the changes.
 
 jsonrpsee has been upgraded to use hyper v1.0 and this mainly impacts users that are using
 the low-level API and rely on the `hyper::service::make_service_fn`
-which has been removed, and for now on you need to manage the socket yourself.
+which has been removed, and from now on you need to manage the socket yourself.
 
 The `hyper::service::make_service_fn` can be replaced by the following example template:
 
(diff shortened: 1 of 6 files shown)


pkhry (approved):
lgtm

CHANGELOG.md
```diff
@@ -4,6 +4,192 @@ The format is based on [Keep a Changelog].
...
+    ));
+  }
+}
+```
+
+Also, beware that `tower::service_fn` and `hyper::service::service_fn` are different and it's recommended to use `tower::service_fn` from now.
+
+### Extensions
+
+Because it was not possible or easy to state between RPC middleware layers
```
niklasad1, line 63:
```suggestion
Because it was not possible or easy to share state between RPC middleware layers
```

CHANGELOG.md
```diff
@@ -4,6 +4,192 @@ The format is based on [Keep a Changelog].
...
 
+## [v0.23.0] - 2024-06-06
+
+This is a new breaking release, and let's go through the changes.
+
+### hyper v1.0
+
+jsonrpsee has been upgraded to use hyper v1.0 and this mainly impacts users that are using
+the low-level API and rely on the `hyper::service::make_service_fn`
+which has been removed, and for now on you need to manage the socket yourself.
```
jsdw, line 15:
```suggestion
which has been removed, and from now on you need to manage the socket yourself.
```

CHANGELOG.md
```diff
@@ -4,6 +4,192 @@ The format is based on [Keep a Changelog].
...
+    tokio::spawn(jsonrpsee::server::serve_with_graceful_shutdown(
+      sock,
+      svc,
+      stop_handle.clone().shutdown(),
+    ));
+  }
+}
+```
+
+Also, beware that `tower::service_fn` and `hyper::service::service_fn` are different and it's recommended to use `tower::service_fn` from now.
```
jsdw, line 59:
```suggestion
Also, be aware that `tower::service_fn` and `hyper::service::service_fn` are different and it's recommended to use `tower::service_fn` from now.
```

CHANGELOG.md
```diff
@@ -4,6 +4,192 @@ The format is based on [Keep a Changelog].
...
+	#[method(name = "with_ext", with_extensions)]
+	async fn f(&self) -> bool;
+}
+
+impl RpcServer for () {
+	async fn async_method(&self) -> u16 {
+		12
+	}
+
+	// NOTE: ext is injected here in the API
```
jsdw, line 109:
```suggestion
	// NOTE: ext is injected just after self in the API
```

jsdw (approved):
Nice one!

PR #1412, opened 2024-06-18
fix: remove sockaddrs field from `Target`

### LINKED ISSUE #1411: bug(transport): it should not try to resolve sockaddrs when using `build_with_stream`
I've been using and testing the `build_with_stream` method added in https://github.com/paritytech/jsonrpsee/pull/1168, but as `Target` been used in both [`build_with_stream`](https://github.com/paritytech/jsonrpsee/blob/47f6b173452785d6f926284207b2d32df461b30b/client/transport/src/ws/mod.rs#L311-L322) and [`try_connect`](https://github.com/paritytech/jsonrpsee/blob/47f6b173452785d6f926284207b2d32df461b30b/client/transport/src/ws/mod.rs#L429-L485) it's resolving the sockaddrs when creating `Target` from the given `Url`, and leading to two problems:
- it's possibly leaking data as it resolves through a standard connection and not through the preferred data stream passed as a parameter.
- in connections, to `.onion` for example, it leads to an error because it tries to resolve it to sockaddrs which is expected to not be possible, and it's one of the reasons to use the specified data stream, in this case for example `arti_client::DataStream`.

I do see two possible fixes for the problem:
- remove the `sockaddrs` resolution from the `Target` data structure, and do it only on `try_connect_over_tcp`, which is the only place it's being used.
- receive the `Target` as a parameter, with the `sockaddrs` field being an `Option<>` leading to the user to possibly do the resolution when needed. It would also require upstreaming the parameter to ws-client methods as well.

oleonardolima:
fixes #1411

### DIFF
--- a/client/transport/src/ws/mod.rs
+++ b/client/transport/src/ws/mod.rs
@@ -326,7 +326,7 @@ impl WsTransportClientBuilder {
 		&self,
 		uri: Url,
 	) -> Result<(Sender<Compat<EitherStream>>, Receiver<Compat<EitherStream>>), WsHandshakeError> {
-		let mut target: Target = uri.try_into()?;
+		let mut target: Target = uri.clone().try_into()?;
 		let mut err = None;
 
 		// Only build TLS connector if `wss` in URL.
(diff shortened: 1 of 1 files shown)


client/transport/src/ws/mod.rs
```diff
@@ -340,7 +340,8 @@ impl WsTransportClientBuilder {
 			tracing::debug!(target: LOG_TARGET, "Connecting to target: {:?}", target);
```
oleonardolima, line 340: The `sockaddrs` were being logged as part of `Target`, should we keep it in the logs ?

niklasad1: It's quite noisy and the hostname is probably sufficient.

To debug such things i.e, that we may resolve the "wrong host" is harder but I'm fine omitting that.

niklasad1 (approved):
LGTM

client/transport/src/ws/mod.rs
oleonardolima: Sure, that makes sense.

oleonardolima:
> LGTM

I was thinking about how to prevent regressions on this (as it could "leak" connection data), and I'll try to create a test for it in a new commit, then it should be ready to go.

niklasad1:
@oleonardolima are you happy with this PR without the extra tests you mentioned?

oleonardolima:
> @oleonardolima are you happy with this PR without the extra tests you mentioned?

Yes, I can add it later in another PR.

PR #1416, opened 2024-06-21
add `RpcModule::remove`

polachok:
Currently it's not possible to remove or replace a method. This is useful for overriding methods from an RpcModule with manual implementation or another module.

### DIFF
--- a/core/src/server/rpc_module.rs
+++ b/core/src/server/rpc_module.rs
@@ -243,11 +243,6 @@ impl Methods {
 		}
 	}
 
-	// Removes the method callback for a given name, if it exists.
-	fn remove(&mut self, name: &'static str) -> Option<MethodCallback> {
-		self.mut_callbacks().remove(name)
-	}
-
 	/// Helper for obtaining a mut ref to the callbacks HashMap.
 	fn mut_callbacks(&mut self) -> &mut FxHashMap<&'static str, MethodCallback> {
 		Arc::make_mut(&mut self.callbacks)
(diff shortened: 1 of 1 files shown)


niklasad1 (approved):

core/src/server/rpc_module.rs
```diff
@@ -552,6 +557,11 @@ impl<Context: Send + Sync + 'static> RpcModule<Context> {
 		)
 	}
 
+	/// Removes method if it exists.
+	pub fn remove_method(&mut self, method_name: &'static str) -> Option<MethodCallback> {
```
niklasad1, line 561: I would want add something to actually know whether it's a method call or subscription that was removed.

If call was a subscription the `unsubscribe method` should be removed well because it will be "dangling"... but that is not easy to implement currently.

Thus, I think the way forward is to clearly document that it's the caller's responsible to remove both for methods subscriptions i.e, both the "subscribe method" and "unsubscribe method" which is acceptable to me.

core/src/server/rpc_module.rs
```diff
@@ -243,6 +243,11 @@ impl Methods {
 		}
 	}
 
+	// Removes the method callback for a given name, if it exists.
+	fn remove(&mut self, name: &'static str) -> Option<MethodCallback> {
+		self.mut_callbacks().remove(name)
+	}
```
jsdw, line 249: Why not just put this single line into the `remove_method` call that calls it?

jsdw (approved):

PR #1419, opened 2024-06-28
refactor(client+transport)!: split `tls` into `tls` and`tls-rustls-platform-verifier` features

### LINKED ISSUE #1407: `0.23.0` `tls` feature relies on `rustls-platform-verifier`, bringing other unwanted dependencies
In `0.23.0` the new `tls` feature brings all: `["tokio-rustls", "rustls-pki-types", "rustls-platform-verifier", "rustls"]`, which ends up bringing `rustls-native-certs`, and then `openssl-probe`. Even when using `CertificateStore::Custom` it's still requires `tls` feature, which will require `rustls-platform-verifier`.
 
Please let me know if missed something, but is there a way to still use `tls` without depending completely on `rustls-platform-verifier` ?

oleonardolima:
fixes  #1407

### DIFF
--- a/client/transport/Cargo.toml
+++ b/client/transport/Cargo.toml
@@ -44,7 +44,6 @@ futures-channel = { version = "0.3.14", default-features = false, optional = tru
 [features]
 tls = ["rustls", "tokio-rustls", "rustls-pki-types"]
 tls-rustls-platform-verifier = ["tls", "rustls-platform-verifier"]
-default = ["tls-rustls-platform-verifier"]
 
 ws = [
     "base64",
--- a/client/transport/Cargo.toml
+++ b/client/transport/Cargo.toml
@@ -42,7 +42,9 @@ gloo-net = { version = "0.5.0", default-features = false, features = ["json", "w
 futures-channel = { version = "0.3.14", default-features = false, optional = true }
 
 [features]
-tls = ["tokio-rustls", "rustls-pki-types", "rustls-platform-verifier", "rustls"]
+tls = ["rustls", "tokio-rustls", "rustls-pki-types"]
+tls-rustls-platform-verifier = ["tls", "rustls-platform-verifier"]
+default = ["tls-rustls-platform-verifier"]
 
 ws = [
     "base64",
--- a/client/transport/src/ws/mod.rs
+++ b/client/transport/src/ws/mod.rs
@@ -61,7 +61,6 @@ pub type CustomCertStore = rustls::ClientConfig;
 #[cfg(feature = "tls")]
 #[derive(Debug, Clone)]
 pub enum CertificateStore {
-	#[cfg(feature = "tls-rustls-platform-verifier")]
 	/// Native.
 	Native,
 	/// Custom certificate store.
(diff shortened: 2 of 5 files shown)


tvolk131:
Thanks for pushing this through Leonardo!

oleonardolima:
I guess it will need another feature `tls-custom`, in order to properly handle the `Default` trait for both WsTranportBuilder and WsClientBuilder

client/transport/src/ws/mod.rs
```diff
@@ -614,7 +614,15 @@ impl TryFrom<url::Url> for Target {
 #[cfg(feature = "tls")]
 fn build_tls_config(cert_store: &CertificateStore) -> Result<tokio_rustls::TlsConnector, WsHandshakeError> {
 	let config = match cert_store {
+		#[cfg(feature = "tls-rustls-platform-verifier")]
 		CertificateStore::Native => rustls_platform_verifier::tls_config(),
+		#[cfg(not(feature = "tls-rustls-platform-verifier"))]
+		CertificateStore::Native => {
```
niklasad1, line 620: I don't think it's necessary to add such complicated feature-gated stuff for such a niche feature.

Not super elegant to throw an error here with "the wrong configuration" when it could be a compile error but I'm okay with this.

client/transport/Cargo.toml
```diff
@@ -42,7 +42,9 @@ gloo-net = { version = "0.5.0", default-features = false, features = ["json", "w
 futures-channel = { version = "0.3.14", default-features = false, optional = true }
 
 [features]
-tls = ["tokio-rustls", "rustls-pki-types", "rustls-platform-verifier", "rustls"]
+tls = ["rustls", "tokio-rustls", "rustls-pki-types"]
+tls-rustls-platform-verifier = ["tls", "rustls-platform-verifier"]
+default = []
```
niklasad1, line 47:
```suggestion
```

niklasad1 (approved):

client/transport/src/ws/mod.rs
niklasad1: are you okay with my changes @oleonardolima?

oleonardolima: As we removed the `Default` implementation now I think it's fine.

client/transport/Cargo.toml
```diff
@@ -42,7 +42,9 @@ gloo-net = { version = "0.5.0", default-features = false, features = ["json", "w
 futures-channel = { version = "0.3.14", default-features = false, optional = true }
 
 [features]
-tls = ["tokio-rustls", "rustls-pki-types", "rustls-platform-verifier", "rustls"]
+tls = ["rustls", "tokio-rustls", "rustls-pki-types"]
+tls-rustls-platform-verifier = ["tls", "rustls-platform-verifier"]
+default = ["tls-rustls-platform-verifier"]
```
niklasad1, line 47: remove

oleonardolima: Oh, I made some mistake while rebasing for local tests, added your suggestions again.

lexnv (approved):
LGTM! Thanks for contributing! 🙏

oleonardolima:
@niklasad1 just to have an idea, when are you planning to cut a new release/patch?

niklasad1:
> @niklasad1 just to have an idea, when are you planning to cut a new release/patch?

Good question, I don't think that we can't make a patch release unfortunately because code with the explicit  `--feature tls` will break existing code but I haven't planned another release "soon" and I would like to fix a few more things before releasing v1.0 (I'll be AFK for a few weeks).

Do you need another release soon? We could also release v0.24, what do you say @lexnv?

oleonardolima:
Thanks! Yes, a new release soon would be helpful, as it would allow the usage of onion addresses on our side.

PR #1436, opened 2024-08-01
fix(client): support notifications with no params

### LINKED ISSUE #1435: Unparseable notification
I am getting this message, which I _do not_ understand.
```log
Unparseable message: {"jsonrpc":"2.0","method":"notify_klippy_disconnected"}
```
Why is this not parseable ? It is json and it has both the `jsonrpc` field and the `method`?
There is a few like this, but also a bunch that are working.

```toml
jsonrpsee = { version = "0.24.1", features = [
  "tokio",
  "ws-client",
  "jsonrpsee-core",
] }
```

niklasad1:
Close #1435

Technically notifications can omit the params field according to JSON-RPC spec which this fixes.
However, it's quite weird to emit notifications without data in them in such scenarios we just push `null` to the notification handler.

In contrary, subscriptions requires params field where the subscription ID is sent in every message and that's why I didn't change the notification type i.e, just a type alias `Notification<'a, Option<serde_json::Value>>`

### DIFF
--- a/client/ws-client/src/tests.rs
+++ b/client/ws-client/src/tests.rs
@@ -191,6 +191,26 @@ async fn notification_handler_works() {
 }
 
 #[tokio::test]
+async fn notification_no_params() {
+	let server = WebSocketTestServer::with_hardcoded_notification(
+		"127.0.0.1:0".parse().unwrap(),
+		server_notification_without_params("no_params"),
+	)
+	.with_default_timeout()
+	.await
+	.unwrap();
+
+	let uri = to_ws_uri_string(server.local_addr());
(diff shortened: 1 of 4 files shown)


Jomik (approved):
This looks great.
Use case for me is listening to state on my 3d printer.
https://moonraker.readthedocs.io/en/latest/web_api/#klippy-disconnected

It could have been implemented as a single method with a state enum, but they chose to implement multiple notifications, one for each state.

niklasad1:
Aight, I see thanks for the explanation

lexnv (approved):

jsdw (approved):

Jomik:
Do you have an estimate of when you will create a release with this fix? It is currently interferring with my progress 😅

niklasad1:
I'll release it today

PR #1437, opened 2024-08-01
server: syncronize ws ping/pong messages

### LINKED ISSUE #1433: server: add synchronization of ping/pong mechanism
Currently jsonrpsee doesn't synchronize ping/pong instead it assumes that a pong should be seen at the configured interval however if the server is really busy like handling plenty of concurrent connections/messages.

Then it's possible that send operation takes longer than configured interval and that peer is regarded as `inactive`.

Thus, we should modify the current implementation such that jsonrpsee measures the time from that the ping is send until the pong received (associate each ping with pong)

niklasad1:
Close https://github.com/paritytech/jsonrpsee/issues/1433

### DIFF
--- a/server/src/server.rs
+++ b/server/src/server.rs
@@ -240,8 +240,8 @@ pub struct TowerServiceBuilder<RpcMiddleware, HttpMiddleware> {
 	pub(crate) conn_id: Arc<AtomicU32>,
 	/// Connection guard.
 	pub(crate) conn_guard: ConnectionGuard,
-	/// Remote address.
-	pub(crate) remote_addr: Option<SocketAddr>,
+	/// IP address.
+	pub(crate) ip_addr: Option<IpAddr>,
 }
 
 /// Configuration for batch request handling.
@@ -463,7 +463,7 @@ pub struct Builder<HttpMiddleware, RpcMiddleware> {
 	server_cfg: ServerConfig,
 	rpc_middleware: RpcServiceBuilder<RpcMiddleware>,
 	http_middleware: tower::ServiceBuilder<HttpMiddleware>,
-	remote_addr: Option<SocketAddr>,
+	ip_addr: Option<IpAddr>,
 }
 
 impl Default for Builder<Identity, Identity> {
@@ -472,7 +472,7 @@ impl Default for Builder<Identity, Identity> {
 			server_cfg: ServerConfig::default(),
 			rpc_middleware: RpcServiceBuilder::new(),
 			http_middleware: tower::ServiceBuilder::new(),
-			remote_addr: None,
+			ip_addr: None,
 		}
 	}
 }
@@ -484,8 +484,8 @@ impl Builder<Identity, Identity> {
 	}
 
 	/// Set the address of the remote peer.
-	pub fn set_remote_addr(mut self, remote_addr: SocketAddr) -> Self {
-		self.remote_addr = Some(remote_addr);
+	pub fn set_ip_addr(mut self, ip_addr: IpAddr) -> Self {
+		self.ip_addr = Some(ip_addr);
 		self
 	}
 }
@@ -507,7 +507,7 @@ impl<RpcMiddleware, HttpMiddleware> TowerServiceBuilder<RpcMiddleware, HttpMiddl
 				conn_id,
 				conn_guard: self.conn_guard,
 				server_cfg: self.server_cfg,
-				remote_addr: self.remote_addr,
+				ip_addr: self.ip_addr,
 			},
 			on_session_close: None,
 		};
@@ -537,7 +537,7 @@ impl<RpcMiddleware, HttpMiddleware> TowerServiceBuilder<RpcMiddleware, HttpMiddl
 			http_middleware: self.http_middleware,
 			conn_id: self.conn_id,
 			conn_guard: self.conn_guard,
-			remote_addr: self.remote_addr,
+			ip_addr: self.ip_addr,
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


server/src/transport/ws.rs
```diff
@@ -274,62 +286,104 @@ enum Receive<S> {
...
+
+#[derive(Debug, Copy, Clone)]
+enum KeepAlive {
+	Ping(Instant),
+	Pong(Instant),
+}
+
+async fn ping_pong_task(mut rx: mpsc::Receiver<KeepAlive>, max_inactive_limit: Duration, max_inactive: usize) {
+	let mut polling_interval = IntervalStream::new(interval(max_inactive_limit));
+	let mut pending_pings: VecDeque<Instant> = VecDeque::new();
```
niklasad1, line 332: The only annoyance with this approach is that this may become big if the max_inactive_limit is way bigger than the ping interval however that's configurable thingy and I don't think it's a real issue as we also regard received messages as pong.

server/src/server.rs
```diff
@@ -1049,6 +1087,15 @@ where
 
 		request.extensions_mut().insert::<ConnectionId>(conn.conn_id.into());
 
+		if let Some(ip_addr) = self.inner.ip_addr {
+			// Only insert the remote address if it's not already set.
+			// We expect servers deployed behind a reverse proxy to set the remote address
+			// themselves otherwise the remote address will be the address of the reverse proxy.
+			if request.extensions().get::<IpAddr>().is_none() {
+				request.extensions_mut().insert(ip_addr);
```
niklasad1, line 1095: Added this for better logging but I can revert this

jsdw: I'm probably being dumb but: how does setting thr IP address of the machine running this server help with logging? To help differentiate when logs from lots of servers are mingled together?

server/src/transport/ws.rs
```diff
@@ -229,11 +238,17 @@ async fn send_task(
 			// Received message.
 			Either::Left((Some(response), not_ready)) => {
 				// If websocket message send fail then terminate the connection.
+				let now = Instant::now();
 				if let Err(err) = send_message(&mut ws_sender, response).await {
 					tracing::debug!(target: LOG_TARGET, "WS send error: {}", err);
 					break;
 				}
 
+				if now.elapsed() > Duration::from_secs(30) {
```
jsdw, line 247: Is there a wway to avoid this check unless the log level is set to warn?

server/src/transport/ws.rs
```diff
@@ -246,12 +261,24 @@ async fn send_task(
 			// Handle timer intervals.
 			Either::Right((Either::Left((_instant, _stopped)), next_rx)) => {
 				stop = _stopped;
+				let now = Instant::now();
 				if let Err(err) = send_ping(&mut ws_sender).await {
 					tracing::debug!(target: LOG_TARGET, "WS send ping error: {}", err);
 					break;
 				}
 
+				if now.elapsed() > Duration::from_secs(30) {
```
jsdw, line 270: Same; can we avoid this check when log level isn't warn?

server/src/transport/ws.rs
```diff
@@ -246,12 +261,24 @@ async fn send_task(
...
+					tracing::warn!(target: LOG_TARGET, "Send ping was slow {}s, peer={:?}", now.elapsed().as_secs(), remote_addr);
+					break;
+				}
+
 				rx_item = next_rx;
+
+				let ping_tx = ping_tx.clone();
+				tokio::spawn(async move {
+					ping_tx.send(KeepAlive::Ping(Instant::now())).await.ok();
+				});
```
jsdw, line 280: If there is no ping_rx (ie because no ping config, so it's dropped ie on https://github.dev/paritytech/jsonrpsee/blob/705148f0c05dca1bf945e2387a866561a58beac5/server/src/transport/ws.rs#L92), I guess we are just doing this work for no gain. Is it worth putting an if around the task spawning to avoid in this case or does it over complicate things?

server/src/server.rs
niklasad1: It's useful for debug logs if your server has lots connections and maybe only certain peers/ip addrs are slow but again I think it's better to revert this from this PR..

server/src/transport/ws.rs
```diff
@@ -268,68 +295,128 @@ async fn send_task(
...
+						// Both pong and data are considered as a response to the ping.
+						// So we might get more responses than pings that's why it's possible
+						// that the pending_pings may be empty.
+						if let Some(start) = pending_pings.pop_front() {
+							// Calculate the round-trip time (RTT) of the ping/pong.
+							// We adjust for the time it took to send to this task.
+							let elapsed = start.elapsed().saturating_sub(end.elapsed());
+
+							if elapsed >= max_inactive_limit {
+								missed_pings += 1;
```
jsdw, line 405: What decrements this again? If somebody misses a few pings but then is ok and replies to the next 100 pings after that, I'd expect this counter to be reset to 0 again, else it might gradually accumulate until it hits the limit.

I'd prob reset it each time a successful ping comes in, eg an `else` on this?

server/src/transport/ws.rs
niklasad1: good catch, I can make it an option instead.

server/src/transport/ws.rs
```diff
@@ -38,7 +39,6 @@ pub(crate) async fn send_message(sender: &mut Sender, response: String) -> Resul
 }
 
 pub(crate) async fn send_ping(sender: &mut Sender) -> Result<(), SokettoError> {
-	tracing::debug!(target: LOG_TARGET, "Send ping");
```
niklasad1, line 41: I removed this, it's quite useless and possible get it by enabling soketto logs

PR #1440, opened 2024-08-12
feat(server): add `Methods::extensions/extensions_mut`

niklasad1:
The RpcModule/Methods may need to inject extensions which could be needed by the RPC implementations.

For example in substrate I have moved the `deny unsafe stuff` to an extension but the unit tests fails because it's needs be added by the RpcModule.

Thus, this PR adds additional API to be used for unit testing a RpcModule. For instance one may inject some arbitary data which can't be unit tested unless it's explictly injected by the RpcModule.

### DIFF
--- a/core/src/server/rpc_module.rs
+++ b/core/src/server/rpc_module.rs
@@ -213,6 +213,7 @@ impl Debug for MethodCallback {
 #[derive(Default, Debug, Clone)]
 pub struct Methods {
 	callbacks: Arc<FxHashMap<&'static str, MethodCallback>>,
+	extensions: Option<Extensions>,
 }
 
 impl Methods {
@@ -366,11 +367,17 @@ impl Methods {
 		subscription_permit: SubscriptionPermit,
 	) -> RawRpcResponse {
 		let (tx, mut rx) = mpsc::channel(buf_size);
+		// The extensions is always empty when calling the method directly because decoding an JSON-RPC
+		// request doesn't have any extensions.
 		let Request { id, method, params, mut extensions, .. } = req;
 		let params = Params::new(params.as_ref().map(|params| params.as_ref().get()));
 		let max_response_size = usize::MAX;
 		let conn_id = ConnectionId(0);
 		extensions.insert(conn_id);
+
+		if let Some(other_ext) = self.extensions.as_ref() {
+			extensions.extend(other_ext.clone());
+		}
 
 		let response = match self.method(&method) {
 			None => MethodResponse::error(id, ErrorObject::from(ErrorCode::MethodNotFound)),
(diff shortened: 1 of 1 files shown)


jsdw (approved):

core/src/server/rpc_module.rs
```diff
@@ -469,6 +476,48 @@ impl Methods {
 	pub fn method_names(&self) -> impl Iterator<Item = &'static str> + '_ {
 		self.callbacks.keys().copied()
 	}
+
+	/// Insert extensions to the methods.
+	///
+	/// This only affects direct calls to the methods and subscriptions.
```
niklasad1, line 482:
```suggestion
	/// This only affects direct calls to the methods and subscriptions
```

core/src/server/rpc_module.rs
```diff
@@ -469,6 +476,48 @@ impl Methods {
...
+	///     ext.insert(33_u64);
+	///
+	///     module.with_extensions(ext);
+	///
+	///     let magic: u64 = module.call("magic_multiply", [1_u64]).await.unwrap();
+	///     assert_eq!(magic, 33);
+	/// }
+	/// ```
+
+	pub fn with_extensions(&mut self, extensions: Extensions) {
```
niklasad1, line 514:
```suggestion
	pub fn insert_extensions(&mut self, extensions: Extensions) {
```

core/src/server/rpc_module.rs
```diff
@@ -366,22 +367,26 @@ impl Methods {
...
-		let Request { id, method, params, mut extensions, .. } = req;
+		// The extensions is always empty when calling the method directly because decoding an JSON-RPC
+		// request doesn't have any extensions.
+		let Request { id, method, params, .. } = req;
 		let params = Params::new(params.as_ref().map(|params| params.as_ref().get()));
 		let max_response_size = usize::MAX;
 		let conn_id = ConnectionId(0);
-		extensions.insert(conn_id);
+		let mut ext = self.extensions.clone();
+
```
niklasad1, line 377:
```suggestion
```

core/src/server/rpc_module.rs
```diff
@@ -366,22 +367,25 @@ impl Methods {
...
-		let Request { id, method, params, mut extensions, .. } = req;
+		// The extensions is always empty when calling the method directly because decoding an JSON-RPC
+		// request doesn't have any extensions.
+		let Request { id, method, params, .. } = req;
 		let params = Params::new(params.as_ref().map(|params| params.as_ref().get()));
 		let max_response_size = usize::MAX;
 		let conn_id = ConnectionId(0);
-		extensions.insert(conn_id);
+		let mut ext = self.extensions.clone();
+		ext.insert(conn_id);
```
niklasad1, line 377: we could actually inject this dummy connection in the constructor but I don't want to add the overhead for stuff that isn't using it.

core/src/server/rpc_module.rs
```diff
@@ -213,6 +213,7 @@ impl Debug for MethodCallback {
 #[derive(Default, Debug, Clone)]
 pub struct Methods {
 	callbacks: Arc<FxHashMap<&'static str, MethodCallback>>,
+	extensions: Extensions,
```
niklasad1, line 216: This is cheap, the default is just Option::None,

jsdw (approved):
That looks cleaner than before; like it!

PR #1443, opened 2024-08-16
feat: expose ConnectionGuard as request extension

### LINKED ISSUE #1438: Expose current connections count as request extension
I would like to have the number of active connections exposed as a request extension, so it can be available in the middleware layer to be tracked using our custom metrics.

I cannot find any other way to retrieve this information.

https://github.com/paritytech/jsonrpsee/blob/6c293950fc0eb7c3c2b08d9491de6f33b84047c8/server/src/server.rs#L1048

dinhani:
Fixes #1438

Since ConnectionGuard keeps the internal information inside an Arc, it should be cheap to clone it for each request.

### DIFF
--- a/server/src/server.rs
+++ b/server/src/server.rs
@@ -1047,7 +1047,9 @@ where
 		let curr_conns = max_conns - conn_guard.available_connections();
 		tracing::debug!(target: LOG_TARGET, "Accepting new connection {}/{}", curr_conns, max_conns);
 
-		request.extensions_mut().insert::<ConnectionId>(conn.conn_id.into());
+		let req_ext = request.extensions_mut();
+		req_ext.insert::<ConnectionGuard>(conn_guard.clone());
+		req_ext.insert::<ConnectionId>(conn.conn_id.into());
 
 		let is_upgrade_request = is_upgrade_request(&request);
 
--- a/tests/tests/integration_tests.rs
+++ b/tests/tests/integration_tests.rs
@@ -207,7 +207,7 @@ async fn ws_method_call_works_over_proxy_stream() {
 }
 
 #[tokio::test]
-async fn extensions_with_different_ws_clients() {
+async fn connection_id_extension_with_different_ws_clients() {
 	init_logger();
 
 	let server_addr = server().await;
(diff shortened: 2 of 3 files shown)


niklasad1:
Thanks

Can you just create a simple test for this? Similar to https://github.com/paritytech/jsonrpsee/blob/master/tests/tests/integration_tests.rs#L210-#L225 but you need to add an new handler that actually fetches the ConnGuard from the extension.

server/src/server.rs
```diff
@@ -1047,7 +1047,9 @@ where
 		let curr_conns = max_conns - conn_guard.available_connections();
 		tracing::debug!(target: LOG_TARGET, "Accepting new connection {}/{}", curr_conns, max_conns);
 
-		request.extensions_mut().insert::<ConnectionId>(conn.conn_id.into());
+		let req_ext = request.extensions_mut();
+		req_ext.insert::<ConnectionGuard>(conn_guard.clone());
+		req_ext.insert::<ConnectionId>(conn.conn_id.into());
```
niklasad1, line 1052: Cool, we could probably merge the ConnectionID and ConnectionGuard as some point to another type e.g. ConnectionDetails or something

tests/tests/integration_tests.rs
```diff
@@ -225,6 +225,29 @@ async fn extensions_with_different_ws_clients() {
 	assert_ne!(connection_id, second_connection_id);
 }
 
+#[tokio::test]
+async fn connection_guard_extension_with_different_ws_clients() {
```
niklasad1, line 229: it's sufficient with one client but this does hurt, thanks :)

niklasad1 (approved):

dinhani:
@niklasad1 
I did not notice code was being tested via JSON-RPC interface.

I tried to put the new test together with the one you pointed, but I think it would become too complex, so I created a new test and renamed the other one to make it clear which extension each one is testing.

niklasad1:
> I did not notice code was being tested via JSON-RPC interface.

> I tried to put the new test together with the one you pointed, but I think it would become too complex, so I created a new test and renamed the other one to make it clear which extension each one is testing.

Yeah, you made the right call by creating another test. All good, just run ` cargo +stable fmt --all` on your PR which will make the CI happy

mayconamaro:
that's nice the PR is already created <3 
thank you @dinhani! 

@niklasad1 could you suggest another reviewer with write access for this PR?

lexnv (approved):

PR #1449, opened 2024-09-04
fix: use error(transparent) for wrapped errors when possible

### LINKED ISSUE #1447: core::client::Error double prints error message
https://github.com/paritytech/jsonrpsee/blob/9b66e9c0bdad95385bd29696d50c6a0deffa0b08/core/src/client/error.rs#L36-L38

this both prints the error message on display (via `#[error("{0}")]`) and includes the exact same message as the [`Error::source`](https://doc.rust-lang.org/stable/std/error/trait.Error.html#method.source) (via `#[from]`). this results in the error being printed twice when showing the full stack trace of an error:

```
ErrorObject { code: ServerError(-1), message: "error", data: None }: ErrorObject { code: ServerError(-1), message: "error", data: None }
```

(printed using [this](https://github.com/unionlabs/union/blob/dbf7b57e1c09998714b2ee33c1e5b7edf57642a3/lib/unionlabs/src/lib.rs#L343-L355), but any error tracer will have the same issue).

`#[error(transparent)]` should be used here, since no additional context is provided currently anyways.

niklasad1:
This changes the jsonrpsee's error to use transparent for wrapped errors for more idiomatic backtraces to avoid print errors etc more than once.

Close #1447

### DIFF
--- a/client/http-client/src/tests.rs
+++ b/client/http-client/src/tests.rs
@@ -25,7 +25,6 @@
 // DEALINGS IN THE SOFTWARE.
 
 use crate::types::error::{ErrorCode, ErrorObject};
-
 use crate::HttpClientBuilder;
 use jsonrpsee_core::client::{BatchResponse, ClientT, IdKind};
 use jsonrpsee_core::params::BatchRequestBuilder;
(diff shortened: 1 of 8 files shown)


jsdw (approved):
LGTM!

Issue #1451, opened 2024-09-11
Misleading message in jsonrpsee_server::transport::http::response::method_not_allowed


dcfreire:
Currently jsonrpsee returns "Used HTTP Method is not allowed. POST or OPTIONS is required\n" as the body of the response to any request that is not POST. Either it should actually accept OPTIONS requests, or return "Used HTTP Method is not allowed. POST is required\n".

I understand that a CORS layer can be added to handle OPTIONS requests, but it is not required to reach this message.

niklasad1:
Closed by #1452

PR #1452, opened 2024-09-11
chore: change transport::http::response::method_not_allowed message

dcfreire:
#1451


niklasad1 (approved):

jsdw (approved):

jsdw:
I thougght options might be in the error for CORS but that's just middleware now anyway right?

niklasad1:
> I thougght options might be in the error for CORS but that's just middleware now anyway right?

Yes, correct

niklasad1:
The function is only used https://github.com/paritytech/jsonrpsee/blob/master/server/src/transport/http.rs#L84-#L109 and this changed error message is "more accurate"

PR #1468, opened 2024-10-04
fix(server): count http calls in connection guard

### LINKED ISSUE #1467: `ConnectionState` dropped before HTTP request is finished
I noticed the [`ConnectionState`](https://github.com/paritytech/jsonrpsee/blob/1bfea409145dfa11a76fd1e7b3586ab11fdafe26/server/src/server.rs#L259) struct is being dropped before the request is finished when executing HTTP requests.

I have a custom middleware that uses the `ConnectionGuard` exposed via request extensions, and the number of active requests returned was 0 when it was supposed to be 1.

After some investigation, I found the [`conn`](https://github.com/paritytech/jsonrpsee/blob/1bfea409145dfa11a76fd1e7b3586ab11fdafe26/server/src/server.rs#L1044) instance is held active only for websocket requests, but for HTTP requests handled in the [`else` ](https://github.com/paritytech/jsonrpsee/blob/1bfea409145dfa11a76fd1e7b3586ab11fdafe26/server/src/server.rs#L1135) branch, it is not used, causing the `ConnectionState` and the semaphore permit to be dropped before the request finished executing.

niklasad1:
Close #1467 

The ConnectionGuard was dropped before actually waiting for the http future to resolved which this fixes by enforcing RpcService to take ownership of ConnectionState.

### DIFF
--- a/server/src/server.rs
+++ b/server/src/server.rs
@@ -1150,13 +1150,15 @@ where
 				let rp =
 					http::call_with_service(request, batch_config, max_request_size, rpc_service, max_response_size)
 						.await;
+				// NOTE: The `conn guard` must be held until the response is processed
+				// to respect the `max_connections` limit.
 				drop(conn);
 				Ok(rp)
 			})
 		} else {
-			let rp = http::response::denied();
-			drop(conn);
-			Box::pin(async { Ok(rp) })
+			// NOTE: the `conn guard` is dropped when this function which is fine
+			// because it doesn't rely on any async operations.
+			Box::pin(async { Ok(http::response::denied()) })
 		}
 	}
 }
--- a/server/src/server.rs
+++ b/server/src/server.rs
@@ -1146,12 +1146,17 @@ where
 				RpcServiceCfg::OnlyCalls,
 			));
 
-			Box::pin(
-				http::call_with_service(request, batch_config, max_request_size, rpc_service, max_response_size)
-					.map(Ok),
-			)
+			Box::pin(async move {
+				let rp =
+					http::call_with_service(request, batch_config, max_request_size, rpc_service, max_response_size)
+						.await;
+				drop(conn);
+				Ok(rp)
+			})
 		} else {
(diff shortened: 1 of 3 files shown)


lexnv (approved):

server/src/server.rs
```diff
@@ -1146,12 +1146,17 @@ where
...
+					http::call_with_service(request, batch_config, max_request_size, rpc_service, max_response_size)
+						.await;
+				drop(conn);
+				Ok(rp)
+			})
 		} else {
-			Box::pin(async { http::response::denied() }.map(Ok))
+			let rp = http::response::denied();
+			drop(conn);
+			Box::pin(async { Ok(rp) })
```
jsdw, line 1159: I guess this is just a tidy up and isn't changing any semantics right?

server/src/server.rs
```diff
@@ -1146,12 +1146,17 @@ where
...
 
-			Box::pin(
-				http::call_with_service(request, batch_config, max_request_size, rpc_service, max_response_size)
-					.map(Ok),
-			)
+			Box::pin(async move {
+				let rp =
+					http::call_with_service(request, batch_config, max_request_size, rpc_service, max_response_size)
+						.await;
+				drop(conn);
```
jsdw, line 1153: And this is the main bit; droping here to pull it into the async function after the await.

Perhaps worth a comment to make clear we are dropping it for this reaosn?

server/src/server.rs
niklasad1: Yupp, it was to make it more verbose that it wasn't any async code but I can revert it if you want.

jsdw (approved):

server/src/server.rs
niklasad1: I added a comment for this instead. I'm not super happy with the drop stuff here because it's easy to mess up but fine for now.

pkhry (approved):

PR #1470, opened 2024-10-08
feat(http client builder): Add max_concurrent_requests setter

sydhds:


lexnv (approved):

niklasad1 (changes requested):
This "feature" has been removed and has no effect anymore.

We/I forgot to remove the `max_concurrent_request` field on the builder because it was not ergonomic to use in async to code to wait until one would get `MaxSlotsExceeded` ...

For now you enforce the "limit" yourself for example with a semaphore/channel or something like that. Ideally with a channel bounded channel to provide backpressure which is much more ergonomic to deal with in async code.

However, I reckon that the API is not really uniform anymore compared to the "ws-client/async-client" APIs which provides  `max_concurrent_requests`. So we could add "it back" to http-client but with backpressure

niklasad1:
Thanks for the PR but I have created another PR that fixes this properly https://github.com/paritytech/jsonrpsee/pull/1473

PR #1473, opened 2024-10-09
http client: add `max_concurrent_requests`

niklasad1:
I decided to add back `max_concurrent_requests` to provide a uniform API because the ws-client already provides it.
This is only opt-in because it may be annoying with the extra sempahore overhead which some use-cases may not want..

Superseeds #1470


jsdw (approved):
I'd generally lean towards saying "if this can be implemented outside of the library then it should be" but for the sake of consistency I don't mind it :)

niklasad1:
> I'd generally lean towards saying "if this can be implemented outside of the library then it should be" but for the sake of consistency I don't mind it :)

Yes, that was the intention by removing it but I prefer to keep the APIs somewhat consistent instead after thinking about it

PR #1477, opened 2024-10-17
chore(deps): upgrade tower-http to 0.6

hanabi1224:
This PR bumps the version of `tower-http` crate which is only used in examples and tests(thus this change should be non-breaking).


niklasad1:
Hi, thanks for your PR but I merged the dependabot PR instead which was already approved

I'll merge in tower 0.5 as well but no new release planned yet because I want to check whether if there something else that we want to get in because it's a breaking change

PR #1478, opened 2024-10-17
[ci] Move benchmark job to github-action

alvicsam:
PR moves benchmark job from gitlab to github. The workflow uses the same machine as it was in gitlab. Also in this PR I suggest removing `on: schedule` condition and run benchmarks on each push to master but I can revert and return schedule. I tested this on a fork and unfortunately there were some [regressions](https://paritytech-stg.github.io/jsonrpsee/bench/dev2/) but only after one attempt. Still I'd like to merge it and check the results in long run. 

cc @niklasad1 
close https://github.com/paritytech/ci_cd/issues/1048


AndWeHaveAPlan (approved):

niklasad1:
I'm generally fine with this because this library is not under active development and the benches are not as critical anymore

> Also in this PR I suggest removing on: schedule condition and run benchmarks on each push to master but I can revert and return schedule

I think we can do both but perhaps both on each push to master and just run it once per week.

alvicsam:
@niklasad1 sure, added the schedule.

niklasad1 (approved):

PR #1483, opened 2024-10-28
chore(server): remove useless ServerConfigBuilder

koushiro:
The `ServerConfigBuilder` is useless, so I remove it, 
but this will cause a **breaking change**, because `ServerConfig` exposes a `builder` method.

### DIFF
--- a/server/src/server.rs
+++ b/server/src/server.rs
@@ -203,30 +203,6 @@ pub struct ServerConfig {
 	pub(crate) tcp_no_delay: bool,
 }
 
-#[derive(Debug, Clone)]
-pub struct ServerConfigBuilder {
-	/// Maximum size in bytes of a request.
-	max_request_body_size: u32,
-	/// Maximum size in bytes of a response.
-	max_response_body_size: u32,
-	/// Maximum number of incoming connections allowed.
-	max_connections: u32,
-	/// Maximum number of subscriptions per connection.
-	max_subscriptions_per_connection: u32,
-	/// Whether batch requests are supported by this server or not.
-	batch_requests_config: BatchRequestConfig,
-	/// Enable HTTP.
-	enable_http: bool,
-	/// Enable WS.
-	enable_ws: bool,
-	/// Number of messages that server is allowed to `buffer` until backpressure kicks in.
-	message_buffer_capacity: u32,
-	/// Ping settings.
-	ping_config: Option<PingConfig>,
-	/// ID provider.
-	id_provider: Arc<dyn IdProvider>,
-}
-
 /// Builder for [`TowerService`].
 #[derive(Debug, Clone)]
 pub struct TowerServiceBuilder<RpcMiddleware, HttpMiddleware> {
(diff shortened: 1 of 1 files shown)


server/src/server.rs
```diff
@@ -563,6 +438,12 @@ impl<HttpMiddleware, RpcMiddleware> Builder<HttpMiddleware, RpcMiddleware> {
 		self
 	}
 
+	/// Set the maximum number of connections allowed. Default is 1024.
```
niklasad1, line 441: I don't understand why you changed these on the Builder?  It seems like you just moved the APIs around or am I missing something?

I would be easier if you just reverted these

niklasad1 (commented):
nice catch, I would just want you to revert the changes you made on the ServerBuilder which seems unrelated to this PR unless I'm missing something?!

server/src/server.rs
koushiro: I just adjusted the order of the methods to make them consistent with the order of the fields in `ServerConfig` to avoid missing some methods.

niklasad1: It's okay this time because it's rather small change but generally I prefer to keep each PR as precise/small as possible.

niklasad1 (approved):

niklasad1 (changes requested):
Hey again, thanks for your PR but I had proper look and it's not possible to configure the `ServerConfig` except the default one, https://docs.rs/jsonrpsee-server/latest/jsonrpsee_server/struct.ServerConfig.html

Thus, I prefer to keep it as it is right now alternatively move all the setters to the `ServerConfig` but I think it's quite clean to have all settings on the `ServerConfigBuilder`.

Thus, the `ServerConfigBuilder` is not useless

koushiro:
> Hey again, thanks for your PR but I had proper look and it's not possible to configure the `ServerConfig` except the default one, https://docs.rs/jsonrpsee-server/latest/jsonrpsee_server/struct.ServerConfig.html
> 
> Thus, I prefer to keep it as it is right now alternatively move all the setters to the `ServerConfig` but I think it's quite clean to have all settings on the `ServerConfigBuilder`.
> 
> Thus, the `ServerConfigBuilder` is not useless

`ServerBuilder` does all things that `ServerConfigBuilder` can do, and `ServerConfigBuilder` doesn't even have a `build` method to generate `ServerConfig`. I really don't know what `ServerConfigBuilder` is for.

niklasad1:
Yes, you are right the `fn build` was/is missing I noticed it myself and therefor opened https://github.com/paritytech/jsonrpsee/pull/1484

koushiro:
For downstream developers, the jsonrpc server can only be constructed through the `Builder` type. 
Moreover, the existing `Builder` implementation doesn't have a method to accept a constructed `ServerConfig` (built by `ServerConfigBuilder`).
So the changes in  #1484 are useless for me, as the `ServerConfigBuilder` cannot be used by the `Builder`.

koushiro:
My point is that if we decide to keep `ServerConfigBuilder`, why not just delete all the config setter methods of `Builder` and use the following way, which can avoid a lot of repeated setter implementations (and if it can also avoid adding new setter methods to `ServerConfigBuilder` later, but forget to add corresponding methods to `Builder`)

```rust
async fn demo() -> anyhow::Result<()> {
	let builder = Builder::with_config(
		ServerConfig::builder()
			.max_request_body_size(TEN_MB_SIZE_BYTES)
                         // ...
			.build()
	)
	.set_rpc_middleware(...)
	.set_http_middleware(...)
	.build(...)
	.await?;
	// ...
}
```

niklasad1:
That would be neat and I'm not against removing such repeating setters but I have no good answer where to draw to line for such thing i.e, what to have in the `ServerConfig` vs ServerBuilder, or in some other type.

If you could write up an issue with a suggestion that would be great

PR #1492, opened 2024-11-07
refactor(server): make `ProxyGetRequestLayer` http middleware support multiple path-method pairs

koushiro:
**BREAKING CHNAGES**:

- change `ProxyGetRequestLayer::new` method to support multiple path-method pairs
- remove `ProxyGetRequest::new` method, which is useless

### DIFF
--- a/server/Cargo.toml
+++ b/server/Cargo.toml
@@ -15,27 +15,28 @@ publish = true
 
 [dependencies]
 futures-util = { version = "0.3.14", default-features = false, features = ["io", "async-await-macro"] }
-jsonrpsee-types = { workspace = true }
-jsonrpsee-core = { workspace = true, features = ["server", "http-helpers"] }
-tracing = "0.1.34"
-serde = "1"
-serde_json = { version = "1", features = ["raw_value"] }
-soketto = { version = "0.8", features = ["http"] }
-tokio = { version = "1.23.1", features = ["net", "rt-multi-thread", "macros", "time"] }
-tokio-util = { version = "0.7", features = ["compat"] }
-tokio-stream = { version = "0.1.7", features = ["sync"] }
-hyper = { version = "1.3", features = ["server", "http1", "http2"] }
-hyper-util = { version = "0.1", features = ["tokio", "service", "tokio", "server-auto"] }
 http = "1"
 http-body = "1"
 http-body-util = "0.1.0"
-tower = { workspace = true, features = ["util"] }
-thiserror = "2"
-route-recognizer = "0.3.1"
+hyper = { version = "1.3", features = ["server", "http1", "http2"] }
+hyper-util = { version = "0.1", features = ["tokio", "service", "tokio", "server-auto"] }
+jsonrpsee-core = { workspace = true, features = ["server", "http-helpers"] }
+jsonrpsee-types = { workspace = true }
 pin-project = "1.1.3"
+route-recognizer = "0.3.1"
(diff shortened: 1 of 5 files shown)


koushiro:
@niklasad1 @jsdw PTAL

server/Cargo.toml
```diff
@@ -15,27 +15,28 @@ publish = true
 
 [dependencies]
 futures-util = { version = "0.3.14", default-features = false, features = ["io", "async-await-macro"] }
-jsonrpsee-types = { workspace = true }
```
niklasad1, line 18: did you sort the dependencies here or why is this changed?

koushiro: Yes, this is automatically done by my vscode extension if I change the `Cargo.toml`.

niklasad1 (approved):
Looks good
I guess it may be useful for users that want to expose several endpoints.

server/Cargo.toml
lexnv: dq: that sounds useful, what extension are you using?

server/src/middleware/http/proxy_get_request.rs
```diff
@@ -53,29 +55,40 @@ pub struct InvalidPath(String);
 /// See [`ProxyGetRequest`] for more details.
 #[derive(Debug, Clone)]
 pub struct ProxyGetRequestLayer {
-	path: String,
-	method: String,
+	// path => method mapping
+	methods: Arc<FxHashMap<String, String>>,
```
lexnv, line 59: I think a plain `HashMap` would work as well, don't that would bring a noticeable performance hit 
```suggestion
	methods: Arc<HashMap<String, String>>,
```

server/src/middleware/http/proxy_get_request.rs
```diff
@@ -53,29 +55,40 @@ pub struct InvalidPath(String);
...
-		if !path.starts_with('/') {
-			return Err(InvalidPath(path));
+	/// The request `GET /path` is redirected to the provided method.
+	/// Fails if the path does not start with `/`.
+	pub fn new<P, M>(methods: impl IntoIterator<Item = (P, M)>) -> Result<Self, InvalidPath>
+	where
+		P: Into<String>,
+		M: Into<String>,
+	{
+		let methods = methods
```
lexnv, line 72: I believe we can check for path duplicates. Currently, we'll keep in the hashmap just the last entry over which we iterate.

For example, we currently accept:

```rust
[("/health", "system_health"), ("/health", "system2")]
```

This is a misconfig from the user perspective, atm the implementation keeps the `/health` -> `system2` instead of returning an error

lexnv (approved):
LGTM! Just some minor notes on deduplication of paths for better user errors
Thanks for contributing 🙏

server/src/middleware/http/proxy_get_request.rs
niklasad1: We could also just use a `Vec` here because we don't expect that many endpoints exposed anyway. 
HashMap is probably a bit overkill but either will work...

server/Cargo.toml
koushiro: https://taplo.tamasfe.dev/

server/src/middleware/http/proxy_get_request.rs
koushiro: Added check and error type for duplicate path

lexnv (approved):
Nice 👍

PR #1512, opened 2025-01-07
fix: preserve the URI query in ProxyGetRequest::call

dcfreire:
#1513

### DIFF
--- a/server/src/middleware/http/proxy_get_request.rs
+++ b/server/src/middleware/http/proxy_get_request.rs
@@ -149,8 +149,12 @@ where
 			(Some(method), &Method::GET) => {
 				// RPC methods are accessed with `POST`.
 				*req.method_mut() = Method::POST;
-				// Precautionary remove the URI.
-				*req.uri_mut() = Uri::from_static("/");
+				// Precautionary remove the URI path.
+				*req.uri_mut() = if let Some(query) = req.uri().query() {
+					Uri::from_str(&format!("/?{}", query)).expect("Valid uri; qed")
+				} else {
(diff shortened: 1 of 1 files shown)


server/src/middleware/http/proxy_get_request.rs
```diff
@@ -149,8 +150,12 @@ where
 			(Some(method), &Method::GET) => {
 				// RPC methods are accessed with `POST`.
 				*req.method_mut() = Method::POST;
-				// Precautionary remove the URI.
-				*req.uri_mut() = Uri::from_static("/");
+				// Precautionary remove the URI path.
```
niklasad1, line 153: //cc @lexnv do you remember why we removed the URI in the first place?

niklasad1 (approved):

server/src/middleware/http/proxy_get_request.rs
```diff
@@ -149,8 +150,12 @@ where
...
+				*req.uri_mut() = if let Some(query) = req.uri().query() {
+					Uri::from_str(&format!("/?{}", query)).expect("Valid uri; qed")
```
niklasad1, line 155:
```suggestion
					Uri::from_str(&format!("/?{}", query)).expect("The query valid and comes from a valid URI; qed")
```

server/src/middleware/http/proxy_get_request.rs
lexnv: I don't quite remember 🤔

lexnv (approved):

PR #1514, opened 2025-01-08
http server: propagate extensions in http response

### LINKED ISSUE #1509: Pass response contexts from RPC middleware to HTTP middleware
I am developing a rate-limiting RPC middleware. Currently, jsonrpsee framework utilizes `MethodResponse` Extensions in the RPC layer and `http::Response` Extensions in the HTTP layer, with these extensions operating independently of each other.

I would like to propagate context information from the RPC layer to the HTTP layer to achieve functionality such as returning a status code other than 200 (429 in my case) or, at the very least, adding an HTTP header to indicate rate limiting based on data in contexts.

What would be the best approach to transfer context (Extensions) information from a jsonrpsee RPC middleware to a hyper HTTP middleware?

niklasad1:
Close #1509 

This PR propagates the extensions set by the RPC middleware such that users can inspect them via HTTP middleware and perhaps indicate a HTTP error or something.

The handling of batch requests are not ideal because because the if the same type is used more than once then the 
most recent wins in the old data is overwritten. Because the http::Extensions works based on the rust type ids and it's not possible to merge two types in a clean way I think i.e, just overwrite them with the most recent one...

I had in mind that the user could inject a Vec or another collection that each call RPC call/request can push to it but we can't do that internally in jsonrpsee without some hacks like merging the rust collection types ourselves such as `Vec<String>, Vec<usize>` and so on...

For now, I don't have a good solution except make user responsible to "inject unique types" because it may be overwritten in a batch but annoying....

We could introduce another representation of the extensions internally and just merge them before constructing the response though...


lexnv (approved):

pkhry (approved):

server/src/tests/http.rs
```diff
@@ -42,6 +52,66 @@ fn init_logger() {
...
+	type Response = S::Response;
+	type Error = BoxError;
+	type Future = Pin<Box<dyn Future<Output = Result<Self::Response, Self::Error>> + Send + 'static>>;
+
+	fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+		self.service.poll_ready(cx).map_err(Into::into)
+	}
+
+	fn call(&mut self, request: HttpRequest<B>) -> Self::Future {
+		tracing::info!("{:?}", request);
```
niklasad1, line 101:
```suggestion
```

PR #1516, opened 2025-01-09
send original error in ProxyGetRequest

### LINKED ISSUE #1515: Errors to requests that go through ProxyGetRequest are always internal error
Errors to requests that go through ProxyGetRequest are always internal error instead of reflecting the original error returned by the rpc method.

dcfreire:
closes #1515

### DIFF
--- a/proc-macros/tests/ui/incorrect/rpc/rpc_empty_bounds.stderr
+++ b/proc-macros/tests/ui/incorrect/rpc/rpc_empty_bounds.stderr
@@ -17,7 +17,6 @@ error[E0277]: the trait bound `<Conf as Config>::Hash: Clone` is not satisfied
   |
   = note: required for `Result<<Conf as Config>::Hash, ErrorObject<'_>>` to implement `IntoResponse`
   = note: this error originates in the attribute macro `rpc` (in Nightly builds, run with -Z macro-backtrace for more info)
-  |
 9 | #[rpc(server, client, namespace = "foo", client_bounds(), server_bounds())]
   |
 
--- a/proc-macros/tests/ui/incorrect/rpc/rpc_empty_bounds.stderr
+++ b/proc-macros/tests/ui/incorrect/rpc/rpc_empty_bounds.stderr
@@ -17,7 +17,6 @@ error[E0277]: the trait bound `<Conf as Config>::Hash: Clone` is not satisfied
   |
   = note: required for `Result<<Conf as Config>::Hash, ErrorObject<'_>>` to implement `IntoResponse`
   = note: this error originates in the attribute macro `rpc` (in Nightly builds, run with -Z macro-backtrace for more info)
-help: consider removing this method call, as the receiver has type `&Self` and `&Self: Clone` trivially holds
   |
 9 | #[rpc(server, client, namespace = "foo", client_bounds(), server_bounds())]
   |
--- a/proc-macros/tests/ui/incorrect/rpc/rpc_empty_bounds.stderr
+++ b/proc-macros/tests/ui/incorrect/rpc/rpc_empty_bounds.stderr
@@ -17,6 +17,10 @@ error[E0277]: the trait bound `<Conf as Config>::Hash: Clone` is not satisfied
   |
   = note: required for `Result<<Conf as Config>::Hash, ErrorObject<'_>>` to implement `IntoResponse`
   = note: this error originates in the attribute macro `rpc` (in Nightly builds, run with -Z macro-backtrace for more info)
+help: consider removing this method call, as the receiver has type `&Self` and `&Self: Clone` trivially holds
+  |
+9 | #[rpc(server, client, namespace = "foo", client_bounds(), server_bounds())]
+  |
 
 error[E0277]: the trait bound `<Conf as Config>::Hash: DeserializeOwned` is not satisfied
  --> tests/ui/incorrect/rpc/rpc_empty_bounds.rs:9:1
--- a/proc-macros/tests/ui/incorrect/rpc/rpc_empty_bounds.stderr
+++ b/proc-macros/tests/ui/incorrect/rpc/rpc_empty_bounds.stderr
@@ -2,7 +2,7 @@ error[E0277]: the trait bound `<Conf as Config>::Hash: Serialize` is not satisfi
  --> tests/ui/incorrect/rpc/rpc_empty_bounds.rs:9:1
   |
 9 | #[rpc(server, client, namespace = "foo", client_bounds(), server_bounds())]
-  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Serialize` is not implemented for `<Conf as Config>::Hash`, which is required by `Result<<Conf as Config>::Hash, ErrorObject<'_>>: IntoResponse`
+  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Serialize` is not implemented for `<Conf as Config>::Hash`
   |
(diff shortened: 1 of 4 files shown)


dcfreire:
I tried using an enum instead of deserializing in two places, but it did not work, serde kept failing with "data did not match any variant of untagged enum". I think it has something to do with untagged enums expecting all parameters of the json object to be in the variant's struct, while structs will deserialize even if the json contains extra fields.

niklasad1:
> I tried using an enum instead of deserializing in two places, but it did not work, serde kept failing with "data did not match any variant of untagged enum". I think it has something to do with untagged enums expecting all parameters of the json object to be in the variant's struct, while structs will deserialize even if the json contains extra fields.

Yes, we are using serde_json::value::RawValue which doesn't work with untagged enums https://github.com/serde-rs/json/issues/779

server/src/middleware/http/proxy_get_request.rs
```diff
@@ -179,16 +179,27 @@ where
...
+						#[serde(borrow)]
+						error: &'a serde_json::value::RawValue,
+					}
+
+					let response = if let Ok(payload) = serde_json::from_slice::<SuccessResponse>(&bytes) {
 						http::response::ok_response(payload.result.to_string())
 					} else {
-						http::response::internal_error()
+						serde_json::from_slice::<ErrorResponse>(&bytes)
+							.and_then(|payload| serde_json::from_str::<ErrorObject>(&payload.error.to_string()))
```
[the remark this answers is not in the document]
dcfreire: I changed the implementation of Display for ErrorObject to do this, in my opinion the previous Display implementation derived by thiserror looked more like a debug message.

types/src/error.rs
```diff
@@ -96,6 +95,12 @@ impl<'a> ErrorObject<'a> {
 	}
 }
 
+impl fmt::Display for ErrorObject<'_> {
+	fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+		write!(f, "{}: {}", self.code(), self.message())
```
niklasad1, line 100: Please include "data" as well if it's `Some` but you could just use `serde_json::to_string(self)` as Display impl

server/src/middleware/http/proxy_get_request.rs
```diff
@@ -197,8 +197,8 @@ where
 						serde_json::from_slice::<ErrorResponse>(&bytes)
 							.and_then(|payload| serde_json::from_str::<ErrorObject>(&payload.error.to_string()))
 							.map_or_else(
-								|_| http::response::internal_error(),
-								|error| http::response::error_response(error.to_owned()),
+								|_| http::response::error_response(ErrorCode::InternalError.to_string()),
```
niklasad1, line 200: would be better to use `serde_json::to_string(&ErrorCode::InternalError)` or just change the Display impl to what I suggested :point_down:

server/src/middleware/http/proxy_get_request.rs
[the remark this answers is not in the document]
niklasad1: ok, I see didn't remember that we already used that the internal error but it is intended to be used for JSON-RPC calls not general HTTP calls ^^

types/src/error.rs
dcfreire: Doing that returns a json too though. `"{\"code\":-32603,\"message\":\"Internal error\"}"`

server/src/middleware/http/proxy_get_request.rs
dcfreire: Using serde_json only returns the error code, without the message

types/src/error.rs
niklasad1: What is wrong with that? It's just the jsonrpc headers that I don't want in in that response

server/src/middleware/http/proxy_get_request.rs
niklasad1: aight, you can do `serde_json(&ErrorObject::from(ErrorCode::InternalError))` my bad :)

niklasad1:
If you merge master then UI tests should pass I fixed it right now, new stable rust version broke it

types/src/error.rs
dcfreire: Ohhh, ok. My bad, I thought your suggestion was to just return the message without any json object, since the ok_response only returns the result string

dcfreire: For that end I just need to modify the error_response function I implemented then.

dcfreire: Done

niklasad1 (approved):

proc-macros/tests/ui/incorrect/rpc/rpc_empty_bounds.stderr
```diff
@@ -17,6 +17,10 @@ error[E0277]: the trait bound `<Conf as Config>::Hash: Clone` is not satisfied
   |
   = note: required for `Result<<Conf as Config>::Hash, ErrorObject<'_>>` to implement `IntoResponse`
   = note: this error originates in the attribute macro `rpc` (in Nightly builds, run with -Z macro-backtrace for more info)
+help: consider removing this method call, as the receiver has type `&Self` and `&Self: Clone` trivially holds
```
niklasad1, line 20:
```suggestion
```

proc-macros/tests/ui/incorrect/rpc/rpc_empty_bounds.stderr
```diff
@@ -17,6 +17,8 @@ error[E0277]: the trait bound `<Conf as Config>::Hash: Clone` is not satisfied
   |
   = note: required for `Result<<Conf as Config>::Hash, ErrorObject<'_>>` to implement `IntoResponse`
   = note: this error originates in the attribute macro `rpc` (in Nightly builds, run with -Z macro-backtrace for more info)
+9 | #[rpc(server, client, namespace = "foo", client_bounds(), server_bounds())]
```
niklasad1, line 20:
```suggestion
```

proc-macros/tests/ui/incorrect/rpc/rpc_empty_bounds.stderr
```diff
@@ -17,6 +17,7 @@ error[E0277]: the trait bound `<Conf as Config>::Hash: Clone` is not satisfied
   |
   = note: required for `Result<<Conf as Config>::Hash, ErrorObject<'_>>` to implement `IntoResponse`
   = note: this error originates in the attribute macro `rpc` (in Nightly builds, run with -Z macro-backtrace for more info)
+  |
```
niklasad1, line 20:
```suggestion
```

server/src/middleware/http/proxy_get_request.rs
```diff
@@ -179,16 +179,25 @@ where
...
+					struct SuccessResponse<'a> {
 						#[serde(borrow)]
 						result: &'a serde_json::value::RawValue,
 					}
 
-					let response = if let Ok(payload) = serde_json::from_slice::<RpcPayload>(&bytes) {
+					#[derive(serde::Deserialize, serde::Serialize, Debug)]
+					struct ErrorResponse<'a> {
+						#[serde(borrow)]
+						error: &'a serde_json::value::RawValue,
```
niklasad1, line 191:
```suggestion
						error: ErrorObject<'a>,
```

dcfreire: I applied this by mistake, if it is ok here to directly deserialize to the error object, why not also dirrectly deserialize the reulst in the success response?

dcfreire: Also I think in this case we can just do `serde_json::from_slice::<ErrorObject>(&bytes)` and not have the ErrorResponse struct at all

carneiro-cw: > Also I think in this case we can just do serde_json::from_slice::<ErrorObject>(&bytes)
This is incorrect, my bad

dcfreire: > Also I think in this case we can just do serde_json::from_slice::<ErrorObject>(&bytes)

This is incorrect, my bad

lexnv (approved):

PR #1521, opened 2025-01-22
feat: client rpc middleware

### LINKED ISSUE #1503: client: add `RpcServiceT` middleware
jsonrpsee has a specific RPC middleware trait for the server and we should expose a similar API for the client as well for a uniform API. It may be useful for similar reasons but from the client-side.

Basically, the task is to move the `RpcServiceT` and associated layers except the Server's RpcService impl to `core` and change the return type to a generic one to work for both the client and server.

niklasad1:
Close https://github.com/paritytech/jsonrpsee/issues/1503

This PR unifies the RPC service API to be used by both clients and servers, introducing a common trait (RpcServiceT) that supports method calls, notifications, and batch requests along with middleware support. Key changes include:

- Extending a unified RpcServiceT trait with batch and notification methods.
- Extending the HttpClient with generic middleware parameters and integrating RPC middleware.
- Refactoring middleware layers and associated components (e.g., logger, either) to support the new API.

Old trait:

```rust
/// Similar to the [`tower::Service`] but specific for jsonrpsee and
/// doesn't requires `&mut self` for performance reasons.
pub trait RpcServiceT<'a> {
	/// The future response value.
	type Future: Future<Output = MethodResponse> + Send;

	/// Process a single JSON-RPC call it may be a subscription or regular call.
	///
	/// In this interface both are treated in the same way but it's possible to
	/// distinguish those based on the `MethodResponse`.
	fn call(&self, request: Request<'a>) -> Self::Future;
}
```

Proposed change:

```rust
/// Represent a JSON-RPC service that can process JSON-RPC calls, notifications, and batch requests.
///
/// This trait is similar to [`tower::Service`] but it's specialized for JSON-RPC operations.
///
/// The response type is a future that resolves to a `Result<R, E>` mainly because this trait is
/// intended to by used by both client and server implementations.
///
/// In the server implementation, the error is infallible but in the client implementation, the error
/// can occur due to I/O errors or JSON-RPC protocol errors.
pub trait RpcServiceT {
	/// The error type.
	type Error: std::fmt::Debug;

	/// The response type
	type Response: ToJson;

	/// Processes a single JSON-RPC call, which may be a subscription or regular call.
	fn call<'a>(&self, request: Request<'a>) -> impl Future<Output = Result<Self::Response, Self::Error>> + Send + 'a;

	/// Processes multiple JSON-RPC calls at once, similar to `RpcServiceT::call`.
	///
	/// This method wraps `RpcServiceT::call` and `RpcServiceT::notification`,
	/// but the root RPC service does not inherently recognize custom implementations
	/// of these methods.
	///
	/// As a result, if you have custom logic for individual calls or notifications,
	/// you must duplicate that implementation in this method or no middleware will be applied
	/// for calls inside the batch.
	fn batch<'a>(&self, requests: Batch<'a>) -> impl Future<Output = Result<Self::Response, Self::Error>> + Send + 'a;

	/// Similar to `RpcServiceT::call` but processes a JSON-RPC notification.
	fn notification<'a>(
		&self,
		n: Notification<'a>,
	) -> impl Future<Output = Result<Self::Response, Self::Error>> + Send + 'a;
}
```

The reason is that the underlying client service will serialize the request which won't support Batches unless we add with a dedicate API i.e, we would need to send them one-by-one.


core/src/middleware/layer/either.rs
```diff
@@ -31,7 +31,7 @@
 //! work to implement tower::Layer for
 //! external types such as future::Either.
 
-use crate::middleware::rpc::RpcServiceT;
+use crate::middleware::{Notification, RpcServiceT};
 use jsonrpsee_types::Request;
 
 /// [`tower::util::Either`] but
```
jsdw, line 37: Random nit: ,issing closing "\`" on RpcServiceT on the line below

jsdw:
Overall I'm good with this change. The pros are basically allowing middlewares to be shared between client and server, but need to make sure that we don't have to mangle one side or the other to fit this trait into both places. As we said, possible making the response type generic as well as the error type would make that easier, but you'd know best :)

core/src/middleware/mod.rs
```diff
@@ -0,0 +1,348 @@
...
+	///
+	/// This method wraps `RpcServiceT::call` and `RpcServiceT::notification`,
+	/// but the root RPC service does not inherently recognize custom implementations
+	/// of these methods.
+	///
+	/// As a result, if you have custom logic for individual calls or notifications,
+	/// you must duplicate that logic here.
+	///
+	// TODO: Investigate if the complete service can be invoked inside `RpcService`.
+	fn batch(&self, requests: Batch<'a>) -> Self::Future;
```
jsdw, line 232: I'm still struggling to parse the comment with certainty, but probably because I am not familiar enough with this `RpcServiceT`.
 
Are you basically saying: If a batch request is made, it will trigger this `batch()` method but won't trigger the individual `call` and `notification` methods for each item in the batch. Thus, if you want to run some logic for a call or notification in a batch as well as not in a btch, it needs to be defined here as well as in `call`/`notification` methods.

I wonder whether there is a way for `batch` to have a default implementation so that, if it's not implemented, then `call`/`notification` _are_ called automatically for each item in the batch?

Eg

```rust
fn batch(&self, requests: Batch<'a>) -> Self::Future {
    let mut responses = vec![];
    for entry in requests.inner {
        match entry {
            BatchEntry::Call(call) => {
                responses.push(self.call(call).await);
            },
            BatchEntry::Notification(notif) => {
                responses.push(self.notification(notif).await);
            }
        }
    }
    merge_responses_into_batch_response(responses)
}
```

That way, people just impl `call` or `notification` if they want to handle those things, and only need to impl `batch` if they care about catching the batch as a whole and eg adding or removing things from it or checking its size or whatever, because otherwise the logic in call/notification will work whether they are batched or not.

examples/examples/jsonrpsee_as_service.rs
```diff
@@ -91,6 +95,14 @@ where
 			ResponseFuture::future(self.inner.call(req))
 		}
 	}
+
+	fn batch(&self, batch: Batch<'a>) -> Self::Future {
+		ResponseFuture::future(self.inner.batch(batch))
+	}
```
jsdw, line 101: I guess this is a good place where havign a default impl for `batch` would be nice.

I assume that as it stands, `trusted_call` requires auth headers if sent individually, but if sent as part of a batch the user would avoid needing any auth since the `batch` method bypasses that logic?

examples/examples/jsonrpsee_server_low_level_api.rs
```diff
@@ -98,6 +101,14 @@ where
 		}
 		.boxed()
 	}
+
+	fn batch(&self, batch: Batch<'a>) -> Self::Future {
+		Box::pin(self.service.batch(batch))
```
jsdw, line 106: Same, ie no rate limit for the batch call?

examples/examples/jsonrpsee_as_service.rs
jsdw: Actually, if the middleware is a part of a layer like this one (ie it wraps an inner servicxe), then users should impl `batch` to route to the inner services's `batch` call, else nothing else would know about the `batch` any more. ie

```rust
struct MyLayer<S>(S)

impl <S: RpcServiceT> RpcServiceT for MyLayer<S> {
    fn call(&self, request: Request<'a>) -> Self::Future {
        println!("do something with call, then forward to inner");
        self.inner.call(request);
    }
    fn notification(&self, notif) -> Self::Future {
        println!("do something with notification, then forward to inner");
        self.inner.notification(notif);
    }
    fn batch(&self, batch) -> Self::Future {
        println!("do something with batch, then forward to inner");
        self.inner.batch(batch);
    }
} 
```

Forgetting to impl `batch` here would mean that the inner servuce would only have `call` and `notification` methods called, ie it would lose awareness of the batch, which isn't what we'd want from a middleware

So I guess for most (all?) middlewares (ie ones that wrap other middlewares) we would need to impl batch manually anyway...

Maybe there can just be a helper method or something so that we can do something like:

```rust
struct MyLayer<S>(S)

impl <S: RpcServiceT> RpcServiceT for MyLayer<S> {
    fn call(&self, request: Request<'a>) -> Self::Future {
        println!("do something with call, then forward to inner");
        self.inner.call(request);
    }
    fn notification(&self, notif) -> Self::Future {
        println!("do something with notification, then forward to inner");
        self.inner.notification(notif);
    }
    fn batch(&self, batch) -> Self::Future {
        println!("do something with batch, then forward to inner");

        let batch_response = handle_calls_and_notifs_in_batch(self);

        if batch_response.is_error() {
            batch_response
        } else {
            self.inner.batch(batch);
        }
    }
} 
```

Not sure!

jsdw: The problem with the above idea though is that we fully resolve each call/notificaiton using the inner service and then do the same again with the batch, so that's no good either!

I guess we can't have any default impl, and people just need to write the logic in a reusable way eg

```rust
struct MyLayer<S>(S)

impl <S: RpcServiceT> RpcServiceT for MyLayer<S> {
    fn call(&self, request: Request<'a>) -> Self::Future {
        handle_call(&request);
        self.inner.call(request);
    }
    fn notification(&self, notif) -> Self::Future {
        handle_notification(&request)
        self.inner.notification(notif);
    }
    fn batch(&self, batch) -> Self::Future {
        // Maybe we can have a simple helper/method on Batch that just iterates the 
        // batch and runs some logic on each request/notif, returning the first error or () if all good? 
        if let Err(err) = handle_entries_in_batch(handle_call, handle_notification) {
            return err
        }

        self.inner.batch(batch);
    }
}

// middleware specific logic
fn handle_call(..) {}
fn handle_notification(..) {}
```

client/http-client/src/client.rs
```diff
@@ -377,34 +378,15 @@ where
...
-			Ok(Err(e)) => {
-				return Err(Error::Transport(e.into()));
-			}
-		};
-
-		// NOTE: it's decoded first to `JsonRawValue` and then to `R` below to get
-		// a better error message if `R` couldn't be decoded.
-		let response = ResponseSuccess::try_from(serde_json::from_slice::<Response<&JsonRawValue>>(&body)?)?;
+		let request = Request::borrowed(method, params.as_deref(), id.clone());
+		let rp =
```
niklasad1, line 382: /cc @jsdw this into_method_call() is the annoyance I was talking about

core/src/middleware/mod.rs
```diff
@@ -0,0 +1,353 @@
...
+	Call(Request<'a>),
+	/// A JSON-RPC notification.
+	Notification(Notification<'a>),
+	/// The representation of an invalid request and kept to maintain the batch order.
+	///
+	/// WARNING: This is an internal state and should be NOT be used by external users
+	///
+	#[serde(skip)]
+	#[doc(hidden)]
+	InvalidRequest(Id<'a>),
```
niklasad1, line 148: /cc @jsdw this is annoying internal state I was talking about

examples/examples/rpc_middleware_client.rs
```diff
@@ -0,0 +1,178 @@
...
+	}
+}
+
+impl<S> Metrics<S> {
+	pub fn new(service: S) -> Self {
+		Self { service, metrics: Arc::new(Mutex::new(InnerMetrics::default())) }
+	}
+}
+
+// NOTE: We are MethodResponse as the response type here to be able to inspect the response
```
niklasad1:
```suggestion
// NOTE: We are using MethodResponse as the response type here to be able to inspect the response
```

client/http-client/src/client.rs
jsdw: Aah gotcha! Yeah the only "nice" fix would be if we can use `impl Future<Output = Self::CallResponse>` etc in the trait and then have a different type for method, call and batch response or whatever.

This is all good for now though; hard to avoid without overcomplicating things unless `impl Trait` works ok!

server/src/server.rs
```diff
@@ -1257,53 +1275,40 @@ where
...
 						Ok(err) => err.id,
 						Err(_) => Id::Null,
 					};
 
-					if let Err(too_large) =
-						batch_response.append(MethodResponse::error(id, ErrorObject::from(ErrorCode::InvalidRequest)))
-					{
-						return Some(too_large);
-					}
+					batch.push(BatchEntry::InvalidRequest(id));
```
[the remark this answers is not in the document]
jsdw: Ah I see now!

Yeah I don't think it's too bad having this variant in BatchEntry which represents an erroneous entry I guess. You could also preerve a `Box<RawValue>` for any entry that can't be parsed and have the variant be something like just `Invalid` I guess! That way, middleware can handle the invalid looking entry in some way if it chooses. But yeah, no strong feelings on that; just thinking about preserving data in middlewares I guess.

Alternately, you give the middleware a `Batch::from_batch_entries` and it could be that `BatchEntry` doesn't support/expose any "invalid" stuff, but internally in `Batch` we store eg `Option<BatchEntry>` to denote any valid ones (or `Result<BatchEntry, InvalidEntry>` if we want to preserve stuff). Then we can expose whatever it is we want to the middlewares without "polluting" a public enum.

core/src/tracing.rs
```diff
@@ -99,7 +99,7 @@ pub mod server {
 }
 
 /// Find the next char boundary to truncate at.
-fn truncate_at_char_boundary(s: &str, max: usize) -> &str {
+pub fn truncate_at_char_boundary(s: &str, max: usize) -> &str {
```
jsdw, line 102: Is this useful outside the crate now then? Is it exposed to help people writing custom loggers?

examples/examples/core_client.rs
```diff
@@ -42,7 +42,7 @@ async fn main() -> anyhow::Result<()> {
 	let uri = Url::parse(&format!("ws://{}", addr))?;
 
 	let (tx, rx) = WsTransportClientBuilder::default().build(uri).await?;
-	let client: Client = ClientBuilder::default().build_with_tokio(tx, rx);
+	let client: Client<_> = ClientBuilder::default().build_with_tokio(tx, rx);
```
jsdw, line 45: Shame about the new generic; I guess the default param doesn't help here!

examples/examples/http.rs
```diff
@@ -29,13 +29,41 @@ use std::time::Duration;
...
+use jsonrpsee::core::middleware::{Batch, Notification, RpcServiceBuilder, RpcServiceT};
 use jsonrpsee::http_client::HttpClient;
 use jsonrpsee::rpc_params;
 use jsonrpsee::server::{RpcModule, Server};
+use jsonrpsee::types::Request;
 use tower_http::LatencyUnit;
 use tower_http::trace::{DefaultMakeSpan, DefaultOnResponse, TraceLayer};
 use tracing_subscriber::util::SubscriberInitExt;
 
+struct Logger<S>(S);
```
jsdw, line 41: Clean! There is a Logger middleware implemented in the `rpc_middleware.rs` example too. What's the thinking abotu adding one here?

examples/examples/rpc_middleware.rs
```diff
@@ -56,68 +61,110 @@ use jsonrpsee::ws_client::WsClientBuilder;
 pub struct CallsPerConn<S> {
```
jsdw, line 61: Maybe improve the comment eg say what this middleware will do?

examples/examples/rpc_middleware.rs
```diff
@@ -56,68 +61,110 @@ use jsonrpsee::ws_client::WsClientBuilder;
...
+		Box::pin(self.service.batch(batch))
+	}
+
+	fn notification(&self, n: Notification<'a>) -> Self::Future {
+		Box::pin(self.service.notification(n))
+	}
 }
 
 #[derive(Clone)]
 pub struct GlobalCalls<S> {
```
jsdw, line 102: Same; a comment about what this middleware does would make the example clearer I reckon

examples/examples/rpc_middleware.rs
```diff
@@ -56,68 +61,110 @@ use jsonrpsee::ws_client::WsClientBuilder;
...
+	}
+
+	fn notification(&self, n: Notification<'a>) -> Self::Future {
+		Box::pin(self.service.notification(n))
+	}
 }
 
 #[derive(Clone)]
-pub struct Logger<S>(S);
+pub struct Logger<S> {
```
jsdw, line 144: And here (though it's more obvious what this one does from the name, but hey)

examples/examples/rpc_middleware.rs
```diff
@@ -132,7 +179,14 @@ async fn main() -> anyhow::Result<()> {
 	let url = format!("ws://{}", addr);
 
 	for _ in 0..2 {
-		let client = WsClientBuilder::default().build(&url).await?;
+		let global_cnt = Arc::new(AtomicUsize::new(0));
+		let rpc_middleware = RpcServiceBuilder::new()
```
jsdw, line 183: Maybe a comment here like

```rust
// Make use of the middleware we defined in the client
```

examples/examples/rpc_middleware.rs
```diff
@@ -149,11 +203,11 @@ async fn run_server() -> anyhow::Result<SocketAddr> {
 	let global_cnt = Arc::new(AtomicUsize::new(0));
 
 	let rpc_middleware = RpcServiceBuilder::new()
```
jsdw, line 205: And then something like:

```rust
// We can also use the same middleware in the server, too..
```

To emphasise how we can now use the same middleware in both places

examples/examples/rpc_middleware_modify_request.rs
```diff
@@ -57,6 +59,14 @@ where
 
 		self.0.call(req)
 	}
+
+	fn batch(&self, batch: Batch<'a>) -> Self::Future {
+		self.0.batch(batch)
```
jsdw, line 64: A valid impl might be good here to show that you can't ignore methods in batches either?

(or a comment like `// Check for the same method in batches too`).

Implementing the logic for `call` and `batch` in these exmaples might lead to some ideas or whatever to make it easier to reuse logic across the two places, because that's a new challenge for people with this API that we might be able to expose some helper method or whatever to help with :)

examples/examples/rpc_middleware_rate_limiting.rs
```diff
@@ -126,6 +129,14 @@ where
 			ResponseFuture::future(self.service.call(req))
 		}
 	}
+
+	fn batch(&self, batch: Batch<'a>) -> Self::Future {
+		ResponseFuture::future(self.service.batch(batch))
```
jsdw, line 134: Batches aren't rate limited; also need an impl or comment here :)

examples/examples/server_with_connection_details.rs
```diff
@@ -53,13 +51,23 @@ struct LoggingMiddleware<S>(S);
...
-	fn call(&self, request: jsonrpsee::types::Request<'a>) -> Self::Future {
+	fn call(&self, request: Request<'a>) -> Self::Future {
 		tracing::info!("Received request: {:?}", request);
 		assert!(request.extensions().get::<ConnectionId>().is_some());
 
 		self.0.call(request)
 	}
+
+	fn batch(&self, batch: Batch<'a>) -> Self::Future {
+		self.0.batch(batch)
```
jsdw, line 65: Comment / impl to handle batches

proc-macros/tests/ui/correct/basic.rs
```diff
@@ -152,11 +152,21 @@ pub async fn server() -> SocketAddr {
...
+		type Error = S::Error;
+		type Response = S::Response;
 
 		fn call(&self, mut request: jsonrpsee::types::Request<'a>) -> Self::Future {
 			request.extensions_mut().insert(self.connection_id);
 			self.inner.call(request)
 		}
+
+		fn batch(&self, batch: Batch<'a>) -> Self::Future {
+			self.inner.batch(batch)
```
jsdw, line 164: Also here?

jsdw (commented):
I thoguht the comments were added individually but her; VSCode is a bit buggy leaving them!

Looks good overall to me! Some things to ponder, most which we chatted about already but just to sum up here:
- Can we improve the `RpcServiceTrait` to use async fn or `impl Future` by eg moving lifetime to inner methods from trait.
    - If this is possible; does it make sense to have a separate type for Call/Batch/Notification so  we don't have an enum to extract/expect thngs in
- When implementing `RpcServiceT`, can we do something to help people about needing to duplicate logic across calls & batches. A helper fn maybe? Doing it in exampels might lead to a good pattern or whatever to use here
- Are you happy that we're testing batch/notification enough or anything we should add here?
- Can that `InvalidId` variant on the batch enum be removed and instead we get a `Result` back and then hide things inside the `Batch` struct and give whatever API we want back? (or just expose the individual results). Should the error include the RawValue that we tried and failed to deserialize so that middlewares can log/report the failure with the value that failed or whatever?
- Should we use `Box<RawValue>` in more places rather than `String`?

client/http-client/src/client.rs
```diff
@@ -473,22 +451,17 @@ where
...
-	S: Service<HttpRequest, Response = HttpResponse<B>, Error = TransportError> + Send + Sync + Clone,
-	<S as Service<HttpRequest>>::Future: Send,
-	B: http_body::Body<Data = Bytes> + Send + Unpin + 'static,
-	B::Data: Send,
-	B::Error: Into<BoxError>,
+	for<'a> S: RpcServiceT<'a, Error = Error, Response = MethodResponse> + Send + Sync,
 {
 	/// Send a subscription request to the server. Not implemented for HTTP; will always return
 	/// [`Error::HttpNotImplemented`].
-	#[instrument(name = "subscription", fields(method = _subscribe_method), skip(self, _params, _subscribe_method, _unsubscribe_method), level = "trace")]
```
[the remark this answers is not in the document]
niklasad1: yeah, I remove them because it added duplicate spans both this one and the logging middleware

client/http-client/src/transport.rs
```diff
@@ -119,10 +117,10 @@ impl HttpTransportClientBuilder<Identity> {
 			certificate_store: CertificateStore::Native,
 			max_request_size: TEN_MB_SIZE_BYTES,
 			max_response_size: TEN_MB_SIZE_BYTES,
-			max_log_length: 1024,
 			headers: HeaderMap::new(),
 			service_builder: tower::ServiceBuilder::new(),
 			tcp_no_delay: true,
+			request_timeout: Duration::from_secs(60),
```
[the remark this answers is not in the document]
niklasad1: reverted this, let's keep in the client

Cargo.toml
```diff
@@ -98,3 +98,6 @@ rust_2024_compatibility = { level = "warn", priority = -1 }
 missing_docs = { level = "warn", priority = -1 }
 missing_debug_implementations = { level = "warn", priority = -1 }
 missing_copy_implementations = { level = "warn", priority = -1 }
+
+[workspace.lints.clippy]
+manual_async_fn = { level = "allow", priority = -1 }
```
niklasad1, line 103:
```suggestion
manual_async_fn = { level = "allow", priority = -1 }

```

core/src/tracing.rs
niklasad1: I moved the this the rpc logger stuff so not needed to be public

examples/examples/core_client.rs
niklasad1: it's fixed now, needless type hint

examples/examples/http.rs
niklasad1: my bad, let's revert this

examples/examples/jsonrpsee_as_service.rs
```diff
@@ -70,26 +75,97 @@ struct AuthorizationMiddleware<S> {
...
+impl<S> RpcServiceT for AuthorizationMiddleware<S>
 where
-	S: Send + Clone + Sync + RpcServiceT<'a>,
+	S: Send + Clone + Sync + RpcServiceT<Response = MethodResponse, Error = Infallible> + 'static,
+	S::Error: Into<BoxError> + Send + 'static,
+	S::Response: Send,
 {
-	type Future = ResponseFuture<S::Future>;
+	type Error = S::Error;
+	type Response = S::Response;
```
jsdw, line 113: This could be a followup thing to try out but: I wonder whether it's worth now having a separate response type for each call, so that there is no enum?

eg

``rust
type CallResponse = ...
type BatchResponse = ...
type NotificationResponse = ...
```

Comment (jsdw) [reply examples/examples/jsonrpsee_as_service.rs:113]: (though maybe this prevents the middleware from doing certain things, ie returning a MethodResponse::Error in response to a call or whatever?)

File: examples/examples/jsonrpsee_as_service.rs [hunk 25]
```diff
@@ -70,26 +75,97 @@ struct AuthorizationMiddleware<S> {
...
+		let entries: Vec<_> = batch
+			.into_iter()
+			.filter_map(|entry| match entry {
+				Ok(BatchEntry::Call(req)) => {
+					if self.auth_method_call(&req) {
+						Some(Ok(BatchEntry::Call(req)))
+					} else {
+						// If the authorization header is missing, we return
+						// a JSON-RPC error instead of an error from the service.
+						Some(Err(ErrorResponse::new(req.id, auth_reject_error())))
```
jsdw, line 143: How does `Err(ErrorResponse::new(req.id, auth_reject_error()))` differ from `Ok(MethodResponse::error(req.id, auth_reject_error()))`? Do they both lead to the same behaviour (ie a JSONRPC error being returned for that call)?

examples/examples/jsonrpsee_server_low_level_api.rs
```diff
@@ -89,14 +90,47 @@ where
...
+		async move {
+			let mut lock = count.lock().await;
+
+			if *lock >= 10 {
+				let _ = state.try_send(());
+				Ok(MethodResponse::error(Id::Null, ErrorObject::borrowed(-32000, "RPC rate limit", None)))
+			} else {
+				let batch_len = batch.len();
+				let rp = service.batch(batch).await;
+				*lock += batch_len;
```
jsdw, line 116: What if the batch len is like 100?

Would it make sense to do something like "if *lock + batch.len() >= 10"?

examples/examples/jsonrpsee_as_service.rs
```diff
@@ -70,26 +75,97 @@ struct AuthorizationMiddleware<S> {
...
 
-			ResponseFuture::future(self.inner.call(req))
-		} else {
-			ResponseFuture::future(self.inner.call(req))
-		}
+	fn notification<'a>(
+		&self,
+		n: Notification<'a>,
+	) -> impl Future<Output = Result<Self::Response, Self::Error>> + Send + 'a {
+		self.inner.notification(n)
```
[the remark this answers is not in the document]
jsdw: `Ok(MethodResponse::notification())`?

examples/examples/rpc_middleware_rate_limiting.rs
```diff
@@ -78,54 +78,73 @@ impl<S> RateLimit<S> {
...
+
+impl<S> RpcServiceT for RateLimit<S>
+where
+	S: Send + RpcServiceT<Response = MethodResponse> + 'static,
+	S::Error: Send,
+{
+	type Error = S::Error;
+	type Response = S::Response;
+
+	fn call<'a>(&self, req: Request<'a>) -> impl Future<Output = Result<Self::Response, Self::Error>> + Send + 'a {
```
jsdw, line 116: I haven't tried Rust 2024 yet but it has this new `use` syntax, so I wonder whether it's more correct to have `+ use<'a>` in the response rather than `+ 'a`? I always forget the subtleties around why `+ 'a` isn't quite the same thing!

tests/tests/metrics.rs
```diff
@@ -88,14 +86,24 @@ where
...
+	fn batch<'a>(&self, batch: Batch<'a>) -> impl Future<Output = Result<Self::Response, Self::Error>> + Send + 'a {
+		self.service.batch(batch)
+	}
+
+	fn notification<'a>(
+		&self,
+		n: Notification<'a>,
+	) -> impl Future<Output = Result<Self::Response, Self::Error>> + Send + 'a {
+		self.service.notification(n)
 	}
```
jsdw, line 107: If these aren't necessary, maybe better to `panic!("not used in tests")` or something, so it's super clear in case we ever add a test which makes a batch call or whatever and get confused why nothing changes?

types/src/error.rs
```diff
@@ -145,6 +145,8 @@ pub const SERVER_IS_BUSY_CODE: i32 = -32009;
 pub const TOO_BIG_BATCH_REQUEST_CODE: i32 = -32010;
 /// Batch response limit was exceed.
 pub const TOO_BIG_BATCH_RESPONSE_CODE: i32 = -32011;
+/// Middleware error code.
+pub const MIDDLEWARE_ERROR_CODE: i32 = -32012;
```
jsdw, line 149: I'm a bit on the fence wondering whether there should be an error specific to middleware, or whether middlewares should emit "normal" errors or use custom codes or whatever. Like, when would one want to return a "middleware error" rather than an "auth error" or "rate limit error" or something?

types/src/lib.rs
```diff
@@ -44,6 +44,95 @@ pub mod response;
...
 
 pub use error::{ErrorCode, ErrorObject, ErrorObjectOwned};
+pub use http::Extensions;
 pub use params::{Id, InvalidRequestId, Params, ParamsSequence, SubscriptionId, TwoPointZero};
-pub use request::{InvalidRequest, Notification, NotificationSer, Request, RequestSer};
+pub use request::{InvalidRequest, Notification, Request};
 pub use response::{Response, ResponsePayload, SubscriptionPayload, SubscriptionResponse, Success as ResponseSuccess};
+
+/// Deserialize calls, notifications and responses with HTTP extensions.
+pub mod deserialize_with_ext {
```
jsdw, line 53: These only appear to be used in one place, so I wonder if it's worth just having the logic there rather than exposing it form this crate (which makes it a public API)? Maybe it's useful for others though and not just an internal thing?

jsdw (approved):
Nice, LGTM! left a few comments/nits that are worth looking at before merging but nothing too major :)

types/src/error.rs
niklasad1: This shouldn't occur since we are using `Infallible` all over the place but if someone decides to return error from "their own middleware impl" this may occur.

We can probably just do a custom error and remove this, probably for the best.

niklasad1: enforced this requiring Error = Infallible in the server

examples/examples/jsonrpsee_as_service.rs
niklasad1: Let's file an issue for this, it's not trivial to break up the method response into different types in the server

examples/examples/jsonrpsee_as_service.rs
niklasad1: This must be outdated code but returning an error from the RpcServiceT would lead to that we would actually catch in the server/RpcService impl and convert into a JSON-RPC error which is annoying.

I have restricted this now such that all RpcServiceT must use Error = Infallible, it much better than that we provide an error message from the Debug/String impl from the Error type.

examples/examples/jsonrpsee_server_low_level_api.rs
niklasad1: yupp :)

server/src/middleware/rpc.rs
```diff
@@ -138,13 +137,65 @@ impl<'a> RpcServiceT<'a> for RpcService {
...
+				Ok(MethodResponse::from_batch(batch_rp.finish()))
+			}
+		}
+	}
+
+	fn notification<'a>(
+		&self,
+		_: Notification<'a>,
+	) -> impl Future<Output = Result<Self::Response, Self::Error>> + Send + 'a {
+		async move { Ok(MethodResponse::notification()) }
```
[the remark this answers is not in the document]
niklasad1: It's simply ignore by server but since we are ACK:ing on HTTP it could make sense to set-up the extensions in the response.

examples/examples/rpc_middleware_rate_limiting.rs
niklasad1: we need https://github.com/rust-lang/rust/pull/138128 for that to work with impl Trait<Self::Foo>

server/src/middleware/rpc.rs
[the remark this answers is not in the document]
jsdw: Ah yeah gotcha!

examples/examples/rpc_middleware_rate_limiting.rs
jsdw: Ooh I didn't realise it wasn't stable for traits yet, but yay for that PR having merged!

examples/examples/jsonrpsee_as_service.rs
jsdw: Ah I see it's BatchEntry and BatchEntryErr; that works for me and feels clearer :)

examples/examples/jsonrpsee_as_service.rs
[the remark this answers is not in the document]
niklasad1: it's doesn't matter that much but better the call the inner service if extensions are provided in the notification..

Issue #1520, opened 2025-01-21
reduce MSRV to 1.74.0


king-11:
cargo msrv generally returns a version higher that what actually works, I tried running things with 1.74.0 and it worked fine maybe the ci can have an additional rust:1.74.0 based step.

reproduction
```bash
sed -i 's/1\.74\.1/1.74.0/g' Cargo.toml
rustup override set 1.74.0
cargo build
cargo test
```

king-11:
@niklasad1 any thoughts on having an msrv step in ci?

king-11:
Sure I can work on that :)

king-11:
@niklasad1 is it possible to get an minor release with new MSRV we make use of jsonrpsee in our open source library and aren't able to upgrade it to latest due to MSRV restrictions on our project till `1.74.0` only.

niklasad1:
sure

niklasad1:
Released v0.24.8 with this fix

king-11:
Thanks a lot :) appreciate it

PR #1522, opened 2025-01-24
ci: add msrv check job

king-11:
related to #1520

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -111,6 +111,7 @@ jobs:
         uses: actions-rs/toolchain@v1.0.7
         with:
           profile: minimal
+          # Check that tests pass using our MSRV
           toolchain: 1.74.0
           override: true
 
@@ -120,8 +121,8 @@ jobs:
       - name: Cargo test
         uses: actions-rs/cargo@v1.0.3
         with:
-          command: test
-          args: --workspace
+          command: check
+          args: --all-targets
 
   tests_ubuntu:
     name: Run tests Ubuntu
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,6 +100,29 @@ jobs:
       - name: Check unused dependencies
         run: cargo machete
 
+  check-msrv:
+    name: Check
+    runs-on: ubuntu-latest
+    steps:
+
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -100,6 +100,29 @@ jobs:
       - name: Check unused dependencies
         run: cargo machete
 
+  check-msrv:
+    name: Check
```
niklasad1, line 104:
```suggestion
    name: Check MSRV
```

.github/workflows/ci.yml
```diff
@@ -100,6 +100,29 @@ jobs:
       - name: Check unused dependencies
         run: cargo machete
 
+  check-msrv:
```
jsdw, line 103: Am I missing how this is checking MSRV? It looks like all it does is run `carrgo test`, same as the below workflow.

.github/workflows/ci.yml
```diff
@@ -100,6 +100,29 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+
+      - uses: actions/checkout@v4.2.2
+
+      - name: Install Rust 1.74.0 toolchain
+        uses: actions-rs/toolchain@v1.0.7
+        with:
+          profile: minimal
+          toolchain: 1.74.0
```
jsdw, line 114: I missed this at first so maybe just to make it obvious:

```suggestion
          # Check that tests pass using our MSRV:
          toolchain: 1.74.0
```

niklasad1 (commented):
Looks overall good but the UI tests won't pass for all rustc compilers down to 1.74.0

Either just type-check the code with 1.74.0 or disable the tests in the `proc-macro-crate`.

king-11:
@niklasad1 what do you mean by type check?

niklasad1:
> @niklasad1 what do you mean by type check? As in build?

`cargo check --all-targets` but cargo build would work as well doesn't matter to me

.github/workflows/ci.yml
king-11: updated

king-11:
should be good now sorry for delay was in phase of changing cities

niklasad1 (approved):

PR #1530, opened 2025-02-25
Add assert for server: set_message_buffer_capacity

### LINKED ISSUE #1526: server: `set_message_buffer_capacity` missing assert for capacity==0
In jsonrpsee-master/server/src/server.rs, may be the code should add an assert! like comment requires.
```rust
/// # Panics
	///
	/// Panics if the buffer capacity is 0.
	///
	pub fn set_message_buffer_capacity(mut self, c: u32) -> Self {
		self.message_buffer_capacity = c;
		self
	}
```

YichiZhang0613:
Fix #1526

### DIFF
--- a/server/src/server.rs
+++ b/server/src/server.rs
@@ -458,6 +458,7 @@ impl ServerConfigBuilder {
 	/// Panics if the buffer capacity is 0.
 	///
 	pub fn set_message_buffer_capacity(mut self, c: u32) -> Self {
+		assert!(c > 0);
 		self.message_buffer_capacity = c;
 		self
 	}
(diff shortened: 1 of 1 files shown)


niklasad1 (approved):

server/src/server.rs
```diff
@@ -458,6 +458,7 @@ impl ServerConfigBuilder {
 	/// Panics if the buffer capacity is 0.
 	///
 	pub fn set_message_buffer_capacity(mut self, c: u32) -> Self {
+		assert!(c > 0);
```
jsdw, line 461:
```suggestion
		assert!(c > 0, "buffer capacity must be set to >0");
```

Maybe?

jsdw (approved):

Issue #1524, opened 2025-02-09
feature request: make request ID generation on client configurable


torkelrogstad:
I want to be able to generate my own request IDs on the client. Currently this is done by `RequestIdManager`, which monotonically increases IDs by 1 for each request. I'm writing a CLI application. This application does precisely one RPC call before exiting. This means that for every invocation of the CLI application the request ID is 0, which is kind of meaningless. 

It'd be nice to be able to specify my own request ID creator, in which I could generate a random number that actually would give each request a unique identifier.

PanGan21:
Hey @niklasad1  I am wondering if this issue is open for contribution! Let me know!

niklasad1:
@PanGan21 Yes, this issue is open for contribution.

PanGan21:
Hey @niklasad1  I did some work on this in #1535 . I would appreciate some feedback 🙂

PR #1535, opened 2025-03-12
feat(request): make request ID generation on client configurable

PanGan21:
The following pr:
- Adds `Custom` variant on `IdKind`
- Adds an example that sets up a custom request id generator

Related: #1524


niklasad1 (commented):
Hey hey,

Thanks for your PR and it looks generally good but I had in mind to add a the "custom generator" to the IdKind enum to avoid adding extra APIs for this.  After this change one can configure the `id_format` with `ClientBuilder::id_format` and set a custom generator with `ClientBuilder::with_request_id_generator`. This means that `with_request_id_generator` overrides the configured id_format.

Thus, I would really like to unify them and perhaps replace these APIs with `set_id_generator` or something similar.

I reckon that we may need to refactor things and remove the  `next_id APIs` with a similar generator impl as you added.

So, WDYT?

PanGan21:
> Hey hey,
> 
> 
> 
> Thanks for your PR and it looks generally good but I had in mind to add a the "custom generator" to the IdKind enum to avoid adding extra APIs for this.  After this change one can configure the `id_format` with `ClientBuilder::id_format` and set a custom generator with `ClientBuilder::with_request_id_generator`. This means that `with_request_id_generator` overrides the configured id_format.
> 
> 
> 
> Thus, I would really like to unify them and perhaps replace these APIs with `set_id_generator` or something similar.
> 
> 
> 
> I reckon that we may need to refactor things and remove the  `next_id AP` with a similar generator impl as you added.
> 
> 
> 
> So, WDYT?

Thanks for the direction!
I will start the refactor as you described! 
If necessary I might ask for some pointers since I am new to the repo.

niklasad1:
> If necessary I might ask for some pointers since I am new to the repo.

All good, just ping us if you are blocked or need some extra help :)

core/src/client/mod.rs
```diff
@@ -485,14 +486,17 @@ pub enum IdKind {
...
 	/// Number.
 	Number,
+	/// Custom generator.
+	Custom(IdGeneratorFn),
 }
 
 impl IdKind {
-	/// Generate an `Id` from number.
+	/// Generate an `Id` from number or from a registered generator.
 	pub fn into_id(self, id: u64) -> Id<'static> {
```
PanGan21, line 495: Hey @niklasad1 
I tried to not refactor too much as a first approach.
For now the user can set the custom generator with `ClientBuilder::id_format` but I believe the `into_id` function is  unnecessarily receiving a new id as an argument in case the IdKind is custom.

In your opinion how would `into_id` function signature change?

niklasad1: Do you think we can remove this method completely and replace `IdKind::Number` and `IdKind::String` with generators as well but that takes state/has state?

It could also be an iterator-like API where we just call next() on it.

PanGan21: So the user when calling `ClientBuilder::id_format` should provide a generator or the default one for Number and String right?

niklasad1: yes but we may need to change the name for it but I'm not sure.

PanGan21: The only reason I dislike the Iterator pattern here is that we are forced to return an Option from next so also from next_id when there is no real need to return an option

niklasad1: You don't need use the Iterator API, if you have something better in mind go ahead (it was just an example)

It's really up to you what you think is the best.

core/src/client/mod.rs
```diff
@@ -485,16 +496,8 @@ pub enum IdKind {
 	String,
 	/// Number.
 	Number,
-}
-
-impl IdKind {
```
PanGan21, line 490: I tried changing this to:

```
enum IdKind {
    Number(NumberGenerator),
    String(StringGenerator),
    Custom(CustomGenerator),
}

impl IdKind {
    fn next_id(&mut self) -> Id<'static> {
        match self {
            Self::Number(n) => n.next(),
            Self::String(s) => s.next(),
            Self::Custom(g) => g.next(),
        }
    }
}
```

It seems that the ergonomics are not very good here. I faced the following issues:
- `IdKind` needs to implement Copy trait because it comes all the way from the Client as a requirement. The issue is that if it has `CurrentId` state, it cannot implement `Copy` trait due to `AtomicUsize` usage
- In case I used u64 instead of `AtomicUsize`, `next_rqeuest_id` and `into_id` should accept a mutable reference which forces mutable references all the way to the client.

client/http-client/src/client.rs
```diff
@@ -418,8 +418,8 @@ where
 		let id_range = generate_batch_id_range(id, batch.len() as u64)?;
 
 		let mut batch_request = Vec::with_capacity(batch.len());
-		for ((method, params), id) in batch.into_iter().zip(id_range.clone()) {
-			let id = self.id_manager.as_id_kind().into_id(id);
+		for ((method, params), _id) in batch.into_iter().zip(id_range.clone()) {
```
niklasad1, line 421: revert this, I don't see reason to change this.

for batches we are generating (next_id..next_id+batch_len)

PanGan21: I am working on some improvements on the design to make batching possible again

niklasad1: Ok, sorry about that I didn't realize it until I saw the code again

PanGan21: No problem at all. I made it a member function of `RequestId`. Let me know what you think

core/src/client/async_client/mod.rs
```diff
@@ -536,8 +536,8 @@ impl ClientT for Client {
 		let id_range = generate_batch_id_range(id, batch.len() as u64)?;
 
 		let mut batches = Vec::with_capacity(batch.len());
-		for ((method, params), id) in batch.into_iter().zip(id_range.clone()) {
-			let id = self.id_manager.as_id_kind().into_id(id);
+		for ((method, params), _id) in batch.into_iter().zip(id_range.clone()) {
```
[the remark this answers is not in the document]
PanGan21: Fixed and refactored batch id generation

client/http-client/src/client.rs
```diff
@@ -414,12 +414,11 @@ where
 			None => None,
 		};
 		let batch = batch.build()?;
-		let id = self.id_manager.next_request_id();
-		let id_range = generate_batch_id_range(id, batch.len() as u64)?;
+		let id_range = self.id_manager.generate_batch_id_range(batch.len());
```
niklasad1, line 417: Hey again,

So it doesn't make sense to have this `id_range` anymore because we are using it to check whether all 
requests in a batch was replied to.

Now in the code the actual ID is fetched from the `id_manager`, thus it's not necessarily true that the `id_range` and `self.id_manager.next_request_id()` contains the same ids.

Additionally, because it's possible to use a custom id generator we can't really on these ID ranges anymore anyway. 

I reckon that we need to remove the `generate_batch_id_range` completely and just push the generated IDs to a Vec such as:

```rust
// this will contain the order of the request IDs in the batch
let mut ids = Vec::new();

for (method, params) in batch.into_iter() {
     let id = self.id_manager.next_request_id();
	 batch_request.push(RequestSer {
              jsonrpc: TwoPointZero,
		      id,
			  method: method.into(),
			  params: params.map(StdCow::Owned),
	});
}

// use ids to check whether to response has responded to all ids in the batch
```

The reason why we need this is because a server can submit respond to each call in the batch in arbitrary order.

niklasad1: The actual logic to match a batch response to the batch request must be also be changed because one assumption is that the every ID is generated from a integer and can be parsed as one which is used to index into the response.

An additionally test with an unordered batch response would be great to make sure everything works as intended.

PanGan21: Thanks for the feedback - makes sense.
According to that, the usage of `Range` type in `BatchMessage` and in related functions, should also be changed right?  It should probably be replaces with something like `Vec<Id<'static>>`.

niklasad1: yes, correct

PanGan21: FYI the refactor is in progress but I am facing some issues with the lifetimes. I need to either use `Vec<Id<'a>>` which means that a lifetime parameter will be added across a big part of the client or use `Vec<Id<'static>>` which is always not optimal.

niklasad1: just do the naive implementation now with `Vec<Id<'static>>` the default one is still number and cloning it should be "cheap".

core/src/client/async_client/manager.rs
```diff
@@ -223,14 +220,23 @@ impl RequestManager {
...
-			_ => None,
 		}
+
+		if let Some(key) = matched_key {
+			if let Some((_key, state)) = self.batches.remove_entry(&key) {
+				return Some(state);
+			}
+		}
+
+		None
```
PanGan21, line 239: Most likely this could improved if a different data structure is used to store batches on `RequestManager`

client/http-client/src/client.rs
PanGan21: Updated and test added

niklasad1:
cool cool, I'll have a proper look next week thanks for working on this

niklasad1:
Sorry @PanGan21 I'll take a look tomorrow :pray:

core/src/client/async_client/helpers.rs
```diff
@@ -53,36 +54,29 @@ pub(crate) struct InnerBatchResponse {
...
-			return Err(InvalidRequestId::NotPendingRequest(format!("{:?}", range)));
+			return Err(InvalidRequestId::NotPendingRequest(format!("{:?}", ids)));
 		}
 	};
 
-	for _ in range {
-		let err_obj = ErrorObject::borrowed(0, "", None);
-		responses.push(Err(err_obj));
-	}
+	let mut responses_map: BTreeMap<Id<'static>, Result<_, ErrorObject>> =
```
niklasad1, line 67: We can't use a BTreeMap here because there is nothing that prevents an ID (duplicate ids) to be used more than once with a custom ID generator

core/src/client/async_client/manager.rs
```diff
@@ -223,14 +220,23 @@ impl RequestManager {
...
 	/// Returns `Some` if the subscription was completed otherwise `None`.
-	pub(crate) fn complete_pending_batch(&mut self, batch: Range<u64>) -> Option<BatchState> {
-		match self.batches.entry(batch) {
-			Entry::Occupied(request) => {
-				let (_digest, state) = request.remove_entry();
-				Some(state)
+	pub(crate) fn complete_pending_batch(&mut self, batch: Vec<Id<'static>>) -> Option<BatchState> {
+		let mut matched_key = None;
+
+		for (key, _) in self.batches.iter() {
```
niklasad1, line 226: yeah, this is inefficient/annoying compared to the old code :thinking:

core/src/client/async_client/manager.rs
```diff
@@ -223,14 +220,23 @@ impl RequestManager {
...
+			if key.len() == batch.len() && batch.iter().all(|id| key.contains(id)) {
```
niklasad1, line 227: you could just use Vec<T> partialeq/eq impl here.

niklasad1 (commented):
Thanks for your PR (sorry about slow review though)

I'm a bit annoyed how inefficient the implementation becomes with custom id generator, by using `Vec<Id>` all over the place and to search for element in the batch response.

It's not possible to use BTreeMap/HashMap because we can't assume that duplicated IDs are not used by custom request ID generator.

So, what we could do is to sort `Vec<Id>` and then binary search through it

However, I'm not convinced that this annoying implementation is worth the effort for such a niche feature though

//cc @jsdw what do you think? Or any smarter ideas?

core/src/client/async_client/manager.rs
```diff
@@ -223,14 +220,23 @@ impl RequestManager {
...
+				matched_key = Some(key.clone());
+				break;
 			}
-			_ => None,
 		}
+
+		if let Some(key) = matched_key {
+			if let Some((_key, state)) = self.batches.remove_entry(&key) {
```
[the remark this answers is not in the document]
PanGan21: Fixed. Not using retain but a similar approach. Let me know if that works for you

PR #1540, opened 2025-03-20
chore: fix `http request body` log 

hai-rise:
Randomly found a log labelling a request body with "response" today :pray:.


niklasad1 (approved):
thanks, this is used by both the server and client so make sense to omit the response part

PR #1545, opened 2025-04-07
unify usage of JSON via Box<RawValue>

niklasad1:
Builds on-top of #1521 

This PR unifies that Box<RawValue> is used everywhere instead of having untypes JSON Strings with a the native Rust String type.

Some additional changes that I got annoyed/needed for this work properly:
- Rename StringError -> SubscriptionError
- SubscriptionError now supports both representing errors as Strings or JSON
- SubscriptionMessage, don't serialize internally let the user take care of that. This makes it easier to understand the API


benches/helpers.rs
```diff
@@ -150,8 +151,8 @@ pub async fn ws_server(handle: tokio::runtime::Handle) -> (String, jsonrpsee::se
 			UNSUB_METHOD_NAME,
 			|_params, pending, _ctx, _| async move {
 				let sink = pending.accept().await?;
-				let msg = SubscriptionMessage::from("Hello");
-				sink.send(msg).await?;
+				let json = RawValue::from_string("\"Hello\")".to_string()).unwrap();
```
jsdw, line 154: Nit: I wonder whether something like this API would be nicer:

https://docs.rs/serde_json/latest/serde_json/value/fn.to_raw_value.html

So eg `serde_json::value::to_raw_value("hello").unwrap()`

core/src/server/rpc_module.rs
```diff
@@ -326,14 +325,16 @@ impl Methods {
...
+	///     use jsonrpsee::core::JsonRawValue;
 	///     use futures_util::StreamExt;
 	///
 	///     let mut module = RpcModule::new(());
 	///     module.register_subscription("hi", "hi", "goodbye", |_, pending, _, _| async {
 	///         let sink = pending.accept().await?;
 	///
 	///         // see comment above.
-	///         sink.send("one answer".into()).await?;
+	///         let msg = JsonRawValue::from_string("\"one answer\"".into()).unwrap();
```
jsdw, line 336: same; maybe `to_raw_value` would be nicer to avoid the quotes etc

core/src/server/rpc_module.rs
```diff
@@ -422,12 +423,13 @@ impl Methods {
...
 	/// async fn main() {
 	///     use jsonrpsee::{RpcModule, SubscriptionMessage};
-	///     use jsonrpsee::core::{EmptyServerParams, RpcResult};
+	///     use jsonrpsee::core::{EmptyServerParams, RpcResult, JsonRawValue};
 	///
 	///     let mut module = RpcModule::new(());
 	///     module.register_subscription("hi", "hi", "goodbye", |_, pending, _, _| async move {
 	///         let sink = pending.accept().await?;
-	///         sink.send("one answer".into()).await?;
+	///         let msg = JsonRawValue::from_string("\"one answer\"".into()).unwrap();
```
jsdw, line 431: And here

proc-macros/src/lib.rs
```diff
@@ -247,7 +247,11 @@ pub(crate) mod visitor;
 ///                CloseResponse::None => SubscriptionCloseResponse::None,
 ///                // Send a close response as an ordinary subscription notification
 ///                // when the subscription is terminated.
-///                CloseResponse::Failed => SubscriptionCloseResponse::Notif("failed".into()),
+///                CloseResponse::Failed => {
+///                     const FAILED: &str = "\"Failed\"";
```
jsdw, line 251: `serde_json::value::to_raw_value` again to aboud double quotes?

core/src/error.rs
```diff
@@ -24,13 +24,49 @@
...
+	Json(Box<RawValue>),
+}
+
+/// Error returned when a subscription fails.
+///
+/// It's recommended to use `SubscriptionErr::from_json` to create a new instance of this error
+/// because using the `String` representation may not very ergonomic for clients to parse.
 #[derive(Debug)]
-pub struct StringError(pub(crate) String);
+pub struct SubscriptionError(pub(crate) InnerSubscriptionErr);
```
jsdw, line 41: I'm wondering; do we need both enums?

`from_string` could turn a string into a JSON encoded raw value, and from_json for more arbitrary error types.

Then, the generic From impl could also serialize any arbitrary type `T: Serialize` to JSON too if that's what we want to hand back?

(Right now, it's a bit confusing to me that we can convert arbitrary T's into strings and then show them as JSOn strings in an error. Is this used anywhere? Maybe I'm missing a case when we need this!)

jsdw:
Mostly LGTM!

I wonder if we can avoid all of the double quoted strings by using `serde_json::value::to_raw_value("some string")` instead (I didn't comment every place this was done because there are quite a lot :D).

Also a small Q about the subscriptionerror!

core/src/error.rs
niklasad1: I would be happy to remove the String stuff completely because it could be a fotgun.

I think https://github.com/paritytech/jsonrpsee/blob/add0baf6b01bd41064b25ad9ae6289de40f336ab/tests/tests/rpc_module.rs#L502-#L568 explains the different behavior JSON vs ToString

benches/bench.rs
```diff
@@ -113,7 +113,7 @@ pub fn jsonrpsee_types_v2(crit: &mut Criterion) {
 	// Construct the serialized array request using the `RawValue` directly.
 	crit.bench_function("jsonrpsee_types_array_params_baseline", |b| {
 		b.iter(|| {
-			let params = serde_json::value::RawValue::from_string("[1, 2]".to_string()).unwrap();
+			let params = serde_json::value::to_raw_value("[1, 2]").unwrap();
```
niklasad1, line 116:
```suggestion
			let params = serde_json::value::to_raw_value(&[1, 2]).unwrap();
```

core/src/params.rs
```diff
@@ -107,7 +108,7 @@ mod params_builder {
 		}
 
 		/// Finish the building process and return a JSON compatible string.
```
niklasad1, line 110:
```suggestion
		/// Finish the building process and return a JSON object.
```

core/src/traits.rs
```diff
@@ -50,7 +50,7 @@ use serde_json::value::RawValue;
 /// impl ToRpcParams for ManualParam {
 ///     fn to_rpc_params(self) -> Result<Option<Box<RawValue>>, serde_json::Error> {
 ///         // Manually define a valid JSONRPC parameter.
-///         RawValue::from_string("[1, \"2\", 3]".to_string()).map(Some)
+///         serde_json::value::to_raw_value("[1,2,3]").map(Some)
```
niklasad1, line 53:
```suggestion
///         serde_json::value::to_raw_value(&[1,2,3]).map(Some)
```

core/src/error.rs
niklasad1: you are proposing something like:

```git
--- a/core/src/error.rs
+++ b/core/src/error.rs
@@ -27,46 +27,14 @@
 use serde::Serialize;
 
-#[derive(Debug)]
-pub(crate) enum InnerSubscriptionErr {
-       String(String),
-       Json(Box<RawValue>),
-}
-
 /// Error returned when a subscription fails.
-///
-/// It's recommended to use `SubscriptionErr::from_json` to create a new instance of this error
-/// because using the `String` representation may not very ergonomic for clients to parse.
 #[derive(Debug)]
-pub struct SubscriptionError(pub(crate) InnerSubscriptionErr);
-
-impl Serialize for SubscriptionError {
-       fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
-       where
-               S: serde::Serializer,
-       {
-               match &self.0 {
-                       InnerSubscriptionErr::String(s) => serializer.serialize_str(s),
-                       InnerSubscriptionErr::Json(json) => json.serialize(serializer),
-               }
-       }
-}
+pub struct SubscriptionError(pub(crate) Box<RawValue>);
 
-impl<T: ToString> From<T> for SubscriptionError {
+impl<T: Serialize> From<T> for SubscriptionError {
        fn from(val: T) -> Self {
-               Self(InnerSubscriptionErr::String(val.to_string()))
-       }
-}
-
-impl SubscriptionError {
-       /// Create a new `SubscriptionErr` from a JSON value.
-       pub fn from_json(json: Box<RawValue>) -> Self {
-               Self(InnerSubscriptionErr::Json(json))
-       }
-
-       /// Create a new `SubscriptionErr` from a String.
-       pub fn from_string(s: String) -> Self {
-               Self::from(s)
+               let json = serde_json::value::to_raw_value(&val);
+               Self(json.unwrap_or_default())
        }
 }
```

It would be a PITA do require every error type returned from subscription callback to satisfy `T: Serialize` but would be nicer and more understable for use.

core/src/server/subscription.rs
```diff
@@ -522,21 +496,21 @@ pub struct SubscriptionState<'a> {
...
+pub(crate) fn sub_message_to_json(msg: SubscriptionMessage, sub_id: &SubscriptionId, method: &str) -> Box<RawValue> {
 	match msg.0 {
 		SubscriptionMessageInner::Complete(msg) => msg,
-		SubscriptionMessageInner::NeedsData(result) => {
-			let sub_id = serde_json::to_string(&sub_id).expect("valid JSON; qed");
-			format!(
-				r#"{{"jsonrpc":"2.0","method":"{method}","params":{{"subscription":{sub_id},"{result_or_err}":{result}}}}}"#,
-			)
-		}
+		SubscriptionMessageInner::NeedsData(result) => serde_json::value::to_raw_value(&SubscriptionResponse::new(
```
niklasad1, line 502: This was actually faster than the manual stuff above

```bash
jsonrpsee_types_serialize_sub_manual
                        time:   [177.53 ns 177.85 ns 178.23 ns]
  3 (3.00%) high severe

jsonrpsee_types_serialize_sub_serde
                        time:   [90.732 ns 91.579 ns 92.648 ns]
  3 (3.00%) high mild
  2 (2.00%) high severe

```

core/src/error.rs
jsdw: Makes sense after our chat; thanks :)

jsdw (approved):
Nice! Liking the to_raw_values and such :D

PR #1556, opened 2025-04-15
fix: version parsing in `scripts/publish.sh` 

mdqst:
Previously, the `read_toml` function always read the version from the root `Cargo.toml`, regardless of the current crate being processed in the loop. This was due to referencing `$DIR/Cargo.toml`, which pointed to the original working directory before `cd` into the sub-crate.

Switched to using `./Cargo.toml` to correctly read the version from each crate's local `Cargo.toml`.

Now it properly fetches the version for each submodule instead of reusing the root version.

### DIFF
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -6,7 +6,6 @@
 set -eu
 
 ORDER=(types proc-macros core client/http-client client/transport client/ws-client client/wasm-client server jsonrpsee)
-DIR=$(pwd)
 
 function read_toml () {
 	NAME=""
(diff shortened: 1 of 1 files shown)


niklasad1:
Ok, cool basically the issue is that we `DIR=$pwd` at start-up then using after doing `cd foo` it's not accurate anymore.

All good thanks for spotting this.

niklasad1 (approved):

niklasad1:
@mdqst bonus in this PR would be to remove https://github.com/paritytech/jsonrpsee/blob/master/scripts/publish.sh#L9 as well :)

niklasad1 (approved):

PR #1561, opened 2025-04-17
fix(examples): use SubscriptionError as error type in `drop_oldest_when_lagging`

mdqst:


niklasad1 (commented):
Hey, thanks for your PR

Can you please run `cargo +stable fmt --all` on this PR to make it easier to review?
Looks like you replaced tabs with spaces.

If it just changes the return error type, should be good to go.

niklasad1 (approved):

niklasad1:
Okey, this is just wrong we don't have a SubscriptionError exposed and this doesn't work.

Thus, closing this PR and feel free to improve the example when you have working code.

PR #1564, opened 2025-04-19
middleware: RpcServiceT distinct return types for notif, batch, call

niklasad1:
This PR modifies to RpcServiceT to:

```rust
pub trait RpcServiceT {
	/// Response type for `RpcServiceT::call`.
	type MethodResponse: ToJson;
	/// Response type for `RpcServiceT::notification`.
	type NotificationResponse: ToJson;
	/// Response type for `RpcServiceT::batch`.
	type BatchResponse: ToJson;

	/// Processes a single JSON-RPC call, which may be a subscription or regular call.
	fn call<'a>(&self, request: Request<'a>) -> impl Future<Output = Self::MethodResponse> + Send + 'a;

	/// Processes multiple JSON-RPC calls at once, similar to `RpcServiceT::call`.
	///
	/// This method wraps `RpcServiceT::call` and `RpcServiceT::notification`,
	/// but the root RPC service does not inherently recognize custom implementations
	/// of these methods.
	///
	/// As a result, if you have custom logic for individual calls or notifications,
	/// you must duplicate that implementation in this method or no middleware will be applied
	/// for calls inside the batch.
	fn batch<'a>(&self, requests: Batch<'a>) -> impl Future<Output = Self::BatchResponse> + Send + 'a;

	/// Similar to `RpcServiceT::call` but processes a JSON-RPC notification.
	fn notification<'a>(&self, n: Notification<'a>) -> impl Future<Output = Self::NotificationResponse> + Send + 'a;
}
```

This is an improvement because it's now possible to return different futures for the async operations but the downside is that it's harder to chain error because it's not an associated type anymore.

Thus, for instance in the client-side impls we are wrapping the responses in Result<T, E> where the concrete error type must be returned instead of E: Into<BoxError>`

### DIFF
--- a/client/http-client/src/client.rs
+++ b/client/http-client/src/client.rs
@@ -35,8 +35,8 @@ use crate::{HttpRequest, HttpResponse};
 use hyper::body::Bytes;
 use hyper::http::{Extensions, HeaderMap};
 use jsonrpsee_core::client::{
-	BatchResponse, ClientT, Error, IdKind, MethodResponse, RequestIdManager, Subscription, SubscriptionClientT,
-	generate_batch_id_range,
+	BatchResponse, ClientT, Error, IdKind, MiddlewareBatchResponse, MiddlewareMethodResponse, MiddlewareNotifResponse,
+	RequestIdManager, Subscription, SubscriptionClientT, generate_batch_id_range,
 };
 use jsonrpsee_core::middleware::layer::{RpcLogger, RpcLoggerLayer};
 use jsonrpsee_core::middleware::{Batch, RpcServiceBuilder, RpcServiceT};
@@ -351,7 +351,12 @@ impl HttpClient<HttpBackend> {
 
 impl<S> ClientT for HttpClient<S>
 where
-	S: RpcServiceT<Error = Error, Response = MethodResponse> + Send + Sync,
+	S: RpcServiceT<
+			MethodResponse = Result<MiddlewareMethodResponse, Error>,
+			BatchResponse = Result<MiddlewareBatchResponse, Error>,
+			NotificationResponse = Result<MiddlewareNotifResponse, Error>,
+		> + Send
+		+ Sync,
 {
 	fn notification<Params>(&self, method: &str, params: Params) -> impl Future<Output = Result<(), Error>> + Send
 	where
@@ -363,13 +368,9 @@ where
 				None => None,
 			};
 			let params = params.to_rpc_params()?.map(StdCow::Owned);
+			let fut = self.service.notification(Notification::new(method.into(), params));
 
-			run_future_until_timeout(
-				self.service.notification(Notification::new(method.into(), params)),
-				self.request_timeout,
-			)
-			.await
-			.map_err(|e| Error::Transport(e.into()))?;
+			run_future_until_timeout(fut, self.request_timeout).await.map_err(|e| Error::Transport(e.into()))?;
 			Ok(())
 		}
 	}
@@ -392,8 +393,7 @@ where
 				self.request_timeout,
 			)
 			.await?
-			.into_method_call()
-			.expect("Method call must return a method call reponse; qed");
+			.into_response();
 
 			let rp = ResponseSuccess::try_from(method_response.into_inner())?;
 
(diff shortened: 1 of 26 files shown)


client/http-client/src/client.rs
```diff
@@ -477,7 +476,12 @@ where
 
 impl<S> SubscriptionClientT for HttpClient<S>
 where
-	S: RpcServiceT<Error = Error, Response = MethodResponse> + Send + Sync,
+	S: RpcServiceT<
+			MethodResponse = Result<MiddlewareMethodResponse, Error>,
+			BatchResponse = Result<MiddlewareBatchResponse, Error>,
+			NotificationResponse = Result<MiddlewareNotifResponse, Error>,
+		> + Send
```
jsdw, line 483: Nit: Since `S` isn't used in the impl, can we remove most/all of these constraints here?

core/src/client/error.rs
```diff
@@ -66,4 +66,8 @@ pub enum Error {
 	/// The error returned when registering a method or subscription failed.
 	#[error(transparent)]
 	RegisterMethod(#[from] RegisterMethodError),
+	#[error("Fetch from backend")]
+	/// Internal error state when the underlying channel is closed
+	/// and the error reason needs to be fetched from the backend.
+	FetchFromBackend,
```
jsdw, line 72: Nit: comment below attr unlike above :)

core/src/middleware/mod.rs
```diff
@@ -296,14 +296,15 @@ impl<'a> BatchEntry<'a> {
...
-	type Error: std::fmt::Debug;
-
-	/// The response type
-	type Response: ToJson;
+	/// Response type for `RpcServiceT::call`.
+	type MethodResponse: ToJson;
+	/// Response type for `RpcServiceT::notification`.
+	type NotificationResponse: ToJson;
+	/// Response type for `RpcServiceT::batch`.
+	type BatchResponse: ToJson;
```
jsdw, line 304: iirc you said to me that `ToJson` is only needed in middlewares. If I got this right, perhaps only the middleware that needs it should have this `ToJson` constraint on things and the service itself shouldn't require it?

client/http-client/src/client.rs
niklasad1: It's needed unfortunately because the trait SubscriptionClientT: ClientT and ClientT which requires the concrete types to satisfy the backend/server impl :(

core/src/middleware/mod.rs
niklasad1: good catch, fixed now

client/http-client/src/client.rs
jsdw: Aah gotcha!

core/src/middleware/mod.rs
```diff
@@ -314,13 +315,10 @@ pub trait RpcServiceT {
...
 
 	/// Similar to `RpcServiceT::call` but processes a JSON-RPC notification.
-	fn notification<'a>(
-		&self,
-		n: Notification<'a>,
-	) -> impl Future<Output = Result<Self::Response, Self::Error>> + Send + 'a;
+	fn notification<'a>(&self, n: Notification<'a>) -> impl Future<Output = Self::NotificationResponse> + Send + 'a;
 }
 
 /// Interface for types that can be serialized into JSON.
```
jsdw, line 324: Nit: I'd be tempted to put this with the loggerm iddleware since that's where it's used, and then have the impls on specific types right next to it. That'll make it clear what the logger middleware will work with right off the bat :)

examples/examples/jsonrpsee_as_service.rs
```diff
@@ -51,10 +51,54 @@ use jsonrpsee::ws_client::{HeaderValue, WsClientBuilder};
...
+
+	fn layer(&self, inner: S) -> Self::Service {
+		Identity(inner)
+	}
+}
+
+#[derive(Clone)]
+struct Identity<S>(S);
+
+impl<S> RpcServiceT for Identity<S>
```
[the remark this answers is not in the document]
jsdw: (Though I realise this is because of the impl Future and not the splitting of return types!)

jsdw (approved):
Nice! Just a couple of tiny nits :)

PR #1572, opened 2025-05-13
chore: make RpcService pub

mattsse:
part of pub api via https://github.com/paritytech/jsonrpsee/blob/720591fd2697576ab80267bfe90352c412125547/client/http-client/src/client.rs#L329-L329

### DIFF
--- a/client/http-client/src/lib.rs
+++ b/client/http-client/src/lib.rs
@@ -37,6 +37,7 @@
 
 mod client;
 mod rpc_service;
+pub use rpc_service::RpcService;
 
 /// HTTP transport.
 pub mod transport;


emhane (approved):

jsdw:
Maybea adumb question but; where in your code do you need to name this `RpcService` type? I looked at the original issue and it wasn't clear offhand!

mattsse:
@jsdw if you do something like

image

then you can't name that type because that type is private

lexnv:
dq: Would using the ClientT trait or SubscriptionT (or both) trait solve the problem here? 
https://github.com/paritytech/jsonrpsee/blob/720591fd2697576ab80267bfe90352c412125547/client/http-client/src/client.rs#L352

https://github.com/paritytech/jsonrpsee/blob/720591fd2697576ab80267bfe90352c412125547/client/http-client/src/client.rs#L477

mattsse:
@lexnv yeah that was the workaround we went with, the downside of this is that because this is also not object safe, it effecitvely requires generics to replace something that previously used the named type like `struct C {client: HttpClient<...>}`

jsdw:
Thankyou for your responses! I'll branch off this PR to merge your change and also document RpcService since it'll become public. It makes sense to me to expose it. My only hesitation is that the `RpcService` name is overloaded in several places so it has the potential to cause confusion by exposing another one, but it'll have to do for now :)

jsdw:
Closing this; will merge #1574 to handle it; thanks for the PR! :)

PR #1585, opened 2025-07-01
increase max websocket frame size

tsite:
Increase the max frame size so that the jsonrpsee client can decode large responses when max_response_size is set

### DIFF
--- a/client/transport/src/ws/mod.rs
+++ b/client/transport/src/ws/mod.rs
@@ -497,6 +497,7 @@ impl WsTransportClientBuilder {
 			Ok(ServerResponse::Accepted { .. }) => {
 				tracing::debug!(target: LOG_TARGET, "Connection established to target: {:?}", target);
 				let mut builder = client.into_builder();
+				builder.set_max_frame_size(usize::MAX);
 				builder.set_max_message_size(self.max_response_size as usize);
 				let (sender, receiver) = builder.finish();
 				Ok((Sender { inner: sender, max_request_size: self.max_request_size }, Receiver { inner: receiver }))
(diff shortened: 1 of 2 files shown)


client/transport/src/ws/mod.rs
```diff
@@ -497,6 +497,7 @@ impl WsTransportClientBuilder {
 			Ok(ServerResponse::Accepted { .. }) => {
 				tracing::debug!(target: LOG_TARGET, "Connection established to target: {:?}", target);
 				let mut builder = client.into_builder();
+				builder.set_max_frame_size(usize::MAX);
```
lexnv, line 500: We'll need to expose this option on the builder itself. This way we are not opinionated on some specifics, and developers can control the maximum frame without us needing to create a different release.

Please note that setting this to `usize::MAX` might be problematic, as it creates a possible entry for exploits. I'm not aware what the underlying code is doing but pressume a similar situation that could DoS the service:

```rust
let s = socket.read_frame_size();
if s > max_frame (in this case usize::MAX) {
    return TooBig;
}

// s here might be usize::MAX or very close to it, such that resources of the node are exhausted.
let frame_buffer = Buffer::allocate(s);
```

I would see the code as:

```rust
	/// Set the maximum size of a frame in bytes. Default is None.
	pub fn max_frame_size(mut self, size: u32) -> Self {
        // self.max_response_size is None by default.
		self.max_response_size = Some(size);
		self
	}
```

Then in this code we can just use the builder parameters set above:

```rust
// Use the max frame size if any, otherwise let the underlying code use appropriate defaults.
if let Some(max_frame_size) = self.max_frame_size {
    builder.set_max_frame_size(max_frame_size);
}
```

tsite: makes sense - I updated the code to make the max frame size customizable

jsdw (approved):

jsdw (approved):

lexnv (approved):
Nice!

PR #1588, opened 2025-07-11
chore: expose `TowerServiceNoHttp` type

koushiro:
- expose `TowerServiceNoHttp` type
- improve some trait bounds

PR #1593, opened 2025-08-10
chore: make  `TowerServiceNoHttp` pub

0xOsiris:
This PR makes `TowerServiceNoHttp` pub. This is useful as `Server` directly bounds `HttpMiddleware` with respect to `TowerServiceNoHttp`, so constraining a generic tower `Layer` within the `Server` is not possible


jsdw:
Will close this in favour of #1588; thank you though!

PR #1627, opened 2026-02-13
refactor: replace soketto with yawc for per-message deflate support

ifdario:
Migrate from soketto to yawc across the entire codebase (server, client transport, ws-client, test-utils). yawc provides RFC 7692 permessage-deflate compression with balanced compression enabled by default, near-zero-copy frame processing, and SIMD-optimized masking.

Key changes:
- Replace soketto handshake/connection with yawc WebSocket::upgrade (server) and WebSocket::handshake_with_request (client)
- Remove tokio-util compat layer (yawc uses tokio I/O natively)
- Add configurable WsOptions (compression level, payload limits) to ServerConfigBuilder, WsTransportClientBuilder, and WsClientBuilder
- Re-export DeflateOptions and WsOptions from server and client crates
- Update test mocks to use yawc with safe Arc<Mutex<Option>> pattern
- Adapt max request body size test for yawc's connection-terminating behavior on oversized payloads

### DIFF
--- a/server/src/server.rs
+++ b/server/src/server.rs
@@ -39,8 +39,8 @@ use crate::transport::{http, ws};
 use crate::utils::deserialize_with_ext;
 use crate::{Extensions, HttpBody, HttpRequest, HttpResponse, LOG_TARGET};
 
+use futures_util::StreamExt;
 use futures_util::future::{self, Either, FutureExt};
-use futures_util::io::{BufReader, BufWriter};
 use hyper::body::Bytes;
 use hyper_util::rt::{TokioExecutor, TokioIo};
 use jsonrpsee_core::id_providers::RandomIntegerIdProvider;
@@ -53,10 +53,8 @@ use jsonrpsee_types::error::{
 	BATCHES_NOT_SUPPORTED_CODE, BATCHES_NOT_SUPPORTED_MSG, ErrorCode, reject_too_big_batch_request,
 };
 use jsonrpsee_types::{ErrorObject, Id};
-use soketto::handshake::http::is_upgrade_request;
 use tokio::net::{TcpListener, TcpStream, ToSocketAddrs};
 use tokio::sync::{OwnedSemaphorePermit, mpsc, watch};
-use tokio_util::compat::TokioAsyncReadCompatExt;
 use tower::layer::util::Identity;
 use tower::{Layer, Service};
 use tracing::{Instrument, instrument};
@@ -170,7 +168,7 @@ where
 }
 
 /// Static server configuration which is shared per connection.
-#[derive(Debug, Clone)]
+#[derive(Clone)]
 pub struct ServerConfig {
 	/// Maximum size in bytes of a request.
 	pub(crate) max_request_body_size: u32,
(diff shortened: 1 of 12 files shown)


jsdw:
Thankyou for your PR. For full context, I can see that `yawc` is a library that you authored or at least contribute heavily too, and that you've opened several similar PRs. I appreciate the graft!

That said, while I'm curious to see how far this PR would get on our CI run, we don't have much capacity for reviewing such a change (especially as we don't require it for the current jsonrpsee use cases), and we also own `soketto` and have the internal knowledge to fix and tweak it as needed. Thus, while I am curious, I would not hold your breath re this being merged and if it is, it may take a while! :)

ifdario:
Hello @jsdw 

I have been planning this PRs for quite some time, waiting for yawc to be stable enough. I consider that yawc has reached this point.
yawc is quite curated and it has been battle tested. Passes all autobahn tests and many other thorough tests. It is design with easy fixability (is that a word?) in mind. Also the library is pretty much organized in simple components that work independently with sans-io design in mind.
Having said that, the main purpose I performed all of this PRs is the services I use ([vector](https://github.com/vectordotdev/vector) and [nanoreth](https://github.com/hl-archive-node/nanoreth)) they both use websocket libraries that generally do not support compression. In your case, I see [soketto](https://github.com/paritytech/soketto/blob/master/src/extension/deflate.rs#L20) does. The main difference between our crates being [yawc](https://github.com/infinitefield/yawc) supports compression written fully in Rust. Among other features, yawc also supports streaming, stream-decompression, full-duplex concurrency (through Sink/StreamExt traits) and simd masking. Also, if it is of any interest, wasm too.

Feel free to leave the PR open if you see no use case or you would like to leave it for the future.

niklasad1:
Would be cool if you could run the benchmarks on this branch and compare to master, it seems that you don't have any benches against soketto in yawc.

Nice that all tests passes but it's up to @jsdw to decide as the solo maintainer of jsonrpsee these days :grinning:

loyd:
I've tried this yawc-based implementation (without performance comparison yet); it works pretty stably, including compression. Soketto looks abandoned: the last changes were made 2 years ago, and critical compression issues have persisted there for years (https://github.com/paritytech/soketto/issues/49).

However, I don't think maintainers will replace the implementation, since Soketto is developed by the same org. So, this is a conflict of interest.

server/src/server.rs
```diff
@@ -541,6 +588,29 @@ impl ServerConfigBuilder {
...
+	///
+	/// // Disable compression
+	/// let ws_opts = WsOptions::default().without_compression();
+	/// let builder = ServerConfigBuilder::default().set_ws_options(ws_opts);
+	///
+	/// // High compression
+	/// let ws_opts = WsOptions::default().with_high_compression();
+	/// let builder = ServerConfigBuilder::default().set_ws_options(ws_opts);
+	/// ```
+	pub fn set_ws_options(mut self, options: yawc::Options) -> Self {
```
loyd, line 609: It's a leaky abstraction:
1. If `yawc` is updated (it's still `v0.x`), it will affect the version of jsonrpsee. For instance, this `yawc::Options` is not even `non_exhaustive`, so even adding a new option is a breaking change (raising questions about the current state of `yawc` API).
2. It makes it more complicated to switch backend or support a new one.

I can't say that `jsonrpsee` is a well-designed API, quite the opposite. But it would be great to avoid unfounded breaking changes in the future

server/src/tests/ws.rs
```diff
@@ -56,14 +56,15 @@ async fn can_set_the_max_request_body_size() {
...
+	assert!(client.send_request_text(req).await.is_err());
 
-	// Max request body size should not override the max response body size
+	// After an oversized request, the connection is closed; verify with a new connection
+	// that normal-sized requests still work fine.
+	let mut client2 = WebSocketTestClient::new(addr).await.unwrap();
 	let req = r#"{"jsonrpc":"2.0","method":"anything","id":1}"#;
-	let response = client.send_request_text(req).await.unwrap();
+	let response = client2.send_request_text(req).await.unwrap();
 	assert_eq!(response, ok_response(JsonValue::String("a".repeat(100)), Id::Num(1)));
```
loyd, line 69: I think the integration test (like https://github.com/paritytech/jsonrpsee/pull/1632/changes#diff-8f746fec0e7d22931bdfb823ebec4c4079430e50bce5b4a795f9972207d239cdR1609) should be added to check compression

server/src/server.rs
```diff
@@ -541,6 +588,29 @@ impl ServerConfigBuilder {
 		self
 	}
 
+	/// Set custom WebSocket options such as compression settings.
+	///
+	/// By default, balanced compression is enabled. The `max_payload_read` setting
```
loyd, line 593: > By default, balanced compression is enabled

I think custom `Options` can have this in the `Default` instance instead of `.unwrap_or()` handling in the code.

loyd, line 593: > from these options will be overridden by [`ServerConfigBuilder::max_request_body_size`].

... which is set by default. One more point that it shouldn't accept `yawc::Options` directly

jsdw:
Apologies; I am leaving Parity myself tomorrow and so will pass on mention of these PRs internally and hopefully somebody else can pick them up!

PR #1629, opened 2026-04-08
Fix clippy and compiler warnings across client and server crates

merkle-maren:
Warnings surfaced while working on #1628. Merging these separately - they're independent from that work and trivial to review on their own.

### Changes

- Removed unused lifetime parameter `'a` from `process_connection` in the server crate.
- Fixed `mismatched_lifetime_syntaxes` warnings in `http-client` and `ws-client` test helpers by using explicit `'static` lifetimes consistently.
- Suppressed dead code warnings in `core::client` for types only used when `async-client` or `async-wasm-client` features are enabled (`FrontToBack`, `BatchMessage`, `SubscriptionSender`, etc.).

### Motivation

`cargo clippy` and `cargo test -p jsonrpsee-http-client` produced ~14 compiler warnings. Mix of an unused lifetime, inconsistent lifetime syntax, and dead code from types gated behind feature flags not active in all compilation contexts.

Clean builds should be clean. Warnings that linger get ignored, and ignored warnings hide real problems.

### Test plan

- `cargo clippy --all-targets --all-features` — zero warnings.
- `cargo test --features tls` — zero warnings.

### DIFF
--- a/client/http-client/src/tests.rs
+++ b/client/http-client/src/tests.rs
@@ -252,9 +252,9 @@ async fn batch_request_out_of_order_response() {
 }
 
 async fn run_batch_request_with_response<T: Send + DeserializeOwned + std::fmt::Debug + Clone + 'static>(
-	batch: BatchRequestBuilder<'_>,
+	batch: BatchRequestBuilder<'static>,
 	response: String,
-) -> Result<BatchResponse<T>, ClientError> {
+) -> Result<BatchResponse<'static, T>, ClientError> {
 	let server_addr = http_server_with_hardcoded_response(response).with_default_timeout().await.unwrap();
 	let uri = format!("http://{server_addr}");
 	let client = HttpClientBuilder::default().build(&uri).unwrap();
--- a/client/http-client/src/tests.rs
+++ b/client/http-client/src/tests.rs
@@ -251,10 +251,10 @@ async fn batch_request_out_of_order_response() {
 	assert_eq!(response, vec!["hello".to_string(), "goodbye".to_string(), "here's your swag".to_string()]);
 }
 
-async fn run_batch_request_with_response<T: Send + DeserializeOwned + std::fmt::Debug + Clone + 'static>(
-	batch: BatchRequestBuilder<'static>,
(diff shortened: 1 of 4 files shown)


core/src/client/mod.rs
```diff
@@ -63,6 +63,7 @@ pub(crate) struct SubscriptionLagged(Arc<RwLock<bool>>);
 /// Owned version of [`RawResponse`].
 pub type RawResponseOwned = RawResponse<'static>;
 
+#[allow(dead_code)]
```
jsdw: All of these `#[allow(dead_code)]`s I would guess (without checking myself) are feature flag related, and the relevant things should be hidden when the feature(s) they are used for aren't enabled, rather then ignored via dead_code (which might legitimately mean they aren't needed anywhere)

client/http-client/src/tests.rs
```diff
@@ -252,9 +252,9 @@ async fn batch_request_out_of_order_response() {
 }
 
 async fn run_batch_request_with_response<T: Send + DeserializeOwned + std::fmt::Debug + Clone + 'static>(
-	batch: BatchRequestBuilder<'_>,
+	batch: BatchRequestBuilder<'static>,
```
jsdw: I would have thought that adding `'_` fo the `BatchResponse` type would also work here, but the proper equivalent would be adding a `'a` lifetime in the generic params and using that. Elided / unnamed lifetimes aren't the same as static lifetimes.

client/ws-client/src/tests.rs
```diff
@@ -440,9 +440,9 @@ async fn is_connected_works() {
 }
 
 async fn run_batch_request_with_response<T: Send + DeserializeOwned + std::fmt::Debug + Clone + 'static>(
-	batch: BatchRequestBuilder<'_>,
+	batch: BatchRequestBuilder<'static>,
```
jsdw: As above

merkle-maren:
Removed all `#[allow(dead_code)]`.

These trigger warnings on `cargo test -p jsonrpsee-http-client` but not on `cargo test`. I couldn't find a feature flag or conditional compilation path that accounts for the difference. Rather than suppress warnings we don't fully understand, I've dropped the annotations and fixed the lifetime warnings directly.

Open to a better approach if you know the exact cfg gate here.

PR #1632, opened 2026-04-21
feat: enable WS deflate in soketto

loyd:
This PR introduces the new feature `ws-deflate` and extends server/client configs with `ws_compression: bool`. It doesn't replace soketto with yawc, only enables existing extension in soketto if the corresponding feature is enabled.

Alternative to https://github.com/paritytech/jsonrpsee/pull/1627

cc @jsdw

### DIFF
--- a/client/transport/Cargo.toml
+++ b/client/transport/Cargo.toml
@@ -60,6 +60,7 @@ ws = [
     "tracing",
     "url",
 ]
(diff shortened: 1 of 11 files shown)


loyd:
Oh, it's also affected by https://github.com/paritytech/soketto/issues/49

PR #1634, opened 2026-05-21
server: expose subscription id on PendingSubscriptionSink

DenzelPenzel:
## Summary

Expose the generated subscription id on `PendingSubscriptionSink`.

The subscription id is generated before the subscription callback is invoked and is already used by `accept()` to build the subscription response. However, it is currently only publicly accessible after accepting the subscription through `SubscriptionSink::subscription_id()`.

Some servers need to register per-subscription state before accepting the subscription. Otherwise, follow-up RPC calls using the subscription id can race with server-side registration.

This adds a small getter mirroring the existing `SubscriptionSink::subscription_id()` API.

### DIFF
--- a/core/src/server/subscription.rs
+++ b/core/src/server/subscription.rs
@@ -266,6 +266,11 @@ impl PendingSubscriptionSink {
 		}
 	}
 
+	/// Get the subscription ID.
+	pub fn subscription_id(&self) -> SubscriptionId<'static> {
+		self.uniq_sub.sub_id.clone()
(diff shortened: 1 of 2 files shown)


lexnv (approved):

lexnv:
I would also tackle the CI issues, not entirely sure if related to this PR 🙏 

Ping me if you'd like to land this in a release soon

AndreiEres (approved):
Good!

PR #1642, opened 2026-06-09
fix: make domain name resolution async

citizen-stig:
So faulty resovler won't starve runtime.

try_connect_over_tcp resolved DNS with uri.socket_addrs(|| None) (initial connect and redirect) → std::net::ToSocketAddrs → blocking getaddrinfo on the current Tokio worker thread, outside the timeout. The connection_timeout only ever wrapped TcpStream::connect (inside connect() via tokio::select!).

Include resolution into timeout

### DIFF
--- a/client/transport/src/ws/mod.rs
+++ b/client/transport/src/ws/mod.rs
@@ -531,8 +531,8 @@ impl WsTransportClientBuilder {
 ///
 /// This replaces `url::Url::socket_addrs`, which performs *blocking* `getaddrinfo`
 /// resolution (via `std::net::ToSocketAddrs`) on the calling thread. Inside the async client
-/// that means a Tokio worker thread can be blocked for as long as the system resolver takes
-/// `tokio::net::lookup_host` runs the  lookup on Tokio's blocking thread pool instead,
+/// that means a Tokio worker thread can be blocked for as long as the system resolver takes.
+/// `tokio::net::lookup_host` runs the lookup on Tokio's blocking thread pool instead,
 /// and IP literals are turned into a `SocketAddr` directly without any DNS lookup.
 ///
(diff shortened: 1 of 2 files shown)


bkolad (approved):

citizen-stig:
Hey @lexnv , if you can take a look on this PR too, that would be fantastic. Failures look the same as in #1643 . Formatting is fixed.

citizen-stig:
@DenzelPenzel I also would like to bump this PR

client/transport/src/ws/mod.rs
```diff
@@ -530,6 +527,42 @@ impl WsTransportClientBuilder {
...
+		let port =
+			uri.port_or_known_default().ok_or_else(|| WsHandshakeError::Url("No port number in the URL".into()))?;
+
+		// NOTE: match on `uri.host()` (the `url::Host` enum) rather than `host_str()`. The latter
+		// returns IPv6 hosts *with* brackets (e.g. `"[::1]"`), which `getaddrinfo` rejects, whereas
+		// the enum exposes a parsed `Ipv6Addr`. This mirrors what `url::Url::socket_addrs` does
+		// internally, so IP literals and default ports behave exactly as before.
+		match uri.host() {
+			Some(url::Host::Domain(domain)) => {
+				tokio::net::lookup_host((domain, port)).await.map(|addrs| addrs.collect()).map_err(|e| {
```
DenzelPenzel, line 552: timeout stops waiting for `lookup_host`, but it does not cancel the underlying `spawn_blocking` DNS lookup. wdty?

citizen-stig: That's true, but I don't know how it can addressed. 
But current approach makes situation a little better than on main, because on main broken resolver will stall whole runtime, while here we will have task sitting on blocking pool thread.

PR #1645, opened 2026-06-29
proc-macros/tests: Fix CI

lexnv:


server/src/tests/ws.rs
```diff
@@ -735,6 +735,10 @@ async fn ws_server_backpressure_works() {
 				let bp = serde_json::value::to_raw_value(&2).unwrap();
 
 				let mut msg = n.clone();
+				// Only notify the client once that backpressure was hit. The signal channel has a
+				// capacity of one and the client reads it a single time, so re-sending on every
+				// timeout would eventually block this task forever and stall the subscription.
```
michalkucharczyk: Var is self-explaining I would say...
```suggestion
```

michalkucharczyk (approved):

lrubasze (approved):

PR #1646, opened 2026-07-01
fix: an in-progress RPC call is not aborted when the websocket connection is closed

hanabi1224:
This PR fixes the isseue and adds some tests

### DIFF
--- a/tests/tests/integration_tests.rs
+++ b/tests/tests/integration_tests.rs
@@ -58,7 +58,7 @@ use tokio::time::interval;
 use tokio_stream::wrappers::IntervalStream;
 use tower_http::cors::CorsLayer;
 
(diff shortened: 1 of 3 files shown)


tests/tests/integration_tests.rs
```diff
@@ -289,6 +289,48 @@ async fn http_method_call_works() {
...
+	assert!(
+		tokio::time::timeout(
+			Duration::from_millis(100),
+			client.request::<String, _>("blocking_hello_with_cleanup_notif", rpc_params![]),
+		)
+		.await
+		.is_err()
+	);
+	drop(client);
+	notified.await
```
hanabi1224, line 327: This test times out without the fix

